PR #2, opened 2019-03-29
Load gtfs, and /api/stop-list endpoint

skyqrose:
Runs a genserver (`gtfs.ex`) that fetches the gtfs file from the cdn, and stores the files in its state.

Right now, it only stores the ids from `stops.txt`, but more can be added easily.

There's an endpoint `/api/stop-list` that returns all the stops.


paulswartz:
I thought the ticket for this was routes/trips/stop_times?

skyqrose:
Yeah, I guess? I hadn't gotten that far. This was mostly to test that the whole pipeline works, and I'll likely rip it out and replace it with an endpoint that does something useful once I write that more complicated piece. But I guess I should have started with routes instead of stops.

arkadyan:
At a real quick glance this is looking like a great start. @paulswartz can correct me if I'm misinterpretting, but I think he was speaking less about the endpoint, and more about https://app.asana.com/0/1112935048846093/1114820119722033 asking for routes, trips, and stops to all be ingested from GTFS. Does that seem like a good amount for this PR?

paulswartz:
@arkadyan is correct. Stop data (at the moment) isn't an important set of data to import. I wanted to be sure we're focused on getting to the MVP.

skyqrose:
Okay. I'll close this PR and reopen it once I have all those in gtfs.

config/config.exs
```diff
@@ -7,6 +7,9 @@
 # General application configuration
 use Mix.Config
 
+config :skate,
+  gtfs_url: "https://cdn.mbta.com/MBTA_GTFS.zip"
```
arkadyan: I know you are still working on this, so I'll wait on a full review, but this one line jumped out at me. Is this something that should be configured outside of the code base? Maybe we can discuss using a `.env` file or ENV vars or some other such solution, I'm not too particular.

paulswartz: :+1: that's also one of the things called out in the ticket. Should we do a grooming next week to make sure everyone's on the same page w/ these tickets?

PR #3, opened 2019-04-01
Load GTFS

skyqrose:
* Can optionally specify the url to get GTFS with the `SKATE_GTFS_URL` env variable.
* Loads `routes.txt`, `stops.txt`, `stop_times.txt`, and `trips.txt`. (Just ids for now, but adding new columns and files is easy.)
* `/api/routes` endpoint just to make sure everything works.
* Easy way to create mocked state for testing the queries.
* It's a little slow to load the `stop_times` (~75 seconds, which is comparable to how long the api takes to load them) so the gen server with real data is turned off when testing.


lib/gtfs.ex
```diff
@@ -0,0 +1,165 @@
...
+  # Client functions
+
+  @spec start_link(String.t()) :: GenServer.on_start()
+  def start_link(url) do
+    GenServer.start_link(__MODULE__, {:url, url}, name: __MODULE__)
+  end
+
+  @spec start_mocked(mocked_files) :: GenServer.on_start()
+  def start_mocked(mocked_files) do
+    GenServer.start_link(__MODULE__, {:mocked_files, mocked_files})
```
arkadyan: Seems like we aren't using it right now, but would it make sense to include `name: __MODULE__` here too as in line 52 for consistency?

lib/gtfs.ex
```diff
@@ -0,0 +1,165 @@
...
+  def gtfs(opts) do
+    server = opts[:server] || __MODULE__
+    GenServer.call(server, :state)
+  end
+
+  # GenServer callbacks
+
+  @impl true
+  def init(files_source) do
+    case fetch_files(files_source) do
```
arkadyan: This works, but `with` is another option for this as well (and same in `fetch_files`).

lib/gtfs/route.ex
```diff
@@ -0,0 +1,24 @@
...
+  ]
+
+  @derive Jason.Encoder
+
+  defstruct [
+    :id
+  ]
+
+  @spec from_csv_row(%{required(String.t()) => String.t()}) :: t()
+  def from_csv_row(row) do
```
arkadyan: They are trivial now, but could it be helpful in the future to have tests for these `from_csv_row` functions?

lib/skate_web/controllers/route_controller.ex
```diff
@@ -0,0 +1,8 @@
+defmodule SkateWeb.RouteController do
+  use SkateWeb, :controller
+
+  def index(conn, _params) do
+    routes = Gtfs.Query.all_routes()
+    json(conn, %{data: routes})
```
arkadyan: Could `routes` be a more descriptive key than `data`?

lib/gtfs.ex
skyqrose: The goal here was to be able to have multiple mocked genservers running independently. They would be accessed by their id, and the `name: __MODULE__` would be reserved for the genserver with real data.

lib/gtfs.ex
skyqrose: `with` will return the failing value, right? In this case, we want control over that value, so we can return `{:stop, ...}` instead of whatever `HTTPoison` returns.

lib/gtfs/route.ex
skyqrose: Hmm... I think these functions are too trivial to be worth it, but it would be good to test the data loading process, probably as a whole.
Something like
```
server = start_mocked(%{a bunch of files})
assert gtfs(server) == %__Gtfs__{all the data}
```
I'll add that into the next PR.

lib/gtfs.ex
arkadyan: Ah right, of course.

lib/gtfs.ex
```diff
@@ -0,0 +1,165 @@
...
+        state = parse_files(files)
+        Logger.info(fn -> "Successfully loaded gtfs" end)
+        {:ok, state}
+    end
+  end
+
+  @impl true
+  def handle_call(:state, _from, state) do
+    {:reply, state, state}
+  end
```
paulswartz: You probably want calls to return particular parts of the state, rather than the whole thing. Let the GenServer encapsulate the state, and have the clients request particular pieces they're looking for. In this case, something like `handle_call(:routes, _from, state), do: {:reply, state.routes, state}`

lib/gtfs.ex
```diff
@@ -0,0 +1,165 @@
...
+          "stop_times.txt",
+          "stops.txt",
+          "trips.txt"
+        ]
+
+        unzip_files(zip_binary, file_list)
+
+      response ->
+        Logger.warn(fn -> "Unexpected response from #{url} : #{inspect(response)}" end)
+        {:err, response}
```
paulswartz: `{:error, ...}`

lib/skate/application.ex
```diff
@@ -14,6 +16,14 @@ defmodule Skate.Application do
       # {Skate.Worker, arg},
     ]
 
+    # Don't start the Gtfs gen server when testing, because it's slow to start
```
paulswartz: It's slow to start because of the processing that's happening at startup. You probably won't need this if you do the `init` refactoring.

config/config.exs
```diff
@@ -7,6 +7,9 @@
 # General application configuration
 use Mix.Config
 
+config :skate,
+  gtfs_url: System.get_env("SKATE_GTFS_URL") || "https://cdn.mbta.com/MBTA_GTFS.zip"
```
paulswartz: This works in local development, but won't in production as the configuration is built at compile time.

Instead, you can override the configuration in `Skate.Application.start/2`, which is called at runtime:
```elixir
def start(_, _) do
  override_config!()
end

defp override_config do
  for {application_key, env_var} <- [gtfs_url: "SKATE_GTFS_URL"] do
    if value = System.get_env(env_var) do
      Application.put_env(:skate, application_key, value)
    end
  end
  :ok
end
```

lib/skate_web/controllers/route_controller.ex
skyqrose: You'd know it's routes because it's from the `/routes` endpoint. This is just to get around the restriction that you can only return objects as the root of a json endpoint. Since all the endpoints will need this, I think it will be easier to always use `data:` rather than try to remember which endpoint uses which field name. This is consistent with the way the v3 api uses the `data` field in the json api format.

lib/gtfs.ex
arkadyan: You can use `else` with `with`, so you'd have something along the lines of:

    with {:files, files} <- fetch_files(files_source) do
      <do work>
    else
      {:err, err} ->
        {:stop, err}
    end

(Optional, but returning data as a tuple (i.e. `{:files, files}`) can make it better for deciphering positive results from errors in some cases.)

lib/skate_web/controllers/route_controller.ex
arkadyan: Ok, that makes sense.

lib/gtfs.ex
paulswartz: You can return `:not_loaded` or an empty list if the data hasn't loaded yet. It's an OTP anti-pattern to do long processing in `init/1`, as the supervisor can't do anything while it waits for the process to start. See https://medium.com/@derek.kraan2/avoid-these-otp-supervision-performance-pitfalls-c3fe64e70f9e for more.

lib/gtfs.ex
paulswartz: I suppose, but even then you probably want to provide helper functions to `Gtfs.Query`. Otherwise, now you've got two modules which need to know about the shape of the state here, rather than it being encapsulated by `Gtfs`. It also sets you up better when there's more data here, as calls won't need to copy the entire state to another process when a request comes in.

config/config.exs
skyqrose: Really? Configuring it like this is what Glides does to have different databases for dev and prod. https://github.com/mbta/glides/blob/master/config/prod.exs#L72

lib/gtfs.ex
skyqrose: That's why I clearly defined the shape of the data in `Gtfs.t()`. I think that sharing type between `Gtfs.Query` and `Gtfs` is less trouble than trying to keep query logic organized in `gtfs.ex`.

Will returning the state require copying all the data, or would it be a reference to shared memory?

lib/gtfs.ex
skyqrose: This article alleviates my concern that loading asynchronously would give other processes incomplete information: https://elixirschool.com/blog/til-genserver-handle-continue/

> This approach ensures that our GenServer won’t handle any other messages until `handle_continue/2` is finished. No more race conditions!

lib/gtfs/route.ex
skyqrose: Nvm. Added it to this PR.

config/config.exs
paulswartz: Glides is deployed differently, as it was built before the deployment code was added. It's not great to leave it like this, but if you do, please add a comment so that future-us knows why the deploys don't pick up the configuration change.

lib/gtfs.ex
paulswartz: It's copied. From http://erlang.org/doc/efficiency_guide/processes.html#process-messages

> All data in messages between Erlang processes is copied, except for refc binaries on the same Erlang node.

If you want to keep is this way, I'd flip how the state/querying logic is split.
* Gtfs.Query: responsible for managing the struct, and returning results based on that struct (no process)
* Gtfs: responsible for fetching/keeping the Gtfs.Query struct, and calls functions in Gtfs.Query to return results.

For example:
```elixir
defmodule Gtfs do
  def routes(server \\ __MODULE__) do
    GenServer.call(server, :routes)
  end

  def handle_continue(:fetch, state) do
    with {:ok, %{body: body}} <- HTTPoison.get(url()) do
      new_state = Gtfs.Query.from_zip_file(body)
      {:noreply, new_state}
    else
      _ -> {:stop, :normal, state}
  end

  def handle_call(:routes, _from, state) do
    routes = Gtfs.Query.routes(state)
    {:reply, routes, state}
  end
end

defmodule Gtfs.Query do
  defstruct [:routes, :trips, ...]

  def from_zip_file(zip_file) do
    %__MODULE__{routes: ...}
  end

  def routes(%__MODULE__{routes: routes}), do: routes
end
```

config/config.exs
skyqrose: [Follow up task](https://app.asana.com/0/1112935048846093/1116881667899749)
I'll do it this sprint.

lib/gtfs.ex
skyqrose: Discussion continued in slack: https://mbtace.slack.com/archives/CESHREAS1/p1554216526039500

config/config.exs
skyqrose: Ended up getting around to this before this PR merged, so I've included it here.

lib/gtfs.ex
skyqrose: Addressed by moving `Gtfs.Query` into `Gtfs`. There now needs to be two functions per query, and the queries are mixed in the same file as the genserver. That's fine for now. If it becomes too unwieldy when we have many more queries, we can look at splitting them again.

test/gtfs_test.exs
```diff
@@ -0,0 +1,75 @@
...
+      trips: [
+        %Trip{
+          id: "red-trip",
+          route_id: "Red"
+        }
+      ]
+    }
+
+    {:ok, pid} = Gtfs.start_mocked(csvs)
+    assert Gtfs.state(pid) == {:loaded, expected}
```
paulswartz: This seems like a duplicate test of the `all_routes` test below? You're adding a function that's only used for testing, so you could save the test and the extra function by removing both :)

skyqrose: I was thinking this would be for testing the csv parsing and any preprocessing, but I guess that that will all be tested by the tests for each query.

skyqrose:
`mix coveralls.json` is failing and I don't know why.
Here's the output from semaphore:
```
==> skate
Compiling 20 files (.ex)
Generated skate app
...

  1) test all_routes all_routes (GtfsTest)
     test/gtfs_test.exs:9
     ** (EXIT from #PID<0.2457.0>) an exception was raised:
         ** (ErlangError) Erlang error: :timeout_value
             (stdlib) gen_server.erl:389: :gen_server.loop/7
             (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3



  2) test GET / (SkateWeb.PageControllerTest)
     test/skate_web/controllers/page_controller_test.exs:4
     ** (ArgumentError) argument error
     code: conn = get(conn, "/")
     stacktrace:
       (stdlib) :ets.lookup(SkateWeb.Endpoint, :secret_key_base)
       (skate) lib/phoenix/endpoint.ex:632: SkateWeb.Endpoint.config/2
       (elixir) lib/map.ex:752: Map.update!/3
       (skate) lib/skate_web/endpoint.ex:1: SkateWeb.Endpoint.call/2
       (phoenix) lib/phoenix/test/conn_test.ex:235: Phoenix.ConnTest.dispatch/5
       test/skate_web/controllers/page_controller_test.exs:5: (test)



Finished in 0.06 seconds
1 doctest, 4 tests, 2 failures

Randomized with seed 463422
```

Here's the output when I run it locally:
```
16:12 skate % mix coveralls.json
.....

Finished in 0.2 seconds
1 doctest, 4 tests, 0 failures

Randomized with seed 157032
```

What might be different between the environments?

paulswartz:
This looks good. However, it doesn't completely implement https://app.asana.com/0/1112935048846093/1114820119722037 which was also looking for hyperlinks. That can happen in a separate PR.

PR #4, opened 2019-04-03
unsorted list of checkpoints

skyqrose:
Asana Task:
  - [Calculate a static route ladder for a single route/direction from GTFS - Part 1 - Unsorted time points](https://app.asana.com/0/1112935048846093/1114820119722039)
  - [Calculate list of bus routes (and whatever other information is needed for UI)](https://app.asana.com/0/1112935048846093/1114820119722037)

3 new endpoints:
  - `/api/routes/:route_id` (List of checkpoints on the route)
Screen Shot 2019-04-03 at 10 58 24
  - `/ladder` (List of routes)
Screen Shot 2019-04-03 at 10 58 41
  - `/ladder/:route_id` (List of checkpoints on the route)
Screen Shot 2019-04-03 at 10 58 49

I still don't have controller tests because I think of these as kind of throwaway endpoints.

### DIFF
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -3,6 +3,7 @@ defmodule Gtfs do
   require Logger
 
   alias Gtfs.Route
+  alias Gtfs.RoutePattern
   alias Gtfs.Stop
   alias Gtfs.StopTime
   alias Gtfs.Timepoint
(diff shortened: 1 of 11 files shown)


lib/gtfs/checkpoint.ex
```diff
@@ -0,0 +1,3 @@
+defmodule Gtfs.Checkpoint do
```
paulswartz: I know that they're in a field called `checkpoint_id` in the GTFS file, but to avoid confusion they should be called Timepoints internally & externally.

PR #6, opened 2019-04-03
sort timepoints

skyqrose:
Asana Task: [Calculate a static route ladder for a single route/direction from GTFS - Part 2 - Sort the time points](https://app.asana.com/0/1112935048846093/1116725298640264)

Sets the order on `/ladder/:route_id` and `/api/routes/:route_id`.

- Splits the route_patterns by direction (requires adding `direction.ex` and related fields)
- Merges all the variants for each direction.
- Merges the two directions

Uses the same algorithm as the api (except the api makes an effort to put long branches after shorter ones, which I haven't done here). There are lots of situations where the algorithm finds it ambiguous what order is better. As an example, the 225 timepoint for the Quintree Mall (only used on the short turn variant) is listed as outside of Columbian Sq (the extension's terminal).

To provide an order that feels natural to inspectors, we'll probably need to either adapt the algorithm based on our domain knowledge (like the api's branch handling), or provide manual hints (like [the api's fix for the recent plimptonville ordering bug](https://github.com/mbta/api/pull/21)). I haven't made any attempt at those yet. I figure we'll start thinking about that after we have some experience with this order.

Open question: should we return the stops in inbound-to-outbound order or outbound-to-inbound order? I arbitrarily chose outbound-to-inbound.


lib/gtfs.ex
```diff
@@ -73,16 +73,38 @@ defmodule Gtfs do
...
-    timepoint_ids =
+    route_patterns_by_direction =
       gtfs_data.route_patterns
       |> Enum.filter(fn route_pattern -> route_pattern.route_id == route_id end)
-      |> Enum.map(fn route_pattern -> route_pattern.representative_trip_id end)
-      |> Enum.flat_map(fn trip_id -> gtfs_data.stop_times[trip_id] end)
-      |> Enum.map(fn stop_time -> stop_time.timepoint_id end)
-      |> Enum.uniq()
-      |> List.delete("")
+      |> Enum.group_by(fn route_pattern -> route_pattern.direction_id end)
```
paulswartz: might be a personal preference, but this can also be `Enum.group_by(& &1.direction_id)`

lib/gtfs.ex
```diff
@@ -73,16 +73,38 @@ defmodule Gtfs do
...
+
+    timepoints_by_direction =
+      Helpers.map_values(route_patterns_by_direction, fn route_patterns ->
```
paulswartz: similarly, `Helpers.map_values(route_patterns_by_direction, &timepoints_for_route_patterns(&1, gtfs_data))`

lib/gtfs.ex
```diff
@@ -73,16 +73,38 @@ defmodule Gtfs do
...
+  @spec timepoints_for_route_patterns([RoutePattern.t()], t()) :: [Timepoint.id()]
+  defp timepoints_for_route_patterns(route_patterns, gtfs_data) do
+    route_patterns
+    |> Enum.map(fn route_pattern ->
+      stop_times = gtfs_data.stop_times[route_pattern.representative_trip_id]
+
+      stop_times
+      |> Enum.sort_by(fn stop_time -> stop_time.stop_sequence end)
+      |> Enum.map(fn stop_time -> stop_time.timepoint_id end)
+      |> Enum.filter(fn timepoint_id -> timepoint_id != nil end)
```
paulswartz: It might be simpler to filter these out during the parsing stage.

lib/helpers.ex
```diff
@@ -0,0 +1,46 @@
...
+      merge_two_lists(previously_defined_orders, next_order)
+    end)
+  end
+
+  @spec merge_two_lists([any()], [any()]) :: [any()]
+  defp merge_two_lists(list1, list2) do
+    edits = List.myers_difference(list1, list2)
+
+    edits
+    |> Enum.flat_map(fn {_action, elements} -> elements end)
```
paulswartz: This is a built-in: `&elem(&1, 1)`

lib/gtfs.ex
skyqrose: It could. I have a personal preference for named arguments, since they give more of a hint of what's going on. That's not really necessary here, but since I switch between operating on route_patterns, directions, and stop_times in one function, I think it helps a bit.

lib/gtfs.ex
skyqrose: Same as above. I have a slight personal preference for named arguments.

lib/gtfs.ex
skyqrose: I've give that a shot.

lib/helpers.ex
skyqrose: Same preference for named arguments as before, but here, it's important, since it's otherwise unclear what's going on (unless you happen to have read the `myers_difference` docs recently.)

lib/gtfs.ex
skyqrose: I've pushed it here: https://github.com/mbta/skate/commit/7e52f1fc67560ea77e8a0edfe56300de07f71aef

I'm not totally sold on it. It is less code, mostly because we don't need to store stop_times anymore. But it takes this file a step away from the simple model of "gtfs that we make queries against", and if we ever need the rest of the stop_times in the future, this will be net more complicated.

I'm leaning in favor of the preprocessing, but want a second opinion before including it in the PR.

paulswartz: I don't think "gtfs that we make queries against" should be the model. GTFS is the source data, but when it's ingested, we should be processing it into structures that make it easy to query. The internal models should be designed to match the domain, not GTFS: it's why we should be using Timepoint.

An example from a different application is Concentrate, which has several servers each maintaining a view of part of the GTFS: https://github.com/mbta/concentrate/blob/master/lib/concentrate/filter/gtfs/supervisor.ex PickupDropOff is the one which looks at stop times if you want a more direct example.

skyqrose: Okay. I'll merge it in.

lib/helpers.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Helpers do
```
paulswartz: Having the GTFS parsing as a top-level module makes sense, but this seems likely to conflict with another module from somewhere else. As it's currently only used by the GTFS parsing/state, maybe `lib/gtfs/helpers.ex`? You can use `alias GTFS.Helpers` in the relevant modules to avoid any other code changes.

lib/gtfs/helpers.ex
```diff
@@ -1,8 +1,8 @@
-defmodule Helpers do
+defmodule Gtfs.Helpers do
   @doc """
   Apply a function to all of the vaues in a map
 
-  iex> Helpers.map_values(%{a: 1, b: 2}, fn x -> x+1 end)
+  iex> Gtfs.Helpers.map_values(%{a: 1, b: 2}, fn x -> x+1 end)
```
paulswartz: No need to change, but you can do `alias Gtfs.Helpers` at the top of `test/gtfs/helpers_test.exs` and then you wouldn't need to update the documentation here. The doctests are generated as tests in the module and inherit any import or aliases in there.

PR #8, opened 2019-04-05
/_health endpoint

skyqrose:
Asana Task: [/_health endpoint](https://app.asana.com/0/1112935048846093/1117444455227122)

Makes a `/_health` endpoint
Returns 200 if healthy
Returns 503 if the gtfs genserver hasn't loaded yet.

It's a little weird since the genserver will timeout if it's not loaded, rather than returning false, and I don't know the best way to express that.

Another issue: if you catch a genserver timeout, the reply will still happen later, and can clog up the clients mailbox, as mentioned here https://cultivatehq.com/posts/genserver-call-timeouts/ . Unfortunately, I couldn't find much information on this, how to fix it, or whether it's actually a problem, so I haven't tried to do anything about it.

Perhaps a way to avoid the timeout issues would be to have a separate genserver whose only job is to keep track of whether the gtfs genserver is awake, but that seems like a lot of infrastructure for this one simple thing.

I'm not really happy with this code. Is there a different approach I should consider?

### DIFF
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -109,9 +109,11 @@ defmodule Gtfs do
     GenServer.start_link(__MODULE__, {:url, url}, name: __MODULE__)
   end
 
-  @spec start_mocked(mocked_files) :: GenServer.on_start()
+  @spec start_mocked(mocked_files) :: pid()
   def start_mocked(mocked_files) do
-    GenServer.start_link(__MODULE__, {:mocked_files, mocked_files})
+    {:ok, pid} =
+      GenServer.start_link(__MODULE__, {:mocked_files, mocked_files})
+    pid
   end
 
   # Initialization (Server)
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -2,6 +2,7 @@ defmodule Gtfs do
   use GenServer
   require Logger
 
+  alias Gtfs.HealthServer
   alias Gtfs.Helpers
 
   alias Gtfs.Route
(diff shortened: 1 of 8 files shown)


lib/gtfs.ex
```diff
@@ -69,6 +84,18 @@ defmodule Gtfs do
...
   @impl true
+  def handle_call(:ready, _from, state) do
+    case state do
+      {:loaded, _gtfs_data} ->
+        {:reply, true, state}
+
+      # This shouldn't happen,
+      # The handle_continue means the server loads before accepting requests
+      :not_loaded ->
+        {:reply, false, state}
```
skyqrose: Saying "This shouldn't happen" in a comment is a big smell. That could be avoided by this
```elixir
  def handle_call(:ready, _from, {:loaded, _gtfs_data} = state) do
      {:reply, true, state}
  end
```
Not having a function case for this is a more normal way to say that it shouldn't happen, but it would feel weird for this function, whose only job is to tell you if it's loaded, not handle the `:not_loaded` case.

paulswartz:
The way the API handles it is to have a Health server, responsible for knowing whether the system is currently healthy or not. The GTFS server can send Health a message when it's ready.

skyqrose:
Okay, I tried adding a separate `HealthWorker` to keep track of the gtfs server's health. It seems a little verbose, but does get rid of the timeout problems.

There's quite a bit of plumbing to get the pids around so that it can be mocked.

test/skate_web/controllers/health_controller_test.exs
```diff
@@ -0,0 +1,19 @@
+defmodule SkateWeb.HealthControllerTest do
+  use SkateWeb.ConnCase
+
+  describe "index" do
+    test "returns 200 when gtfs loads", %{conn: conn} do
```
paulswartz: This works, but isn't really the test you want. HealthController should return 200 when HeathWorker says the application is healthy, and 503 when it is not healthy. The tests for `HealthWorker.ready?/1` can make assertions about the GTFS.

lib/skate/application.ex
```diff
@@ -16,6 +16,7 @@ defmodule Skate.Application do
       SkateWeb.Endpoint,
       # Starts a worker by calling: Skate.Worker.start_link(arg)
       # {Skate.Worker, arg},
+      worker(HealthWorker, []),
```
paulswartz: Now that HealthWorker and GTFS are linked, you should also change the strategy on line 25 to `:one_for_all`. Otherwise, if the GTFS worker crashes, the HealthWorker won't know about it and will continue to report that the system is healthy.

lib/health_worker.ex
```diff
@@ -0,0 +1,52 @@
+defmodule HealthWorker do
```
paulswartz: probably should be `GTFS.HealthWorker`, given the `@moduledoc`.

PR #12, opened 2019-04-10
Filter data imports to only save bus-related data

arkadyan:
This includes:
• routes
• route patterns
• trips
• timepoints

Time gtfs loading.

Asana ticket: [Filter routes to only include buses](https://app.asana.com/0/1112935048846093/1117662445344605)

### DIFF
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -223,7 +223,7 @@ defmodule Gtfs do
     bus_trip_timepoints =
       files["stop_times.txt"]
       |> Csv.parse(&Timepoint.includes_a_checkpoint_and_in_id_set?(&1, bus_trip_ids))
-      |> trip_timepoints_from_csv()
+      |> Timepoint.trip_timepoints_from_csv()
 
     all_stops = Csv.parse(files["stops.txt"], fn _row -> true end, &Stop.from_csv_row/1)
 
@@ -234,17 +234,5 @@ defmodule Gtfs do
       trip_timepoints: bus_trip_timepoints,
       trips: bus_trips
     }
-  end
-
-  @spec trip_timepoints_from_csv([%{optional(String.t()) => String.t()}]) ::
-          %{optional(Trip.id()) => Timepoint.id()}
-  defp trip_timepoints_from_csv(stop_times_csv) do
-    stop_times_csv
-    |> Enum.group_by(fn stop_time_row -> stop_time_row["trip_id"] end)
-    |> Helpers.map_values(fn stop_times_on_trip ->
-      stop_times_on_trip
-      |> Enum.sort_by(fn stop_time_row -> stop_time_row["stop_sequence"] end)
-      |> Enum.map(fn stop_time_row -> stop_time_row["checkpoint_id"] end)
-    end)
   end
 end
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -2,9 +2,9 @@ defmodule Gtfs do
   use GenServer
   require Logger
 
+  alias Gtfs.Csv
   alias Gtfs.HealthServer
   alias Gtfs.Helpers
-
   alias Gtfs.Route
   alias Gtfs.RoutePattern
   alias Gtfs.Stop
@@ -201,18 +201,18 @@ defmodule Gtfs do
 
   @spec parse_files(files()) :: t()
   defp parse_files(files) do
-    bus_routes = parse_csv(files["routes.txt"], &Route.bus_route?/1, &Route.from_csv_row/1)
+    bus_routes = Csv.parse(files["routes.txt"], &Route.bus_route?/1, &Route.from_csv_row/1)
     bus_route_ids = MapSet.new(bus_routes, & &1.id)
 
     bus_route_patterns =
-      parse_csv(
+      Csv.parse(
         files["route_patterns.txt"],
         &RoutePattern.in_id_set?(&1, bus_route_ids),
         &RoutePattern.from_csv_row/1
       )
 
     bus_trips =
-      parse_csv(
+      Csv.parse(
         files["trips.txt"],
         &Trip.in_id_set?(&1, bus_route_ids),
         &Trip.from_csv_row/1
@@ -222,10 +222,10 @@ defmodule Gtfs do
 
     bus_trip_timepoints =
       files["stop_times.txt"]
-      |> parse_csv(&Timepoint.includes_a_checkpoint_and_in_id_set?(&1, bus_trip_ids))
+      |> Csv.parse(&Timepoint.includes_a_checkpoint_and_in_id_set?(&1, bus_trip_ids))
       |> trip_timepoints_from_csv()
 
-    all_stops = parse_csv(files["stops.txt"], fn _row -> true end, &Stop.from_csv_row/1)
+    all_stops = Csv.parse(files["stops.txt"], fn _row -> true end, &Stop.from_csv_row/1)
 
     %__MODULE__{
       routes: bus_routes,
@@ -234,47 +234,6 @@ defmodule Gtfs do
       trip_timepoints: bus_trip_timepoints,
       trips: bus_trips
     }
-  end
-
-  @doc """
-  Takes binary csv data, a function to filter each row, and a function to parse each row, and returns the list of results.
-  The rows will be passed to the parser as maps with string keys and values.
-  e.g. %{"col1" => "1", "col2" => "x"}
-
-  iex> Gtfs.parse_csv("col1,col2\\n1,x\\n2,y\\n3,z", fn row -> row["col2"] != "y" end, fn row -> String.to_integer(row["col1"]) end)
-  [1, 3]
-
-  exposed for testing
-  """
-  @spec parse_csv(
-          binary() | nil,
-          (%{required(String.t()) => String.t()} -> boolean),
-          (%{required(String.t()) => String.t()} -> row_struct)
-        ) ::
-          [row_struct]
-        when row_struct: var
-  def parse_csv(file_binary, row_filter, row_decoder \\ & &1)
-
-  def parse_csv(nil, _row_filter, _row_decoder) do
-    []
-  end
-
-  def parse_csv(file_binary, row_filter, row_decoder) do
-    {:ok, file_stream} =
-      file_binary
-      |> StringIO.open()
-
-    file_stream
-    |> IO.binstream(:line)
-    |> CSV.decode(headers: true)
-    |> Stream.flat_map(fn {:ok, csv_row} ->
-      if row_filter.(csv_row) do
-        [row_decoder.(csv_row)]
-      else
-        []
-      end
-    end)
-    |> Enum.to_list()
   end
 
   @spec trip_timepoints_from_csv([%{optional(String.t()) => String.t()}]) ::
--- a/lib/gtfs.ex
+++ b/lib/gtfs.ex
@@ -136,15 +136,19 @@ defmodule Gtfs do
     case fetch_files(files_source) do
       {:error, error} ->
         Logger.info(fn ->
-          "Error loading gtfs, time=#{Time.diff(Time.utc_now(), start_time, :millisecond)}"
+          "Error loading gtfs, time_in_ms=#{Time.diff(Time.utc_now(), start_time, :millisecond)}"
         end)
+
         {:stop, error}
 
       files ->
         data = parse_files(files)
         state = {:loaded, data}
+
         Logger.info(fn ->
-          "Successfully loaded gtfs, time=#{Time.diff(Time.utc_now(), start_time, :millisecond)}"
+          "Successfully loaded gtfs, time_in_ms=#{
+            Time.diff(Time.utc_now(), start_time, :millisecond)
+          }"
         end)
 
         if health_server_pid do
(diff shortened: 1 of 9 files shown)


lib/gtfs.ex
```diff
@@ -190,58 +201,86 @@ defmodule Gtfs do
 
   @spec parse_files(files()) :: t()
   defp parse_files(files) do
+    bus_routes = parse_csv(files["routes.txt"], &Route.bus_route?/1, &Route.from_csv_row/1)
+    bus_route_ids = bus_routes |> Enum.map(& &1.id) |> MapSet.new()
```
paulswartz: `bus_route_ids = MapSet.new(bus_routes, & &1.id)`

lib/gtfs/route_pattern.ex
```diff
@@ -37,4 +37,7 @@ defmodule Gtfs.RoutePattern do
       representative_trip_id: row["representative_trip_id"]
     }
   end
+
+  @spec member_of_routes?(MapSet.t(id()), %{required(String.t()) => String.t()}) :: boolean
+  def member_of_routes?(route_ids, row), do: Enum.member?(route_ids, row["route_id"])
```
paulswartz: - Since this is a function on the route_pattern, the route_pattern (here, `row`) should be the first argument: maybe naming it something like `in_id_set?(row, set)` makes that clearer?
- You can ensure the argument is a MapSet (per the type) with `(row, %MapSet{} = route_ids)` and then simplify the body to `MapSet.member?(route_ids, row["route_id"])`

lib/gtfs/trip.ex
```diff
@@ -27,4 +27,7 @@ defmodule Gtfs.Trip do
       route_id: row["route_id"]
     }
   end
+
+  @spec member_of_routes?(MapSet.t(id()), %{required(String.t()) => String.t()}) :: boolean
+  def member_of_routes?(route_ids, row), do: Enum.member?(route_ids, row["route_id"])
```
paulswartz: similiar comments here as to `member_of_routes?` above

lib/gtfs.ex
```diff
@@ -190,58 +201,86 @@ defmodule Gtfs do
...
+      )
+
+    bus_trips =
+      parse_csv(
+        files["trips.txt"],
+        &Trip.member_of_routes?(bus_route_ids, &1),
+        &Trip.from_csv_row/1
+      )
+
+    bus_trip_ids = bus_trips |> Enum.map(& &1.id) |> MapSet.new()
```
paulswartz: `bus_trip_ids = MapSet.new(bus_trips, & &1.id)`

test/gtfs_test.exs
```diff
@@ -68,11 +128,20 @@ defmodule GtfsTest do
     test "merges stops from both directions, flipping direction 0" do
       pid =
         Gtfs.start_mocked(%{
+          "routes.txt" => [
+            "route_id,route_type",
+            "route,3"
```
skyqrose: Potential trip hazard: If you forget to specify the `route_type` column in a test, (reasonably likely, since we don't include it in the final `Route.t()` struct), the route will be silently dropped. I can imagine losing a couple hours debugging something like that.

Should we make `bus_route?` crash if it's missing this column to make the problem easier to spot, or accept all routes as bus routes if the column, or something like that?

lib/gtfs.ex
```diff
@@ -99,7 +99,7 @@ defmodule Gtfs do
   defp timepoints_for_route_patterns(route_patterns, gtfs_data) do
     route_patterns
     |> Enum.map(fn route_pattern -> route_pattern.representative_trip_id end)
-    |> Enum.map(fn trip_id -> gtfs_data.trip_timepoints[trip_id] end)
+    |> Enum.map(fn trip_id -> Map.get(gtfs_data.trip_timepoints, trip_id, []) end)
```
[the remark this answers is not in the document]
arkadyan: Yes, needs a rebase.

arkadyan:
@skyqrose At first it slowed things down drastically due to all the `member?` checks, but then when I changed `bus_route_ids` and `bus_trip_ids` from Lists to MapSets it resulted in a slight overall speed up: very roughly from the order of 120k ms to 85k-90k ms.

lib/gtfs.ex
```diff
@@ -190,58 +201,86 @@ defmodule Gtfs do
...
+        files["trips.txt"],
+        &Trip.member_of_routes?(bus_route_ids, &1),
+        &Trip.from_csv_row/1
+      )
+
+    bus_trip_ids = bus_trips |> Enum.map(& &1.id) |> MapSet.new()
+
+    bus_trip_timepoints =
+      files["stop_times.txt"]
+      |> parse_csv(&Timepoint.includes_a_checkpoint_and_member_of_trips?(bus_trip_ids, &1))
```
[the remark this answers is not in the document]
arkadyan: I like that solution, so the parse function would automatc combine the filters with an `&&`. That does seem cleaner and more readable. Let me try it out.

lib/gtfs/timepoint.ex
```diff
@@ -4,4 +4,16 @@ defmodule Gtfs.Timepoint do
   """
 
   @type id :: String.t()
+
+  @type csv_row :: %{required(String.t()) => String.t()}
```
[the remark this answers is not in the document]
arkadyan: 👍 Refactored both this and the parse function into a new `Gtfs.Csv` module.

lib/gtfs/csv.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @doc """
+  Takes binary csv data, a function to filter each row, and a function to parse each row, and returns the list of results.
+  The rows will be passed to the parser as maps with string keys and values.
+  e.g. %{"col1" => "1", "col2" => "x"}
+
+  iex> Gtfs.Csv.parse("col1,col2\\n1,x\\n2,y\\n3,z", fn row -> row["col2"] != "y" end, fn row -> String.to_integer(row["col1"]) end)
+  [1, 3]
+
+  exposed for testing
```
skyqrose: This comment is now out of date.

lib/gtfs/route.ex
```diff
@@ -15,10 +17,13 @@ defmodule Gtfs.Route do
...
+  @spec from_csv_row(Csv.row()) :: t()
+  def from_csv_row(route_row) do
     %__MODULE__{
-      id: row["route_id"]
+      id: route_row["route_id"]
     }
   end
+
+  @spec bus_route?(Csv.row()) :: boolean
+  def bus_route?(route_row), do: route_row["route_type"] == "3"
```
skyqrose: Sorry, I meant the function names could get a `_row` suffix, so `bus_route_row?`. This isn't a strong opinion. I'm just bringing it up again cuz the old comment got lost in the new commits.

lib/gtfs/csv.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Gtfs.Csv do
+  @moduledoc """
+  Manage parsing data from GTFS CSV files.
+  """
+
+  @type row :: %{required(String.t()) => String.t()}
```
skyqrose: I like this, and I like pulling this into another file.

lib/gtfs/timepoint.ex
```diff
@@ -3,5 +3,32 @@ defmodule Gtfs.Timepoint do
...
+
   @type id :: String.t()
+
+  @type id_set :: MapSet.t(id())
+
+  @spec includes_a_checkpoint_and_in_id_set?(Csv.row(), id_set) :: boolean
+  def includes_a_checkpoint_and_in_id_set?(timepoint_row, id_set),
+    do: includes_a_checkpoint?(timepoint_row) && in_id_set?(timepoint_row, id_set)
+
+  @spec trip_timepoints_from_csv([%{optional(String.t()) => String.t()}]) ::
```
skyqrose: This can use the new `Csv.row()`

lib/gtfs/trip.ex
```diff
@@ -20,11 +21,14 @@ defmodule Gtfs.Trip do
...
     %__MODULE__{
-      id: row["trip_id"],
-      route_id: row["route_id"]
+      id: trip_row["trip_id"],
+      route_id: trip_row["route_id"]
     }
   end
+
+  @spec in_id_set?(Csv.row(), MapSet.t(id())) :: boolean
+  def in_id_set?(trip_row, id_set), do: MapSet.member?(id_set, trip_row["route_id"])
```
skyqrose: This is checking whether the route is in a set of routes, but the type and names say it's for a set of trips.

I know Paul just suggested these names, but I think the `member_of_routes?(row, route_ids)` name was clearer. (This argument order is better, though.)

(Ditto for `route_pattern.ex`)

paulswartz: It's not a MapSet of Routes, but of route IDs. `in_route_id_set?` maybe?

lib/gtfs/route.ex
arkadyan: Oh, I misunderstood. Yeah, that makes sense.

lib/gtfs/timepoint.ex
arkadyan: The only difference here is `optional` vs `required`. Is this ever really `optional`? Just want to double check.

skyqrose: I guess the difference is whether a csv with no columns is a csv. I don't think it matters.

lib/gtfs/csv.ex
skyqrose: I just mean the `exposed for testing` line. It's now exposed for regular use in `gtfs.ex`.

arkadyan: Oh right, very true!

lib/gtfs/trip.ex
arkadyan: I like that idea.

test/gtfs_test.exs
arkadyan: Added a check.

lib/gtfs/route_pattern.ex
```diff
@@ -37,4 +38,8 @@ defmodule Gtfs.RoutePattern do
       representative_trip_id: row["representative_trip_id"]
     }
   end
+
+  @spec in_route_id_set?(Csv.row(), MapSet.t(id())) :: boolean
```
skyqrose: Should have type `MapSet.t(Route.Id())`
```suggestion
  @spec in_route_id_set?(Csv.row(), MapSet.t(Route.id())) :: boolean
```

lib/gtfs/trip.ex
```diff
@@ -20,11 +21,14 @@ defmodule Gtfs.Trip do
...
-  @spec from_csv_row(%{required(String.t()) => String.t()}) :: t()
+  @spec from_csv_row(Csv.row()) :: t()
   def from_csv_row(row) do
     %__MODULE__{
       id: row["trip_id"],
       route_id: row["route_id"]
     }
   end
+
+  @spec in_route_id_set?(Csv.row(), MapSet.t(id())) :: boolean
```
skyqrose: Should have type `MapSet.t(Route.Id())`
```suggestion
  @spec in_route_id_set?(Csv.row(), MapSet.t(Route.id())) :: boolean
```

lib/gtfs/timepoint.ex
```diff
@@ -3,5 +3,32 @@ defmodule Gtfs.Timepoint do
...
+      stop_times_on_trip
+      |> Enum.sort_by(fn stop_time_row -> stop_time_row["stop_sequence"] end)
+      |> Enum.map(fn stop_time_row -> stop_time_row["checkpoint_id"] end)
+    end)
+  end
+
+  @spec includes_a_checkpoint?(Csv.row()) :: boolean
+  defp includes_a_checkpoint?(row), do: row["checkpoint_id"] != ""
+
+  @spec in_trip_id_set?(Csv.row(), id_set) :: boolean
```
skyqrose: Should have type `MapSet.t(Trip.Id())`
```suggestion
  @spec in_trip_id_set?(Csv.row(), MapSet.t(Trip.id())) :: boolean
```

Same for `includes_a_checkpoint_and_in_trip_id_set`

lib/gtfs/route_pattern.ex
arkadyan: Good catch

arkadyan:
@skyqrose and @paulswartz thanks for all the great feedback. I think I've addressed everything.

lib/gtfs/route_pattern.ex
```diff
@@ -37,4 +38,8 @@ defmodule Gtfs.RoutePattern do
       representative_trip_id: row["representative_trip_id"]
     }
   end
+
+  @spec in_route_id_set?(Csv.row(), MapSet.t(Route.id())) :: boolean
+  def in_route_id_set?(row, route_id_set),
```
skyqrose: Do you think having `row` or `csv` somewhere in this function name would reduce risk of misusing it?
(Same for `Trip.in_route_id_set?`, `Timepoint.in_trip_id_set?`, and `Timepoint.includes_a_checkpoint?`)

arkadyan: As in something like, `csv_row_in_route_id_set?/2`?

arkadyan: Updated

skyqrose: I'd lean toward the shorter `row_in_route_id_set?` or `row_has_timepoint?`, but yeah.

PR #13, opened 2019-04-10
select and deselect routes. move state and reducer to its own file

skyqrose:
Asana Task: [build UI for selecting routes (wireframe-level version)](https://app.asana.com/0/1112935048846093/1114820119722041)

Kapture 2019-04-10 at 13 36 51


assets/src/state.ts
```diff
@@ -0,0 +1,75 @@
+import { Dispatch as ReactDispatch } from "react"
+import { Route, RouteId } from "./skate.d"
+
+export interface State {
+  routes: null | Route[]
```
arkadyan: This `null | Route[]` is repeated in routePicker - maybe worth making a named type for it?

assets/src/state.ts
```diff
@@ -0,0 +1,75 @@
...
+  switch (action.type) {
+    case "SET_ROUTES":
+      return {
+        ...state,
+        routes: action.payload.routes,
+      }
+    case "SELECT_ROUTE":
+      return {
+        ...state,
+        selectedRouteIds: [...state.selectedRouteIds, action.payload.routeId],
```
arkadyan: Is it worth it to ensure that duplicate values aren't entered here (or are "unique"-ified away)? Or is that overkill since our interface effectively prevents it?

assets/tests/state.test.ts
```diff
@@ -0,0 +1,47 @@
+import { Route } from "../src/skate.d"
+import * as State from "../src/state"
```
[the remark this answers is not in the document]
skyqrose: As we get more and more actions, that import list would get pretty long. I did it this way so that the common variables (`initialState` and `reducer`) would be available without needing to update this line every time we add a new action.

Thoughts on the tradeoffs? Is it worth it?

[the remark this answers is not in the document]
arkadyan: Yeah, it looks a little weird to me (like it's a static function call), so I might not use this style but I also don't really care at all.

assets/src/state.ts
skyqrose: I considered that. The failure mode is that you'd see multiple ladders for the same route, which I think is fine. And if there were a way in the interface to re-add a route, that'd be a bigger problem, and I'd rather not try to quietly sweep it under the rug here.

arkadyan: Fair enough

assets/src/state.ts
skyqrose: I tried out defining `export type LoadingRoutes = null | Route[]` in `state.ts`. The indirection and verbosity ended up being worse than the duplication.

PR #14, opened 2019-04-10
Fix prettier trailing comma configuration

arkadyan:
Prettier was occasionally inserting commas that tslint said shouldn't be there.

### DIFF
--- a/assets/.prettierrc
+++ b/assets/.prettierrc
@@ -1,5 +1,4 @@
 {
   "semi": false,
-  "singleQuote": false,
-  "trailingComma": "es5"
+  "singleQuote": false
 }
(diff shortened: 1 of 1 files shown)


skyqrose:
VSCode has a setting `prettier.tslintIntegration` that says it will read from `tslint.json` config when enabled. Can you try enabling that and removing the line from `.prettierrc`, and see if that also fixes the problem? If it does, that'd be better than trying to keep `tslint.config` and `.prettierrc` in sync.

arkadyan:
@skyqrose Oh great. I honestly don't remember where I was seeing the issue in the first place, but removing this setting seems like a good thing to do anyway, so I just pulled the line.

PR #15, opened 2019-04-11
Update js-yaml to resolve a DoS vulnerability

arkadyan:


arkadyan:
@skyqrose Yes, I ran `npm audit fix` to fix this case. Looks like `tar` will need a manual intervention.

PR #17, opened 2019-04-11
Display a route ladder for each selected route

arkadyan:
Asana ticket: [build UI for route ladder - name in center (wireframe-level version)](https://app.asana.com/0/1112935048846093/1114820119722043)

Screen Shot 2019-04-11 at 13 25 19


assets/src/skate.d.ts
```diff
@@ -3,3 +3,13 @@ export interface Route {
...
 export type RouteId = string
+
+export interface Timepoint {
+  id: TimepointId
+}
+
+export type TimepointId = string
+
+export interface TimepointsByRoute {
+  [routeId: string]: Timepoint[]
```
skyqrose: Can be the `RouteId` type instead of `string`.

assets/css/app.scss
```diff
@@ -1 +1,50 @@
-/* This file is for your main application css. */
+.m-app {
+  display: flex;
+}
+
+.m-route-ladders {
```
skyqrose: Let's put this in a `route_ladders.scss` file

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useEffect } from "react"
+import { fetchTimepointsForRoute } from "../api"
+import { Route, Timepoint, TimepointsByRoute } from "../skate"
+import { Dispatch, setTimepointsForRoute } from "../state"
+import Loading from "./loading"
+
+interface Props {
+  route: Route
+  dispatch: Dispatch
+  timepointsByRoute: TimepointsByRoute
```
skyqrose:
```suggestion
  timepoints: Timepoint[] | null
```
If we do the lookup before we call this component, then it doesn't need to know anything about other ladders.

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,59 @@
...
+const RouteLadder = ({ route, dispatch, timepointsByRoute }: Props) => {
+  const timepoints = timepointsByRoute[route.id]
+
+  useEffect(() => {
+    if (!timepointsByRoute[route.id]) {
+      fetchTimepointsForRoute(route.id).then((newTimepoints: Timepoint[]) =>
+        dispatch(setTimepointsForRoute(route.id, newTimepoints))
+      )
+    }
+  })
```
skyqrose: This effect needs something to make sure it's not called multiple times, since it could be rendered multiple times before we get the response.

Alternatively, maybe this should live in the reducer, in the case for `selectRoute`. That's how Elm does it: the reducer both updates the state and initiates any new side effects. I don't know if the React convention is to put side effects in the components or the reducer.

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,59 @@
...
+import Loading from "./loading"
+
+interface Props {
+  route: Route
+  dispatch: Dispatch
+  timepointsByRoute: TimepointsByRoute
+}
+
+const TimepointStop = () => (
+  <div className="m-route-ladder__timepoint-stop">
```
skyqrose:
```suggestion
  <div className="m-route-ladder__stop">
```
BEM doesn't need the whole hierarchy. I don't think having `timepoint-` adds any understanding, so I'd lean slightly against including it.

(`__timepoint-name` does help disambiguate it from `__route-name`, so this comment wouldn't apply there.)

assets/tests/components/routeLadders.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      { id: "BLTAL" },
+      { id: "ROXBS" },
+      { id: "FRNPK" },
+      { id: "GHALL" },
+      { id: "LATAC" },
+      { id: "WARWL" },
+      { id: "DUDLY" },
+      { id: "LOUIS" },
+      { id: "MALCX" },
+      { id: "RUGG" },
```
skyqrose: We don't need this much sample data. 2 or 3 for each route would be fine. Using less data would make it easier to add fields to the timepoint type, and easier to scroll through this file once it has lots of tests.

(Same for `routeLadder.test.tsx`)

assets/src/skate.d.ts
arkadyan: Typescript doesn't allow index signature parameter types to be type aliases.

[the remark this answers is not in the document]
arkadyan: Re `routeId`: typescript doesn't allow index signature parameter types to be type aliases.

Re `TimepointsByRouteId`: 👍

assets/src/components/routeLadder.tsx
arkadyan: 👍 I actually started to do this at one point but it got lost in the shuffle. Added a function.

assets/css/app.scss
```diff
@@ -1 +1,50 @@
...
+
+.m-route-ladder__timepoint {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+}
+
+.m-route-ladder__timepoint-stop {
+  height: 60px;
+  width: 20px;
```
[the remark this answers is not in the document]
arkadyan: I was trying to match the svg attributes which are in `px`, but I discovered that if I set the `line-height` instead then I don't need to set `height` and `width`.

assets/src/components/routeLadder.tsx
arkadyan: Good call. I added a null state for loading and updated the check to only care about `undefined`.

Hooks need to be in React components so we can't move it to the reducer, but we should keep in mind the ability to create custom hooks if we need to share or abstract them in the future. (I don't think there's value in doing that here at this point.)

arkadyan:
@skyqrose I added a test to verify that we fetch timepoint data when we render a route ladder for which we don't already have data. I think I've addressed the rest of the feedback as well.

assets/tests/components/routeLadder.test.tsx
```diff
@@ -0,0 +1,114 @@
+import React from "react"
+import renderer, { act } from "react-test-renderer"
+import RouteLadder from "../../src/components/routeLadder"
+import { Route } from "../../src/skate"
+import { TimepointsForRouteId } from "../../src/state"
+
+declare global {
+  interface Window {
+    /* eslint-disable typescript/no-explicit-any */
+    fetch: (uri: string) => Promise<any>
```
skyqrose: Could this be `Promise<Response>` in order to avoid the `eslint-disable`? Or would that require specifying fields in the mock that we don't care about?

assets/src/state.ts
```diff
@@ -66,10 +109,35 @@ export const reducer = (state: State, action: Action): State => {
...
   }
 }
+
+export type TimepointsForRouteId = (
+  routeId: RouteId
+) => LoadableTimepoints | undefined
+
+export const timepointsForRouteId = (
+  state: State
+): TimepointsForRouteId => routeId => state.timepointsByRoute[routeId]
```
skyqrose: What's the difference between the names `timepointsByRoute`, `TimepointsByRouteId` and `[tT]imepointsForRouteId`? Could they all have the same name?

Is using a function to wrap the object so that we can get a real type on the function because of the weird typing restriction on objects?

assets/tests/components/routeLadder.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+test("fetches timepoints for this route if we don't yet have them", () => {
+  const mockDispatch = jest.fn()
+  const route: Route = { id: "28" }
+  const timepointsForRouteId: TimepointsForRouteId = () => undefined
+
+  window.fetch = () =>
+    Promise.resolve({
+      json: () => ({
+        data: [{ id: "28" }, { id: "39" }, { id: "71" }],
```
skyqrose: This looks like `Route[]`, not `Timepoint[]`. The endpoint we're calling should be a list of timepoints.
(Same for the other test.)

assets/tests/components/routeLadder.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <RouteLadder
+          route={route}
+          dispatch={mockDispatch}
+          timepointsForRouteId={timepointsForRouteId}
+        />
+      )
+      .toJSON()
+  })
+
+  expect(mockDispatch.mock.calls.length).toBe(1)
```
skyqrose: Shouldn't we have two calls? One to say we're loading, and a second one once the api call returns?

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,66 @@
...
+  setLoadingTimepointsForRoute,
+  setTimepointsForRoute,
+  TimepointsForRouteId,
+} from "../state"
+import Loading from "./loading"
+
+interface Props {
+  route: Route
+  dispatch: Dispatch
+  timepointsForRouteId: TimepointsForRouteId
```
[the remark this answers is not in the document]
arkadyan: No this is a function to retrieve the timepoints given a route id, not the full object.

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  useEffect(() => {
+    if (timepoints === undefined) {
+      dispatch(setLoadingTimepointsForRoute(route.id))
+
+      fetchTimepointsForRoute(route.id).then((newTimepoints: Timepoint[]) =>
+        dispatch(setTimepointsForRoute(route.id, newTimepoints))
+      )
+    }
+  }, [])
```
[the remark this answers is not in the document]
arkadyan: This doesn't prevent duplicate calls, the `timepoints === undefined` check above does.

assets/src/components/routeLadder.tsx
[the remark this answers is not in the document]
arkadyan: Replaced this as we discussed.

assets/src/state.ts
arkadyan: Fixed all this as discussed.

assets/tests/components/routeLadder.test.tsx
arkadyan: Oops, yeah…

assets/tests/components/routeLadder.test.tsx
arkadyan: The second one happens asynchronously so the test finishes before that happens. But I tested the reducer to test handling that data.

assets/tests/components/routeLadder.test.tsx
arkadyan: It appears that you have to add a bunch of extra functions we don't care about to support that, so I think it's not worth it…

assets/src/state.ts
skyqrose: Just so it's written down:
"timepoints for route" is for functions / actions etc that work on a single route.
"timepoints by route id" is the map of many route ids to their timepoints.
"timepoints for route id" no longer exists.
"timepoints by route" no longer exists.

assets/tests/components/routeLadders.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Route, TimepointsByRouteId } from "../../src/skate"
+
+const dispatch = () => undefined
+
+test("renders a route ladder", () => {
+  const routes: Route[] = [{ id: "1" }, { id: "28" }]
+  const timepointsByRouteId: TimepointsByRouteId = {
+    "1": [{ id: "WASMA" }, { id: "MELWA" }, { id: "HHGAT" }],
+    "28": [{ id: "MATPN" }, { id: "WELLH" }, { id: "MORTN" }],
+  }
```
skyqrose: Is it worth testing a `null` case and an `undefined` case here?

assets/tests/components/routeLadder.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const timepoints = null
+
+  window.fetch = () =>
+    Promise.resolve({
+      json: () => ({
+        data: ["MATPN", "WELLH", "MORTN"],
+      }),
+      ok: true,
+      status: 200,
+    })
```
skyqrose: This test shouldn't call fetch, so this isn't needed.

Alternatively, for this test and the previous, could we make `window.fetch` a `jest.fn` so we can check that it was called the right number of times?

assets/tests/components/routeLadder.test.tsx
skyqrose: If the promise doesn't resolve, before the test finishes, does its return value matter?

Is there a way to make the test wait for the promise? This page has some options: https://jestjs.io/docs/en/asynchronous

assets/src/components/routeLadder.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {
+  Dispatch,
+  setLoadingTimepointsForRoute,
+  setTimepointsForRoute,
+} from "../state"
+import Loading from "./loading"
+
+interface Props {
+  route: Route
+  timepoints: LoadableTimepoints | undefined
```
skyqrose: Thoughts on moving this `|` to `skate.d` so that the `undefined`=not asked and `null`=loading concepts are documented in one place?

One way to do that might be to change `LoadableTimepoints` to `Timepoint[] | null | undefined`, and then `TimepointsByRouteId` use `Timepoint[] | null` instead of `LoadableTimepoints`

assets/tests/components/routeLadder.test.tsx
arkadyan: Trouble is we aren't calling the promise, we are triggering it by interacting with the DOM. But I think demostrating that the fetch process starts seems like a sufficient test here to me.

assets/src/skate.d.ts
```diff
@@ -8,9 +8,9 @@ export interface Timepoint {
   id: TimepointId
 }
 
-// A null value (as opposed to undefined) indicates that we are currently
-// loading the timepoints
-export type LoadableTimepoints = Timepoint[] | null
+// An undefined value indicates that the timepoints need to be loaded
+// A null value indicates that we are currently loading the timepoints
+export type LoadableTimepoints = Timepoint[] | null | undefined
```
paulswartz: Future change: have an actual type for these cases rather than overriding null and undefined.

arkadyan: Yeah, I think that could be worthwhile.

PR #19, opened 2019-04-16
Cache GTFS data in dev and test for faster loading

arkadyan:
Asana ticket: [Speed up GTFS loading](https://app.asana.com/0/1112935048846093/1117099115669805)

Mostly stolen from Alerts.

A cold load takes ~90s. With a cached file that drops to ~2s.


lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,127 @@
...
+  if the the envs are :dev or :test.
+
+    iex> Mix.env
+    :test
+    iex> CacheFile.should_use_file?
+    true
+  """
+  @spec should_use_file?() :: boolean
+  def should_use_file? do
+    EnvHelper.is_env?(:dev) || EnvHelper.is_env?(:test)
```
paulswartz: Rather than checking this, it should be configured in `config/*.exs`. Something like:
```elixir
# config/config.exs
config :skate, Gtfs.CacheFile, use_cache?: false

# config/dev.ex
config :skate, Gtfs.CacheFile, use_cache?: true, cache_filename: "dev_cache.terms"

# config/test.ex
config :skate, Gtfs.CacheFile, use_cache?: true, cache_filename: "test_cache.terms"

def should_use? do
  Application.get_env(:skate, __MODULE__)[:use_cache?]
end

def cache_filename do
  Application.get_env(:skate, __MODULE__)[:cache_filename]
end
```

lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,127 @@
...
+  to allowing caching of the changes.
+  """
+  require Logger
+
+  alias Gtfs.Helpers.EnvHelper
+
+  @directory Path.join([File.cwd!(), "priv/gtfs_cache"])
+
+  @dev_filename "dev_cache.terms"
+  @test_filename "test_cache.terms"
```
paulswartz: Should also be configured in `config/*.exs`

lib/gtfs.ex
```diff
@@ -185,6 +190,31 @@ defmodule Gtfs do
...
+          Logger.info("Loaded gtfs data remote files")
+          CacheFile.save_gtfs(state)
+          state
+      end
+    else
+      Logger.info("Loading gtfs data remote files")
+      state = parse_files(files)
+      Logger.info("Loaded gtfs data remote files")
+      state
+    end
```
paulswartz: - There's some duplicate code in these cases.
- What happens if the shape of the state has changed since the cache was saved?

arkadyan: Re: changes to gtfs data, I was only thinking of this as a speed-up measure in development, so personally I'm happy just saying, "blow away your cache file anytime anything changes". I can write a custom mix task to actually do that, I had meant to do that and forgot.

lib/gtfs.ex
```diff
@@ -185,6 +190,31 @@ defmodule Gtfs do
...
+  end
```
skyqrose: If we use the cache, we end up downloading the files, and then discarding them. What do you think about calling the cache in the fetch step instead of the parse step??

It'd require moving the call to `parse_files` into that function, but I think that's fine. Something like (very roughly)
```elixir
@spec fetch_gtfs(files_source()) :: t() | {:error, any()}
defp fetch_gtfs({:url, url}) do
  if should_cache and cache_exists do
    Cache.load()
  else
    zip_binary = HTTPoison.get()
    zip_binary |> unzip() |> parse()

defp fetch_gtfs({:mocked_files}, mocked_files) do
  mocked_files |> parse()
```

Small benefit: it's easier to not cache the mocked_files, since the cache would only be looked at in the url case.

lib/gtfs/cache_file.ex
skyqrose: This might not work, and it might not be a good idea, but what if we could decide whether to use the cache from config instead of by checking the environment, by having a 3rd file source `{:cached_url, url}`?

Cons:
* Depends on my other comment about moving the cache check into the fetch step.
* Requires specifying a `files_source()` in the config instead of just a `String.t()` url. (This is even harder for specifying the source be environment variable, but perhaps we could just always use the uncached `files_source` in that case.)
* Makes dev+prod more different.
* Makes bad states like trying to use cached data on prod more possible by uncoupling the decision of whether to use the cache from the environment.

Pros:
* Doesn't need the `EnvHelper` and other infrastructure.
* The decision about whether to use the cache or not becomes simpler.
* More testable because it make decisions based on the argument instead of global state.

skyqrose: Could `cache_filename` control whether to use the cache by whether or not it exists, instead of having a separate `use_cache?`?

lib/gtfs.ex
skyqrose: If there's been a change, can we reasonably detect it? Will loading the data into structs fail?

If so, perhaps we could handle a failed load by automatically removing the cache, and downloading and storing new data.

lib/gtfs/cache_file.ex
skyqrose: Ah, hadn't seen Paul's comments.

If the config has `cache_filename` or `use_cache?`, it'd be much easier to create a `files_source()` based on that.

lib/gtfs.ex
arkadyan: I don't know, I know this area less well, but in my experience with Alerts I can say that small changes to this sort of data (e.g. a stop name changing) happen occasionally but would likely not be easily detected (would not fail to load into structs). I never experienced larger failures. Also, knowing that there was a cache that sped things up but that I should just delete if I even thought something might have changed did not feel particularly problematic. Worst case is the app takes longer to load once afterwards.

arkadyan: Refatored out duplicate code.

lib/gtfs/cache_file.ex
arkadyan: Combined both suggestions.

lib/helpers/env_helper.ex
```diff
@@ -0,0 +1,43 @@
...
+    iex> EnvHelper.is_env?(:other)
+    false
+
+  """
+  def is_env?(env_name) do
+    mix_is_loaded?() && Mix.env() == env_name
+  end
+
+  @doc """
+  Returns the current Mix.env or nil.
```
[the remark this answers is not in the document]
arkadyan: I was able to remove this file completely.

lib/helpers/env_helper.ex
```diff
@@ -0,0 +1,43 @@
...
+
+    iex> EnvHelper.env
+    :test
+
+  """
+  def env do
```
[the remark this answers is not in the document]
arkadyan: I was able to remove this file completely.

lib/gtfs.ex
arkadyan: This was a good catch. The latest commit does this - it's a bit hairy to read through, but it seems to work.

Plus the time to load from a cache drops even further to ~80ms!

arkadyan:
Added a mix task to delete the cache file.

lib/gtfs.ex
```diff
@@ -133,33 +134,75 @@ defmodule Gtfs do
...
+          with {:ok, state} <- fetch_and_parse_data(files_source) do
+            CacheFile.save_gtfs(state)
+            {:ok, state}
+          else
+            {:error, error} ->
+              {:error, error}
+          end
+      end
+    else
+      fetch_and_parse_data(files_source)
```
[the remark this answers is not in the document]
arkadyan: Yeah, personally I think that this might make things less clear…

lib/gtfs.ex
```diff
@@ -133,33 +134,75 @@ defmodule Gtfs do
...
+  end
+
+  defp fetch_gtfs(files_source) do
+    if CacheFile.should_use_file?() do
+      Logger.info("Loading gfts data from cached file")
+
+      with {:ok, state} <- CacheFile.load_gtfs() do
+        {:ok, state}
+      else
+        _ ->
```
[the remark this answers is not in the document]
skyqrose: Discussed in person: The answer is not really, we don't care.

lib/gtfs.ex
```diff
@@ -133,34 +134,76 @@ defmodule Gtfs do
...
+      end
+    else
+      fetch_and_parse_data(url)
+    end
+  end
 
-        if health_server_pid do
-          HealthServer.loaded(health_server_pid)
-        end
+  @spec fetch_and_parse_data(String.t()) :: {:ok, t()} | {:error, any()}
```
skyqrose: There are a bunch of functions named `fetch_` here:
`files_from_mocked`, `fetch_files`, `fetch_and_parse_data`, `fetch_gtfs`.
The names don't make it clear which ones might hit the filesystem and which ones might hit the internet.
What do you think about naming them:
`files_from_mocked`, `files_from_url`, `gtfs_from_url`, and `fetch_gtfs`?

lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,118 @@
...
+  def load_gtfs(filepath) when is_binary(filepath) do
+    Logger.info(fn -> "Loading gtfs cache from file #{filepath}" end)
+
+    with {:ok, binary_cache} <- File.read(filepath),
+         {:ok, state} when is_map(state) <- binary_to_term(binary_cache) do
+      Logger.info(fn -> "Loaded gtfs cache from file #{filepath}" end)
+      {:ok, state}
+    else
+      _ ->
+        Logger.info(fn -> "Failed to load gtfs cache from file #{filepath}" end)
```
skyqrose: There's a lot of logging here. It's all really useful for testing the caching system, but long term we probably won't need quite this much. We should think about what log events should stick around.

It's not a huge deal since it's all happens only on startup.

I think there's also room to improve the messages. For example, this could be "Cache file doesn't exist. Downloading gtfs from url".

lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,118 @@
...
+        {:error, :cache_not_loaded}
+    end
+  end
+
+  def load_gtfs(_) do
```
skyqrose: I'm having a hard time following where this case comes from. Is it that if we're on prod, the `cache_filename()` is `nil`? But `generate_filepath()` doesn't have a case for that.

Instead of passing nils around, I'd prefer to notice the `nil` at a higher level. Something like
```elixir
def load_gtfs() do
  case cache_filename() do
    nil -> {:error, "no cache file configured"}
    filename ->
      load_gtfs(generate_filepath(filename)) # we know the path exists now, so load_gtfs/1 doesn't need a case for nil.
```

lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,118 @@
...
+  if the the envs are :dev or :test.
+
+    iex> Mix.env
+    :test
+    iex> CacheFile.should_use_file?
+    true
+  """
+  @spec should_use_file?() :: boolean
+  def should_use_file? do
+    Application.get_env(:skate, __MODULE__)[:cache_filename] != nil
```
skyqrose: We're doing two checks for `nil`. One here to decide whether to attempt the cache, and one when we try to get the filename. There could be bugs if the two checks become slightly different. Is there a way to just have a single `get_filename_if_we_should_cache() :: String.t | nil` for both purposes, so we only need to make one decision?

It's a small thing and I think it'd require a little bit more of a refactor, so this probably isn't worth addressing, but if you see an easy way to do it it might help.

lib/gtfs/cache_file.ex
arkadyan: In that case we don't actually know that the file doesn't exist. That's certainly the most common case, but if anything goes wrong reading the file you'll end up here. But yeah, we can remove logger statements if we find them to be annoying.

lib/gtfs/cache_file.ex
arkadyan: Removed this case, there wasn't a specific intention behind it.

lib/gtfs/cache_file.ex
arkadyan: Removed the other case.

arkadyan:
@skyqrose I think I've addressed your comments if you want to take another look.

test/mix/tasks/cache.clean_test.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Mix.Tasks.Cache.CleanTest do
```
paulswartz: This does not appear to be testing anything: https://codecov.io/gh/mbta/skate/compare/8c56a4beab5aeac5fb15c6fa060c057137e12124...7bd5f55fe020d2348fc4edc0c767f60c21ca0252/diff#D11-10

lib/gtfs/cache_file.ex
```diff
@@ -0,0 +1,113 @@
...
+      Logger.info(fn -> "Loaded gtfs cache from file #{filepath}" end)
+      {:ok, data}
+    else
+      _ ->
+        Logger.info(fn -> "Failed to load gtfs cache from file #{filepath}" end)
+        {:error, :cache_not_loaded}
+    end
+  end
+
+  @doc """
```
paulswartz: This isn't tested: https://codecov.io/gh/mbta/skate/compare/8c56a4beab5aeac5fb15c6fa060c057137e12124...7bd5f55fe020d2348fc4edc0c767f60c21ca0252/diff#D9-62

test/mix/tasks/cache.clean_test.ex
arkadyan: The filename had the wrong extension so it wasn't being included in the test suite.

arkadyan:
Thanks again for your help with the test coverage, @paulswartz!

PR #20, opened 2019-04-16
Add close button to route ladder header

arkadyan:
Deselects the route.

Asana ticket: [Add X button to route ladder header](https://app.asana.com/0/1112935048846093/1118437179466238)

### DIFF
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -2,6 +2,7 @@ import React, { useEffect } from "react"
 import { fetchTimepointsForRoute } from "../api"
 import { LoadableTimepoints, Route, Timepoint } from "../skate"
 import {
+  deselectRoute,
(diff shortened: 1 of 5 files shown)


assets/src/components/routeLadder.tsx
```diff
@@ -14,6 +15,19 @@ interface Props {
   dispatch: Dispatch
 }
 
+const Header = ({ route, dispatch }: { route: Route; dispatch: Dispatch }) => (
```
skyqrose: For my own learning: When would you use a component, created with `<Header ...args... />` and when would you use a function `{header(route, dispatch)}`?

arkadyan: Lately I've been moving towards using more and more of the former because they feel more React-y and obviously like components. As far as I know though it is just a style thing. 🤷‍♀️

PR #21, opened 2019-04-18
Style the route ladders section

arkadyan:
Asana ticket: [Prettify the route ladders](https://app.asana.com/0/1112935048846093/1118883223380539)

Screen Shot 2019-04-18 at 12 17 25


ingridpierre:
SUPER EXCITING!!! This was so fun to pull down.

Sorry for the weird photos, I can't screenshot from these locked tablets:
IMG_2591
- Looks like there's some weird default styling on the X's that needs to be removed.
- Also seems to be some clipping and strangeness at the ends of the lines, kind of hard to see in this photo, sorry. Looks like the stop circles and the borders should be a little larger. I accomplished what I had in Sketch by having one be 8px wide and the other 6px with a 1px border.
- I think we can reduce the width between route ladders to be like 2 or 1 rem apart, the headers separate them well so we don't need to be this generous.
- Agree with Sky about shifting the timepoints up a smidge to center them.

as a positive aside: as you add routes, pinch-zooming works super well to get everything in view, shockingly legible even with like 7-8 routes up on these lil screens.

skyqrose:
The 3d x is because the x is unicode rather than an svg, and I guess the tablet has a font makes it look 3d. I think that could be addressed by using an svg instead of unicode or by making sure the fonts we specify include the x so that they're used instead of the tablet's default.

ingridpierre:
Oop let me get you an svg right now then.

ingridpierre:
https://drive.google.com/file/d/1HnWVb-u5AFGX4wn3t3xpuN3a2M9DsfrI/view?usp=sharing

try this icon, let me know if that works?

assets/css/_route_ladder.scss
```diff
@@ -28,24 +43,46 @@
...
 
 .m-route-ladder__timepoint {
-  align-items: center;
   display: flex;
   justify-content: center;
+  margin-top: -3px;
+
+  &:first-of-type {
+    margin-top: 0;
+    padding-top: 3px;
```
skyqrose: Is this needed? Shouldn't the top timepoint be bumped up just as much as the others?

arkadyan: The `padding` bit is an attempt to fix a bug on the tablets where the circle is being cutoff. I'm working on seeing this on the tablet now so I can determine if this actually fixes it or not…

skyqrose:
@arkadyan do you need another review from me?

arkadyan:
@skyqrose no, thanks for checking. I'm just waiting to be able to test and then fix a final bug on the tablet.

assets/css/_route_ladder.scss
```diff
@@ -1,25 +1,40 @@
...
+
+.m-route-ladder__close-icon {
+  display: inline-block;
+  height: 0.625rem;
+  width: 0.625rem;
+
+  svg {
+    stroke: $dark;
+    fill: $dark;
+  }
```
skyqrose: This all seems pretty general. Could this be applied to all close icons, so we don't need to set the custom `className` for each one?

arkadyan: Would you add something like `m-close-icon` to the svg itself, and set the css on that then?

arkadyan: Or maybe we even just set this for all svg elements until we have one that differs?

arkadyan: Made the svg bit generic.

arkadyan:
@ingridpierre I think the latest version (on https://skate-dev.mbtace.com) fixes the circle & line drawing bugs.

assets/css/app.scss
```diff
@@ -1,6 +1,19 @@
...
 
+body {
+  background-color: $pale-grey;
+  font-family: "Inter UI", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-size: 16px;
+}
+
+svg {
+  stroke: $dark;
+  fill: $dark;
```
skyqrose: Having this selector this generic might cause problems, because everywhere we use every svg everywhere, including the route ladder roads+stops, has to override it. I was thinking of replacing `.m-route-ladder__close-icon` with `.m-icon__close` or something like that. Not a strong preference.

arkadyan: Yeah, true. In that case I might say let's wait till we have a second close button to refactor it out.

PR #22, opened 2019-04-18
Add a dispatch context and provider

arkadyan:


assets/src/contexts/dispatchContext.ts
```diff
@@ -0,0 +1,8 @@
+import { createContext } from "react"
+import { Dispatch } from "../state"
+
+// tslint:disable-next-line: no-empty
+const noopDispatch: Dispatch = () => {}
```
skyqrose: Why is this needed?

arkadyan: `createContext` needs a default value. My reading of the documentation is that this is mostly to support testing, and that this default value is never used when you are using a provider around your component.

PR #25, opened 2019-04-24
Test fetching routes when the app mounts

arkadyan:
@skyqrose Is it worth just pulling this in since it stands alone?


skyqrose:
Closing in favor of #28, which contains this, as well as a more complete setup for testing effects.

PR #26, opened 2019-04-24
 Add a route ID filter

arkadyan:
Asana ticket: [Add filter by route ID to the route picker](https://app.asana.com/0/1112935048846093/1118437179466244)


assets/src/components/routePicker.tsx
```diff
@@ -50,37 +68,76 @@ const SelectedRoutesList = ({
   )
 }
 
+const RouteFilterInterface = ({
```
skyqrose: What do you think about having a `routeFilter.tsx` that exports the `<RouteFilterInterface>` and the `useRouteFilter` hook?

assets/src/hooks/useRouteFilter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface RouteFilter {
+  filterType: FilterType
+  filterText: string
+  handleTypeChange: (event: React.FormEvent<HTMLSelectElement>) => void
+  handleTextInput: (event: React.FormEvent<HTMLInputElement>) => void
+  clearTextInput: () => void
+}
+
+const isFilterType = (str: string): str is FilterType => str === "id"
```
skyqrose: Since we only have one filter type now, probably shouldn't set up all the infrastructure for switching types yet, and just assume that the text is for matching ids. If you think it'd be a lot of work to rip it out and add it back in later, then I guess it's fine, but I wouldn't be surprised if the requirements for filtering by something else like stop are so complicated that a lot of this logic would need to change anyway.

assets/src/components/routePicker.tsx
```diff
@@ -9,23 +11,39 @@ interface Props {
...
-    ) : (
+const RoutePicker = ({ routes, selectedRouteIds }: Props) => {
+  const routeFilter = useRouteFilter()
+
+  const filterRoutes = (allRoutes: Route[]): Route[] => {
+    const { filterType, filterText } = routeFilter
+
+    switch (filterType) {
+      case "id":
+        return allRoutes.filter((route: Route) => route.id.includes(filterText))
```
skyqrose: If you continue to support different types of filters, this `filterRoutes` function should be defined inside `useRouteFilter.ts`. (If you simplify it so there are no filter types, the simple call to `.filter()` is fine living here.)

assets/tests/components/routePicker.test.tsx
```diff
@@ -2,85 +2,121 @@ import { mount } from "enzyme"
...
-  const routes = [{ id: "id" }]
-
-  const routePicker = mount(
-    <DispatchProvider dispatch={mockDispatch}>
+const handleTextInputSpy = jest.fn()
+const mockRouteFilter: RouteFilter = {
+  filterType: "id",
+  filterText: "",
+  handleTypeChange: jest.fn(),
+  handleTextInput: handleTextInputSpy,
```
skyqrose: This is a mock, not a spy. (A spy preserves the real implementation.)

I don't think this needs its own variable, though. You could do `handleTextInput: jest.fn()` and `expect(mockRouteFilter.handleTextInput).toHaveBeenCalled()`

assets/tests/components/routePicker.test.tsx
```diff
@@ -2,85 +2,121 @@ import { mount } from "enzyme"
...
+  clearTextInput: jest.fn(),
+}
+const mockUseRouteFilter = jest.fn()
+mockUseRouteFilter.mockReturnValue(mockRouteFilter)
```
skyqrose: Can be on one line. `const mockUseRouteFilter = jest.fn(() => mockRouteFilter)`

assets/src/hooks/useRouteFilter.ts
arkadyan: Yeah, this is something I might think about handling later, but in this case I wanted to show the type picker in the view in order to get the layout correct, and it seemed more important to have the structure backing the view rather than leaving it "floating". But this is also why I was expecting the filter function to have logic that varies based on the type.

assets/src/components/routePicker.tsx
arkadyan: I had the same thought too, and was initially attempting to do that. The thing I was getting hung up on is that`RouteFilterInterface` requires the state values that are returned by `useRouteFilter`. If you can think of something I'm missing to make that work I'd be happy to hear it.

ingridpierre:
image
I was bummed by how the border radius and the clear icon were interacting (like a peg in a square hole), so I tweaked it down and took off the border. Changed the radius and removed border;  I think it looks a bit better this way ^^

Let's only add the black border when this element is in focus.

arkadyan:
@ingridpierre I think I made adjustments that match your changes, please take a look. I actually added a transparent border when the input is not focused, just to avoid the buttons below jumping down by a few pixels whenever you clicked into the input element.

ingridpierre:
Are you finding that the keyboard doesnt dismiss after you press "go" or enter? I think if we can make it go away that would be helpful, currently it's kinda hard to dismiss once you have it up.

arkadyan:
@ingridpierre I guess I hadn't done that much for whatever reason, but now that you point that out it is annoying. I added a new change that I think adds the behavior you are asking for. It's on dev so you can test on the tablet.

arkadyan:
@ingridpierre added the new icon

ingridpierre:
new icon and changes lgtm! 👍 thanks @arkadyan

assets/src/components/routePicker.tsx
skyqrose: The `RouteFilter` type would also be exported from `routeFilter.ts`, and `RoutePicker` would pass them in. Exact same setup as now, just moving the code into a different file.

assets/tests/components/routePicker.test.tsx
```diff
@@ -2,85 +2,120 @@ import { mount } from "enzyme"
...
+
+    routePicker
+      .find(".m-route-picker__selected-routes-button")
+      .first()
+      .simulate("click")
+
+    expect(mockDispatch).toHaveBeenCalledWith(deselectRoute("id"))
+  })
+
+  test("inputting filter text updates the route filter", () => {
```
skyqrose: There's more functionality that's not tested yet: Changing types, and clearing the input.

assets/tests/hooks/useRouteFilter.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    })
+  })
+
+  test("returns a RouteFilter", () => {
+    expect(typeof routeFilter.filterType).toBe("string")
+    expect(typeof routeFilter.filterText).toBe("string")
+    expect(routeFilter.handleTypeChange).toBeInstanceOf(Function)
+    expect(routeFilter.handleTextInput).toBeInstanceOf(Function)
+    expect(routeFilter.clearTextInput).toBeInstanceOf(Function)
+  })
```
skyqrose: Is this necessary? The typescript compiler should take care of this.

arkadyan: No, not necessary, I can remove it.

assets/src/components/routePicker.tsx
arkadyan: Oh I see, yep that works, I just had a blind spot trying to hide the filter data away. Thanks!

assets/tests/components/routePicker.test.tsx
arkadyan: Changing types is waiting til there is a second value to test changing to. Added a test for clearing the input.

arkadyan:
@skyqrose I'm fine with waiting and refactoring on top of #28. Should be straightfoward.

arkadyan:
Depends on #28

skyqrose:
Oops, I meant depends on #29, not #28

arkadyan:
@skyqrose rebased on master

assets/css/_route_picker.scss
```diff
@@ -24,6 +24,71 @@
   overflow-y: scroll;
 }
 
+.m-route-filter__type {
```
skyqrose: With a `routeFilter.tsx` more isolated, this should definitely be in its own `_route_filter.scss`

assets/src/hooks/useRouteFilter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return {
+    filterType,
+    filterText,
+    handleTypeChange,
+    handleTextInput,
+    clearTextInput,
+  }
+}
+
+export const RouteFilterInterface = ({
```
skyqrose: naming:

The component is named `RouteFilterInterface` and the object returned by the hook is named `RouteFilter`, which has confused me a couple of times. I think it might be clearer if the component was named simply `RouteFilter` (which makes it more like other components. We don't name the Route Picker `RoutePickerInterface`). The object could then be `Props` maybe? It's used as props but also has other uses. Maybe `RouteFilterData`?

arkadyan: RouteFilterData makes sense to me.

PR #27, opened 2019-04-25
Set up a channel for sending vehicles to client

arkadyan:
This is purely the server-side work pulled out from the work on [build UI for displaying/updating a table of vehicles and their locations](https://app.asana.com/0/1112935048846093/1114820119722045)

### DIFF
--- a/lib/skate_web/channels/vehicles_channel.ex
+++ b/lib/skate_web/channels/vehicles_channel.ex
@@ -1,6 +1,11 @@
 defmodule SkateWeb.VehiclesChannel do
   use SkateWeb, :channel
 
+  alias Phoenix.Socket
+  alias Realtime.{Server, Vehicle}
+
+  @spec join(topic :: String.t(), payload :: map, socket :: Socket.t()) ::
+          {:ok, reply :: %{vehicles: Server.vehicles()}, Socket.t()}
   def join("vehicles:" <> route_id, _message, socket) do
     vehicles = Realtime.Server.subscribe(route_id)
     {:ok, %{vehicles: vehicles}, socket}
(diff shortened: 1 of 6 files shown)


arkadyan:
@skyqrose is it ok with you if I pull this server-side work out early? I don't think there's really a risk of serious merge conflicts with the work on doing adding vehicle data, but it seems like it could be nice to see the entire picture at once.

lib/skate_web/channels/vehicles_channel.ex
```diff
@@ -0,0 +1,18 @@
+defmodule SkateWeb.VehiclesChannel do
+  use SkateWeb, :channel
+
+  def join("vehicles:" <> route_id, _message, socket) do
+    socket = assign(socket, :route_id, route_id)
```
paulswartz: Do you need the route_id? I don't see it used elsewhere.

arkadyan:
Adding tests for `VehiclesChannel` now.

lib/skate_web/channels/vehicles_channel.ex
arkadyan: Nope, doesn't look like it.

arkadyan:
Added tests

lib/skate_web/channels/vehicles_channel.ex
```diff
@@ -1,8 +1,12 @@
 defmodule SkateWeb.VehiclesChannel do
   use SkateWeb, :channel
 
+  alias Phoenix.Socket
+  alias Realtime.{Server, Vehicle}
+
+  @spec join(topic :: String.t(), payload :: map, socket :: Socket.t()) ::
```
paulswartz: These already have types: https://github.com/phoenixframework/phoenix/blob/v1.4.3/lib/phoenix/channel.ex#L288

So you can simplify to `@impl Phoenix.Channel`

arkadyan: Great, thanks, I didn't know that.

paulswartz: Anything that has a behaviour works this way: GenServer, GenStage, and Plug are some other examples off the top of my head.

PR #28, opened 2019-04-25
move fetchTimepoints to App. test effects.

skyqrose:
We finally have a way of testing effects that I feel
* tests everything we have adequately
* keeps tests isolated enough that future new tests won't trip over existing tests
* provides a structure that we can copy when testing future effects

This PR
* Moves the fetchTimepoints effect into App, so we can consolidate all effects at the top level.
* Tests for all the effects in App so far.

It doesn't add [the sockets task](https://app.asana.com/0/1112935048846093/1114820119722045) yet, but by providing a good testing framework, the socket tests will be easy enough to build on top of this PR.

---

Each effect is given its own custom hook, so we can test it independently from the App component.  That testing uses the `testHook` component from #26, which will need to be rebased on top of this.

Mocking is done by
```ts
// app.test.tsx
import { fetchRoutes } from "../../src/api"
jest.mock("../../src/api", () => ({
  __esModule: true,
  fetchRoutes: jest.fn(() => new Promise(() => {})),
}))
// In the test
    const mockFetchRoutes: jest.Mock = fetchRoutes as jest.Mock
    expect(mockFetchRoutes).toHaveBeenCalled()
```

The mock is shared within the file, but `"clearMocks": true` in the jest config (in `package.json`) prevents interference, as long as all tests are good about using `act` to make sure they're done with any effects by the end of the test.

So far, no tests need custom implementations of mocked functions. If we do, we'll have to make sure that the implementation is appropriately reset to the default mocked implementation.

This doesn't test that dispatch is called after the promises resolve. Should it?

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -1,8 +1,8 @@
-import React, { useEffect, useReducer } from "react"
-import * as Api from "../api"
+import React, { useReducer } from "react"
+import { useFetchRoutes } from "../hooks/useFetchRoutes"
+import { useFetchTimepoints } from "../hooks/useFetchTimepoints"
 import DispatchProvider from "../providers/dispatchProvider"
-import { Route } from "../skate.d"
-import { initialState, reducer, setRoutes } from "../state"
+import { initialState, reducer } from "../state"
 import RouteLadders from "./routeLadders"
 import RoutePicker from "./routePicker"
 
@@ -10,11 +10,8 @@ const App = (): JSX.Element => {
   const [state, dispatch] = useReducer(reducer, initialState)
   const { routes, selectedRouteIds, timepointsByRouteId } = state
 
-  useEffect(() => {
-    Api.fetchRoutes().then((newRoutes: Route[]) =>
-      dispatch(setRoutes(newRoutes))
-    )
-  }, [])
+  useFetchRoutes(dispatch)
+  useFetchTimepoints(selectedRouteIds, timepointsByRouteId, dispatch)
 
   const selectedRoutes = (routes || []).filter(route =>
     selectedRouteIds.includes(route.id)
--- a/assets/src/hooks/useFetchTimepoints.ts
+++ b/assets/src/hooks/useFetchTimepoints.ts
@@ -1,0 +1,26 @@
+import { useEffect } from "react"
+import { fetchTimepointsForRoute } from "../api"
+import { RouteId, Timepoint, TimepointsByRouteId } from "../skate"
+import {
+  Dispatch,
+  setLoadingTimepointsForRoute,
+  setTimepointsForRoute,
+} from "../state"
+
+export const useFetchTimepoints = (
+  selectedRouteIds: RouteId[],
+  timepointsByRouteId: TimepointsByRouteId,
+  dispatch: Dispatch
+): void => {
+  useEffect(() => {
+    selectedRouteIds.forEach((routeId: RouteId) => {
+      if (!(routeId in timepointsByRouteId)) {
+        dispatch(setLoadingTimepointsForRoute(routeId))
+
+        fetchTimepointsForRoute(routeId).then((newTimepoints: Timepoint[]) =>
+          dispatch(setTimepointsForRoute(routeId, newTimepoints))
+        )
+      }
+    })
+  }, [selectedRouteIds])
+}
--- a/assets/tests/components/app.test.tsx
+++ b/assets/tests/components/app.test.tsx
@@ -4,7 +4,5 @@ import App from "../../src/components/app"
 
 test("renders", () => {
   const tree = renderer.create(<App />).toJSON()
-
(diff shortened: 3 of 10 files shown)


assets/src/components/app.tsx
```diff
@@ -34,4 +38,28 @@ const App = (): JSX.Element => {
   )
 }
 
+export const useFetchRoutes = (dispatch: Dispatch): void => {
```
arkadyan: What do you think about moving these each to its own file in `hoooks/` or something like that? I worry that this App component file might get more and more sprawling as we continue to add more of these. As long as the hooks are named well (which I think these are) I think we can understand the behavior where they are used while keeping each file more focused.

assets/tests/components/app.test.tsx
```diff
@@ -1,10 +1,87 @@
...
+    const mockFetchRoutes: jest.Mock = Api.fetchRoutes as jest.Mock
+    act(() => {
+      testHook(() => {
+        useFetchRoutes(jest.fn())
+      })
+    })
+    expect(mockFetchRoutes).toHaveBeenCalledTimes(1)
+  })
+
+  test("fetches routes exactly once", () => {
```
arkadyan: What about adding some context about when this might happen or what we are trying to avoid, e.g. something along the lines of, "doesn't re-fetch routes on subsequent renders"?

assets/src/hooks/useFetchTimepoints.ts
```diff
@@ -0,0 +1,26 @@
...
+    selectedRouteIds.forEach((routeId: RouteId) => {
+      if (!(routeId in timepointsByRouteId)) {
+        dispatch(setLoadingTimepointsForRoute(routeId))
+
+        fetchTimepointsForRoute(routeId).then((newTimepoints: Timepoint[]) =>
+          dispatch(setTimepointsForRoute(routeId, newTimepoints))
+        )
+      }
+    })
+  }, [selectedRouteIds])
```
paulswartz: this seems like it also depends on `timepointsByRouteId`?

assets/src/hooks/useFetchTimepoints.ts
```diff
@@ -0,0 +1,26 @@
...
+  selectedRouteIds: RouteId[],
+  timepointsByRouteId: TimepointsByRouteId,
+  dispatch: Dispatch
+): void => {
+  useEffect(() => {
+    selectedRouteIds.forEach((routeId: RouteId) => {
+      if (!(routeId in timepointsByRouteId)) {
+        dispatch(setLoadingTimepointsForRoute(routeId))
+
+        fetchTimepointsForRoute(routeId).then((newTimepoints: Timepoint[]) =>
```
paulswartz: why is fetching the data something defined here in the effect, rather than something handled by an event in `dispatch`?

assets/src/hooks/useFetchTimepoints.ts
skyqrose: It does, but we don't care about rerunning the effect if `timeointsByRouteId` changes. And if we do include it as a dependency, the effect will rerun (without doing anything) whenever we receive new data.

Given that, should `timepointsByRouteId` still be in the list of dependencies?

assets/src/hooks/useFetchTimepoints.ts
skyqrose: The reducer has to be pure, so it can't trigger the fetch.

assets/src/hooks/useFetchTimepoints.ts
arkadyan: My instinct is that `timepointsByRouteId` shouldn't be a dependency. It's true that it's used here, but in effect it is the result of this function, so we know that if it has changed it is because we just changed it. It is something we should keep in our heads because it will be a concern if `timepointsByRouteId` is ever altered by something else in the future, but I can't imagine a scenario for that happening right now.

assets/src/hooks/useFetchTimepoints.ts
paulswartz: What happens when there's an error?

skyqrose: Right now, we don't handle that case. If we want to handle it in the future, we'd add a `SetFetchErrorAction`, and then add a case to the `.then(`, which would call dispatch with that action.

paulswartz: Can we log something to `console.error` so that it's not failing silently?

arkadyan: `fetchTimepointsForRoute` does have a `catch` block that's logging the error to `console.error`. In general our error handling and UI needs a bunch of thought - I'm going to make a ticket to at least have a place holder.

arkadyan: Adapting an existing ticket: https://app.asana.com/0/1112935048846093/1117662445344607. Please feel free to add anything there that would be helpful.

PR #29, opened 2019-04-26
Socket on frontend. Hooks maintain their own state.

skyqrose:
Screen Shot 2019-04-26 at 18 36 53

* Moves state into the data-fetching hooks.

This simplifies `app.ts` and `state.ts`, since they don't have to deal with all the tiny details like deleting channels for routes we aren't using anymore. It does ruin our "one central place for state", which makes me not 100% sold on this arrangement, but I think the clean abstraction makes it worth it.

* Uses `react-hooks-testing-library` instead of our home-grown version.
* Opens a socket from the frontend
* Shows the vehicles in a simple list under each route ladder.


assets/tests/hooks/useVehiclesSocket.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      (selectedRouteIds: RouteId[]) => useVehiclesSocket(selectedRouteIds),
+      { initialProps: ["1"] }
+    )
+    rerender([]) // Deselect the route
+
+    expect(mockChannel.leave).toHaveBeenCalledTimes(1)
+  })
+})
+
+describe("reducer", () => {
```
skyqrose: Is this test needed?

It's peeking pretty deeply at implementation details (as opposed to the `useVehiclesSocket` test, which doesn't need to know anything about how the hook stores its internal state). If we got rid of it, we could get rid of all the exports from `useVehiclesSocket.ts` except for the hook itself, which would be nice. But until we figure out how to test data arriving, this tests some code that the more blackbox testing can't.

assets/src/hooks/useFetchTimepoints.ts
```diff
@@ -1,26 +1,44 @@
...
-  setLoadingTimepointsForRoute,
-  setTimepointsForRoute,
-} from "../state"
 
-export const useFetchTimepoints = (
-  selectedRouteIds: RouteId[],
-  timepointsByRouteId: TimepointsByRouteId,
-  dispatch: Dispatch
-): void => {
+const useFetchTimepoints = (
```
paulswartz: Since this is now a hook, what about `useTimepoints(RouteId): Timepoint[] | null` instead? The map seems like overkill.

assets/src/hooks/useVehiclesSocket.ts
```diff
@@ -0,0 +1,175 @@
...
+          dispatch(setChannelForRoute(routeId, channel))
+        }
+      })
+    }
+  }, [socket, selectedRouteIds])
+
+  return vehiclesByRouteId
+}
+
+export default useVehiclesSocket
```
paulswartz: What about splitting this into two hooks? One to maintain the socket, and the other to maintain a channel for each route?

assets/src/skate.d.ts
```diff
@@ -1,9 +1,17 @@
+export interface ByRouteId<T> {
```
arkadyan: 👍

assets/tests/hooks/useFetchRoutes.test.ts
```diff
@@ -13,23 +12,18 @@ jest.mock("../../src/api", () => ({
 describe("useFetchRoutes", () => {
```
arkadyan: Testing that this returns a list of Routes seems like it could be useful.

assets/src/hooks/useVehiclesSocket.ts
```diff
@@ -0,0 +1,175 @@
...
+      return {
+        ...state,
+        vehiclesByRouteId: {
+          ...state.vehiclesByRouteId,
+          [action.payload.routeId]: action.payload.vehicles,
+        },
+      }
+    case "REMOVE_ROUTE":
+      const {
+        [action.payload.routeId]: channel,
```
arkadyan: Completely a style thing, but naming it `_ channel` (and `_ vehicles` below) would indicate that you aren't interested in using the value of this variable.

assets/tests/hooks/useVehiclesSocket.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    mockSocketChannel.mockImplementationOnce(() => mockChannel)
+
+    const { rerender } = renderHook(() => useVehiclesSocket(["1"]))
+
+    // Needs to be kicked to do the effects again after the socket initializes
+    rerender()
+
+    expect(mockSocketChannel).toHaveBeenCalledTimes(1)
+    expect(mockSocketChannel).toHaveBeenCalledWith("vehicles:1")
+    expect(mockChannel.join).toHaveBeenCalledTimes(1)
```
arkadyan: Can we also test the resulting `vehiclesByRouteId` value for each of these tests?

assets/src/hooks/useFetchTimepoints.ts
skyqrose: React internals depend on hooks always being called at the top level. You can't put them in a loop. [explanation](https://reactjs.org/docs/hooks-rules.html)

You could call the hook with a single argument if the hook was called directly from the RouteLadder component instead of from the App component, which is similar to how we used to do it, but I like having the data live top level.

paulswartz: It wouldn't be in a loop: each component which needed the timepoint data would call `const timepoints = useTimepoints(routeId)` with the routeId it receives as an argument.

assets/src/hooks/useVehiclesSocket.ts
skyqrose: Mmm. That was we could reuse the socket elsewhere, and mocking the socket in the tests becomes easier. I like it.

assets/tests/hooks/useFetchRoutes.test.ts
skyqrose: Unfortunately, that requires figuring out how to trigger effects when resolving promises without `act` acting up. I investigated this, but couldn't quite figure it out.

Perhaps I could test it with the `mockState` helper from `useFetchTimepoints.test.ts`?

assets/tests/hooks/useVehiclesSocket.test.ts
skyqrose: I wish. I haven't figured out how to get promises to safely cause effects by resolving in tests yet. That'd make our tests a lot better, so if you think it's worth it I can put more time into figuring that out.

assets/src/hooks/useFetchRoutes.ts
```diff
@@ -1,10 +1,13 @@
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { fetchRoutes } from "../api"
 import { Route } from "../skate"
-import { Dispatch, setRoutes } from "../state"
 
-export const useFetchRoutes = (dispatch: Dispatch): void => {
+const useFetchRoutes = (): Route[] | null => {
```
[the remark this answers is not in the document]
skyqrose: I agree that that would read better, but would it be clear that it's making network calls?

assets/tests/hooks/useFetchRoutes.test.ts
arkadyan: I don't know if it would work, but one idea is to return a mock result that wasn't a promise, but rather just an object that included a `then` function that returned a value.

assets/src/hooks/useFetchRoutes.ts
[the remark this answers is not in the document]
paulswartz: I'm not sure that the location of the route information is important to the callers of this function. Is there some other option to get the routes which doesn't involve a network call?

arkadyan:
@skyqrose I'm fine with waiting and refactoring on top of #28.

assets/tests/hooks/useVehiclesSocket.test.ts
arkadyan: I'm happy to trust your judgement and put off perfect tests in favor of moving forward if that seems like the most practical thing for now.

assets/src/hooks/useFetchTimepoints.ts
skyqrose: I want to keep it at the top level so that e.g. if the Ladder and the VehicleProperties components both need the timepoints, they can use the same data rather than each having to do their own fetch.

paulswartz: Isn't avoiding an extra fetch the responsibility of the hook? Both of the components would use `const timepoints = useTimepoints(routeId)` and the hook gives them the data.

assets/tests/hooks/useFetchRoutes.test.ts
skyqrose: That worked! 🎊

assets/src/hooks/useFetchTimepoints.ts
skyqrose: Hooks don't really have their own state. They store their state in their component. So if two separate components use the same hook, those two instances of the hook wouldn't share any state.

I suppose the hook could abstract away a reference to some shared mutable state, but that defeats the purpose of using React / FRP.

assets/tests/hooks/useVehiclesSocket.test.ts
skyqrose: I managed to test the return results (by synchronously calling the handler passed into `.on()` and `.receive()`), so testing the reducer implementation is no longer necessary. Removed this.

assets/tests/hooks/useVehiclesSocket.test.ts
skyqrose: Got this to work! By synchronously calling the handlers passed to `.on` and `.receive`, (similar to your synchronously resolving the promise suggestion on the other comment).

PR #31, opened 2019-04-30
Capture the Cache.Clean.run/1 output while testing

arkadyan:
No longer prints to the terminal, plus tests that it prints.


arkadyan:
@skyqrose Inspired by your last PR.

PR #33, opened 2019-05-01
use bare timepoint id instead of timepoint object/struct

skyqrose:
We're both working on things that will be a little easier on top of this.

### DIFF
--- a/assets/tests/api.test.ts
+++ b/assets/tests/api.test.ts
@@ -52,7 +52,7 @@ describe("fetchRoutes", () => {
 })
 
 describe("fetchTimepointsForRoute", () => {
-  test("fetches a list of timepoints for a route", done => {
+  test("fetches a list of timepointIds for a route", done => {
     window.fetch = () =>
       Promise.resolve({
         json: () => ({
@@ -62,12 +62,8 @@ describe("fetchTimepointsForRoute", () => {
         status: 200,
       })
 
-    fetchTimepointsForRoute("28").then(timepoints => {
-      expect(timepoints).toEqual([
-        { id: "MATPN" },
-        { id: "WELLH" },
-        { id: "MORTN" },
-      ])
+    fetchTimepointsForRoute("28").then(timepointIds => {
+      expect(timepointIds).toEqual(["MATPN", "WELLH", "MORTN"])
       done()
     })
   })
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -1,10 +1,10 @@
-import { Route, RouteId, Timepoint, TimepointId } from "./skate.d"
+import { Route, RouteId, TimepointId } from "./skate.d"
 
 interface RoutesResponse {
   data: Route[]
 }
 
-interface TimepointsForRouteResponse {
+interface TimepointsResponse {
   data: TimepointId[]
 }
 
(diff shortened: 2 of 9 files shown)


assets/tests/api.test.ts
```diff
@@ -63,11 +63,7 @@ describe("fetchTimepointsForRoute", () => {
       })
 
     fetchTimepointsForRoute("28").then(timepoints => {
-      expect(timepoints).toEqual([
-        { id: "MATPN" },
-        { id: "WELLH" },
-        { id: "MORTN" },
-      ])
+      expect(timepoints).toEqual(["MATPN", "WELLH", "MORTN"])
```
arkadyan: Maybe `timepointIds` instead of `timepoints` just for consistency?

paulswartz:
Can one of you speak to why you're going in this direction? It seems the opposite of what I would expect (expanding from an ID to an object).

arkadyan:
Given https://app.asana.com/0/1112935048846093/1120900265272092, do we want to hold off on this change?

skyqrose:
I think this change will make code that both of us are working on easier, so I don't think we should hold off on it.

A timepoint struct might be
```ex
@spec Timepoint.t :: %__MODULE__{
  id: Timepoint.id(),
  name: String.t(),
  stop_ids: [Stop.id()]
}
```

If we ever want those fields in the future, we could use a struct for them, but we're not right now.

In [#32 adding next timepoint to vehicles] (https://github.com/mbta/skate/pull/32/commits/6ddddee1e843ea47a63c46e74973a380131d1e01#diff-5dacdb8cc9a59c482d42c932f5926b80R14) , we were kind of heading towards including a single stop id in the timepoint and using it to represent a stop time that happens to be at a timepoint. Having a stop time struct, with an optional field for timepoint_id, would be a better way to represent that, and this PR is a move toward that representation.

With a struct with just one field, it was getting unclear as to what should be a timepoint struct and what should just be an id reference to that struct, and having `%Timepoint{id: timepoint_id}` structs all over the place was getting a little cumbersome.

paulswartz:
> Having a stop time struct, with an optional field for timepoint_id, would be a better way to represent that, and this PR is a move toward that representation.

This seems like the opposite of that: you're moving from an object to a single value? If you need to refactor towards the future goal of having a struct for all the stop times, a better path might be keeping the existing structure, but renaming it to `StopTime` and the existing `id` -> `timepointId`.

skyqrose:
Even if we had a `StopTime` type, all of the places that use `Timepoint` right now on the frontend should be using `TimepointId`. We can add the `StopTime` type once we need it on the frontend, but doing this change now allows my current work on [Add “Clear All” and “Select All” buttons to route picker](https://app.asana.com/0/1112935048846093/1118437179466240) to not need to juggle the difference between `TimepointId` and `Timepoint`.

paulswartz:
It seems like this is going to make your life easier and @arkadyan 's harder, but since he approved, who am I to stand in the way? :)

skyqrose:
Discussion continued in Slack: https://mbtace.slack.com/archives/CESHREAS1/p1556821548028800

skyqrose:
#34 doesn't need this, and [Add “Clear All” and “Select All” buttons to route picker](https://app.asana.com/0/1112935048846093/1118437179466240) needs some redesign and probably won't touch timepoints for a while, so I'm closing this.

PR #34, opened 2019-05-02
Add status towards current timepoint to vehicle data

arkadyan:
Include the percent we are there from the previous timepoint.

Asana ticket: [Add status towards current timepoint to vehicle data in backend](https://app.asana.com/0/1112935048846093/1119382502909570)


lib/realtime/vehicle.ex
```diff
@@ -64,20 +76,122 @@ defmodule Realtime.Vehicle do
...
+    |> past_current_future()
+  end
+
+  @spec split_on_stop_id([StopTime.t()], Stop.id()) :: {[StopTime.t()], [StopTime.t()]}
+  defp split_on_stop_id(stop_times, stop_id),
+    do: Enum.split_while(stop_times, &(&1.stop_id != stop_id))
+
+  @spec past_current_future({[StopTime.t()], [StopTime.t()]}) ::
+          {[StopTime.t()], StopTime.t() | nil, [StopTime.t()]}
+  defp past_current_future({past, [current | future]}), do: {Enum.reverse(past), current, future}
```
skyqrose: I don't think `current` needs to be split out from `future`. If the vehicle is `:in_transit_to`, then the current stop is in the future.

Right now, the only place one is used without the other is in the counting, but current should be included in future there to so that a bus whose next stop is a timepoint counts to that timepoint instead of to the next.

lib/realtime/vehicle.ex
```diff
@@ -64,20 +76,122 @@ defmodule Realtime.Vehicle do
...
+  defp past_current_future({past, future}), do: {Enum.reverse(past), nil, future}
+
+  @spec count_to_timepoint([StopTime.t()]) :: non_neg_integer()
+  defp count_to_timepoint(stop_times) do
+    count = Enum.find_index(stop_times, &is_a_timepoint?(&1))
+
+    if is_number(count), do: count + 1, else: 0
```
skyqrose: If there is no future or past timepoint, returning the length of the list might be a good default. It'd be like the first/last stop is a timepoint.

lib/realtime/vehicle.ex
```diff
@@ -64,20 +76,122 @@ defmodule Realtime.Vehicle do
...
+  def percent_of_the_way_to_next_timepoint(stop_times, stop_id) do
+    {past_stop_times, current_stop_time, future_stop_times} =
+      split_stop_times(stop_times, stop_id)
+
+    next_timepoint_stop_time =
+      Enum.find([current_stop_time | future_stop_times], &is_a_timepoint?(&1))
+
+    {
+      percent_of_the_way(
+        count_to_timepoint(past_stop_times),
```
skyqrose: After moving `current` into the future, the `+1` should move from counting to here.

lib/realtime/vehicle.ex
```diff
@@ -64,20 +76,122 @@ defmodule Realtime.Vehicle do
...
+      label: label,
+      timestamp: timestamp,
+      direction_id: direction_id,
+      route_id: route_id,
+      trip_id: trip_id,
+      current_stop_status: current_stop_status,
+      stop_id: stop_id,
+      current_timepoint_status:
+        timepoint_status(current_stop_status, next_timepoint_stop_time, stop_id),
+      timepoint_id: next_timepoint_id(next_timepoint_stop_time),
```
skyqrose: could be `next_timepoint_stop_time && next_timepoint_stop_time.timepoint_id` instead of needing its own function.

lib/gtfs/stop_time.ex
```diff
@@ -0,0 +1,55 @@
...
+  def trip_stop_times_from_csv(stop_times_csv) do
+    stop_times_csv
+    |> Enum.group_by(fn stop_time_row -> stop_time_row["trip_id"] end)
+    |> Helpers.map_values(fn stop_times_on_trip ->
+      stop_times_on_trip
+      |> Enum.sort_by(fn stop_time_row -> stop_time_row["stop_sequence"] end)
+      |> Enum.map(fn stop_time_row ->
+        %__MODULE__{
+          stop_id: stop_time_row["stop_id"],
+          timepoint_id: stop_time_row["checkpoint_id"]
```
skyqrose: stop times that aren't at a timepoint will have `""` here, so you'll need to catch that and turn it into `nil`.

lib/gtfs/stop_time.ex
```diff
@@ -0,0 +1,55 @@
...
+        }
+      end)
+    end)
+  end
+
+  @spec row_includes_a_checkpoint?(Csv.row()) :: boolean
+  def row_includes_a_checkpoint?(row), do: row["checkpoint_id"] != ""
```
skyqrose: This function is now unused.

test/channels/vehicles_channel_test.exs
```diff
@@ -30,6 +45,20 @@ defmodule SkateWeb.VehiclesChannelTest do
...
+        Application.put_env(:realtime, :stop_times_on_trip_fn, real_stop_times_on_trip_fn)
+      end)
+
+      Application.put_env(:realtime, :stop_times_on_trip_fn, fn _trip_id ->
+        [
+          %StopTime{stop_id: "6553", timepoint_id: "tp1"},
+          %StopTime{stop_id: "6554", timepoint_id: ""},
+          %StopTime{stop_id: "6555", timepoint_id: "tp2"}
+        ]
+      end)
```
skyqrose: Why are these needed?

test/gtfs/data_test.exs
```diff
@@ -0,0 +1,121 @@
...
+      route_patterns: [],
+      stops: [],
+      trip_stop_times: %{},
+      trips: []
+    }
+
+    assert Data.all_routes(data) == [%Route{id: "1"}, %Route{id: "2"}]
+  end
+
+  test "all_trips/1 returns all the routes" do
```
skyqrose: typo

test/gtfs/data_test.exs
```diff
@@ -0,0 +1,121 @@
...
+  alias Gtfs.Trip
+
+  test "all_routes/1 returns all the routes" do
+    data = %Data{
+      routes: [%Route{id: "1"}, %Route{id: "2"}],
+      route_patterns: [],
+      stops: [],
+      trip_stop_times: %{},
+      trips: []
+    }
```
skyqrose: (Applies to all the tests in this file)

Instead of creating the internal struct here, it might be nice to test something like
```ex
{:mocked_files, %{"routes.txt": [...]}
|> parse_files()
|> all_routes()
```
so the test doesn't depend on the internal structure, and is more similar to how it's used. I think it'd even mean you can put the struct in a private `defpstruct`

test/gtfs/route_pattern_test.exs
```diff
@@ -0,0 +1,122 @@
...
+                 id: "3",
+                 route_id: "10",
+                 direction_id: 1,
+                 representative_trip_id: "20"
+               }
+             ]
+    end
+  end
+
+  describe "by_direction/1" do
```
skyqrose: This `describe` has an `assert` directly inside of it. It should either have a `test` inside of it or be a `test` instead of a `describe`.

lib/realtime/vehicle.ex
arkadyan: Dialyzer was complaining about that, that's what I had initially.

test/channels/vehicles_channel_test.exs
arkadyan: Decoding vehicle data makes a call to get stop times on the trip: https://github.com/mbta/skate/blob/mss-vehicle-timepoint-data/lib/realtime/vehicle.ex#L93

test/gtfs/data_test.exs
arkadyan: I'm not sure I follow. This isn't internal sturcture, this is what Data is.

lib/realtime/vehicle.ex
```diff
@@ -64,20 +76,122 @@ defmodule Realtime.Vehicle do
...
+    stop_id = json["vehicle"]["stop_id"]
+
+    stop_times_on_trip =
+      try do
+        stop_times_on_trip_fn.(trip_id)
+      catch
+        # Handle Gtfs server timeouts gracefully
+        :exit, _ ->
+          []
+      end
```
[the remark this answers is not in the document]
arkadyan: I guess the question is what are we trying to improve? This is already catching an existing timeout. If this is timing out it also means that the site isn't loadable in the browser yet anyway, so no one will every see any effects from this.

lib/realtime/vehicle.ex
arkadyan: This percentage calculation doesn't worry about in transit to vs at. Including the current in next would make the result wrong when it is at a station (see the 40% test), and when it's in transit as well, because it will give the percentage value for the previous stop. Let's see how this level of granularity works in practice. If we want more position detail we can always adjust the calculation in the future to take in transit vs at into account.

arkadyan:
I should have noted you'll need to run `mix cache.clean` before running this branch.

lib/realtime/vehicle.ex
skyqrose: Discussed in person.

arkadyan: 👍 updated

test/gtfs/data_test.exs
skyqrose: Instead of testing that given some gtfs files, parsing creates the Data we expect; and then that given some Data, the query generates the answer we expect; we could do it it in one step by testing that given some gtfs files, parsing and then querying generates the answer we expect.

This would make the tests less dependent on the implementation details of the `Data.t()` format, but typing the gtfs files in the test might be more annoying than typing the data structs.

arkadyan: It might be more annoying, and I also think that for everything other than the parse function we don't want to be thinking in terms of those files. Those are meaningless to us once they are parsed, we are operating on our own data struct for everything once we are going.

lib/realtime/vehicle.ex
```diff
@@ -1,15 +1,22 @@
...
-          route_id: Gtfs.Route.id(),
-          trip_id: Gtfs.Trip.id(),
-          current_status: current_status(),
-          stop_id: Gtfs.Stop.id()
+          direction_id: Direction.id(),
+          route_id: Route.id(),
+          trip_id: Trip.id(),
+          current_stop_status: current_status(),
+          stop_id: Stop.id(),
+          current_timepoint_status: current_status(),
```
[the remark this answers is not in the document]
arkadyan: Updated this structure as discussed.

lib/realtime/vehicle.ex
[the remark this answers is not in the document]
skyqrose: K. I'm fine leaving this as is.

lib/skate_web/router.ex
```diff
@@ -26,13 +26,13 @@ defmodule SkateWeb.Router do
   scope "/", SkateWeb do
     pipe_through [:redirect_prod_http, :browser]
 
-    get "/", PageController, :index
+    get("/", PageController, :index)
   end
 
   scope "/api", SkateWeb do
-    pipe_through :api
+    pipe_through(:api)
```
skyqrose: Why does this get `()`, but the `pipe_through` above doesn't? Was this automatic formatting?

lib/realtime/vehicle.ex
```diff
@@ -64,18 +76,103 @@ defmodule Realtime.Vehicle do
...
   end
 
+  @spec percent_of_the_way_to_next_timepoint([StopTime.t()], Stop.id()) ::
+          {non_neg_integer(), StopTime.t() | nil} | nil
+  def percent_of_the_way_to_next_timepoint([], _stop_id), do: {0, nil}
+
+  def percent_of_the_way_to_next_timepoint(stop_times, stop_id) do
+    {past_stop_times, future_stop_times} = Enum.split_while(stop_times, &(&1.stop_id != stop_id))
+
+    next_timepoint_stop_time = Enum.find(future_stop_times, &is_a_timepoint?(&1))
```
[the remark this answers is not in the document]
arkadyan: I thought we'd discussed cases where a route didn't have timepoints (at least in sections?), and so we were treating the final stop as the next timepoint. Am I confusing things?

arkadyan:
@skyqrose I'm not sure I'm quite following about the `timepoint_id` type. Everwhere that interacts with that field will need to deal with the fact that it could be a string or could be nil. Is it a question of naming? E.g. would calling it `optional_timepoint_id` or something like that be more clear?

lib/realtime/vehicle.ex
```diff
@@ -64,18 +76,103 @@ defmodule Realtime.Vehicle do
...
+
+  @spec is_a_timepoint?(StopTime.t()) :: boolean
+  defp is_a_timepoint?(nil), do: false
+  defp is_a_timepoint?(stop_time), do: stop_time.timepoint_id != ""
+
+  @spec percent_of_the_way(non_neg_integer(), non_neg_integer()) :: non_neg_integer()
+  defp percent_of_the_way(past_count, future_count) do
+    Kernel.trunc(
+      past_count / (past_count + future_count) *
+        100
```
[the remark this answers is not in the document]
arkadyan: I like keeping it so that the function name describes the intent.

arkadyan:
@skyqrose fixed the timepoint check that was causing the percents to all be 100.

lib/skate_web/endpoint.ex
```diff
@@ -1,46 +1,50 @@
...
   # Set :encryption_salt if you would also like to encrypt it.
-  plug Plug.Session,
+  plug(Plug.Session,
     store: :cookie,
     key: "_skate_key",
     signing_salt: "jkUgGkwy"
+  )
 
-  plug SkateWeb.Router
+  plug(SkateWeb.Router)
```
skyqrose: I noticed you got rid of the `(` `)` formatting in `router.ex`. Did you also want to revert the formatting here?

arkadyan: Ok

lib/realtime/vehicle.ex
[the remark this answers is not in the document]
skyqrose: Oh yeah, I had forgotten about that case.

My current opinion is to lean slightly toward making the whole thing `nil` in that case, because without a timepoint id, it'd be difficult to do anything useful with the percentage on the frontend. But I'd be fine either way.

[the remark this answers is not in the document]
arkadyan: I think it's the opposite, I think this lets you do something with it on the frontend. With this info you can plot progress towards the end of the line. You couldn't do that if the whole struct was nil.

test/realtime/vehicle_test.exs
```diff
@@ -0,0 +1,150 @@
...
+                 status: :in_transit_to,
+                 timepoint_id: "tp2",
+                 percent_of_the_way_to_timepoint: 100
+               }
+             }
+    end
+  end
+
+  describe "percent_of_the_way_to_next_timepoint/3" do
+    test "returns 0 if the stop is a timepoint, plus the timepoint" do
```
skyqrose: `0` -> `100`

PR #35, opened 2019-05-02
Configure prod site to redirect from http to https

arkadyan:
Asana ticket: [Set up a redirect from http to https](https://app.asana.com/0/1112935048846093/709916117691678)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,6 +14,8 @@ config :skate,
 
 config :skate, Gtfs.CacheFile, cache_filename: nil
 
+config :skate, :redirect_http?, false
+
 # Configures the endpoint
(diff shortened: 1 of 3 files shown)


arkadyan:
@paulswartz does the port still need to be open for this on the dev server?

arkadyan:
@skyqrose yep, good call. Added that.

paulswartz:
I'd actually have said no: API requests to the wrong URL should fail, as they're automated. The reason for redirecting is to allow people to type `skate-dev.mbtace.com` into their browser and have it work: that's not a concern for the API side.

arkadyan:
@paulswartz Ok, fair. I can pull that back out.

PR #38, opened 2019-05-03
Add a button to reverse the order of the route ladder timepoints

arkadyan:
Asana ticket: [Add “Reverse” button to route ladders](https://app.asana.com/0/1112935048846093/1118437179466246)

Screen Shot 2019-05-03 at 13 12 40


assets/src/components/routeLadder.tsx
```diff
@@ -58,23 +58,40 @@ const Vehicle = ({ vehicle }: { vehicle: Vehicle }) => (
   </ul>
 )
 
-const RouteLadder = ({ route, timepoints, vehicles }: Props) => (
-  <div className="m-route-ladder">
-    <Header route={route} />
+const RouteLadder = ({ route, timepoints, vehicles }: Props) => {
+  const [shouldReverseStops, setShouldReverseStops] = useState(false)
```
skyqrose: When we start drawing the vehicles, we'll want a sense of "is direction_id 0 upwards or downwards", and not just "is this the same order or opposite order from the server".

Is that something we can define now? Can we document that the api returns stops in the 0 to 1 order, and then have this be `"0to1" | "1to0"`?

arkadyan: Ok, fine with me. I might suggest using something like `ZeroToOne` and `OneToZero` so that they can be keys in an `enum`.

arkadyan: Updated

ingridpierre:
Sorry for the delay, @arkadyan lgtm! 👍

PR #39, opened 2019-05-06
Add a collapse/show button to route picker

arkadyan:
Asana ticket: [Add collapse/show button to route picker](https://app.asana.com/0/1112935048846093/1118437179466242)

Screen Shot 2019-05-06 at 16 53 21

PR #40, opened 2019-05-07
Add an initial Vehicle Properties Panel

arkadyan:
Asana ticket: [Add an initial Vehicle Properties Panel](https://app.asana.com/0/1112935048846093/1120388911202260)

Screen Shot 2019-05-07 at 15 02 14

### DIFF
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -1,0 +1,70 @@
+.m-vehicle-properties-panel {
+  background-color: $pale-grey;
+  height: 100%;
+  max-width: 100%;
+  position: fixed;
+  right: 0;
+  top: inherit;
+  width: 23.5rem;
+  z-index: 1001;
+
+  hr {
+    border: none;
+    border-top: solid 2px $lighter-grey;
+    margin: 0 1.25rem;
+  }
+}
+
+.m-vehicle-properties-panel__header {
+  margin-top: 0.5rem;
+  min-height: 2rem;
+
+  .m-close-button {
+    position: absolute;
+    right: 0.875rem;
+    top: 0.875rem;
+  }
+}
+
+dl.m-vehicle-properties-panel__vehicle-properties {
+  margin-bottom: 2rem;
+
+  dt,
+  dl {
+    width: 50%;
+  }
+
+  dt {
+    font-size: 0.8125rem;
+    font-weight: bold;
+    text-align: right;
+    text-transform: uppercase;
+  }
+
+  dd {
+    font-size: 0.875rem;
+    margin-left: 0.5rem;
+  }
+}
+
+.m-vehicle-properties-panel__vehicle-property {
+  display: flex;
+}
+
+.m-vehicle-properties-panel__close {
+  background-color: $white;
+  border-radius: 4px;
+  display: block;
+  margin: 2rem auto 0;
+  padding: 0.5rem 3rem;
+}
+
+.m-vehicle-properties-panel__modal-overlay {
+  background-color: transparentize($dark, 0.59);
+  height: 100%;
+  left: inherit;
+  position: fixed;
+  top: inherit;
+  width: 100%;
+  z-index: 1000;
+}
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -7,12 +7,6 @@
   top: inherit;
   width: 23.5rem;
   z-index: 1001;
-
-  hr {
-    border: none;
-    border-top: solid 2px $lighter-grey;
-    margin: 0 1.25rem;
-  }
 }
 
 .m-vehicle-properties-panel__header {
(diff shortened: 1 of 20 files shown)


arkadyan:
Dependent on https://github.com/mbta/skate/pull/39

assets/src/components/closeButton.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { closeIcon } from "../helpers/icon"
+
+interface Props {
+  onClick: () => void
+}
+
+const CloseButton = ({ onClick }: Props) => (
```
skyqrose: 👍

assets/src/state.ts
```diff
@@ -1,12 +1,14 @@
 import { Dispatch as ReactDispatch } from "react"
-import { RouteId } from "./skate.d"
+import { RouteId, Vehicle } from "./skate.d"
 
 export interface State {
   selectedRouteIds: RouteId[]
+  selectedVehicle?: Vehicle
```
skyqrose: For the vehicle properties panel to be auto-updating, this should be an id, and `app.tsx` should look up the vehicle from its up-to-date list before passing to `<VehicleProperties>`

```suggestion
  selectedVehicleId?: VehicleId
```

assets/css/_vehicle_properties_panel.scss
```diff
@@ -0,0 +1,70 @@
...
+  background-color: $pale-grey;
+  height: 100%;
+  max-width: 100%;
+  position: fixed;
+  right: 0;
+  top: inherit;
+  width: 23.5rem;
+  z-index: 1001;
+
+  hr {
```
skyqrose: Should the `hr` get its own class to make selecting it easier? (Same for `dl`, `dt`, and `dd` below which also have tag selectors.)

arkadyan: Yeah, I always wonder when then gets more annoying than it's worth, but sure, can do.

ingridpierre:
I know this panel has like 3 ways to close it now, but can we add a swipe gesture as well? (Swipe right inside the browser closes panel) I think that could make this feel extra nice on the tablets.

arkadyan:
Added a swipe gesture. You have to swipe right on the panel itself and it will close.

arkadyan:
Added tests to cover the swipe gesture support. @ingridpierre and @skyqrose could you both please review again when you have a chance?

assets/src/helpers/detectSwipe.ts
```diff
@@ -0,0 +1,79 @@
...
+  el.addEventListener("touchend", handleTouchEnd)
+
+  return () => {
+    el.removeEventListener("touchstart", handleTouchStart)
+    el.removeEventListener("touchmove", handleTouchMove)
+    el.removeEventListener("touchend", handleTouchEnd)
+  }
+}
+
+const firstTouchPosition = (event: TouchEvent) => {
```
skyqrose: The function signature could get a return value `: Position`

assets/src/helpers/detectSwipe.ts
```diff
@@ -0,0 +1,79 @@
+import { SwipeDirection } from "../skate.d"
+
+const MIN_X = 30 // Min x swipe for horizontal swipe
+const MAX_X = 30 // Max x difference for vertical swipe
+const MIN_Y = 50 // Min y swipe for vertical swipe
+const MAX_Y = 60 // Max y difference for horizontal swipe
```
skyqrose: Did you determine these numbers by testing on the tablet?

assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -13,9 +14,22 @@ const VehiclePropertiesPanel = ({ selectedVehicle }: Props) => {
...
+    }
+
+    return detectSwipe("m-vehicle-properties-panel", handleSwipe)
+  })
+
   return (
     <>
-      <div className="m-vehicle-properties-panel">
+      <div
+        id="m-vehicle-properties-panel"
```
skyqrose: If you swipe over the grayed overlay, does that close the panel? Would we want that? If the answers are no and yes, could we move the handler up a level so it covers both?

assets/src/helpers/detectSwipe.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+const detectSwipe = (elementId: string, callback: Callback) => {
+  let startingPosition: Position
+  let endingPosition: Position
+
+  const handleTouchStart = (event: TouchEvent) => {
+    event.preventDefault()
+    startingPosition = firstTouchPosition(event)
+  }
```
skyqrose: Does `handleTouchStart` also need to set `endingPosition` to avoid a bug? If you `touchstart` and `touchend` without any `touchmove`s inbetween, could you get an `endingPosition` that's unset or has a value from the previous swipe?

Alternatively, can you get the endingPosition out of the event in `handleTouchEnd` and not need a `touchmove` handler or stateful `let endingPosition` at all?

assets/src/helpers/detectSwipe.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (Math.abs(movement.x) > MIN_X && Math.abs(movement.y) < MAX_Y) {
+    // Horizontal movement
+    swipeDirection = movement.x > 0 ? "Right" : "Left"
+  } else if (Math.abs(movement.y) > MIN_Y && Math.abs(movement.x) < MAX_X) {
+    // Vertical movement
+    swipeDirection = movement.y > 0 ? "Down" : "Up"
+  }
+
+  return swipeDirection
+}
```
skyqrose: I think you could get away without the `let` by
```ts
const directionOfMovement = (movement: Position): SwipeDirection | undefined => {
   if (...) {
    return movement.x > 0 ? "Right" : "Left"
  } else if (...) {
    return movement.y > 0 ? "Down" : "Up"
  } else {
    return undefined
  }
}
```

assets/src/components/vehiclePropertiesPanel.tsx
arkadyan: No and no. You only need to tap in the grayed overlay to close the panel.

assets/src/helpers/detectSwipe.ts
arkadyan: No, they were cargo culted from a random example on the web. We can tweak them if we find a need.

assets/src/helpers/detectSwipe.ts
```diff
@@ -0,0 +1,79 @@
...
+const MAX_Y = 60 // Max y difference for horizontal swipe
+
+type Callback = (swipeDirection: SwipeDirection) => void
+
+interface Position {
+  x: number
+  y: number
+}
+
+const detectSwipe = (elementId: string, callback: Callback) => {
```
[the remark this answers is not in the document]
arkadyan: Added the return type. I think the test remains useful to guard against regression.

assets/src/helpers/detectSwipe.ts
arkadyan: We can play with endingPosition more too if we want, but it doesn't really matter since if they haven't moved it won't be a swipe.

arkadyan:
@skyqrose If we end up wanting more gestures I think a library might be in order. For this one limited usecase it seemed like overkill.

ingridpierre:
lgtm! 👍

PR #44, opened 2019-05-14
semaphore build scripts

skyqrose:
Asana Task: none

The semaphore configuration currently has a long config with one line per command. This causes two problems which are solved by putting the commands into a script in source control:

* When SSHing into a semaphore worker, running the setup requires copypasting all the commands (usually one by one, and there are a lot of them), which makes debugging semaphore builds much more difficult. Running `source semaphore/setup.sh` on the worker is much easier. (I ran into this recently when investigating why prettier behaved differently locally vs on semaphore.)

* When making changes to the semaphore config, it applies globally to all builds, even though you usually want the config to be in sync with one branch, so other branches will fail the build until they get the source code that matches the semaphore config. By putting the semaphore config in source control, any config changes will only apply to the branch that needs them until the branch merges.

Unfortunately, that config/source being out of sync problem affects this PR. My plan is:
* Get approval for this PR
* Change the semaphore config to call these new scripts. All branches will fail except this one.
* Merge this once semaphore passes.
* Rebase any open PRs.

~~This PR also updates the `NODEJS_VERSION` in `semaphore/setup.sh` from `10.15` to `11.10` so that semaphore uses the same version as we do locally (specified in `.tool-versions` `nodejs 11.10.0`)~~ Semaphore comes with 10.15 preinstalled, but not 11.10


semaphore/setup.sh
```diff
@@ -0,0 +1,14 @@
+# Fail if any command fails
+set -e
```
paulswartz: You'll need this for all the scripts.

skyqrose:
Updated semaphore config. Rebuilding...

PR #45, opened 2019-05-14
Add a Directions link to the Vehicle Properties Panel

skyqrose:
Asana Task: [Add a Directions link to the Vehicle Properties Panel](https://app.asana.com/0/1112935048846093/1122642346801333)

Screen Shot 2019-05-14 at 13 58 43


assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -50,6 +54,10 @@ const VehiclePropertiesPanel = ({ selectedVehicle }: Props) => {
 
         <hr className="m-vehicle-properties-panel__divider" />
 
+        <a className="m-vehicle-properties-panel__link" href={directionsUrl}>
+          Directions to vehicle
```
arkadyan: In the design the copy says, "Directions".

skyqrose: In the design, it's under "Current Stop" so it reads more like "directions to the current stop I just mentioned". Without the current stop, it needs a little more context to be clear what "Directions" means.

Though you're about to do the current stop, so I guess I can change it back and the context will become clear soon.

arkadyan: Yeah, and maybe you can just throw the "Current Location" title in there too even though there's  no value yet.

skyqrose: On second thought, you're adding "Next Stop", not "Current Location" (which we don't have a plan for implementing yet). Can I leave this as is, and let you figure out the right way to add the context as part of the "Next Stop" task?"

arkadyan: That's fine, but why not just make the link "Directions" so we don't have to change it later? Otherwise we should document that work as needing to be done in the future.

PR #46, opened 2019-05-14
simplify timepoint_status format

skyqrose:
Asana Task: none

#43 didn't end up using the `timepoint_status.status` field (the `fraction_until_timepoint` field was enough), and a couple inspectors have recently said that a "Next Timepoint" feature wouldn't be useful, so I don't think we're likely to end up needing it for anything else, and we can take it out.

#43 also didn't end up using the `fraction_until_timepoint` field if the `timepoint_id` was `null`, so we can make `timepoint_id` always present, which makes a couple of things simpler.

### DIFF
--- a/lib/realtime/vehicle.ex
+++ b/lib/realtime/vehicle.ex
@@ -13,11 +13,9 @@ defmodule Realtime.Vehicle do
   """
   @type timepoint_status ::
           %{
-            status: current_status(),
-            timepoint_id: StopTime.possible_timepoint_id(),
+            timepoint_id: StopTime.timepoint_id(),
             fraction_until_timepoint: float()
           }
-          | nil
 
   @type t :: %__MODULE__{
           id: String.t(),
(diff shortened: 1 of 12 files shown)


arkadyan:
@sky Weren't you saying that some trips don't have timepoints? In those cases won't we have `fraction_until_timepoint` while `timepoint_id` is `null` and we need to know all that in order to draw the vehicle?

skyqrose:
If a trip has no timepoints, `case Enum.find(future_stop_times, &is_a_timepoint?(&1))` will be `nil`, and `timepoint_status` will return `nil`.

lib/realtime/vehicle.ex
```diff
@@ -132,24 +117,28 @@ defmodule Realtime.Vehicle do
     }
   end
 
-  @spec fraction_until_timepoint([StopTime.t()], Stop.id()) ::
-          {float(), StopTime.t() | nil} | nil
-  def fraction_until_timepoint([], _stop_id), do: {0.0, nil}
+  @spec timepoint_status([StopTime.t()], Stop.id()) :: timepoint_status() | nil
+  def timepoint_status([], _stop_id) do
```
arkadyan: Is this extra case necessary? Will `Enum.split_while` and `Enum.find` just handle empty Lists and give you this result without an extra function definition?

PR #47, opened 2019-05-14
Use concentrate to incorporate Busloc data for Vehicles

arkadyan:
Asana ticket: [Ingest Busloc feed](https://app.asana.com/0/1112935048846093/1114820119721035)

Proof of concept embedding Concentrate in order to get enhanced vehicle data from Busloc. I don't imagine the vehicle positions consumer remaining a `Reporter`, that was simply for expedency to use an existing structure to get something working.

@skyqrose and @paulswartz what do you think of this approach? Does this seem like it's on a good track?

Open questions:

* Does all of Concentrate turn into a library? Or a core of it, minus the sink configuration for example maybe?
* What's the best way to enable hooking in consumers? I.e. we want a `VehiclePositionsConsumer` in Skate to replace the `Reporter` I have in the Concentrate side and we want to be able to configure it  subscribe to a hook in the Concentrate GenStage pipeline.
* Others?


paulswartz:
1) I don't see what new functionality this is providing?
2) A lot of the stuff you pulled in (reporters, filters) you probably don't need?
3) Could you have used `concentrate` as a dependency rather than pulling in the individual files? I know we talked about this before, but it makes the diff impossible to use.

paulswartz:
Maybe we should have a meeting later this week to talk about the Concentrate architecture, and how we might incorporate that into Skate?

arkadyan:
@paulswartz A meeting sounds like a good idea. I suspect a lot of this will be easier to work through in person. I'll set something up.

PR #48, opened 2019-05-14
Highlight the selected vehicle icon

arkadyan:
Asana ticket: [The selected vehicle for should be highlighted on the route ladder](https://app.asana.com/0/1112935048846093/1121935738144311)

Move the selection action to the new triangle icons.

### DIFF
--- a/assets/css/_ladder.scss
+++ b/assets/css/_ladder.scss
@@ -21,6 +21,21 @@
   text-transform: uppercase;
 }
 
-.m-ladder__label {
-  font-size: 0.75rem;
+.m-ladder__vehicle {
+  cursor: pointer;
+}
+
+.m-ladder__vehicle-triangle {
+  fill: $blue-grey;
+  stroke: none;
+
+  &.selected {
+    stroke: $dark;
+    stroke-width: 2;
+  }
+}
+
+.m-ladder__vehicle-label {
+  color: $dark;
+  font-size: 0.5625rem;
 }
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -1,7 +1,7 @@
 import React, { useContext, useState } from "react"
 import DispatchContext from "../contexts/dispatchContext"
 import { reverseIcon, reverseIconReversed } from "../helpers/icon"
-import { LoadableTimepoints, Route, Vehicle } from "../skate"
+import { LoadableTimepoints, Route, Vehicle, VehicleId } from "../skate"
 import { deselectRoute, selectVehicle } from "../state"
 import CloseButton from "./closeButton"
 import Ladder, { flipLadderDirection, LadderDirection } from "./ladder"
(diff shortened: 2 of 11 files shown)


arkadyan:
Screen Shot 2019-05-15 at 10 04 54

ingridpierre:
looks good, small tweak: can we make the overlay opacity 20% instead of 41 or what it is now? It makes it a little hard to see the state change currently.

assets/css/_ladder.scss
```diff
@@ -21,6 +21,21 @@
...
+  cursor: pointer;
+}
+
+.m-ladder__vehicle-triangle {
+  fill: $blue-grey;
+  stroke: none;
+
+  &.selected {
+    stroke: $dark;
+    stroke-width: 2;
```
skyqrose: We should decide which styling belongs in css and which styling belongs inline when writing the svg. I like this being here. Maybe stroke widths and colors should be in css, and shapes+positioning should be in svg? Would that be good to start standardizing on?

assets/src/components/routeLadder.tsx
```diff
@@ -25,52 +26,49 @@ const Header = ({ route }: { route: Route }) => {
...
-        <br />
-        <em>fraction before:</em>{" "}
-        {vehicle.timepoint_status &&
-          vehicle.timepoint_status.fraction_until_timepoint}
-      </li>
-    </ul>
-  )
-}
+const Vehicle = ({ vehicle }: { vehicle: Vehicle }) => (
+  <ul className="m-route-ladder__vehicle" style={{ border: "1px solid black" }}>
```
skyqrose: Removing dispatch here will cause merge conflicts with #45 and make it hard/impossible to view overlapped vehicles and vehicles without timepoints (which we're likely to want to view during development). Can we leave it in for now?

assets/tests/components/ladder.test.tsx
```diff
@@ -62,9 +65,116 @@ test("renders a ladder", () => {
...
     .toJSON()
 
   expect(tree).toMatchSnapshot()
 })
+
+test("highlights a selected vehicle", () => {
+  const timepoints = [{ id: "t0" }, { id: "t1" }, { id: "t2" }]
+  const vehicles: Vehicle[] = [
+    {
+      id: "upward",
```
skyqrose: I think this test might only need 1 vehicle, which could make the test shorter and easier to read.

assets/css/_ladder.scss
arkadyan: Yeah, I think that makes sense. I guess I was sort of conceptualizing it as "put everything in CSS except where you can't", but I think that's in essence the same thing as what you are saying without being as clear.

assets/src/components/routeLadder.tsx
arkadyan: Sure

assets/tests/components/ladder.test.tsx
arkadyan: I'm going to leave 2 to demonstrate others not being selected.

arkadyan:
@ingridpierre and @skyqrose made all those updates.

Screen Shot 2019-05-15 at 11 46 08

PR #52, opened 2019-05-21
Fix up Babel preset-env configuration

arkadyan:
Add babel polyfill.

Copy .browserslistrc over from dotcom.

### DIFF
--- a/assets/.browserslistrc
+++ b/assets/.browserslistrc
@@ -1,0 +1,8 @@
+# Attempt to cover old but in use browsers
+# Brings in ios safari 10
+> 0.25%
+
+# Browsers that we explicitly support
+safari >= 4
+ie >= 11
+last 2 versions
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -691,6 +691,15 @@
         "regexpu-core": "^4.1.3"
       }
     },
+    "@babel/polyfill": {
+      "version": "7.4.4",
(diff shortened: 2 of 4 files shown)


assets/.browserslistrc
```diff
@@ -0,0 +1,8 @@
+# Attempt to cover old but in use browsers
+# Brings in ios safari 10
+> 0.25%
+
+# Browsers that we explicitly support
+safari >= 4
+ie >= 11
+last 2 versions
```
skyqrose: We may be able to be more explicit about the browsers we support. E.g. I know rail OCC uses Firefox 41, but idk about bus OCC. What browser is on the tablets, and will they be represented in this list?

meagonqz: Yeah that makes sense, you could merge this for now and update later with more specific requirements.

Because this currently specifies `> 0.25%` marketshare, recent tablets (at least Android using Chrome or Apple tablet with iOS >= 10) should be fine.

arkadyan: Yeah, that's true. Also, we are targeting this tablet so that's what we should focus on, but if a configuration setting can help allow folks to access the site using their phone or whatever as well that seems like a nice plus.

arkadyan:
Thanks, @meagonqz. I added (and used) `webpack-bundle-analyzer` - that does seem useful.

PR #55, opened 2019-05-22
Add "Next Stop" to the Vehicle Properties Panel

skyqrose:
Asana Task: [Add "Next Stop" to the Vehicle Properties Panel](https://app.asana.com/0/1112935048846093/1120900265272096)

Screen Shot 2019-05-22 at 15 16 10

### DIFF
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -2,7 +2,6 @@
   background-color: $pale-grey;
   height: 100%;
   max-width: 100%;
-  padding: 1rem;
   position: fixed;
   right: 0;
   top: inherit;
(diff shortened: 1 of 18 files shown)


skyqrose:
Screen Shot 2019-05-24 at 14 22 59

Rebased and restyled according to the new design at https://app.zeplin.io/project/5caf86702bbb91be743b94d9/screen/5ce4874154e6df1dff5a1d14

ingridpierre:
@skyqrose There are some differences in the type sizes, however the new UI kit had a couple discrepancies. I've tried where I could to update them in the zeplin mockups and ui kit.

I've detailed what I'm hoping for below, but let me know if you just want to pair and look at things together as well if that's a little easier!


image

**Headers (headsign on vehicle properties, route ladder headers)**
- 1.25 rem
- font-weight: 700

**Body SM devices (in our case, 1280w or less):**
- At 1rem the base font size for our tablets should be 14px.
- If we want to get a jump on getting it to be styled responsively, we should also add a base font size for larger screens (>1280w) of 16px.

**Links, emphasized body type:**
- 1 rem, font-weight: 500

**Buttons, route picker dropdown, incoming box header etc.:**
- font-weight: 700 

**Labels, small type:**
- .875 rem for the grey labels and any type that's small but not miniscule.

**Tiny type (vehicle numbers):**
- .65 rem for super tiny type like vehicle numbers

There's a slightly bigger font size for the vehicle # in the top left of the vehicle properties panel. This is the only place it's used, but that should be 1.5rem and font-weight: 700.

skyqrose:
Okay, tweaked it a bit. Screen Shot 2019-05-29 at 15 17 37

For the rest of the changes (e.g. button styles, the link color that's not in the palette yet, and anything outside the VPP), I've made a separate task: [Use styles from UI kit](https://app.asana.com/0/1112935048846093/1125059554572559)

PR #56, opened 2019-05-22
Use a custom path for the vehicle triangles

arkadyan:
Asana ticket: [Use correct vehicle triangle icon](https://app.asana.com/0/1112935048846093/1123604338536305)

Screen Shot 2019-05-22 at 17 35 41

This also addresses: [Add a gray stroke to the triangle icon and a gray background behind the run number](https://app.asana.com/0/1112935048846093/1123838037403439)

### DIFF
--- a/assets/src/components/ladder.tsx
+++ b/assets/src/components/ladder.tsx
@@ -157,7 +157,7 @@ const LadderVehicle = ({
       onClick={() => dispatch(selectVehicle(vehicle.id))}
     >
       <Triangle isSelected={isSelected} />
-      <VehicleLabel vehicle={vehicle} rotation={rotation} />
+      <VehicleLabel vehicle={vehicle} rotation={rotation} x={0} y={26} />
     </g>
   )
 }
@@ -205,11 +205,15 @@ const Triangle = ({ isSelected }: { isSelected: boolean }) => {
 const VehicleLabel = ({
   vehicle,
   rotation,
+  x,
+  y,
 }: {
   vehicle: Vehicle
   rotation: number
+  x: number
+  y: number
 }) => (
-  <svg className="m-ladder__vehicle-label" x="0" y="26">
+  <svg className="m-ladder__vehicle-label" x={x} y={y}>
     <rect
       className="m-ladder__vehicle-label-background"
       rx="5.5"
--- a/assets/src/components/ladder.tsx
+++ b/assets/src/components/ladder.tsx
@@ -19,7 +19,6 @@ export const flipLadderDirection = (
 const HEIGHT = 500
 const WIDTH = 180
 const CENTER_TO_LINE = 40 // x-distance between the center of the ladder and the center of the line
-const CENTER_TO_VEHICLE = 60 // x-distance between the center of the ladder and the center of the vehicles
 const MARGIN_TOP_BOTTOM = 40 // space between the top of the route and the top of the viewbox
 
 export interface Props {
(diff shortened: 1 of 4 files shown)


assets/src/components/ladder.tsx
```diff
@@ -140,35 +139,23 @@ const LadderVehicle = ({
...
   const isSelected = vehicle.id === selectedVehicleId
 
+  const rotation = vehicleDirection === VehicleDirection.Down ? 180 : 0
+  const centerOfTriangleX = 16
+  const centerOfTriangleY = 13
+
   return (
     <g
       className="m-ladder__vehicle"
+      transform={`translate(${x},${y}) rotate(${rotation},${centerOfTriangleX},${centerOfTriangleY})`}
```
skyqrose: So it looks like the approach is
* draw the triangle at its default location
* draw the text (upside down if needed)
* translate everything to the right spot.

Would it be possible to instead tell the triangle where to draw itself (probably passing in its center), and tell the text where to draw itself?

There's a lot of hardcoded values and translation around working with the coordinates + size of the triangle that I'd like to isolate to within the `Triangle` component.

assets/src/components/ladder.tsx
```diff
@@ -201,29 +188,43 @@ const yForVehicle = (
...
-    <polygon
-      className={`m-ladder__vehicle-triangle ${selectedClass}`}
-      points={pointsString}
-    />
+    <g className="m-ladder__vehicle-icon" transform={`scale(0.625)`}>
+      <path
+        className={`m-ladder__vehicle-triangle ${selectedClass}`}
+        d="m27.34 9.46 16.84 24.54a4.06 4.06 0 0 1 -1 5.64 4.11 4.11 0 0 1 -2.3.71h-33.72a4.06 4.06 0 0 1 -4.06-4.11 4 4 0 0 1 .72-2.24l16.84-24.54a4.05 4.05 0 0 1 5.64-1.05 4 4 0 0 1 1.04 1.05z"
+      />
+    </g>
```
skyqrose: We'll eventually need to reuse this triangle in at least two other places: the vehicle properties panel and the pending box. This isn't a question we need to solve right now, but what's our plan for that?

arkadyan: It is trivial to pull this out into a shared component.

assets/src/components/ladder.tsx
arkadyan: Yes I think this is better because you remove a bunch custom values for positioning items different deponding on which side of the ladder they are on. It also separates concerns nicely - the triangle and text just draw themselves and don't need to know anything about a ladder, the grouping positions those two relative to each other, and then we position and rotate the entire grouping as a whole so everything is always consistent.

If I'm understanding correcly I think the only 2 values specific to the size of the triange are `centerOfTriangleX` and `centerOfTriangleY`. If those were retrieved from the triangle component (maybe separated into its own file) would that seem better?

skyqrose: Instead of fetching `centerOfTriangleX`, could the triangle center itself around `0, 0`, and then the `LadderVehicle` does the translation based on that? That way the triangle doesn't need to know where it's positioned, and the `LadderVehicle` doesn't need to know details about the triangle's size.

The text being drawn upside down and then rotating back up is similar to the text having to know what side it's on (it's just told its location in terms of degrees instead of Up/Down). Could the `LadderVehicle` rotate just the triangle, and leave the text always right side up?

skyqrose: Other ideas:
* Pass the rotation into the triangle instead of the text, and then LadderVehicle doesn't do any rotation.
* Pass the size of the triangle in as an argument to `Triangle`, and calculate the scale factor dynamically, so `LadderVehicle` can do calculations with the width without needing to know the details of the path.

arkadyan: Reseting the center is an interesting idea. I can look again, but I don't know if that's possible in svg. I'm afraid that it might have the effect of cutting off half the triangle, but I can try.

I don't think rotating just the triangle would work because you lose the relativel positioning of the text the text would end up being below the triangle instead of above it.

Passing the size to the triangle could work. But if the goal is to reduce hard-coded values I don't think that would change anything because you'd need to hard code the original size of the vehicle.

arkadyan: Actually centering the triangle at 0,0 doesn't change anything — that's essentially what rotate with 3 params is doing. If you were going to translate individual items, you'd have to translate all of them by the same "negative triangle center" amount, which would not be good. So instead you'd want to translate the grouping before rotating, but that's precisely what this rotate command is doing for you.

skyqrose: I was thinking the `Triangle` could translate itself so its centered on 0, 0, and then `LadderVehicle` translates it again.

`LadderVehicle` would know whether the text should be above or below, and translate the text accordingly.

The goal is to confine any hardcoding that needs to correspond to the values in the triangle path inside the `Triangle` component, and to have any hardcoding about placement in easily understandable and changeable variables.

arkadyan: What you are suggesting adds more hard coding about placement and the size of elements. In order to translate the text above or below you need to know the size of both the triangle and text and maintain absolute positioning for everything. That's where creating a grouping to define a relative context comes in. When you define a local context for positioning you just need to position elements relative to each other and you know the internal relationships are always consistent. Then you can absolutely place the grouping as a whole.

For a more complex example imagine an animation of a stick figure walking across the screen. If you needed to animate that by defining the absolute position of every joint it would be a nightmare. If instead you group your figure and create a relative context you can define the motion for the swinging arms and legs relatively simply (the fixed points never move, for example). Then you can move the entire group around the screen independantly of its internal motion.

skyqrose: I'm okay with hardcoding as long as it's in an appropriate place.

E.g. the `LadderVehicle` has `x = 47 : -77` where it needs to know the width of the triangle in order to keep the two side balanced. I think it'd be better isolated as (psuedocode) `x = 62 : -62` and `Triangle  = <g translate(x=-15)><path/></g>` This would make future changes like "add more space between the line and the triangle" much easier.

`LadderVehicle` is the smallest thing that encompasses both the triangle and the text, and is the appropriate place to maintain the spacing between them.

arkadyan: Ok, I used the width of the triange for the positioning calculation.

ingridpierre:
@arkadyan looks good! 👍

PR #57, opened 2019-05-22
Make semaphore test script fail when any step fails

arkadyan:
Add a bash shebang.

Asana ticket: [🐞 Semaphore builds aren't failing when tests fail](https://app.asana.com/0/1112935048846093/1124224601646246)

A successful failure can be seen here: https://semaphoreci.com/mbta/skate/branches/mss-semaphore-test-failures/builds/2

### DIFF
--- a/semaphore/test.sh
+++ b/semaphore/test.sh
@@ -1,5 +1,7 @@
+#!/usr/bin/env bash
+
 set -e
 
-mix coveralls.json
-pushd assets && npm test && popd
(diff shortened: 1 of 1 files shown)


skyqrose:
Isn't that what `set -e` is supposed to do?

Do any of the other semaphore scripts need this treatment?

arkadyan:
@skyqrose I don't know, but apparently it didn't. If there's a different fix I'm not particular.

PR #59, opened 2019-05-28
Add an incoming box and display incoming vehicles there

arkadyan:
Asana ticket: [Add “INCOMING” box to bottom of route ladders](https://app.asana.com/0/1112935048846093/1118437179466250)

Screen Shot 2019-05-28 at 11 38 50

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -49,12 +49,6 @@
   width: 12em;
 }
 
-.m-incoming-box__header {
-  font-weight: bold;
-  margin-bottom: 0.25rem;
-  text-transform: uppercase;
-}
-
 .m-incoming-box__vehicle {
   display: flex;
   margin-bottom: 0.125rem;
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -50,6 +50,7 @@
 }
 
 .m-incoming-box__vehicle {
+  cursor: pointer;
   display: flex;
   margin-bottom: 0.125rem;
 }
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -36,30 +36,3 @@
     fill: $charcoal-grey;
   }
 }
-
-.m-incoming-box {
-  align-items: center;
-  background-color: $lightest-grey;
-  color: $dark;
-  display: flex;
-  flex-direction: column;
-  font-size: 0.8125rem;
-  min-height: 2.5rem;
-  padding: 0.5rem;
-  width: 12em;
-}
-
-.m-incoming-box__vehicle {
-  cursor: pointer;
-  display: flex;
-  margin-bottom: 0.125rem;
-}
-
-.m-incoming-box__vehicle-icon {
-  margin-right: 0.5rem;
-}
-
-.m-incoming-box__vehicle-icon-svg {
-  height: 14px;
-  width: 18px;
-}
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -39,7 +39,7 @@
 
 .m-incoming-box {
   align-items: center;
-  background-color: transparentize($color: $white, $amount: 0.5);
+  background-color: $lightest-grey;
   color: $dark;
   display: flex;
   flex-direction: column;
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -44,7 +44,7 @@
   display: flex;
   flex-direction: column;
   font-size: 0.8125rem;
-  min-height: 4.25rem;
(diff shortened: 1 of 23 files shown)


ingridpierre:
@arkadyan paraphrasing our convo:

for this pr
- let's shrink the pending box min height to ~2.5 rem
- let's lose the "INCOMING" label to reclaim a little space in the mockup
- Vehicles in the pending box should be clickable

For future work, the addition of the pending box to my mind necessitates us rethinking how we can cram more in there and get more to fit? I'd love to move this up a bit before we move on to other features, because I think it might hurt perception/adoption of our route ladders.
Some ideas/requirements we might discuss:
- What if we could pinch to zoom out on the right view and get more in view
 - or a +/- button to get more or less on screen.
- What if we could resize the heights of ladders using device size/orientation?
- in general how can we cram more into view if possible?

arkadyan:
First 2 tickets addressing the future work of information density:

- [Set height of ladders based on height of screen](https://app.asana.com/0/1112935048846093/1122859757950931) (existing)
- [Make all Incoming Boxes the same height](https://app.asana.com/0/1112935048846093/1125006647738997) (new)

arkadyan:
@ingridpierre I addressed those 3 items if you want to take another look.

assets/css/_route_ladder.scss
```diff
@@ -36,3 +36,30 @@
     fill: $charcoal-grey;
   }
 }
+
+.m-incoming-box {
+  align-items: center;
+  background-color: transparentize($color: $white, $amount: 0.5);
```
skyqrose: Does this need to be transparent? Why not an opaque gray?

assets/src/components/ladder.tsx
```diff
@@ -2,13 +2,28 @@ import React, { useContext } from "react"
...
 
+export enum VehicleDirection {
+  Up,
+  Down,
+}
+
+export const vehicleDirectionOnLadder = (
+  vehicle: Vehicle,
+  ladderDirection: LadderDirection
+) =>
```
skyqrose: I think a return type here would help.
```suggestion
): VehicleDirection =>
```

assets/src/components/routeLadder.tsx
```diff
@@ -14,6 +20,15 @@ interface Props {
   selectedVehicleId: VehicleId | undefined
 }
 
+const partitionVehiclesByRouteStatus = (vehicles: Vehicle[]): Vehicle[][] =>
+  vehicles.reduce(
+    ([onRouteVehicles, incomingVehicles]: Vehicle[][], vehicle: Vehicle) =>
+      vehicle.route_status === "on_route"
+        ? [[...onRouteVehicles, vehicle], incomingVehicles]
+        : [onRouteVehicles, [...incomingVehicles, vehicle]],
+    [[], []]
```
skyqrose: The type doesn't hint at which ones are on_route vs incoming, and I'll probably mix it up at some point. What if this returned `{on_route: [], incoming: []}`?

assets/css/_route_ladder.scss
```diff
@@ -36,3 +36,30 @@
     fill: $charcoal-grey;
   }
 }
+
+.m-incoming-box {
```
skyqrose: This name (`.m-incoming-box` instead of `m-route-ladder__incoming-box`) is a step towards making the incoming box its own module. What do you think about pulling it all the way out into its own `.tsx` and `.scss` files? It'd maintain the one-to-one-to-one relationship between `.tsx` files, `.scss` files, and css class names, and might make the reorganization in the asana task [Make all Incoming Boxes the same height](https://app.asana.com/0/1112935048846093/1125006647738997) easier.

assets/src/components/vehicleIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+  isSelected?: boolean
+  scale?: number
+}) => {
+  const selectedClass = isSelected ? "selected" : ""
+  const scaleAmount = scale || 1.0
+
+  return (
+    <g className="m-vehicle-icon" transform={`scale(${scaleAmount})`}>
+      <path
+        className={`m-vehicle-icon__triangle ${selectedClass}`}
```
skyqrose: The css only references the class `.m-vehicle-icon__triangle`, so the `.m-vehicle-icon` class could be removed.

assets/css/_route_ladder.scss
arkadyan: No, that's just the way it was specified but I removed it.

assets/src/components/vehicleIcon.tsx
arkadyan: That’s true, but I still find it useful as a descriptive name.

arkadyan:
@skyqrose thanks! I think I've addressed everything.

PR #65, opened 2019-06-03
Scale the viewport smaller for the tablets

arkadyan:
Asana ticket: [Scale the viewport smaller](https://app.asana.com/0/1112935048846093/1125366821749133)


ingridpierre:
@arkadyan: @skyqrose is correct, in chrome the tablet viewport is only about ~962 pixels wide instead of 1280, so that ends up being closer to .75. 

Can we try that tomorrow and see what it looks like? What you have here is definitely better already, so forgive me if we end up wanting to keep what's here.

arkadyan:
We decided to stick with 0.8 looking at the results on the tablet.

PR #67, opened 2019-06-05
Ingest Swiftly realtime vehicles data

arkadyan:
Asana ticket: [Ingest Swiftly vehicle position data](https://app.asana.com/0/1112935048846093/1124497435065298}

Remove concentrate input source, it is now redundant.

~~NB: We need to figure out our Swiftly API keys before merging this.~~


lib/gtfs/direction.ex
```diff
@@ -5,7 +5,7 @@ defmodule Gtfs.Direction do
 
   @type t :: %__MODULE__{
           route_id: Route.id(),
-          direction_id: id(),
+          direction_id: id() | nil,
```
skyqrose: When would this be `nil`?

lib/realtime/vehicle.ex
```diff
@@ -102,15 +120,16 @@ defmodule Realtime.Vehicle do
 
     trip_id = VehiclePosition.trip_id(vehicle_position)
     trip = trip_fn.(trip_id)
-    headsign = trip && trip.headsign
+    headsign = VehiclePosition.headsign(vehicle_position) || (trip && trip.headsign)
```
skyqrose: Where does swiftly get its vehicle headsign? Is it just from the trip? Is there any situation where we'd prefer swiftly's vehicle headsign or prefer the GTFS trip headsign?

lib/realtime/vehicle.ex
```diff
@@ -119,8 +138,9 @@ defmodule Realtime.Vehicle do
       timestamp: VehiclePosition.last_updated(vehicle_position),
       latitude: VehiclePosition.latitude(vehicle_position),
       longitude: VehiclePosition.longitude(vehicle_position),
-      direction_id: TripUpdate.direction_id(trip_update),
-      route_id: TripUpdate.route_id(trip_update),
+      direction_id:
+        VehiclePosition.direction_id(vehicle_position) || TripUpdate.direction_id(trip_update),
+      route_id: VehiclePosition.route_id(vehicle_position) || TripUpdate.route_id(trip_update),
```
skyqrose: Same question as above: Does Swiftly get its `route_id` and `direction_id` from the trip, and would we ever prefer Swiftly's vs GTFS's?

If we always use GTFS's, we don't need to parse this field from the swiftly response.

test/concentrate/parser/swiftly_realtime_vehicles_test.exs
```diff
@@ -0,0 +1,92 @@
+defmodule Concentrate.Parser.SwiftlyRealtimeVehiclesTest do
+  use ExUnit.Case, async: true
+
+  import Concentrate.TestHelpers
+
+  alias Concentrate.VehiclePosition
+  alias Concentrate.Parser.SwiftlyRealtimeVehicles
+
+  describe "parse/1" do
+    test "parsing an enhanced VehiclePositions JSON file returns only VehiclePosition or TripUpdate structs" do
```
skyqrose: The test name includes `TripUpdate`, but the assert below doesn't.

skyqrose:
Updated AWS config with the key, so that's no longer a blocker.

assets/src/skate.d.ts
```diff
@@ -45,6 +45,15 @@ export interface Vehicle {
...
   operator_name: string
+  speed: number
+  block_id: string
+  headway_secs: number
+  previous_vehicle_id: string
+  previous_vehicle_schedule_adherence_secs: number
+  previous_vehicle_schedule_adherence_string: string
+  schedule_adherence_secs: number
+  schedule_adherence_string: string
+  scheduled_headway_secs: number
```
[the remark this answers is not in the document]
arkadyan: I don't know for sure yet, but I think we'll end up using at least most of these.

We could make all these optional, same for the backend. For whatever reason we got into a habit of enforcing the presence of everything, but I don't have strong feelings about that.

lib/gtfs/direction.ex
arkadyan: It wouldn't in practice, but this value comes from `id_from_string/1`, which (for other inputs) might return a nil so dialyzer wants to see one here too.

skyqrose: Looks like it's just [this use](https://github.com/mbta/skate/pull/67/files#diff-559bd7d15af65384238b8072a36fafdeR45) in `concentrate/parser/swiftly_realtime_vehicles.ex` that needs the `nil` check in `id_from_string`:
```elixir
def decode_vehicle(vehicle_data) do
  VehiclePosition.new(
    direction_id: vehicle_data |> Map.get("directionId") |> Direction.id_from_string(),
```

If the nil check is handled in that function before the call to `id_from_string` instead of adding a `nil` case to `id_from_string`, I think you could avoid needing `id_from_string` to return nil, so this wouldn't be needed.

lib/concentrate/parser/swiftly_realtime_vehicles.ex
```diff
@@ -0,0 +1,71 @@
...
+      schedule_adherence_string: Map.get(vehicle_data, "schAdhStr"),
+      scheduled_headway_secs: Map.get(vehicle_data, "scheduledHeadwaySecs")
+    )
+  end
+
+  @spec operator_details(String.t() | nil) :: {String.t() | nil, String.t() | nil}
+  defp operator_details(nil), do: {nil, nil}
+
+  defp operator_details(operator_string) do
+    if String.match?(operator_string, ~r/ - /) do
```
paulswartz: You're basically doing the split twice here:
```elixir
case String.split(operator_string, " - ") do
    [name, id] -> {name, id}
    _ -> :no_operator
end
```

PR #68, opened 2019-06-05
Limit route direction data to only the names

arkadyan:
Asana ticket: [Refactor Route directions data structure](https://app.asana.com/0/1112935048846093/1125751680948291)

### DIFF
--- a/test/gtfs_test.exs
+++ b/test/gtfs_test.exs
@@ -21,17 +21,9 @@ defmodule GtfsTest do
       assert Gtfs.all_routes(pid) == [
                %Route{
                  id: "39",
-                 directions: %{
-                   0 => %Gtfs.Direction{
-                     direction_id: 0,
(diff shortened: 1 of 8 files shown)


test/gtfs_test.exs
```diff
@@ -47,7 +39,13 @@ defmodule GtfsTest do
           ]
         })
 
-      refute Enum.member?(Gtfs.all_routes(pid), %Route{id: "Red", directions: %{}})
+      refute Enum.member?(Gtfs.all_routes(pid), %Route{
+               id: "Red",
+               direction_names: %{
+                 0 => "Outbound",
+                 1 => "Inbound"
```
skyqrose: `Red`'s direction names would be `North` and `South`. Could you refute that the id `Red` exists anywhere rather than trying to make a struct that would exactly match?

PR #69, opened 2019-06-06
implement UI kit

skyqrose:
Screen Shot 2019-06-06 at 12 35 21

Asana Task: [Use styles from UI kit](https://app.asana.com/0/1112935048846093/1125059554572559)
[UI Kit](https://app.zeplin.io/project/5caf86702bbb91be743b94d9/screen/5ce836b69544091df5deb871)

Includes hover, focus, and selected styles. Those are best seen by running the app, rather than screenshots.


ingridpierre:
image
Everything is looking really fab, just one thing -  we want the active/selected vehicle to have the mint border and pill background behind the vehicle label. For the hover/focus state, we want the same thing but with that new lightest mint color.

PR #70, opened 2019-06-10
Offset overlapping vehicles horizontally

arkadyan:
Asana ticket: [Offset overlapping vehicles horizontally](https://app.asana.com/0/1112935048846093/1123838037403440)

### DIFF
--- a/assets/src/components/ladder.tsx
+++ b/assets/src/components/ladder.tsx
@@ -21,9 +21,6 @@ export enum LadderDirection {
   OneToZero,
 }
 
-export const widthOfVehicleGroup = 32
-export const heightOfVehicleGroup = 34
-
 export const flipLadderDirection = (
   ladderDirection: LadderDirection
 ): LadderDirection =>
@@ -61,7 +58,7 @@ const Ladder = ({
     timepointStatusY
   )
 
-  const width = 120 + 2 * widthOfLanes * 1.2
+  const width = 120 + 2 * widthOfLanes
   // (0, 0) is in the center of the first timepoint
   const viewBox = [-width / 2, -MARGIN_TOP_BOTTOM, width, HEIGHT].join(" ")
 
--- a/assets/src/components/ladder.tsx
+++ b/assets/src/components/ladder.tsx
@@ -1,8 +1,19 @@
 import React, { useContext } from "react"
 import DispatchContext from "../contexts/dispatchContext"
+import {
+  ladderVehiclePositionsFromVehicles,
+  VehicleDirection,
+} from "../models/ladderVehiclePosition"
 import { Timepoint, Vehicle, VehicleId, VehicleTimepointStatus } from "../skate"
 import { selectVehicle } from "../state"
 import { Orientation, Size, VehicleIconSvgNode } from "./vehicleIcon"
+
+export interface Props {
+  timepoints: Timepoint[]
+  vehicles: Vehicle[]
+  ladderDirection: LadderDirection
+  selectedVehicleId: VehicleId | undefined
+}
 
 // Timepoints come from the API in the ZeroToOne direction
 export enum LadderDirection {
@@ -10,19 +21,8 @@ export enum LadderDirection {
   OneToZero,
 }
 
-export enum VehicleDirection {
-  Up,
-  Down,
-}
-
-export const vehicleDirectionOnLadder = (
-  vehicle: Vehicle,
-  ladderDirection: LadderDirection
-): VehicleDirection =>
-  (vehicle.directionId === 1) ===
-  (ladderDirection === LadderDirection.ZeroToOne)
-    ? VehicleDirection.Down
-    : VehicleDirection.Up
+export const widthOfVehicleGroup = 32
+export const heightOfVehicleGroup = 34
 
 export const flipLadderDirection = (
   ladderDirection: LadderDirection
@@ -32,16 +32,8 @@ export const flipLadderDirection = (
     : LadderDirection.ZeroToOne
 
 const HEIGHT = 500
-const WIDTH = 180
 const CENTER_TO_LINE = 40 // x-distance between the center of the ladder and the center of the line
 const MARGIN_TOP_BOTTOM = 40 // space between the top of the route and the top of the viewbox
-
-export interface Props {
-  timepoints: Timepoint[]
-  vehicles: Vehicle[]
-  ladderDirection: LadderDirection
-  selectedVehicleId: VehicleId | undefined
-}
 
 const Ladder = ({
   timepoints,
@@ -49,8 +41,7 @@ const Ladder = ({
   ladderDirection,
   selectedVehicleId,
 }: Props) => {
-  // (0, 0) is in the center of the first timepoint
-  const viewBox = [-WIDTH / 2, -MARGIN_TOP_BOTTOM, WIDTH, HEIGHT].join(" ")
+  const dispatch = useContext(DispatchContext)
   const timepointSpacingY =
     (HEIGHT - MARGIN_TOP_BOTTOM * 2) / (timepoints.length - 1)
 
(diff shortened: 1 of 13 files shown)


arkadyan:
Screen Shot 2019-06-10 at 11 41 49

assets/src/models/ladderVehiclePosition.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  const gappedLanes: number[] = range(occupiedLanes.length).filter(
+    lane => !occupiedLanes.includes(lane)
+  )
+  if (gappedLanes && gappedLanes.length) {
+    return gappedLanes[0]
+  }
+
+  return Math.max(...occupiedLanes) + 1
```
skyqrose: Could there be a simpler way to implement this by sorting the input and walking through it til you find the first gap?

assets/src/models/ladderVehiclePosition.ts
```diff
@@ -0,0 +1,156 @@
...
+export const byDirectionAndY = (
+  a: LadderVehiclePosition,
+  b: LadderVehiclePosition
+): number => {
+  if (a.vehicleDirection === VehicleDirection.Down) {
+    if (b.vehicleDirection === VehicleDirection.Up) {
+      return -1
+    } else {
+      if (a.y > b.y) {
+        return -1
```
skyqrose: This comparison is a little hard to follow. It could use a quick human description "Down is less than Up. If they're the same direction the vehicle behind is less than the vehicle in front" or something similar (idk if I got the directions right)

What impact does the order of the sorting have? If it's important, that should probably get noted too.

assets/src/models/ladderVehiclePosition.ts
```diff
@@ -0,0 +1,156 @@
...
+        )
+        const lane = firstOpenLane(occupiedLanes)
+
+        return [
+          {
+            ...ladderVehiclePosition,
+            lane,
+          },
+          ...previousLadderVehiclePositions,
+        ]
```
skyqrose: Does this reduce flip the order of the input? Is it important if it does or not?

assets/src/models/ladderVehiclePosition.ts
```diff
@@ -0,0 +1,156 @@
...
+    areOverlapping(previous, ladderVehiclePosition)
+  )
+
+export const areOverlapping = (
+  a: LadderVehiclePosition,
+  b: LadderVehiclePosition
+): boolean => {
+  return (
+    a.vehicleDirection === b.vehicleDirection &&
+    areOverlappingRanges(yRange(a), yRange(b))
```
skyqrose: This could just be `abs(a.y - b.y) < minDistanceBetween`, and you wouldn't need the range stuff.

assets/src/models/ladderVehiclePosition.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export const xInLaneWithOffset = (ladderVehicleHorizontalOffset: number) => ({
+  x,
+  vehicleDirection,
+  lane,
+}: LadderVehiclePosition): number => {
+  const laneOffset = (lane || 0) * ladderVehicleHorizontalOffset
+  return vehicleDirection === VehicleDirection.Up
+    ? x + laneOffset
+    : x - laneOffset
```
skyqrose: I feel like this could be simpler. Like, maybe `x` doesn't need to be part of `LadderVehiclePosition`, and you could do all the x calculations at once based on the `lane` and `vehicleDirection`.

assets/src/components/ladder.tsx
```diff
@@ -60,18 +64,77 @@ const Ladder = ({
...
+        timepointStatusY(
+          vehicle.timepoint_status,
+          timepoints,
+          timepointSpacingY,
+          vehicleDirection
+        ) || -10
+
+      return {
+        vehicle,
+        vehicleDirection,
```
skyqrose: There's some interesting algorithm stuff, but it's pretty tied up in the vehicle stuff. Do you think it'd be possible to split apart? Maybe something like
`putIntoLanes = (input: [{y: number, payload: whatever}], minSpaceBetween: number): [{y: number, payload: whatever, lane: number}]`
Maybe you could partition the up and down vehicle and call it separately on the two sides so that the algorithmic stuff doesn't need to think about `VehicleDirection` at all? I think it'd end up being easier to reason about and test if the lane algorithm didn't need to think about vehicle stuff, and if all the vehicle logic didn't have any lane algorithm mixed in.

assets/src/components/ladderVehicle.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  return (
+    <g
+      className={`m-ladder__vehicle ${selectedClass}`}
+      transform={`translate(${xInLane(ladderVehiclePosition)},${y -
+        centerOfVehicleGroupY}) rotate(${rotation},${centerOfVehicleGroupX},${centerOfVehicleGroupY})`}
+      onClick={() => dispatch(selectVehicle(vehicle.id))}
+    >
+      <VehicleIcon scale={0.625} />
+      <VehicleLabel vehicle={vehicle} rotation={rotation} x={0} y={26} />
```
skyqrose: #72 pulls the label inside the `VehicleIcon`, so this will change quite a bit.

assets/src/components/ladder.tsx
```diff
@@ -1,56 +1,60 @@
...
+  const allLanes: number[] = ladderVehiclePositions.map(
+    ({ lane }: LadderVehiclePosition) => lane || 0
+  )
+  const furthestLane = allLanes.length ? Math.max(...allLanes) : 0
+  return furthestLane + 1
+}
+
+const ladderVehicleHorizontalOffsetForLanes = (maxOccupiedLane: number) => {
+  const cappedLanes = Math.min(maxOccupiedLane, 5)
+  return ((8 - cappedLanes) / 8) * widthOfVehicleGroup
```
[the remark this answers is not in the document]
arkadyan: This is defining the offset for each lane based on the max number of vehicles wide this ladder gets in order to condense the vehicles more as there are more of them across. So, if there are 2 vehicles across it will space them at a wider 75% of the icon width, while at 5 vehicles across it'll be tighter at 37.5%. It is capped at that condensed to ensure you can still click on the edge of any icon.

assets/src/models/ladderVehiclePosition.ts
arkadyan: That was actually my first implementation. It ended up being much less simple because their might not be a gap, or might not be any occuied lanes. This is much more straightforward and these list sizes are never going to be that large.

assets/src/models/ladderVehiclePosition.ts
arkadyan: It does, but it is not important - the sorting only mattered for the logic in that reduce function, nowhere externally.

assets/src/models/ladderVehiclePosition.ts
arkadyan: Nice!

assets/src/components/ladder.tsx
arkadyan: I think if I understand what you are describing it gets to why I wanted to pull out a view model in the first place, to get all the logic out of the components. I think the problem is that this function here doesn't belong here, it belongs in `ladderVehiclePosition`. What if we passed `vehicles` and `timepointStatusY` into a function in `ladderVehiclePosition` that is basically this function, but moving the `x` calculation to aftert we figure out the lanes? I can do that work to show it more concretely, but does that sound like it would be moving in a better direction?

assets/src/models/ladderVehiclePosition.ts
skyqrose: What about:
```
const sortedOccupiedLanes = occupiedLanes.slice().sort()
for (var i = 0; i < sortedOccupiedLanes.length; i++) { 
  if (sortedOccupiedLanes[i] !== i) {
    return i
  }
}
return sortedOccupiedLanes.length
```
It's not very functional, but I think it's still simpler, and should cover if the list is empty (do the loop 0 times, then return 0), and full (do the loop n times, then return n).

assets/src/components/ladder.tsx
skyqrose: I think the view model helps a lot, and moving the `x` calculation to after we figure out the lanes would make that even better.

I think we could also separate the algorithmic part of `putIntoLanes` from the view model. Like, `putIntoLanes`, `overlappingPreviousVehicles`, `firstOpenLane`, and friends don't actually need any information about the vehicle, just its y, so could be pulled into a separate file.

There's a good abstraction boundary there that would make it easier to see the view model and how it's made and used without having to wade through any of the algorithmic stuff.

assets/src/models/ladderVehiclePosition.ts
arkadyan: That works.

assets/src/components/ladder.tsx
arkadyan: Take a fresh look: I extracted all that logic out of the Ladder component and made the internal logic much more pure about what data it requires (with internal interfaces to demonstrate that).

arkadyan: One small additional update: I curried some timepointStatusY variables to avoid passing them around unnecessarily.

arkadyan:
@skyqrose I don't see why this isn't worth a final review at this point. I haven't seen the new ladder width calculation given 4 or 5 vehicles yet, but that's also trivial to tweak later if we discover any issues.

assets/src/components/ladder.tsx
```diff
@@ -1,28 +1,28 @@
...
-export const vehicleDirectionOnLadder = (
-  vehicle: Vehicle,
-  ladderDirection: LadderDirection
-): VehicleDirection =>
-  (vehicle.directionId === 1) ===
-  (ladderDirection === LadderDirection.ZeroToOne)
-    ? VehicleDirection.Down
-    : VehicleDirection.Up
+export const widthOfVehicleGroup = 32
+export const heightOfVehicleGroup = 34
```
skyqrose: These are only used in `ladderVehicle.ts`, and could be moved there so they can be private.

assets/src/models/ladderVehicle.ts
```diff
@@ -0,0 +1,239 @@
...
+  )
+
+  const vehiclesInLane: VehicleInLane[] = putIntoLanes(vehiclesOnLadder)
+
+  const maxOccupiedLane: number = numOccupiedLanes(vehiclesInLane)
+  const ladderVehicleHorizontalOffset: number = horizontalOffsetForLanes(
+    maxOccupiedLane
+  )
+  const xInLane = xInLaneWithOffset(ladderVehicleHorizontalOffset)
+  // const widthOfLanes = maxOccupiedLane * ladderVehicleHorizontalOffset
```
skyqrose: Dead code

assets/src/components/ladder.tsx
```diff
@@ -60,18 +51,69 @@ const Ladder = ({
...
+    timepoints,
+    timepointSpacingY
+  )
+  const { ladderVehicles, widthOfLanes } = ladderVehiclesFromVehicles(
+    vehicles,
+    ladderDirection,
+    timepointStatusY
+  )
+
+  const width = 120 + 2 * widthOfLanes * 1.2
```
skyqrose: With the new more descriptive width calculations, this `* 1.2` scaling shouldn't be necessary, assuming we want constant sized margins.

assets/src/components/ladder.tsx
```diff
@@ -60,18 +51,69 @@ const Ladder = ({
...
-          key={vehicle.id}
-          vehicle={vehicle}
-          timepoints={orderedTimepoints}
-          timepointSpacingY={timepointSpacingY}
-          ladderDirection={ladderDirection}
-          selectedVehicleId={selectedVehicleId}
-        />
-      ))}
+    <svg className="m-ladder" height={HEIGHT} width={width} viewBox={viewBox}>
+      {ladderVehicles.map(ladderVehicle => {
```
skyqrose: This map function is pretty long, and makes it harder to read the list of everything drawn in the SVG. It could be broken out.

assets/src/models/ladderVehicle.ts
```diff
@@ -0,0 +1,239 @@
...
+          ...previousLadderVehiclePositions,
+        ]
+      },
+      []
+    )
+
+const numOccupiedLanes = (vehicles: InLane[]): number => {
+  const allLanes: number[] = vehicles.map(({ lane }) => lane || 0)
+  const furthestLane = allLanes.length ? Math.max(...allLanes) : 0
+  return furthestLane + 1
```
skyqrose: `numOccupiedLanes` and `maxOccupiedLanes` get conflated. Since one of the uses subtracts 1 later, I think this function should get renamed `maxOccupiedLane` and the `+ 1` here removed.

The `- 1` when calculated `widthOfLanes` can be removed then. (The `horizontalOffsetForLanes` also uses `maxOccupiedLane`, and might need to tweak its calculations.)

assets/tests/models/ladderVehicle.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    ]
+
+    const result = putIntoLanes(original)
+
+    expect(result).toEqual(expected)
+  })
+})
+
+describe("byDirectionAndY", () => {
+  test("sorts the arry by direction, and then by y value", () => {
```
skyqrose: typo `arry` -> `array`

assets/src/components/ladder.tsx
arkadyan: Will be fixed in a future PR

PR #71, opened 2019-06-10
Add schedule adherence info to the Vehicle Properties

arkadyan:
[Add the schedule adherence string to the Vehicle Properties](https://app.asana.com/0/1112935048846093/1125952600534844)


arkadyan:
Screen Shot 2019-06-10 at 13 26 17

assets/src/hooks/useVehicles.ts
```diff
@@ -1,6 +1,56 @@
...
+  ScheduleAdherenceStatus,
+  Vehicle,
+  VehicleRouteStatus,
+  VehiclesByRouteId,
+  VehicleStatus,
+  VehicleStopStatus,
+  VehicleTimepointStatus,
+} from "../skate.d"
+
+export interface VehicleData {
```
skyqrose: Doesn't need to be exported.

assets/src/hooks/useVehicles.ts
```diff
@@ -97,12 +147,89 @@ const reducer = (state: State, action: Action): State => {
...
+    }
+  } else {
+    return null
+  }
+}
+
+const vehicleFromData = (vehicleData: VehicleData): Vehicle => ({
+  id: vehicleData.id,
+  label: vehicleData.label,
+  runId: vehicleData.run_id,
```
skyqrose: Oh thank you for doing this. Having `snake_case` scattered throughout the javascript was bugging me.

assets/src/skate.d.ts
```diff
@@ -15,6 +15,8 @@ export interface Route {
 
 export type RouteId = string
 
+export type ScheduleAdherenceStatus = "on-time" | "early" | "late"
```
skyqrose: When would you use a string constant like this vs an enum?

assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -10,6 +10,66 @@ interface Props {
...
+    scheduleAdherenceStatus === "late"
+  ) {
+    return earlyOrLateLabel(vehicle)
+  } else {
+    return ""
+  }
+}
+
+const ScheduleAdherenceLabel = ({ vehicle }: { vehicle: Vehicle }) => (
+  <div className="m-vehicle-properties-panel__schedule-adherencbe-lael">
```
skyqrose: typo in the class name? but this class isn't used anyway, so you can probably get rid of this layer of `<div>`.

assets/css/_vehicle_properties_panel.scss
```diff
@@ -37,6 +37,34 @@
...
+  }
+  &.late {
+    fill: $color-vehicle-late;
+  }
+}
+
+.m-vehicle-properties-panel__schedule-adherence-status-string {
+  font-weight: bold;
+  margin-right: 0.3rem;
+  text-transform: uppercase;
```
skyqrose: Could use `@include font-emph` here to avoid rewriting these text styling properties and stay consistent with the ui kit.  It would end up being 14px instead of 12px (the label would also need to change to `font-body`).

Here's how it'd look using `@include font-emph`.  @ingridpierre thoughts? 
Screen Shot 2019-06-10 at 16 47 34

assets/css/_vehicle_properties_panel.scss
```diff
@@ -37,6 +37,34 @@
...
+.m-vehicle-properties-panel__schedule-adherence {
+  @include font-small;
+  color: $color-font-dark;
+  display: flex;
+  margin-top: 0.7rem;
+}
+
+.m-vehicle-properties-panel__schedule-adherence-status-icon {
+  margin-right: 0.4rem;
+  stroke: none;
```
skyqrose: `stroke: none` isn't needed since it's the default.
Above, `color: $color-font-dark` isn't needed since it's inherited.

ingridpierre:
Can't tell much with the swiftly outage, but a little thing I noticed based on a screenshot above is the inclusion of a "-" for early buses. For this I think just "X min early" and "X min late" is sufficient without the +/- symbol. 

We got some feedback and found that the plus and minus is a bit confusing saying "late" or "early" alone is sufficient.

assets/src/skate.d.ts
arkadyan: I wish enums worked better, but apparently in TS exporting an enum doesn't work properly. (At least in some cases? If you try to export an enum in this file the TS compile fails.) It seems like the advice is just to use strings like this.

assets/css/_vehicle_properties_panel.scss
arkadyan: Removed the stroke. The color is needed, it falls back to black without it.

arkadyan:
@ingridpierre 
Screen Shot 2019-06-11 at 14 03 26

ingridpierre:
Can we also add the "(X min late/early)" to the on-time vehicles, but with the adherence minutes labeled in grey:

image

For this on-time state it's nice to see a little of the adherence info, but i've put it in grey since it's not truly late, just there for informational purposes. As I was looking at the ladder, vehicles kept bouncing in and out of different adherence buckets. It would be easier to track how a vehicle is transitioning in and out of early/on-time/late if we had that label persists through all three states.

arkadyan:
@ingridpierre 
Screen Shot 2019-06-11 at 15 17 53

arkadyan:
@skyqrose and @ingridpierre I think this is ready for a final review, please.

ingridpierre:
Love it, lgtm! 👍

assets/css/_vehicle_properties_panel.scss
skyqrose: Oh, oops, the color should be specified on `body` in `app.scss`, and then it won't be needed here. All the text is black right now.

assets/css/_vehicle_properties_panel.scss
skyqrose: After talking in person, it should stay small.

assets/css/_vehicle_properties_panel.scss
arkadyan: Ah yeah, that's better—moved.

PR #72, opened 2019-06-10
Triangles: variant, bugfix, vehicle properties panel

skyqrose:
Asana Tasks:
[Put variant number in the triangle](https://app.asana.com/0/1112935048846093/1123217669646641)
[🐛vehicle triangle in incoming box is cut off](https://app.asana.com/0/1112935048846093/1125794829210438)
[Add triangle to the Vehicle Properties Panel header](https://app.asana.com/0/1112935048846093/1124847099574986)

Screen Shot 2019-06-10 at 14 51 57
Screen Shot 2019-06-10 at 14 52 31

This PR changes a lot about how we make vehicle icons.
* Pulls the label and (new) variant into the VehicleIcon component.
* Handles up, down, left, and right-facing vehicles. We don't have any horizontal vehicles yet, but the label/variant placement rules are pretty different for them so I wanted to make sure I didn't make that future work more difficult.
* Has separate interfaces for returning an svg node (for the ladder) vs an html node (for the incoming box and VPP).
* The html node calculates the viewbox and width/height needed to tightly fit the icon. (Missing the viewbox and hardcoding the width+height was the source of the clipping bug)

Here's all the possible orientations and directions. Top two rows are as html nodes and bottom rows are as svg nodes. The line in the svg goes through the centers of the triangles, which is where the schedule line is drawn from, and the point that's used to position the triangle on the ladder. E.g. if a vehicle is at a timepoint, the timepoint will be level with that center of the triangle.
Screen Shot 2019-06-10 at 14 48 58
And with the backgrounds highlighted:
Screen Shot 2019-06-10 at 14 49 25

### DIFF
--- a/assets/src/components/vehicleIcon.tsx
+++ b/assets/src/components/vehicleIcon.tsx
@@ -1,16 +1,302 @@
 import React from "react"
 
-const VehicleIcon = ({ scale }: { scale?: number }) => {
-  const scaleAmount = scale || 1.0
+export enum Orientation {
+  Up,
+  Down,
+  Left,
+  Right,
+}
 
+export enum Size {
+  Small,
+  Medium,
+  Large,
+}
+
+export interface Props {
+  size: Size
+  orientation: Orientation
+  label?: string
+  variant?: string | null
+}
+
+/*
+Size of the triangle for calculation of the viewbox and positioning the label
+Sizes are before scaling, relative to the center of the triangle
+*/
+const X_CENTER_TO_SIDE = 22
+const Y_CENTER_TO_POINT = 16
+const Y_CENTER_TO_BASE = 20
+
+export const VehicleIcon = (props: Props) => {
+  const { left, top, width, height } = viewBox(props)
   return (
-    <g className="m-vehicle-icon" transform={`scale(${scaleAmount})`}>
-      <path
-        className={"m-vehicle-icon__triangle"}
-        d="m27.34 9.46 16.84 24.54a4.06 4.06 0 0 1 -1 5.64 4.11 4.11 0 0 1 -2.3.71h-33.72a4.06 4.06 0 0 1 -4.06-4.11 4 4 0 0 1 .72-2.24l16.84-24.54a4.05 4.05 0 0 1 5.64-1.05 4 4 0 0 1 1.04 1.05z"
-      />
+    <svg
+      style={{ width, height }}
+      viewBox={`${left} ${top} ${width} ${height}`}
+    >
+      <VehicleIconSvgNode {...props} />
+    </svg>
+  )
+}
+
+const viewBox = ({
+  size,
+  orientation,
+  label,
+}: Props): { left: number; top: number; width: number; height: number } => {
+  // shrink the viewbox to fit around the triangle and label
+  const scale = scaleForSize(size)
+  const labelBgWidth = label ? labelBackgroundWidth(size) : 0
+  const labelBgHeight = label ? labelBackgroundHeight(size) : 0
+  let left = 0
+  let right = 0
+  let top = 0
+  let bottom = 0
+  switch (orientation) {
+    case Orientation.Up:
+      left = -scale * X_CENTER_TO_SIDE
+      top = -scale * Y_CENTER_TO_POINT
+      right = scale * X_CENTER_TO_SIDE
+      bottom = scale * Y_CENTER_TO_BASE + labelBgHeight
+      break
+    case Orientation.Down:
+      left = -scale * X_CENTER_TO_SIDE
+      top = -scale * Y_CENTER_TO_BASE - labelBgHeight
+      right = scale * X_CENTER_TO_SIDE
+      bottom = scale * Y_CENTER_TO_POINT
+      break
+    case Orientation.Left:
+      left = -scale * Y_CENTER_TO_POINT
+      top = -scale * X_CENTER_TO_SIDE
+      right = scale * Y_CENTER_TO_BASE
+      bottom = scale * X_CENTER_TO_SIDE + labelBgHeight
+      break
+    case Orientation.Right:
+      left = -scale * Y_CENTER_TO_BASE
+      top = -scale * X_CENTER_TO_SIDE
+      right = scale * Y_CENTER_TO_POINT
+      bottom = scale * X_CENTER_TO_SIDE + labelBgHeight
+      break
+  }
+  left = Math.min(left, -labelBgWidth / 2)
+  right = Math.max(right, labelBgWidth / 2)
+  const width = right - left
+  const height = bottom - top
+  return { left, top, width, height }
+}
+
+export const VehicleIconSvgNode = ({
+  size,
+  orientation,
+  label,
+  variant,
+}: Props) => {
+  return (
+    <g className={`m-vehicle-icon m-vehicle-icon${sizeClassSuffix(size)}`}>
+      {label ? (
+        <Label size={size} orientation={orientation} label={label} />
+      ) : null}
+      <Triangle size={size} orientation={orientation} />
+      {variant && variant !== "_" ? (
+        <Variant size={size} orientation={orientation} variant={variant} />
+      ) : null}
     </g>
   )
+}
+
+const Triangle = ({
+  size,
+  orientation,
+}: {
+  size: Size
+  orientation: Orientation
+}) => {
+  const scale = scaleForSize(size)
+  const rotation = rotationForOrientation(orientation)
+  return (
+    <path
+      className="m-vehicle-icon__triangle"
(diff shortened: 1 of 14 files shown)


assets/src/components/vehicleIcon.tsx
```diff
@@ -1,16 +1,299 @@
...
+  Small,
+  Medium,
+  Large,
+}
+
+export interface Props {
+  size: Size
+  orientation: Orientation
+  label?: string
+  variant?: string | null
```
arkadyan: Doesn't the `?` mean you don't need the `| null`?

assets/src/components/vehicleIcon.tsx
```diff
@@ -1,16 +1,299 @@
...
+export const VehicleIcon = ({ size, orientation, label, variant }: Props) => {
+  // shrink the viewbox to fit around the triangle and label
+  const scale = scaleForSize(size)
+  const labelBgWidth = label ? labelBackgroundWidth(size) : 0
+  const labelBgHeight = label ? labelBackgroundHeight(size) : 0
+  let left = 0
+  let right = 0
+  let top = 0
+  let bottom = 0
+  switch (orientation) {
```
arkadyan: If you broke this logic out into component functions could you get rid of the mutable variables?

assets/src/components/vehicleIcon.tsx
```diff
@@ -1,16 +1,299 @@
...
+        size={size}
+        orientation={orientation}
+        label={label}
+        variant={variant}
       />
+    </svg>
+  )
+}
+
+export const VehicleIconSvg = ({
```
arkadyan: Personally `VehicleIcon` returning an `<svg>` and `VehicleIconSvg` returning a `<g>` is not at all what I expected from the names.

assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -20,7 +21,14 @@ const Header = ({
   hideMe: () => void
 }) => (
   <div className="m-vehicle-properties-panel__header">
-    <div className="m-vehicle-properties-panel__label">{vehicle.label}</div>
+    <div className="m-vehicle-properties-panel__label">
+      <VehicleIcon
+        size={Size.Large}
+        orientation={Orientation.Up}
+        label={vehicle.label}
```
arkadyan: It feels like the logic of formatting the label and variant is kind of split between the callers and the component (e.g. the component overrides with a `variant !== "_"` check). What would you think about passing in a vehicle (so the component handles what it means to dispaly a label and variant based on the data), and then a `showLabel` boolean flag or something like that?

assets/src/components/vehicleIcon.tsx
```diff
@@ -1,16 +1,299 @@
...
+const Label = ({
+  size,
+  orientation,
+  label,
+}: {
+  size: Size
+  orientation: Orientation
+  label: string
+}) => {
+  const scale = scaleForSize(size)
```
arkadyan: There's a whole lot of disperse logic around each of these components sizing themselves differently. Do you think it might be possible to compose the entire component at a base size, and then scale that once?

assets/src/components/vehiclePropertiesPanel.tsx
arkadyan: Whether or not that is a good idea, one thing I just thought about that might play into this consideration is that we'll soon be coloring the icon based on the vehicle status (https://app.asana.com/0/1112935048846093/1122642346801331)

assets/src/components/vehicleIcon.tsx
skyqrose: The `?` is just for `undefined`. We can get either an `undefined` if you don't pass a `<VehicleIcon variant={...}` property or a `null` if you do `<VehicleIcon variant={vehicle.via_variant}` but the `via_variant` is `null`.

assets/src/components/vehicleIcon.tsx
skyqrose: Ooh, I like that idea. It'd be much easier.

I'm a little worried that the text sizing won't look good if it's just scaled. E.g. at smaller font sizes, you want the variant letter to take up a bigger fraction of the triangle to be readable, but at big sizes you can have spacious margins around the variant letter.

I'll give it a shot and see how it looks.

assets/src/components/vehicleIcon.tsx
skyqrose: Yeah, I see that. For the caller, the important distinction is whether the component it's using should be part of an svg or whether it can be treated as an html element. Any ideas for names that would make that clear without feeling like the name contradicts the implementation?

assets/src/components/vehiclePropertiesPanel.tsx
skyqrose: Yeah sure. Everywhere that uses the icon has a vehicle associated with it, and all of them draw the variant. inside.

skyqrose: Except ghost buses maybe? Will drawing them become much harder later if we make this assumption that you always have a vehicle?

assets/src/components/vehicleIcon.tsx
skyqrose: You could, and I do that for some functions. For the ones that are left, I felt like the logic is pretty inseparable from the rest of the function.

I guess for this I could pull out `const viewBox(props: Props): {left: number, top: number, width: number, height: number}`. It won't get rid of the `let` but it'll help a bit.

assets/src/components/vehicleIcon.tsx
arkadyan: What about `VehicleIconSvgGroup` and then either `VehicleIconSvg` or just `VehicleIcon`? Just spitballing…

assets/src/components/vehicleIcon.tsx
arkadyan: I see, thanks. I guess the potential problem is via_variant being null instead of undefined. (But not a problem to be addressed right now.)

assets/src/components/vehiclePropertiesPanel.tsx
arkadyan: That's good to consider. The component could always accept an optional vehicle and handle it.

assets/src/components/vehicleIcon.tsx
arkadyan: Yeah, if the let is inside a helper function I generally find that less worrisome.

ingridpierre:
image
very, very small thing. Can we make the variant number in the vehicle properties arrow a lighter weight, maybe 500 or 400? 

It doesn't look as thin as the variant numbers on the vehicles on the route ladders, and the extra emphasis is making too many things bold in the top of the VP panel. When all things are bold it's like nothing is bold.

ingridpierre:
paraphrasing our conversation earlier here:

- adjusting the alignment of the pending vehicles is possible, but not for this ticket
- removing vehicle #s if it is the common pattern (typicality 1)  is also possible, but for another ticket

assets/src/components/vehicleIcon.tsx
skyqrose: Screen Shot 2019-06-11 at 15 47 57

That's a lot better than I expected.

skyqrose: The original for comparison:
Screen Shot 2019-06-10 at 14 49 25

I'm leaning towards no, because the label for the medium size is too big for the ladder and the label for the small ones is too small (we'll use it for turning vehicles)

@ingridpierre what do you think?

skyqrose: Here it is with custom text sizes for each size and the big text unbolded.

Screen Shot 2019-06-12 at 10 17 28

After talking with Ingrid and trying it out on a tablet, the tiny font sizes weren't quite readable with the scaling, so I'll stay with the manually specified sizes.

assets/src/components/vehicleIcon.tsx
skyqrose: I like `VehicleIcon` for the html version. At that point the fact that it's made with an svg can be an opaque implementation detail.

For the svg node, is the fact that it's a `<g>` important to expose? What about `...SvgNode` or `...SvgElement`?

arkadyan: That makes sense. Personally I might vote for `...SvgNode` so there's no confusion about it being "an <svg> element", but whatever you prefer. Naming is hard…

assets/src/components/vehiclePropertiesPanel.tsx
skyqrose: There will also be `SW-OFF` vehicles where the label used isn't the vehicle id or the run, and pending box vehicles where we're not sure how we'll want to handle the status. Until we understand better how all of these will work, I'd lean towards continuing to have the callers decide what fields to pass in, and we can keep this idea in mind for the future.

arkadyan: Interesting, yeah ok.

skyqrose:
With the large variant unbolded. (The label is still bold.)
Screen Shot 2019-06-12 at 11 47 20

Followup tasks:
 - [Don't show variant number if it's the default.](https://app.asana.com/0/1112935048846093/1126836005441035)
 - [Visual bug: pending vehicles with different length labels aren't aligned.](https://app.asana.com/0/1112935048846093/1126836005441033)

skyqrose:
@arkadyan @ingridpierre Ready for a final review.

ingridpierre:
@skyqrose lgtm! 👍

PR #73, opened 2019-06-14
Prioritize Swiftly vehicle trip and route assignments

arkadyan:
Asana ticket: [Prioritize Swiftly vehicle trip assignment in concentrate merge](https://app.asana.com/0/1112935048846093/1126289135392019)

Save data discrepancies any time swiftly and busloc provide differing
trip_id or route_id.

### DIFF
--- a/lib/concentrate/vehicle_position.ex
+++ b/lib/concentrate/vehicle_position.ex
@@ -74,9 +74,9 @@ defmodule Concentrate.VehiclePosition do
         second
         | trip_id:
             swiftly_priority(
-              VehiclePosition.source(second),
+              second.source,
               second.trip_id,
-              VehiclePosition.source(first),
+              first.source,
               first.trip_id
             ),
           stop_id: first_value(second.stop_id, first.stop_id),
@@ -97,9 +97,9 @@ defmodule Concentrate.VehiclePosition do
           direction_id: first_value(second.direction_id, first.direction_id),
           headsign:
             swiftly_priority(
-              VehiclePosition.source(second),
+              second.source,
               second.headsign,
-              VehiclePosition.source(first),
+              first.source,
               first.headsign
             ),
           headway_secs: first_value(second.headway_secs, first.headway_secs),
@@ -116,9 +116,9 @@ defmodule Concentrate.VehiclePosition do
             ),
           route_id:
             swiftly_priority(
-              VehiclePosition.source(second),
+              second.source,
               second.route_id,
-              VehiclePosition.source(first),
+              first.source,
               first.route_id
             ),
           schedule_adherence_secs:
@@ -146,7 +146,7 @@ defmodule Concentrate.VehiclePosition do
       do: first_value(first_value, second_value)
 
     defp merge_sources(first, second) do
-      [VehiclePosition.source(first), VehiclePosition.source(second)]
+      [first.source, second.source]
       |> Enum.sort()
       |> Enum.join("|")
     end
@@ -164,11 +164,11 @@ defmodule Concentrate.VehiclePosition do
           attribute: key,
           sources: [
             %{
-              id: VehiclePosition.source(first),
+              id: first.source,
               value: first_val
             },
             %{
-              id: VehiclePosition.source(second),
+              id: second.source,
               value: second_val
             }
           ]
--- a/lib/concentrate/vehicle_position.ex
+++ b/lib/concentrate/vehicle_position.ex
@@ -4,6 +4,8 @@ defmodule Concentrate.VehiclePosition do
   """
   import Concentrate.StructHelpers
 
+  alias Concentrate.{DataDiscrepancy, VehiclePosition}
+
   defstruct_accessors([
     :id,
     :trip_id,
@@ -33,6 +35,8 @@ defmodule Concentrate.VehiclePosition do
     :schedule_adherence_secs,
     :schedule_adherence_string,
     :scheduled_headway_secs,
+    :source,
+    :data_discrepancies,
     status: :IN_TRANSIT_TO
   ])
 
@@ -68,7 +72,13 @@ defmodule Concentrate.VehiclePosition do
     defp do_merge(first, second) do
       %{
         second
-        | trip_id: first_value(second.trip_id, first.trip_id),
+        | trip_id:
+            swiftly_priority(
+              VehiclePosition.source(second),
+              second.trip_id,
+              VehiclePosition.source(first),
+              first.trip_id
+            ),
           stop_id: first_value(second.stop_id, first.stop_id),
           label: first_value(second.label, first.label),
           license_plate: first_value(second.license_plate, first.license_plate),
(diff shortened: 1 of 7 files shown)


lib/concentrate/vehicle_position.ex
```diff
@@ -4,6 +4,8 @@ defmodule Concentrate.VehiclePosition do
   """
   import Concentrate.StructHelpers
 
+  alias Concentrate.{DataDiscrepancy, VehiclePosition}
```
skyqrose: This file is aliasing itself? Can't it just use `.source` instead of `&VehiclePosition.source/1`? Other callers use functions to access the fields to keep the data type opaque, but in the same file, that's not a concern. If it does need to be opaque, can it use `&source/1` directly, without `VehiclePosition.`?

lib/concentrate/vehicle_position.ex
```diff
@@ -98,17 +108,76 @@ defmodule Concentrate.VehiclePosition do
               second.previous_vehicle_schedule_adherence_string,
               first.previous_vehicle_schedule_adherence_string
             ),
-          route_id: first_value(second.route_id, first.route_id),
+          route_id:
+            swiftly_priority(
```
skyqrose: Are there other associated fields that should be `swiftly_priority`? Headsign maybe?

lib/concentrate/vehicle_position.ex
```diff
@@ -98,17 +108,76 @@ defmodule Concentrate.VehiclePosition do
...
+    defp swiftly_priority(other_source, other_value, "swiftly", value),
+      do: swiftly_priority("swiftly", value, other_source, other_value)
+
+    defp swiftly_priority(_, first_value, _, second_value),
+      do: first_value(first_value, second_value)
+
+    defp merge_sources(first, second) do
+      [VehiclePosition.source(first), VehiclePosition.source(second)]
+      |> Enum.sort()
+      |> Enum.join("|")
```
skyqrose: Do you need a `uniq` in here?

lib/concentrate/vehicle_position.ex
```diff
@@ -98,17 +108,76 @@ defmodule Concentrate.VehiclePosition do
...
+
+    defp swiftly_priority("swiftly", nil, _, value), do: value
+
+    defp swiftly_priority("swiftly", value, _, _), do: value
+
+    defp swiftly_priority(other_source, other_value, "swiftly", value),
+      do: swiftly_priority("swiftly", value, other_source, other_value)
+
+    defp swiftly_priority(_, first_value, _, second_value),
+      do: first_value(first_value, second_value)
```
skyqrose: If merging a `busloc|swiftly` source whose `trip_id` originally came from swiftly, with a more recent `busloc` assignment with a bad `trip_id`, this might take the `busloc` assignment over the `swiftly` one.

lib/concentrate/vehicle_position.ex
```diff
@@ -98,17 +108,76 @@ defmodule Concentrate.VehiclePosition do
...
+    end
+
+    defp discrepancies(first, second) do
+      attributes = [
+        {:trip_id, &VehiclePosition.trip_id/1},
+        {:route_id, &VehiclePosition.route_id/1}
+      ]
+
+      for {key, accessor_fn} <- attributes,
+          {first_val, second_val} <- do_discrepancies(accessor_fn, first, second) do
```
skyqrose: It took me a minute to figure out how `&do_discrepancies/3` makes the for loop skip non-discrepant values. It might be a little more obvious if we could put the `if` front and center.

How about:
```
Enum.flat_map(attributes, fn {key, accessor_fn} ->
  first_val = ...
  second_val = ...
  if first_val != second_val do
    [%DataDiscrepancy{
      ...
    }]
  else
    []
  end
end
```

lib/concentrate/vehicle_position.ex
arkadyan: Added headsign. I think that's all, but we can always add more easily if we find the need.

lib/concentrate/vehicle_position.ex
arkadyan: Updated the source calls but this is still needed for the attribute accessors in the discrepancies function.

lib/concentrate/vehicle_position.ex
paulswartz: You're doing `String.join` here, only to do a split in `comes_from_swiftly?`. Why not make it a `MapSet` for faster membership checking?

lib/concentrate/vehicle_position.ex
```diff
@@ -98,17 +121,94 @@ defmodule Concentrate.VehiclePosition do
...
+            first_value(second.scheduled_headway_secs, first.scheduled_headway_secs),
+          source: merge_sources(first, second),
+          data_discrepancies: discrepancies(first, second)
       }
     end
 
     defp first_value(value, _) when not is_nil(value), do: value
     defp first_value(_, value), do: value
+
+    defp swiftly_priority(source1, value1, source2, value2) do
```
paulswartz: It looks like you could do some short circuiting here:
```elixir
defp swiftly_priority(source1, value1, source2, value2)
defp swiftly_priority(_source1, value1, _source2, nil), do: value1
defp swiftly_priority(_source1, nil, _source2, value2), do: value2
defp swiftly_priority(source1, value1, source2, value2) do
  # now you know that both positions have values
  ...
end
```

lib/concentrate/vehicle_position.ex
arkadyan: Yes, this is better. A false start set me down this path, but I updated it to use a `MapSet`.

arkadyan:
@paulswartz You are correct that the data discrepancies are not being used yet—I added them early since they were so related to the rest of this work. We know we are going to want to surface these discrepancies to the user so we'll need this data, we just don't have design details yet.

test/concentrate/vehicle_position_test.exs
```diff
@@ -23,5 +58,268 @@ defmodule Concentrate.VehiclePositionTest do
...
+
+      expected =
+        new(
+          last_updated: 2,
+          latitude: 2,
+          longitude: 2,
+          trip_id: "busloc_trip",
+          route_id: "busloc_route",
+          sources: MapSet.new(["busloc", "swiftly"]),
+          data_discrepancies: [
```
paulswartz: One of the values being `nil` and the other having a value probably shouldn't be a discrepancy?

test/concentrate/vehicle_position_test.exs
```diff
@@ -23,5 +58,268 @@ defmodule Concentrate.VehiclePositionTest do
...
+          longitude: 2,
+          trip_id: "swiftly_merged_trip",
+          route_id: "swiftly_merged_route",
+          sources: MapSet.new(["busloc", "swiftly"]),
+          data_discrepancies: [
+            %DataDiscrepancy{
+              attribute: :trip_id,
+              sources: [
+                %{id: "swiftly", value: "swiftly_trip"},
+                %{id: "busloc|swiftly", value: "swiftly_merged_trip"}
```
paulswartz: I don't understand how this situation arises: you shouldn't be merging already merged data.

test/concentrate/vehicle_position_test.exs
```diff
@@ -23,5 +58,268 @@ defmodule Concentrate.VehiclePositionTest do
...
+          route_id: "route",
+          sources: MapSet.new(["first", "second"]),
+          data_discrepancies: []
+        )
+
+      assert Mergeable.merge(first, second) == expected
+    end
+  end
+
+  describe "comes_from_swiftly/1" do
```
paulswartz: I thought this was `comes_from_swiftly?/1` (with the question mark) since it returns a boolean?

test/concentrate/vehicle_position_test.exs
arkadyan: Removed

test/concentrate/vehicle_position_test.exs
arkadyan: Good call, updated

test/concentrate/vehicle_position_test.exs
arkadyan: Discussed this condition in person. We'll want to keep watch for false positives around trip boundaries.

PR #77, opened 2019-06-19
Use css grid to make all Incoming Boxes the same height

arkadyan:
Asana ticket: [Make all Incoming Boxes the same height](https://app.asana.com/0/1112935048846093/1125006647738997)

Screen Shot 2019-06-19 at 12 19 05


assets/css/_ladder.scss
```diff
@@ -1,4 +1,6 @@
 .m-ladder {
+  grid-column: span 1;
+  grid-row: 2 / 3;
```
skyqrose: The grid lines can be given names instead of the default indices. Idk what best practice is, though.

assets/css/_ladder.scss
```diff
@@ -1,4 +1,6 @@
 .m-ladder {
+  grid-column: span 1;
```
skyqrose: I think `span 1` is the default. Have you tried without it or `grid-column: auto`?

assets/css/_route_ladder.scss
```diff
@@ -1,6 +1,6 @@
-.m-route-ladder {
-  display: flex;
-  flex-direction: column;
+.m-route-ladder__header-and-controls {
```
skyqrose: What if the header and the controls were separate rows of the grid, instead of having to be wrapped together?

assets/css/_ladder.scss
arkadyan: Named rows definitely seem nice.

PR #78, opened 2019-06-19
Implement logins using Cognito and Active Directory

arkadyan:
Asana ticket: [Implement login](https://app.asana.com/0/1112935048846093/1126294352975821)

Install git in Docker file so we can install dependencies from GitHub.

Deployed to https://skate-dev.mbtace.com/ for testing.


lib/skate_web/templates/layout/app.html.eex
```diff
@@ -21,6 +21,7 @@
     <main role="main" class="container">
       <%= render @view_module, @view_template, assigns %>
     </main>
+    <script id="app-user-token" type="text/plain"><%= assigns[:user_token] %></script>
```
paulswartz: `<script>` seems like a weird type for this: does this pattern come from somewhere else?

lib/skate_web/router.ex
```diff
@@ -35,4 +53,21 @@ defmodule SkateWeb.Router do
...
   end
+
+  defp put_user_token(conn, _) do
+    token = Guardian.Plug.current_token(conn)
+    assign(conn, :user_token, token)
+  end
+
+  defp api_auth(conn, _) do
+    with [token] <- get_req_header(conn, "token"),
+         {:ok, _claims} <- AuthManager.decode_and_verify(token) do
```
losvedir: What are your API requests, and are they related to Cognito somehow?

In signs_ui our API authentication (the requests that update the countdown signs) doesn't use the Cognito pool, since there's no underlying user to redirect to the auth page and back.

While I think it's possible to use Guardian for both cases, right now the `ErrorHandler` seems to be set up only to work with Cognito. But for us, for our "API" requests, since we control the software on both ends, we just compare the value here to an ENV var.

lib/skate_web/templates/layout/app.html.eex
arkadyan: I copied this from dotcom, but they use it for complex data and I just have a string. Maybe a `data-user-token` attribute on the `#app` div would be a better approach?

paulswartz: Or you could make yours an object:
```elixir
<script id="app-user-token" type="application/json">
  <%= Jason.encode_to_iodata!(%{
    token: assigns[:user_token]
  }) %>
</script>
```

lib/skate_web/router.ex
arkadyan: Our API requests are for all the underlying data, so we only want authorized users to be able to access the results. I'm I'm understanding the question correctly they are related to Cognito in that we want to use the JWT token from Cognito to verify that this user is logged in (and thus an MBTA employee). We do take the user to an Active Directory login page before they can access the page to make the API requests. Does that answer the question?

losvedir: Oh, I think I misunderstood the flow. 

So are these API requests because of the SPA architecture, and are conceptually part of the session of the logged in user? Like, they visit the URL, and if they're authenticated it will load the page, which will start up the JS app, which then makes these requests because of actions the user takes?

If that's the case, does it make sense to just use `:ensure_auth` here, since it will be sending the cookie anyway? The alternative you're using here, of using a Bearer token is useful when supporting, say, mobile apps, or sites on multiple domains or the like.

lib/skate_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,27 @@
...
+    expiration = auth.credentials.expires_at
+
+    current_time = System.system_time(:second)
+
+    conn
+    |> Guardian.Plug.sign_in(
+      AuthManager,
+      username,
+      %{},
+      ttl: {expiration - current_time, :seconds}
```
losvedir: Just FYI this is one hour. That was all we needed initially on signs_ui, since mostly people pop it open to check on something, make a change, and then they're done. It's not meant to be left open for extended periods of time. After one hour they have to sign in again. Someone is currently working on getting the Oauth refresh_token flow up and running, though, to prevent that.

assets/src/hooks/useSocket.ts
```diff
@@ -1,11 +1,14 @@
...
 import { useEffect, useState } from "react"
+import { UserToken } from "../skate"
 
-const useSocket = (): Socket | undefined => {
+const useSocket = (userToken?: UserToken): Socket | undefined => {
   const [socket, setSocket] = useState<Socket | undefined>(undefined)
 
   useEffect(() => {
-    const initialSocket = new Socket("/socket")
+    const initialSocket = new Socket("/socket", {
```
losvedir: I see you sending the token on socket connect, but not seeing you check it on socket connect. There should be a Guardian check on the `connect` callback in your socket code (`/channels/user_socket.ex` is the default, unless Skate changed it). See [here](https://github.com/mbta/signs_ui/blob/487d095cf72212d01ba031b7daa00fa21e490575/lib/signs_ui_web/channels/user_socket.ex#L23).

arkadyan: Great catch! Added the authentication and tests proving it.

lib/skate_web/channels/user_socket.ex
```diff
@@ -15,8 +15,11 @@ defmodule SkateWeb.UserSocket do
   #
   # See `Phoenix.Token` documentation for examples in
   # performing token verification on connect.
-  def connect(_params, socket, _connect_info) do
-    {:ok, socket}
+  def connect(%{"token" => token}, socket, _connect_info) do
```
losvedir: Great. The other thing to consider is whether and how often to verify the token is still valid. For instance, in our signs_ui channel that turns signs on/off and puts custom text on them we [verify the token](https://github.com/mbta/signs_ui/blob/de732233f1f6cc94ade4073973b176b9f762d64a/lib/signs_ui_web/channels/signs_channel.ex#L12) on the channel `handle_in` callback. And if it's no longer valid, we respond with `"auth_expired"` and then close the websocket. The frontend has a bit of logic to detect `"auth_expired"` and refresh the page (which, because of `:ensure_auth` and Guardian, will redirect to the Cognito sign in page).

arkadyan: Nice, copied this shamelessly.

lib/skate_web/router.ex
arkadyan: @losvedir your `:browser` suggestion help me find a solution to this, thanks! I just needed to split out the `accepts` plugs.

assets/src/components/appStateWrapper.tsx
```diff
@@ -4,19 +4,31 @@ import useSocket from "../hooks/useSocket"
 import useTimepoints from "../hooks/useTimepoints"
 import useVehicles from "../hooks/useVehicles"
 import DispatchProvider from "../providers/dispatchProvider"
-import { Route, TimepointsByRouteId } from "../skate"
+import { Route, TimepointsByRouteId, UserToken } from "../skate"
 import { initialState, reducer } from "../state"
 import App from "./app"
 
+export const readUserToken = (): UserToken | undefined => {
+  const dataEl = document.getElementById("app")
```
skyqrose: Should this be in a `useEffect`? Will this necessarily exist when we run this code? I think of reading from global state as a side effect, but it still might be safe to do without the hook.

config/config.exs
```diff
@@ -25,6 +25,10 @@ config :skate, SkateWeb.Endpoint,
   render_errors: [view: SkateWeb.ErrorView, accepts: ~w(html json)],
   pubsub: [name: Skate.PubSub, adapter: Phoenix.PubSub.PG2]
 
+config :skate, SkateWeb.AuthManager,
+  issuer: "skate",
+  secret_key: "test key"
```
skyqrose: Instead of having this as the default, should this be specified in `config/test.exs`? What's the key used for, what value should it have in dev, and would you ever want to set it by env variable in dev?

Similar for the `Ueberauth.Strategy.Fake` below. If dev and test and prod need to be different, we might want to specify them in the environment-specific configuration so there's less chance of the insecure configuration accidentally applying in prod.

lib/skate/ueberauth/strategy/fake.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @impl Ueberauth.Strategy
+  def uid(_conn) do
+    "fake_uid"
+  end
+
+  @impl Ueberauth.Strategy
+  def credentials(_conn) do
+    one_hour_in_seconds = 60 * 60
+    expiration_time = System.system_time(:second) + one_hour_in_seconds
```
skyqrose: Will we have to refresh every hour in development? That might be disruptive? Could this be extended?

lib/skate_web/channels/vehicles_channel.ex
```diff
@@ -15,7 +16,28 @@ defmodule SkateWeb.VehiclesChannel do
...
+
+  @spec socket_authenticated?(Phoenix.Socket.t()) :: boolean()
+  defp socket_authenticated?(socket) do
+    token = Guardian.Phoenix.Socket.current_token(socket)
+
+    with {:ok, _claims} <- AuthManager.decode_and_verify(token) do
+      true
+    else
+      _ -> false
+    end
```
skyqrose: Could be `Kernel.match?({:ok, _claims}, AuthManager.decode_and_verify(token))`

assets/src/components/appStateWrapper.tsx
arkadyan: This will necessarily exist when our app runs, this isn't part of the React app, it's the div the app is written in to.

lib/skate/ueberauth/strategy/fake.ex
arkadyan: This is one hour after the last time the page was loaded (as the token is updated each time). Seems unlikely to be a problem in practice, but it also doesn't matter what this is.

assets/src/components/appStateWrapper.tsx
```diff
@@ -4,19 +4,31 @@ import useSocket from "../hooks/useSocket"
...
+  if (!dataEl) {
+    return undefined
```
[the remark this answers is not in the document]
arkadyan: It's true, but I don't think there's any case where that should be possible, so I don't know what would be more useful to do.

assets/src/components/appStateWrapper.tsx
```diff
@@ -4,19 +4,31 @@ import useSocket from "../hooks/useSocket"
...
+  }
+
+  const token = dataEl.dataset.userToken as UserToken
+  return token
+}
+const userToken: UserToken | undefined = readUserToken()
```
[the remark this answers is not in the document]
arkadyan: We could, though that would mean running this logic on each network request, instead of once when the app loads. Does that tradeoff seem worthwhile? I feel like my first instinct is to lean towards no, but I don't feel strongly.

Oh, we'd also need to do it in the socket as well.

lib/skate/ueberauth/strategy/fake.ex
skyqrose: I can imagine opening a bunch of routes, going to a meeting, and wanting to come out of the meeting able to resume without reloading. Let's bump it to 9 hours.

assets/src/components/appStateWrapper.tsx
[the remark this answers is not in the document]
skyqrose: Ah, yeah, we should only read it once. 👍

assets/src/api.ts
```diff
@@ -39,8 +45,8 @@ const parseRouteData = ({ id, direction_names }: RouteData): Route => ({
 const parseRoutesData = (routesData: RouteData[]): Route[] =>
   routesData.map(parseRouteData)
 
-export const fetchRoutes = (): Promise<Route[]> =>
-  fetch("/api/routes")
+export const fetchRoutes = (userToken?: UserToken): Promise<Route[]> =>
+  request("/api/routes", userToken)
```
losvedir: If I understand correctly, I don't think these requests need to send the userToken anymore?

lib/skate/ueberauth/strategy/fake.ex
losvedir: Are you sure it's one hour since the page was loaded? I believe this is one hour since the cookie is set, which only happens when you go through the /auth/cognito -> /auth/cognito/callback flow, which only happens when you try to load the page while not being signed in. (i.e. if you come back in 30 mins, the cookie will still be a valid authenticated session, so you won't go through the request -> callback flow.)

arkadyan: Oh that's interesting, that sounds true. 9 hours it is.

assets/src/api.ts
arkadyan: Good call! I removed these and move the token reading logic into the socket code since it is only needed there.

PR #79, opened 2019-06-21
Set height of ladders based on height of screen

arkadyan:
Asana ticket: [Set height of ladders based on height of screen](https://app.asana.com/0/1112935048846093/1122859757950931)

Depends on https://github.com/mbta/skate/pull/77

New:

new

Old:

old


assets/tests/components/__snapshots__/routeLadders.test.tsx.snap
```diff
@@ -48,90 +48,89 @@ exports[`renders a route ladder 1`] = `
       </button>
     </div>
   </div>
-  <svg
+  <div
```
paulswartz: Does it matter that the class has moved from the SVG to a div?

assets/tests/models/ladderVehicle.test.ts
```diff
@@ -37,128 +37,89 @@ describe("putIntoLanes", () => {
...
         x: 0,
-        y: 20,
+        y: () => 20,
         vehicleDirection: VehicleDirection.Down,
       },
       {
         vehicle,
         x: 0,
-        y: 10,
+        y: () => 10,
```
paulswartz: Should these functions be taking an argument?

assets/src/components/ladder.tsx
```diff
@@ -39,38 +45,70 @@ const Ladder = ({
...
+}: {
+  timepoints: Timepoint[]
+  ladderVehicles: LadderVehicle[]
+  selectedVehicleId?: VehicleId
+  width: number
+  timepointSpacingY: TimepointSpacingYFunc
+  timepointStatusY: TimepointStatusYFunc
+}) => {
+  const dispatch = useContext(DispatchContext)
+  const [height, setHeight] = useState(500)
```
paulswartz: Is the the same 500 as `ESTIMATED_HEIGHT`?

arkadyan:
Unfortunately scaling doesn't help us because we need to draw timepoints and vehicles dependent on the height of the ladder. We essentially want to "scale" vertically but not horizontally, so if we drew at a set size and then scaled the image it would either end up at the wrong width, or would be stretched.

assets/tests/components/__snapshots__/routeLadders.test.tsx.snap
arkadyan: No, none of the properties are specific to the svg, only to the bounding box.

assets/src/components/ladder.tsx
arkadyan: It technically doesn't have to be, but it's the same gist so I switched to that.

assets/tests/models/ladderVehicle.test.ts
arkadyan: It doesn't really matter since we aren't using the arguments and JS doesn't care, but I can add them if that makes things more clear.

assets/src/components/ladder.tsx
```diff
@@ -39,38 +46,70 @@ const Ladder = ({
...
     timepointStatusY
   )
 
   const width = 120 + 2 * widthOfLanes
-  // (0, 0) is in the center of the first timepoint
-  const viewBox = [-width / 2, -MARGIN_TOP_BOTTOM, width, HEIGHT].join(" ")
 
   return (
-    <svg className="m-ladder" height={HEIGHT} width={width} viewBox={viewBox}>
+    <div className="m-ladder">
```
skyqrose: Why does this need to go on a wrapping div? Can the `ref` and the class go together? If the class is on the `svg` directly, that's one fewer layer, and I think a `[0]` can be removed from `useLadderSvgHeight`.

assets/src/hooks/useLadderSvgHeight.ts
```diff
@@ -0,0 +1,35 @@
...
+      const routeLadder: HTMLElement = document.getElementsByClassName(
+        "m-ladder"
+      )[0] as HTMLElement
+      const newHeight = routeLadder.offsetHeight
+      if (newHeight !== height) {
+        setHeight(newHeight)
+        ref.current.setAttribute("width", `${width}`)
+        ref.current.setAttribute("height", `${newHeight}`)
+        // (0, 0) is in the center of the first timepoint
+        ref.current.setAttribute("viewBox", [x, y, width, height].join(" "))
```
skyqrose: Instead of having to reach into the DOM to mutate these, could they be constructed the normal way in `ladder.tsx` based on the mutated `height` value?

assets/src/components/ladder.tsx
```diff
@@ -39,38 +46,70 @@ const Ladder = ({
...
+        selectedVehicleId={selectedVehicleId}
+        width={width}
+        timepointSpacingY={timepointSpacingY}
+        timepointStatusY={timepointStatusY}
+      />
+    </div>
+  )
+}
+
+const LadderSvg = ({
```
skyqrose: If all of this stays in one component, I think a few things would be a little simpler. E.g. `timepointSpacingY` could be a variable calculated after `height = useState()` instead of being a function with its own `TimepointSpacingYFunc`.

What's the benefit of splitting the `LadderSvg` stuff into a separate component, and is it worth the overhead + extra complexity?

assets/src/components/ladder.tsx
```diff
@@ -185,42 +225,43 @@ const LadderTimepoint = ({
...
   scheduledY: number
+  height: number
 }) => (
   <line
     className={`m-ladder__scheduled-line ${
       ladderVehicle.vehicle.scheduleAdherenceStatus
     }`}
     x1={ladderVehicle.x}
-    y1={ladderVehicle.y}
+    y1={ladderVehicle.y(height)}
```
skyqrose: Would it be possible to use the `height` when calculating the `ladderVehicle` for the first time, instead of having to pass it in later? (This applies to a few different places.)

assets/src/models/ladderVehicle.ts
```diff
@@ -220,9 +218,13 @@ const overlappingPreviousVehicles = (
     areOverlapping(previousVehicle, vehicle)
   )
 
+// We need to determine overlapping vehicles before knowing the height of the
```
skyqrose: Why? I know we might have to start with a default height, but once the height is set, we should be able to read that value, right?

assets/src/components/ladder.tsx
arkadyan: We need the div to have something rendered in the dom in order to pull a height off of. That's the height used in generating the svg.

assets/src/hooks/useLadderSvgHeight.ts
arkadyan: It doesn't use the new height if you do it that way. I think it is rendering the svg before the hook is run.

assets/src/components/ladder.tsx
arkadyan: No we don't know the height the first time, it is generated in an layout effect hook after this code is initially run.

assets/src/components/ladder.tsx
arkadyan: Yep, this is better. That was fallout from one of the several approaches I tried.

assets/src/models/ladderVehicle.ts
arkadyan: This went away when we stopped passing height around.

assets/src/hooks/useLadderSvgHeight.ts
skyqrose: If it's using the height in the calculations inside the svg, it should also be able to use the height in the viewbox, right? Calling `setHeight` should trigger a rerender.

If the `setHeight` can trigger the rerender, then I think we could get rid of the `ref`.

assets/src/components/ladder.tsx
```diff
@@ -40,113 +45,119 @@ const Ladder = ({
...
     vehicles,
     ladderDirection,
     timepointStatusY
   )
 
   const width = 120 + 2 * widthOfLanes
-  // (0, 0) is in the center of the first timepoint
-  const viewBox = [-width / 2, -MARGIN_TOP_BOTTOM, width, HEIGHT].join(" ")
+
+  const ref = createRef<SVGSVGElement>()
```
skyqrose: Should this be `useRef` instead of `createRef`? https://stackoverflow.com/questions/54620698/whats-the-difference-between-useref-and-createref

assets/src/hooks/useLadderSvgHeight.ts
```diff
@@ -0,0 +1,35 @@
...
+      ref.current !== null &&
+      // only if there is no viewBox attribute
+      !ref.current.getAttribute("viewBox") &&
+      // only if rendered
+      ref.current.getBBox().width &&
+      ref.current.getBBox().height
+    ) {
+      const routeLadder: HTMLElement = document.getElementsByClassName(
+        "m-ladder"
+      )[0] as HTMLElement
```
skyqrose: Should this element be gotten by `useRef` instead of `document.getElement`?

All of the heights are set based on the height of the first ladder. How does that compare to each ladder checking itself for the height.

assets/src/hooks/useLadderSvgHeight.ts
arkadyan: Try it, it doesn't work.

assets/src/hooks/useLadderSvgHeight.ts
arkadyan: No, the svg doesn't have the full height yet, we need to get the height of the bounding box in order to tell the svg to fill it. It doesn't matter that we use the first ladder div because they are all the same height.

assets/src/components/ladder.tsx
arkadyan: That's interesting, I didn't know that distinction. But `createRef` is what we want because we want a reference to the svg, not to the Ladder component, so we create the ref and attach it to the svg.

assets/src/hooks/useLadderSvgHeight.ts
skyqrose: Got it to work: https://github.com/mbta/skate/commit/be85eca9051538725216e38a675d84754f02c9cc

arkadyan: Thanks! I pulled that in and updated the tests.

assets/src/models/ladderVehicle.ts
```diff
@@ -220,6 +216,9 @@ const overlappingPreviousVehicles = (
...
+// ladder, so we make a reasonable guess.
+// In practice this won't matter except for extreme edge cases.
```
skyqrose: Obsolete comment.

assets/tests/components/ladder.test.tsx
```diff
@@ -6,285 +6,252 @@ import DispatchProvider from "../../src/providers/dispatchProvider"
...
-      <Ladder
-        timepoints={timepoints}
-        vehicles={vehicles}
-        ladderDirection={ladderDirection}
-        selectedVehicleId={undefined}
-      />
-    )
-    .toJSON()
+  beforeEach(() => {
+    SVGSVGElement.prototype.getBBox = () => {
```
skyqrose: Don't we get the BBox of the HTMLDivElement, not the SvgElement? (Same for other tests that use this pattern)

assets/src/components/ladder.tsx
```diff
@@ -203,7 +225,7 @@ const timepointStatusYFromTimepoints = (
       )
     }
   }
-  return null
+  return 0
```
skyqrose: The `|| -10` later means that it doesn't matter whether we fail with `null` or `0`. Did you mean to remove that now that this function won't return `null`?
(I think there's a related bug where trains that are legitimately at `0` get moved to `-10`)

assets/src/components/ladder.tsx
```diff
@@ -40,113 +44,131 @@ const Ladder = ({
...
   const { ladderVehicles, widthOfLanes } = ladderVehiclesFromVehicles(
     vehicles,
     ladderDirection,
     timepointStatusY
   )
 
   const width = 120 + 2 * widthOfLanes
-  // (0, 0) is in the center of the first timepoint
-  const viewBox = [-width / 2, -MARGIN_TOP_BOTTOM, width, HEIGHT].join(" ")
+  const viewBox = [-width / 2, -MARGIN_TOP_BOTTOM, width, height].join(" ")
```
skyqrose: Oops, I didn't mean to remove that comment `// (0, 0) is in the center of the first timepoint`. It's still relevant and probably worth keeping.

assets/src/components/ladder.tsx
arkadyan: That makes it sound like returning `0` and removing the `|| -10` is the better thing to do?

assets/tests/components/ladder.test.tsx
arkadyan: That was related to implementation details you removed and is no longer relevant.

assets/src/components/ladder.tsx
arkadyan: Made that change

PR #80, opened 2019-06-24
Add a FullStory recording snippet

arkadyan:
Only enable FullStory recording in production.

Asana ticket: [Add FullStory session recording](https://app.asana.com/0/1112935048846093/1125337183434468)


arkadyan:
@skyqrose I forgot to mention, I added the `skate-dev.mbtace.com` domain to FullStory, tested that my traffic came through, and then disabled recording on that domain on their side.

Semi-relatedly, if our prod domain ends up being something other than `*.mbta.com` we'll need to enable recording for that domin once it's live.

PR #82, opened 2019-06-26
Display a selected vehicle on top of all others

arkadyan:
Asana ticket: [When you select a vehicle, bring it to the top](https://app.asana.com/0/1112935048846093/1123838037403441)

Unselected:

unselected

Selected:

selected


arkadyan:
@skyqrose I implemented a slight variation on your filter idea (just using a partition instead of two filters). I don't think we should draw the line above any triangles, and I also don't think it's worthwhile to draw the line above other lines.

PR #83, opened 2019-06-28
Show data discrepancies when in debug mode

arkadyan:
At the bottom of the Vehicle Properties panel.

View this by adding `?debug=1` to the URL.

Asana ticket: [Debug mode for data discrepancies](https://app.asana.com/0/1112935048846093/1128997245638196)

debug

### DIFF
--- a/assets/src/hooks/useVehicles.ts
+++ b/assets/src/hooks/useVehicles.ts
@@ -227,7 +227,6 @@ const vehicleFromData = (vehicleData: VehicleData): Vehicle => ({
     vehicleData.schedule_adherence_secs
   ),
   scheduledHeadwaySecs: vehicleData.scheduled_headway_secs,
-  sources: vehicleData.sources,
   dataDiscrepancies: dataDiscrepanciesFromData(vehicleData.data_discrepancies),
   stopStatus: vehicleStopStatusFromData(vehicleData.stop_status),
   timepointStatus: vehicleTimepointStatusFromData(vehicleData.timepoint_status),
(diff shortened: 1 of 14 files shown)


arkadyan:
Related to, but not directly affecting this change: it seems like Busloc might not ever be assigning a route ID or headway. If that is true, we likely will want to remove those properties for the list that are observed for discrepancies.

skyqrose:
Are there any other fields that would be useful to have in the debug mode? Going through all the fields, these are the ones that might be useful and aren't displayed anywhere else:
* timestamp
* block_id
* run_id

assets/src/hooks/useVehicles.ts
```diff
@@ -46,6 +56,8 @@ interface VehicleData {
   schedule_adherence_secs: number
   schedule_adherence_string: string
   scheduled_headway_secs: number
+  sources: string[]
```
skyqrose: Is this field used at all? Each discrepency has its own list of sources already.

arkadyan:
@skyqrose At least for the moment I want to keep this very focused. We have developer tools that give us access to debug all the data, I want this to be a way for everyone to be able to explore differences in assignments.

lib/realtime/vehicle.ex
```diff
@@ -69,6 +71,8 @@ defmodule Realtime.Vehicle do
     :operator_id,
     :operator_name,
     :run_id,
+    :sources,
+    :data_discrepancies,
```
skyqrose: `[]` would be a reasonable and valid default for this field, so I don't think we need to require the field.

PR #86, opened 2019-07-01
Don't include route id and headway in discrepancies

arkadyan:
These are never set by Busloc so they were always showing up.

Ensure that discrepancy check returns a boolean value.


assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -260,7 +260,7 @@ const directionsUrl = (
 &travelmode=driving`
 
 const shouldShowDataDiscrepancies = ({ dataDiscrepancies }: Vehicle) =>
```
skyqrose: This could get `: boolean`.

PR #88, opened 2019-07-03
Sign the user out and redirect home on auth failure

arkadyan:
Asana ticket: [🐞Unauthorized screen sometimes appears after Cognito login](https://app.asana.com/0/1112935048846093/1129718872336334)

Users are sometimes seeing unexpected Ueberauth failures of unknown
provenance. Instead of sending a 403 unauthenticated response, we are
signing them out and sending them to the home page to start the auth
path over again.

We should be on the lookout for users getting trapped in a loop because
of this. If we observe that happening we should rethink this remedy.


arkadyan:
Deployed to dev for testing. I've verified that it doesn't break normal usage, but don't know how to simulate the problem case to test that.

PR #89, opened 2019-07-05
Save selected routes state in local browser storage

arkadyan:
Asana ticket: [Save selected routes interface state in local browser storage](https://app.asana.com/0/1112935048846093/1124828852046158)

### DIFF
--- a/assets/src/components/appStateWrapper.tsx
+++ b/assets/src/components/appStateWrapper.tsx
@@ -1,4 +1,5 @@
-import React, { useReducer } from "react"
+import React from "react"
+import usePersistedStateReducer from "../hooks/usePersistedStateReducer"
 import useRoutes from "../hooks/useRoutes"
 import useSocket from "../hooks/useSocket"
 import useTimepoints from "../hooks/useTimepoints"
@@ -9,7 +10,7 @@ import { initialState, reducer } from "../state"
 import App from "./app"
 
 const AppStateWrapper = (): JSX.Element => {
-  const [state, dispatch] = useReducer(reducer, initialState)
+  const [state, dispatch] = usePersistedStateReducer(reducer, initialState)
   const { selectedRouteIds, selectedVehicleId } = state
 
   const routes: Route[] | null = useRoutes()
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -16,10 +16,10 @@ const usePersistedStateReducer = (
     (initial: State) => ({ ...initial, ...loadState(APP_STATE_KEY) })
   )
 
+  const persistableState = filter(state, PERSISTED_KEYS)
   useEffect(() => {
-    const perstableState = filter(state, PERSISTED_KEYS)
-    saveState(APP_STATE_KEY, perstableState)
-  }, [state])
+    saveState(APP_STATE_KEY, persistableState)
+  }, [persistableState])
 
   return [state, dispatch]
 }
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -1,0 +1,22 @@
+import { useEffect, useReducer } from "react"
+import { loadState, saveState } from "../localStorage"
+import { Dispatch, Reducer, State } from "../state"
+
+const usePersistedStateReducer = (
+  reducer: Reducer,
+  defaultValue: State
+): [State, Dispatch] => {
+  const [state, dispatch] = useReducer(
+    reducer,
+    defaultValue,
+    (initial: State) => ({ ...initial, ...loadState() })
+  )
+
+  useEffect(() => {
+    saveState(state)
+  }, [state])
+
+  return [state, dispatch]
+}
+
+export default usePersistedStateReducer
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -2,6 +2,10 @@ import { useEffect, useReducer } from "react"
 import { loadState, saveState } from "../localStorage"
 import { Dispatch, Reducer, State } from "../state"
 
+const APP_STATE_KEY = "mbta-skate-state"
+
+const PERSISTED_KEYS = ["selectedRouteIds"]
+
 const usePersistedStateReducer = (
   reducer: Reducer,
   defaultValue: State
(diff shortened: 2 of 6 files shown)


assets/src/components/appStateWrapper.tsx
```diff
@@ -9,7 +10,7 @@ import { initialState, reducer } from "../state"
 import App from "./app"
 
 const AppStateWrapper = (): JSX.Element => {
-  const [state, dispatch] = useReducer(reducer, initialState)
+  const [state, dispatch] = usePersistedStateReducer(reducer, initialState)
```
skyqrose: I love how the `usePersistedStateReducer` hook makes this such an easy swap.

assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useReducer } from "react"
+import { loadState, saveState } from "../localStorage"
+import { Dispatch, Reducer, State } from "../state"
+
+const usePersistedStateReducer = (
+  reducer: Reducer,
```
skyqrose: This hook and `localStorage.ts` seem to be halfway between trying to be generic/reusable and being specific to the app state. E.g. the hook takes a `Reducer` as an argument, but there's only one function that could possibly fit that spec, `state.reducer`.

I'd go all or nothing. Either pass everything in (including `saved_keys` and `app_state_key`), and make the types generic, or make everything specific, where the hook has no arguments and is only usable by the top level `<App>`

Making everything specific would probably be easier. But if this hook becomes truly reusable, perhaps we could store the reversal states by giving each route its own key instead of by moving the reversal states to the top level (this may not be valuable, since we want to move the reversal state up anyway for other reasons).

assets/src/localStorage.ts
```diff
@@ -0,0 +1,34 @@
+import { RouteId } from "./skate"
+
+const APP_STATE_KEY = "mbta-skate-state"
+
+const SAVED_KEYS = ["selectedRouteIds"]
+
+interface PersistedState {
+  selectedRouteIds: RouteId[]
```
skyqrose: Having `SAVED_KEYS` and `PersistedState` defined so far away from the full `State` makes me nervous that it will be easy for them to get out of sync. Perhaps, if this file becomes generic/reusable, then these could be defined in `state.ts`.

assets/src/localStorage.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export const saveState = (state: PersistedState) => {
+  const perstableState = filter(state, SAVED_KEYS)
+  const serializedState = JSON.stringify(perstableState)
+  localStorage.setItem(APP_STATE_KEY, serializedState)
+}
+
+export const loadState = (): PersistedState | undefined => {
+  const serializedState = localStorage.getItem(APP_STATE_KEY)
```
skyqrose: Is there any namespace collision between the name of this file and this reference to `localStorage`?

assets/tests/localStorage.test.ts
```diff
@@ -0,0 +1,75 @@
+import { filter, loadState, saveState } from "../src/localStorage"
+
+const APP_STATE_KEY = "mbta-skate-state"
+
+const mockLocalStorage = {
+  getItem: jest.fn(),
+  setItem: jest.fn(),
+}
+Object.defineProperty(window, "localStorage", {
```
skyqrose: Do we need to reset this when the test is done to avoid polluting other tests?

arkadyan: Can't hurt

assets/src/hooks/usePersistedStateReducer.ts
arkadyan: Well, one big goal is to have the same format as `useReducer`. But this comment and the one below are making me think that `APP_STATE_KEY` and `SAVED_KEYS` should be owned by `usePersistedStateReducer` and `localStorage` should be a naive repository.

assets/src/localStorage.ts
arkadyan: No, names of files don't affect anything in JS.

arkadyan:
@skyqrose Only `selectedRouteIds` is saved, so changes to other state fields will make no difference, but yes, if the format of `selectedRouteIds` changes we will need to account for migration of the saved state data. (Which could be as simple as dropping it.)

assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -0,0 +1,26 @@
...
+): [State, Dispatch] => {
+  const [state, dispatch] = useReducer(
+    reducer,
+    defaultValue,
+    (initial: State) => ({ ...initial, ...loadState(APP_STATE_KEY) })
+  )
+
+  useEffect(() => {
+    saveState(APP_STATE_KEY, state, PERSISTED_KEYS)
+  }, [state])
```
skyqrose: If you moved the `filter` to this file, then the filtered object could also be used here and we'd only call `saveState` when the saved part of the state changes.

assets/tests/localStorage.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  beforeAll(() => {
+    Object.defineProperty(window, "localStorage", {
+      value: mockLocalStorage,
+    })
+  })
+
+  afterAll(() => {
+    Object.defineProperty(window, "localStorage", originalLocalStorage)
+  })
```
skyqrose: `filter` is pure, so these tests don't need the `beforeAll` and `afterAll`.

assets/src/hooks/usePersistedStateReducer.ts
skyqrose: Ah, yeah, `localStorage.ts` is looking pretty nice now. 👍 

I don't think having the exact same interface as `useReducer` is necessary (despite complimenting it earlier) if we're not planning on having it be a drop in replacement (which it can't be when this file has hardcoded references to `APP_STATE_KEY`, `PERSISTED_KEYS`, and the `State` and `Reducer` types).

arkadyan:
This doesn't actually use the `setTimepoints` mechanism updated in #91, it just hydrates the initial state.

assets/src/hooks/usePersistedStateReducer.ts
skyqrose: To prevent calling `saveState` on changes we don't care about, this line should also reference `persistableState` instead of `state` (and `persistableState` will need to be calculated before `useEffect`)

Also, typo `perstableState` -> `persistableState`.

PR #93, opened 2019-07-11
Add map to vehicle properties panel

katehedgpeth:
**Asana Ticket**: [Render a map component given a lat-long by fetching tiles from our tile server](https://app.asana.com/0/1112935048846093/1127990888669976)

Screen Shot 2019-07-11 at 4 20 00 PM

### DIFF
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -136,26 +136,3 @@
   width: 100%;
   z-index: 1000;
 }
-
-.m-vehicle-properties-panel__map {
-  height: 30rem;
-  margin-top: 1rem;
-}
-
-.m-vehicle-properties-panel__map-label {
-  @include font-emph;
-
-  svg {
-    overflow: visible;
-  }
-}
-
-.m-vehicle-properties-panel__map-icon {
-  path {
-    fill: $color-primary;
-  }
-
-  svg {
-    overflow: visible;
-  }
-}
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -141,3 +141,18 @@
   height: 500px;
   margin-top: 1rem;
 }
+
+.m-vehicle-properties-panel__map-label {
+  @include font-small;
+  font-weight: bold;
+
+  svg {
+    overflow: visible;
+  }
+}
+
+.m-vehicle-properties-panel__map-icon {
+  path {
+    fill: $color-primary;
+  }
+}
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -3,7 +3,7 @@
   box-shadow: $modal-box-shadow;
   height: 100%;
   max-width: 100%;
-  overflow: scroll;
+  overflow-y: scroll;
   position: fixed;
   right: 0;
   top: inherit;
@@ -138,13 +138,12 @@
 }
 
 .m-vehicle-properties-panel__map {
-  height: 500px;
+  height: 30rem;
   margin-top: 1rem;
 }
 
 .m-vehicle-properties-panel__map-label {
-  @include font-small;
-  font-weight: bold;
+  @include font-emph;
 
   svg {
     overflow: visible;
@@ -154,5 +153,9 @@
 .m-vehicle-properties-panel__map-icon {
   path {
     fill: $color-primary;
+  }
+
+  svg {
+    overflow: visible;
   }
 }
--- a/assets/css/_vehicle_properties_panel.scss
+++ b/assets/css/_vehicle_properties_panel.scss
@@ -3,6 +3,7 @@
   box-shadow: $modal-box-shadow;
   height: 100%;
   max-width: 100%;
+  overflow: scroll;
   position: fixed;
   right: 0;
   top: inherit;
@@ -134,4 +135,9 @@
   top: inherit;
   width: 100%;
   z-index: 1000;
+}
+
+.m-vehicle-properties-panel__map {
+  height: 500px;
+  margin-top: 1rem;
 }
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -1,10 +1,10 @@
 import Leaflet, { Map as LeafletMap, Marker } from "leaflet"
 import React, {
+  MutableRefObject,
   ReactElement,
   useEffect,
   useRef,
   useState,
-  MutableRefObject,
 } from "react"
 import { ScheduleAdherenceStatus } from "../skate"
 
@@ -93,7 +93,9 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
   })
 
   useEffect(() => {
-    if (!containerRef.current) return
+    if (!containerRef.current) {
+      return
+    }
 
     const map = state.map || Leaflet.map(containerRef.current, mapOptions)
 
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -1,29 +1,28 @@
 import Leaflet, { Map as LeafletMap, Marker } from "leaflet"
 import "leaflet-rotatedmarker"
+import { Label, Orientation, Size } from "./vehicleIcon"
 import React, { MutableRefObject, ReactElement, useEffect, useRef } from "react"
 import { renderToStaticMarkup } from "react-dom/server"
-import { ScheduleAdherenceStatus, ViaVariant } from "../skate"
-import VehicleIcon, { Orientation, Size } from "./vehicleIcon"
 
 interface Props {
   bearing: number
   label: string
   latitude: number
   longitude: number
-  scheduleAdherenceStatus: ScheduleAdherenceStatus
-  viaVariant: ViaVariant | null
 }
+
+const iconAnchor: [number, number] = [9, 3]
 
 const Map = ({
   bearing,
-  label,
+  label: labelText,
   latitude,
   longitude,
-  scheduleAdherenceStatus,
-  viaVariant,
 }: Props): ReactElement<HTMLDivElement> => {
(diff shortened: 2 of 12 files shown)


ingridpierre:
Ah so cool to finally see maps in skate @katehedgpeth!

If this is mostly what we can/want to do for the first pass, then maybe at the very least we can shrink the size of the icon and/or change the default zoom of the mini map to be closer. As-is the icon covers a lot of the streets/map context underneath it.

**My bigger concern/hope:**
Since the whole vehicle/label rotates it's hard to read the labels at a certain point.
image

[Can we use this other vehicle SVG?](https://drive.google.com/file/d/1mxr1Ips6Y7cqQl-TO9mC6goCZwGHxAYJ/view?usp=sharing) And I would love it to more match the new style we're using on the latest shuttle map designs:
image

arkadyan:
@katehedgpeth Just a note that we did have a separate [ticket for drawing the vehicle icon](https://app.asana.com/0/1112935048846093/1114820119722047), so if you want to push that work off to a future PR, feel free. Or if you think it's best to just get it all done in one that's great too.

katehedgpeth:
@ingridpierre I've updated to use the new marker and not rotate the label. The label location currently is just a static distance from the marker -- I'm sure we can write a function to set that distance intelligently based on the marker rotation, but if it's ok with you I'd rather do that as a separate PR.

Screen Shot 2019-07-12 at 3 11 50 PM

ingridpierre:
image
 @katehedgpeth seeing some clipping on the vehicle arrow svg

ingridpierre:
image
it also appears that we're rotating around a center point that is outside of the triangle. Or in other words, rotating the arrow like the hands of a clock instead of like a spinning plate in a microwave.

If this is true we probably want the microwave version (centered point of rotation), since as you can see the arrow and label get farther away from each other depending on how much the arrow rotates.

assets/src/components/map.tsx
```diff
@@ -0,0 +1,111 @@
...
+          }
+        ),
+      ],
+      zoom: 16,
+    })
+
+  const vehicleMarker =
+    existingVehicleMarker ||
+    Leaflet.marker([latitude, longitude], {
+      rotationOrigin: `${iconAnchor[0]}px ${iconAnchor[1]}px`,
```
skyqrose: I see this requires another dependency to do. Would it be possible to do this by rotating the SVG instead of the marker? Could that save us from adding this dependency?

assets/src/components/map.tsx
```diff
@@ -0,0 +1,127 @@
...
+  vehicleMarker.setIcon(icon)
+  vehicleLabel.setIcon(label)
+
+  vehicleMarker.setRotationAngle(bearing)
+
+  return { map, vehicleMarker, vehicleLabel }
+}
+
+const Map = (props: Props): ReactElement<HTMLDivElement> => {
+  const mapRef: MutableRefObject<LeafletMap | null> = useRef(null)
```
skyqrose: I'm not clear on all the differences between `useState` and `useRef`, but I think `useState` might be a little cleaner here.

assets/src/components/map.tsx
```diff
@@ -0,0 +1,127 @@
...
+  useEffect(() => {
+    const { map, vehicleMarker, vehicleLabel } = updateMap(
+      props,
+      mapRef.current,
+      vehicleMarkerRef.current,
+      vehicleLabelRef.current
+    )
+    mapRef.current = map
+    vehicleMarkerRef.current = vehicleMarker
+    vehicleLabelRef.current = vehicleLabel
```
skyqrose: I'm having a little trouble keeping track of where side effects happen. I think I see 3 places:
* Here, saving initialized map/marker/label. (Only happens once)
* When creating the Leaflet map, it attaches itself to the `#map` div. (Only once)
* Setting the updated latlng, bearing, etc on the markers. (Every update)

It might be nice to be able to group them all together a little better.

assets/src/components/map.tsx
```diff
@@ -0,0 +1,127 @@
...
+  })
```
skyqrose: I think this effect will happen on every render, instead of every update. That can be prevented with `useEffect(..., [props])`.

It may also be nice to distinguish between initialization effects that only need to happen once (can be set with `useEffect(..., [])`), and those that need to happen every time the data changes.

assets/css/_vehicle_properties_panel.scss
```diff
@@ -3,6 +3,7 @@
   box-shadow: $modal-box-shadow;
   height: 100%;
   max-width: 100%;
+  overflow: scroll;
```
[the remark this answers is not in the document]
katehedgpeth: yeah, overflow-y is probably a better idea

assets/css/_vehicle_properties_panel.scss
```diff
@@ -135,3 +136,23 @@
   width: 100%;
   z-index: 1000;
 }
+
+.m-vehicle-properties-panel__map {
+  height: 500px;
```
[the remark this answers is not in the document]
katehedgpeth: nope, this was sort of a filler value that i forgot to update

assets/css/_vehicle_properties_panel.scss
```diff
@@ -135,3 +136,23 @@
...
+  margin-top: 1rem;
+}
+
+.m-vehicle-properties-panel__map-label {
+  @include font-small;
+  font-weight: bold;
```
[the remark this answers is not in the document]
katehedgpeth: sure, i'll give that a try

assets/src/components/map.tsx
katehedgpeth: originally i had everything inside of an anonymous function in the `useEffect` call, to keep it all consolidated. But I found it very hard to test that anything leaflet-related was happening since they're all side-effects, so i put the update part into its own exportable function to make it easier to test.

assets/src/components/map.tsx
katehedgpeth: I'm not clear on the differences either -- I found a tutorial where they used refs so that's what i went with. i'll try `useState` and see how it works.

assets/src/components/map.tsx
katehedgpeth: I had trouble rotating the marker initially using transform, but the code was pretty different at that point. i'll try it again.

assets/src/components/map.tsx
katehedgpeth: I gave this a shot and i'm getting `Map container not found` errors in the tests, but not when i run the server. Presumably the useEffect callback is getting called earlier in tests? I don't think I know enough about the react lifecycle to pinpoint exactly why that's happening. Do you have any ideas?

katehedgpeth: nevermind, i think that's related to lat/lng not being set in some of the tests.

katehedgpeth:
@skyqrose I was able to resolve the testing issue I encountered by using a ref for the container. I've made a bunch of updates -- please take another look when you have a chance!

assets/src/components/map.tsx
```diff
@@ -0,0 +1,131 @@
...
+          html: `<svg>
+            <text class="m-vehicle-icon__label">${props.label}</text>
+          </svg>`,
+          iconAnchor: [0, -20],
+        }),
+      }).addTo(map)
+
+    const updated = updateMap(props, { map, vehicleIcon, vehicleLabel })
+
+    updateState(updated)
```
skyqrose: `updateMap` mutates `map` `vehicleIcon` and `vehicleLabel`, right? If so, I think you only need to call `updateState` once when they're initialized, and you wouldn't need to return them from `updateMap`.

assets/src/components/map.tsx
```diff
@@ -0,0 +1,131 @@
...
+      </svg>`,
+    iconAnchor,
+  })
+
+  map.setView([latitude, longitude], zoom)
+
+  vehicleIcon.setLatLng([latitude, longitude])
+  vehicleLabel.setLatLng([latitude, longitude])
+
+  vehicleIcon.setIcon(icon)
```
skyqrose: If the `label` changes (won't happen as vehicle ids, but could happen if we change the label to represent run numbers or badge numbers or something else), then it'd need to be set like the vehicleIcon is.

I think defining the label svg up here might also make it a little easier to see how the label svg and icon svg relate to each other. (i.e. how they're positioned relative to the latlng)

assets/src/components/map.tsx
```diff
@@ -0,0 +1,131 @@
...
+            <text class="m-vehicle-icon__label">${props.label}</text>
+          </svg>`,
+          iconAnchor: [0, -20],
+        }),
+      }).addTo(map)
+
+    const updated = updateMap(props, { map, vehicleIcon, vehicleLabel })
+
+    updateState(updated)
+  }, [props])
```
skyqrose: Potential bug: If the container doesn't exist yet, it might wait until the data changes to redo the effect and draw the map. If that's a problem, this could get a dependency on `containerRef` or `containerRef.current`, or maybe use `useLayoutEffect`. Anecdotally, I've only seen a fraction of a second wait for the map to render, which makes me think that this isn't actually a problem, but it might be worth double checking.

assets/src/components/map.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    const vehicleLabel =
+      state.vehicleLabel ||
+      Leaflet.marker([props.latitude, props.longitude], {
+        icon: Leaflet.divIcon({
+          className: "m-vehicle-map__label",
+          html: `<svg>
+            <text class="m-vehicle-icon__label">${props.label}</text>
+          </svg>`,
+          iconAnchor: [0, -20],
```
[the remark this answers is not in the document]
katehedgpeth: these values hadn't been changed since the icon got changed a while back -- i've updated it to be positioned at `[12, -24]` which centers the label underneath the marker.

katehedgpeth:
@skyqrose this is ready for another look when you have a chance

PR #100, opened 2019-07-19
Move data out of priv folder

katehedgpeth:
No ticket. For some reason, prod was claiming that `Application.app_dir(:skate, "/priv/data")` was not a valid folder, so I've moved the data folder up one level and used `File.cwd!` to access it. I've also changed the data functions so that the files are read at compile time instead of at runtime.


lib/realtime/headway.ex
```diff
@@ -1,7 +1,46 @@
 defmodule Realtime.Headway do
+  require Logger
```
arkadyan: Are these Logger requires actually used anywhere? (another below)

PR #101, opened 2019-07-19
Filter out buses whose block is over from the ladder

arkadyan:
Criteria:
* No Swiftly assignment (the existing "is off course" status).
* Block finished more than an hour ago according to the schedule.

We give the buffer so that we don't tag a bus that is on a detour and
running late, and should thus still be shown on the route ladder.

We are preserving the vehicle in the data looking forward to a future
where you might be able to search for vehicles outside of the route
ladder context.

Asana ticket: [🐞 Filter out buses that are likely done with their block](https://app.asana.com/0/1112935048846093/1131363413413336)

Dev (without fix):
Screen Shot 2019-07-19 at 11 42 13

Local (with fix):
Screen Shot 2019-07-19 at 11 42 16


test/gtfs/block_test.exs
```diff
@@ -19,42 +19,62 @@ defmodule Gtfs.BlockTest do
...
-    ]
+  describe "start_time/1" do
+    test "returns the time of the first stop of the first trip for this block" do
+      assert Block.start_time(block()) == 1
+    end
+  end
+
+  describe "end_time/1" do
+    test "returns the time of the last stop of the last trip for this block" do
+      assert Block.end_time(block()) == 2
```
skyqrose: This can just be `[@trip]`, and `block()` isn't needed. The other tests go through the pipeline that you've abstracted into `block()` because they're testing `group_trips_by_block`, not because the pipeline necessary in order to get a block.

lib/util/time.ex
```diff
@@ -147,4 +133,28 @@ defmodule Util.Time do
...
   end
+
+  @doc """
+  Converts a timestamp into a Date.
+
+  iex> Util.Time.date_of_timestamp(1546362000)
+  ~D[2019-01-01]
+  """
+  @spec date_of_timestamp(timestamp()) :: Date.t()
+  def date_of_timestamp(timestamp) do
```
skyqrose: #107 will have a merge conflict with this, since it also makes the same function. (No action needed here. This comment is for my own reference.)

lib/realtime/vehicle.ex
```diff
@@ -214,6 +225,49 @@ defmodule Realtime.Vehicle do
...
+  @doc """
+  Check whether the vehicle is off course. If so, check if the assigned block
+  was scheduled to end over an hour ago. We give the buffer so that we don't
+  tag a bus that is on a detour and late, and should thus still be shown on the
+  route ladder.
+  """
+  @spec active_block?(boolean(), Block.t(), Util.Time.timestamp()) :: boolean()
+  def active_block?(false, _block, _now), do: true
+
+  def active_block?(true, block, now) do
```
skyqrose: Unnamed boolean parameters can be pretty mysterious and easy to call incorrectly.

That parameter may not even be necessary. I think `active_block?` makes sense even for on course buses. This field could get populated for all buses and then the frontend would check `vehicle.isOffCourse && !vehicle.activeBlock`.

(Even if you don't end up using `is_off_course` to calculate `active_block`, I still like having the off course calculations on the backend.)

test/gtfs/block_test.exs
arkadyan: 👍 thanks

lib/realtime/vehicle.ex
arkadyan: Good point. I think it's important to keep all that logic in a single place, but I've added a name to the parameter for more clarity.

PR #103, opened 2019-07-19
 show bunched/gapped colors on route ladders

katehedgpeth:
Asana ticket: [Use the vehicle bunched/gapped property to draw colored lines on the ladder](https://app.asana.com/0/1112935048846093/1130519303015343)

### DIFF
--- a/assets/src/components/ladder.tsx
+++ b/assets/src/components/ladder.tsx
@@ -7,6 +7,7 @@ import {
 } from "../models/ladderVehicle"
 import { Timepoint, Vehicle, VehicleId, VehicleTimepointStatus } from "../skate"
 import { selectVehicle } from "../state"
+import HeadwayLines from "./headwayLines"
 import { Orientation, Size, VehicleIconSvgNode } from "./vehicleIcon"
 
 export interface Props {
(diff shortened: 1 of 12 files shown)


ingridpierre:
1. As a polish item, can we color the stop stroke according to the headway too? This looks a little funky to me:
image
----
2. Can we have a min-height element or something on the line to show severely bunched buses that are right on top of each other?
image
----
3. In addition to coloring the headway calculated vehicle triangles that darker grey, can we make their lines that default light grey as well? The line is generally helpful to show scheduled locations, but we don't want to mix the colors for the different adherence criteria.
**Current:**
image
**Desired:**
image

ingridpierre:
I think we may need to tweak our logic for bunching. Looking at the example below, I believe that this should have shown only the small segment where the vehicles are as bunched, with all the rest of the line being blue for gapped (since there's no vehicle and things are very late to the line).

image

katehedgpeth:
@arkadyan I've made a few updates, could you give this another review?

katehedgpeth:
@ingridpierre I've tweaked the logic for how the down lines are drawn, and i've also added an id to the lines so you can see which bus each line belongs to if you inspect it. Can you give another look?

assets/src/components/ladder.tsx
```diff
@@ -285,4 +297,9 @@ function partition<T>(items: T[], testFn: (value: T) => boolean): T[][] {
   )
 }
 
+const statusClass = (
+  headwaySpacing: HeadwaySpacing,
+  status: VehicleAdherenceStatus
+): string => (headwaySpacing === null || status === "off-course" ? status : "")
```
arkadyan: I missed this before, but is the `status === "off-course"` check necessary? Would it be just the same if it were `headwaySpacing === null ? status : ""`?

katehedgpeth: no -- i originally had it that way, but then realized that off-track vehicles still get a `headwaySpacing` value so were being colored gray instead of black.

arkadyan: Oh I see, I was thinking about it backwards. Thanks.

PR #105, opened 2019-07-22
Use laboratory to allow us to control feature flags

arkadyan:
Go to /_flags to enable or disable active flags.

Asana ticket: [Feature flags](https://app.asana.com/0/1112935048846093/1132105244340199)

### DIFF
--- a/lib/skate_web/templates/page/index.html.eex
+++ b/lib/skate_web/templates/page/index.html.eex
@@ -1,1 +1,1 @@
-<div id="app" data-user-token="<%= assigns[:user_token] %>" data-laboratory-features="<%= Jason.encode!(assigns[:laboratory_features]) %>"></div>
+<div id="app" data-user-token="<%= @user_token %>" data-laboratory-features="<%= Jason.encode!(@laboratory_features) %>"></div>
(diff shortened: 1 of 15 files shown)


katehedgpeth:
do you want to try connecting the front end as part of this PR, or are you planning to do that separately?

arkadyan:
@katehedgpeth That's a good point. I should just do that now.

arkadyan:
@katehedgpeth I added control over the headway colors using the flag. Codecov seems like it's having trouble, but could you take another look when you get a chance, please?

lib/skate_web/templates/page/index.html.eex
```diff
@@ -1 +1 @@
-<div id="app" data-user-token="<%= assigns[:user_token] %>"></div>
+<div id="app" data-user-token="<%= assigns[:user_token] %>" data-laboratory-features="<%= Jason.encode!(assigns[:laboratory_features]) %>"></div>
```
katehedgpeth: `:laboratory_features` will always be assigned, right? in which case it's safe to use `@laboratory_features` in this template

arkadyan:
@katehedgpeth good catch, thanks, I added that check.

PR #108, opened 2019-07-23
identify users in FullStory

katehedgpeth:
Asana ticket: [Connect usernames to FullStory](https://app.asana.com/0/1112935048846093/1130673090389400)

[Fullstory documentation](https://help.fullstory.com/develop-js/137379) recommends not using strings with personal information for the UID, which is why I've encrypted the username for the UID.

Confirmed that this is working:
Screen Shot 2019-07-23 at 11 47 40 AM


katehedgpeth:
@arkadyan I'm not sure how to fix the test coverage -- in the test environment, `AuthManager.Plug.current_resource(conn)` returns `%{}` instead of a string. Do you know if there's some way to change that, like some kind of config or something?

arkadyan:
@katehedgpeth I'm pretty sure those test values are defined by our [Fake](https://github.com/mbta/skate/blob/master/lib/skate/ueberauth/strategy/fake.ex) strategy module. I haven't messed with that much so this is a guess, but I wonder if putting data into the `info/1` return will make it end up in your `current_resource` call?

katehedgpeth:
@arkadyan turns out `%{}` was what was being passed to `AuthManager.encode_and_sign`, which I totally missed.

arkadyan:
@katehedgpeth Oh, good catch! I guess we just didn't care previously.

lib/skate_web/controllers/page_controller.ex
```diff
@@ -2,11 +2,29 @@ defmodule SkateWeb.PageController do
...
+
     _ = Logger.info("uid=#{uid}")
 
-    render(conn, "index.html")
+    conn
+    |> assign(:user_info, user_info(uid))
+    |> render("index.html")
+  end
+
+  defp user_info(id) when is_binary(id) do
```
arkadyan: Is the `when is_binary(id)` guard necessary?

arkadyan:
@katehedgpeth one more thing: should those salts be in env vars as well?

katehedgpeth:
@arkadyan The screenshot is showing the display name. apparently that's ok to keep as PII, according to their documentation. I put the salts into env vars, good catch.

arkadyan:
@katehedgpeth sorry I keep forgetting things, but could you please add those 2 new vars to `.env.example` as well to let folks know they are required?

PR #109, opened 2019-07-23
add tab bar & faq page

katehedgpeth:
Asana tickets: 
[Shuttle Map | Add a tab bar](https://app.asana.com/0/1112935048846093/1130915127668094)
[Put FAQ and Feedback form link into a new About Skate tab](https://app.asana.com/0/1112935048846093/1132105244340201)


ingridpierre:
Did we decide to use the copy from the FAQ document only? I was hoping to have the little card-style CTAs and things to keep the different actions separate and surface feedback links like I showed at design review.

mockup:
https://zpl.io/bAWvqkP

ingridpierre:
Regarding the tabs:
- Can we make the tabs a square ratio?
- Maeg and I were hoping the collapse (</>) arrow button also collapsed the tab bar, so we can save as much space as possible in the collapsed state.

katehedgpeth:
@ingridpierre I forgot there was a mockup, my mistake. Should the FAQ link just point to the google doc? And I'm guessing we want to leave off the "new features" section -- we probably need a strategy for how that's going to be maintained

ingridpierre:
@katehedgpeth The new feature section with the little red dot and everything is just for future reference. We should go with the little message about contacting the help desk for tablet support, FAQ pointing to public Google doc, Feedback form pointing to asana form mbta.com/skate-feedback.

ingridpierre:
@katehedgpeth Also, not sure if this is just on my end of things, but with these changes I can't seem to open vehicle properties anymore.

katehedgpeth:
@ingridpierre I've updated so that the tab bar hides with the route picker. also the vehicle properties panel should show up now.

assets/css/_page.scss
```diff
@@ -0,0 +1,3 @@
+.c-page {
+  margin-left: $tab-bar-width;
```
skyqrose: Instead of positioning everything absolutely, and getting them to fit around each other, would it be better to postition these both as part of the same layout flow, and put them next to each other? There are a lot of things that need to know about `$tab-bar-width` that feel unrelated.

assets/css/_page.scss
```diff
@@ -0,0 +1,3 @@
+.c-page {
```
skyqrose: This means adding a `c-page` class to all the pages' HTML. Another way to do this would be to make this a mixin and include it from the pages' CSS. Which do you think is a better place to do call this?

assets/css/_faq.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.m-faq__link {
+  background-color: $color-primary;
+  border-radius: 4px;
+  color: $white;
+  display: block;
+  padding: 0.5rem;
+  text-align: center;
+  text-decoration: none;
+  width: 100%;
```
skyqrose: Screen Shot 2019-07-24 at 10 26 29

Visual bug: the button has no right margin.

assets/css/_faq.scss
```diff
@@ -0,0 +1,53 @@
...
+  padding: 0.5rem 1rem;
+}
+
+.m-faq__header {
+  @include font-emph;
+  font-weight: bold;
+}
+
+.m-faq__link {
+  background-color: $color-primary;
```
skyqrose: This can include the button mixins from `_ui-kit.scss`.

assets/css/_faq.scss
```diff
@@ -0,0 +1,53 @@
...
+  padding: 2rem;
+
+  @media (min-width: 720px) {
+    width: 50%;
+    height: 100%;
+    position: fixed;
+  }
+
+  @media (min-width: 1200px) {
+    width: 33%;
```
skyqrose: A single `max-width` would be simpler than the media queries and would probably achieve the same goal.

assets/css/_tab_bar.scss
```diff
@@ -0,0 +1,40 @@
...
+  &--active {
+    background: $color-secondary-light;
+  }
+}
+
+.m-tab-bar__icon {
+  fill: $color-secondary-light;
+
+  .m-tab-bar__link--active & {
+    fill: $color-primary;
```
skyqrose: Would `ui-kit` `button-icon` be useful here? It includes hover states.

assets/src/components/app.tsx
```diff
@@ -1,78 +1,95 @@
...
+      <BrowserRoute path="/faq" component={FAQPage} />
+    </div>
+  </BrowserRouter>
+)
 
-      {selectedVehicle && (
-        <VehiclePropertiesPanel
-          selectedVehicle={selectedVehicle}
-          selectedVehicleRoute={vehicleRoute(routes, selectedVehicle)}
+const TabBar = ({
```
skyqrose: Should this get its own file?

katehedgpeth:
@ingridpierre faq is updated

assets/src/components/app.tsx
```diff
@@ -1,78 +1,95 @@
...
+        >
+          <svg
+            viewBox="0 0 48 48"
+            width="24"
+            height="24"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              className="m-tab-bar__icon"
+              d="m24 0a24 24 0 1 0 24 24 24 24 0 0 0 -24-24zm1.71 40.05a3.5 3.5 0 0 1 -4.82 0 3.15 3.15 0 0 1 -1-2.36 3.23 3.23 0 0 1 1-2.38 3.48 3.48 0 0 1 4.82 0 3.24 3.24 0 0 1 1 2.38 3.16 3.16 0 0 1 -1 2.36zm7-20.46a9.55 9.55 0 0 1 -2 2.59c-.79.73-1.57 1.47-2.36 2.22a10 10 0 0 0 -2 2.64 8.43 8.43 0 0 0 -.76 3.96h-4.59a14.32 14.32 0 0 1 .55-4.3 9.62 9.62 0 0 1 1.41-2.91 13.55 13.55 0 0 1 1.84-2c.66-.59 1.27-1.15 1.85-1.7a7.19 7.19 0 0 0 1.35-1.9 5 5 0 0 0 .55-2.45 4.58 4.58 0 0 0 -1.21-3.39 4.45 4.45 0 0 0 -3.27-1.2 4.91 4.91 0 0 0 -3.07 1.12 4.42 4.42 0 0 0 -1.46 3.73h-5.07a8.89 8.89 0 0 1 1.29-5 8.12 8.12 0 0 1 3.47-3 11.37 11.37 0 0 1 4.88-1 9.55 9.55 0 0 1 6.89 2.4 8.47 8.47 0 0 1 2.49 6.38 8 8 0 0 1 -.76 3.81z"
```
skyqrose: All the SVG data makes it harder to see the big picture of how this component works. What do you think about pulling each SVG into a function?

We also have `icon.tsx` that loads icons from files, though that's probably not necessary to hook into here.

assets/src/components/app.tsx
```diff
@@ -1,78 +1,95 @@
...
+            />
+          </svg>
+        </NavLink>
+      </li>
+    </ul>
+  </div>
```
skyqrose: There's no MBTA logo at the bottom. Are you planning on adding that as part of this PR, as part of a future task, or not at all?

assets/src/components/app.tsx
```diff
@@ -1,78 +1,95 @@
...
     <div className="m-app">
-      <RoutePicker routes={routes} selectedRouteIds={selectedRouteIds} />
-
-      <RouteLadders
-        routes={selectedRoutes}
-        timepointsByRouteId={timepointsByRouteId}
-        vehiclesByRouteId={vehiclesByRouteId}
-        selectedVehicleId={selectedVehicleId}
-      />
+      <TabBar {...props} />
```
skyqrose: The `TabBar` only uses one field (`routePickerIsVisible`). I think it's a little cleaner to not pass arguments that won't be used. (Especially if `TabBar` ends up getting its own file.)

assets/css/_tab_bar.scss
```diff
@@ -0,0 +1,40 @@
+.m-tab-bar {
+  background-color: $color-primary;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  // Required for pinning to the left side of the screen on the tablets
+  left: 0;
+  position: fixed;
+  top: inherit;
```
skyqrose: In the Vehicle Properties Panel, you changed `inherit` to `0`. When would you use `0` and when would you use `inherit`?

arkadyan: I think maybe `inherit` seemed required in at least some places to get things (e.g. fixed positioning) to work on the Samsung tablets for reasons I don't fully understand. It would be good to be sure to review all these changes on an actual tablet to be sure everything looks right, especially areas with fixed positioning.

assets/src/components/app.tsx
ingridpierre: https://drive.google.com/file/d/1hkD8ObZrVsKf2FSlAfe4cMaHk6afAj-9/view?usp=sharing I neglected to include this icon in the assets I provided. link above.

ingridpierre:
image
looks like the buttons are extending a little too far. right margins should match the margins on the left.

ingridpierre:
I had Katrina scan the wording, she had a small suggestion:

"We're making updates available all the time." --> We're making updates all the time.

lib/skate_web/router.ex
```diff
@@ -52,6 +52,7 @@ defmodule SkateWeb.Router do
     ]
 
     get "/", PageController, :index
+    get "/faq", PageController, :index
```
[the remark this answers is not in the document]
katehedgpeth: oh, good catch, that's a holdover from when it was just an FAQ. I'll update.

assets/src/components/faqPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+      </div>
+      <div className="m-faq__section">
+        <h2 className="m-faq__header">Feedback</h2>
+        <p>
+          Have feedback, a suggestion, or need other support related to Skate?
+          Use our feedback form and someone on our team will be in touch with
+          you.
+        </p>
+        <p>
+          <a className="m-faq__link" href="" target="_blank">
```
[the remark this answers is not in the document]
katehedgpeth: 🤦‍♀

assets/src/components/routePage.tsx
```diff
@@ -0,0 +1,84 @@
...
+    .reduce((acc, vehicles) => acc.concat(vehicles), [])
+    .find(vehicle => vehicle.id === selectedVehicleId)
+
+const vehicleRoute = (
+  allRoutes: Route[] | null,
+  vehicle: Vehicle | undefined
+): Route | undefined =>
+  (allRoutes || []).find(route => route.id === (vehicle && vehicle.routeId))
+
+const RoutePage = ({
```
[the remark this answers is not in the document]
katehedgpeth: sure, i'm fine with that

assets/css/_page.scss
katehedgpeth: I did try that initially, but ran into problems when trying to get the RouteLadder/TabBar collapse working correctly while also having the TabBar appear _without_ the RouteLadder on the FAQ page. I agree that it's not ideal to have multiple things depend on `$tab-bar-width`, but this was the best solution i found that worked.

assets/css/_faq.scss
katehedgpeth: totally missed those, thanks

assets/css/_faq.scss
katehedgpeth: the problem is on phone-sized screens -- those screens are narrow enough that it looks weird if this page's content isn't full width.

skyqrose: What about `width: 100%; max-width: 600px;`. That'd fill phone screens, wouldn't it? And then the margin on the right would only appear if the screen gets big enough that it's not weird?

katehedgpeth: oh, you're right. i was thinking of using percentages but we can just use a set width.

assets/src/components/aboutPage.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { ReactElement } from "react"
+
+const FAQ = (): ReactElement<HTMLDivElement> => (
```
skyqrose: There's still a few more places where FAQ needs to change to About.

assets/css/_tab_bar.scss
katehedgpeth: I changed `inherit` to `0` in the other files because when it was `inherit` the vehicle properties panel was positioned off of the page. I'm not fully sure if there even needs to be a value for `top` here -- it seems to work fine without it in chrome but i haven't tested without it on the tablets.

assets/css/_about_page.scss
```diff
@@ -0,0 +1,34 @@
...
+.m-about-page__header {
+  @include font-emph;
+  font-weight: bold;
+}
+
+.m-about-page__link {
+  @include button-primary;
+  display: block;
+  text-align: center;
+  text-decoration: none;
```
skyqrose: `text-decoration:none` might be good to drop in `@mixin button-base` so it covers any future links that look like buttons.

Alternatively if we want to be more aggressive, it could go in `base.scss` and apply to all `<a>` tags. (If you do this, it can be removed from `@mixin font-link`.)

assets/src/components/app.tsx
katehedgpeth: T logo has been added

assets/css/_tab_bar.scss
katehedgpeth: no, this is more than a hover state -- this is when the page that the button represents is the active page, so it needs to always be the highlighted color

PR #112, opened 2019-07-25
right swipe on map should not close panel

katehedgpeth:
Asana ticket: [🐞 Right swipe on map should not close the vehicle properties panel](https://app.asana.com/0/1112935048846093/1131945338037530)

This also uncovered a small bug in our detectSwipe module which I fixed.

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -117,7 +117,7 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
 
   return (
     <div
-      id="vehicle-map"
+      id="id-vehicle-map"
       ref={container => (containerRef.current = container)}
       className="m-vehicle-map"
     />
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -117,6 +117,7 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
 
   return (
     <div
+      id="vehicle-map"
       ref={container => (containerRef.current = container)}
       className="m-vehicle-map"
     />
--- a/assets/src/components/vehiclePropertiesPanel.tsx
+++ b/assets/src/components/vehiclePropertiesPanel.tsx
@@ -218,7 +218,14 @@ const VehiclePropertiesPanel = ({
   const hideMe = () => dispatch(deselectVehicle())
 
   useEffect(() => {
-    const handleSwipe = (swipeDirection: SwipeDirection) => {
+    const handleSwipe = (
+      swipeDirection: SwipeDirection,
+      target: HTMLElement | null
+    ) => {
+      if (target && target.id === "vehicle-map") {
+        return
+      }
+
       if (swipeDirection === "Right") {
         hideMe()
       }
--- a/assets/src/components/vehiclePropertiesPanel.tsx
+++ b/assets/src/components/vehiclePropertiesPanel.tsx
@@ -209,6 +209,19 @@ const DataDiscrepancies = ({
   </ul>
 )
 
+export const handleSwipe = (hideMe: () => void) => (
+  swipeDirection: SwipeDirection,
+  target: HTMLElement | null
+) => {
+  if (target && target.id === "id-vehicle-map") {
+    return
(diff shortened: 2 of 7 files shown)


assets/src/helpers/detectSwipe.ts
```diff
@@ -61,10 +64,16 @@ const firstTouchPosition = (event: TouchEvent): Position => {
   return { x: screenX, y: screenY }
 }
 
-const diff = (end: Position, start: Position): Position => ({
-  x: end.x - start.x,
-  y: end.y - start.y,
-})
+const diff = (end?: Position, start?: Position): Position => {
+  if (!start || !end) {
```
skyqrose: Could this explain https://app.asana.com/0/1112935048846093/1132578944531173 and https://app.asana.com/0/1112935048846093/1132578944533173 ?

When would they be undefined? `handleTouchStart` should set them both. Is it ever not called?

assets/src/components/map.tsx
```diff
@@ -117,6 +117,7 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
 
   return (
     <div
+      id="vehicle-map"
```
skyqrose: What do you think about giving ids a `js-` or `id-` prefix so they don't get mixed up with classes used for CSS?

assets/src/helpers/detectSwipe.ts
katehedgpeth: I think what's happening is that the vehicle data gets refreshed between touchStart and touchEnd, and the value of start and end gets reset

assets/src/components/map.tsx
katehedgpeth: sure. i initially used a `js-map` class but then changed my mind and went with an id, but i'm happy to prefix the id with `id-`

assets/src/helpers/detectSwipe.ts
skyqrose: Ah. Should `startingPosition` and `endingPosition` be stored in the VPP state instead of `let` variables, so they don't get forgotten?

assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -218,7 +218,14 @@ const VehiclePropertiesPanel = ({
   const hideMe = () => dispatch(deselectVehicle())
 
   useEffect(() => {
-    const handleSwipe = (swipeDirection: SwipeDirection) => {
+    const handleSwipe = (
+      swipeDirection: SwipeDirection,
+      target: HTMLElement | null
+    ) => {
+      if (target && target.id === "vehicle-map") {
```
skyqrose: This also needs to be updated to `"id-vehicles-map"`

assets/src/helpers/detectSwipe.ts
katehedgpeth: maybe? i don't know how that would interact with the effect lifecycle. this seems to work fine as-is, as far as i have seen.

katehedgpeth:
@skyqrose do you know how to test swipe interactions? i tried this but it didn't seem to work:
```js
const wrapper = mount(<VehiclePropertiesPanel selectedVehicle={vehicle} />)

wrapper
  .find(".m-vehicle-properties-panel")
  .simulate("touchStart", touchEvent({x: 0, y: 0})

wrapper
  .find(".m-vehicle-properties-panel")
  .simulate("touchMove", touchEvent({x: 50, y: 50})

wrapper
  .find(".m-vehicle-properties-panel")
  .simulate("touchStart", touchEvent({x: 100, y: 100})

expect(mockDispatch).toHaveBeenCalledWith(deselectVehicle)

skyqrose:
No clue. @arkadyan was the one who set up the swiping originally.

arkadyan:
@katehedgpeth It's been a while but based on vague recollections and what tests exist in there, I think I figured out how to get "unit" tests working in `detectSwipe.test.tsx`, but wasn't able to figure out how to get anything useful to work in `vehiclePropertiesPanel.test.tsx`.

PR #116, opened 2019-07-29
Filter shuttles from route list

katehedgpeth:
Asana ticket: [Remove shuttle routes from route picker in route ladder view](https://app.asana.com/0/1112935048846093/1131078147057200)

### DIFF
--- a/lib/gtfs/route.ex
+++ b/lib/gtfs/route.ex
@@ -6,6 +6,7 @@ defmodule Gtfs.Route do
 
   @type t :: %__MODULE__{
           id: id(),
+          description: String.t(),
           direction_names: direction_names()
         }
 
@@ -16,6 +17,7 @@ defmodule Gtfs.Route do
 
   @enforce_keys [
     :id,
+    :description,
     :direction_names
   ]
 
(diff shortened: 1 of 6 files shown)


arkadyan:
I think this is going to filter out most of our routes. My understanding is that "Key Bus" is more or less the same as the list of routes we are displaying headway spacing for. I think we just want to blacklist "Rail Replacement Bus" rather than whitelist anything.

katehedgpeth:
I think I am only blacklisting "Rail Replacement Bus", by calling `Enum.reject` [here](https://github.com/mbta/skate/pull/116/files#diff-d233a518f5dba9f8705f9fbbc52ff47bR12) -- or is this incorrect?

arkadyan:
@katehedgpeth Sorry! I think I confused myself. Let me look again but you are right.

skyqrose:
`CSV.parse` takes a filter parameter that lets you filter by a field even before you parse it into an elixir struct. We use this for more performantly filtering out non-bus stop times, and for filtering by `route_type` without having a `route_type` field in the struct.

We could do the same thing here, and not need to add the field to `%Route{}`

skyqrose:
(Oh, wait, we'll eventually want those routes for the shuttle page, so we still need to load them and differentiate in the controller. nvm.)

PR #119, opened 2019-07-30
Add a headway diagram to the Properties Panel

arkadyan:
Asana ticket: [Add a Headway View to the Vehicle Properties Panel](https://app.asana.com/0/1112935048846093/1120980093841761)

Screen Shot 2019-07-29 at 17 02 06

Screen Shot 2019-07-29 at 17 02 21


assets/src/models/vehiclesByRouteId.ts
```diff
@@ -0,0 +1,54 @@
...
+  vehicles: Vehicle[],
+  currentVehicle: Vehicle
+): NextAndPreviousVehicle => {
+  const vehiclesGoingTheSameDirection = byDirection(vehicles)[
+    currentVehicle.directionId
+  ]
+
+  const nextVehicle = vehiclesGoingTheSameDirection.find(
+    vehicle =>
+      vehicle.previousVehicleId === currentVehicle.id && vehicle.isOnRoute
```
katehedgpeth: if I'm reading this correctly, it looks like if the next/previous vehicle is off-route, there will be no next/previous vehicle. should we instead look for the next vehicle after the off-route vehicle?

assets/src/components/vehiclePropertiesPanel.tsx
```diff
@@ -86,7 +104,11 @@ const Header = ({
       <div className="m-vehicle-properties-panel__variant-name">
         {formatRouteVariant(vehicle)}
       </div>
-      <ScheduleAdherence vehicle={vehicle} />
+      {vehicle.headwaySpacing ? (
```
katehedgpeth: does this need to be behind the flag as well?

assets/css/_vehicle_properties_panel.scss
```diff
@@ -77,8 +90,12 @@
   padding: 1rem;
 }
 
+.m-vehicle-properties-panel .m-headawy-diagram,
```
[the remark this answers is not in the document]
arkadyan: Well, for now this headway diagram is only used here, so not technically, but I thought if we ever introduced it anywhere else we probably wouldn't want this style to apply.

assets/css/_headway_diagram.scss
```diff
@@ -0,0 +1,112 @@
+.m-headawy-diagram {
```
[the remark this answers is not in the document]
arkadyan: 😢 indeed

assets/src/components/ladderPageContext.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { useContext } from "react"
```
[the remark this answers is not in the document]
arkadyan: Turns out it does not need to be separate, so I consolidated them.

assets/src/models/vehiclesByRouteId.ts
arkadyan: Yes, that's true. We don't want to show the next or previous vehicle if it's currently in the incoming box. I think if the next vehicle is in the incoming box, the one after that will be too, so I don't think we need to look further. Can you think of a case I'm overlooking?

assets/src/components/vehiclePropertiesPanel.tsx
arkadyan: Good catch! Added that check.

assets/src/models/vehiclesByRouteId.ts
katehedgpeth: oh, i was confusing this with the vehicle being off-course. this seems correct when the vehicel is off-route, but do we still need to check for the vehicle being off-course?

arkadyan: At least for now (https://app.asana.com/0/1112935048846093/1133115437342310 might change it later), we are just taking what Swiftly says was the previous vehicle. So if that is now off course I think that seems fine to still show it.

ingridpierre:
@arkadyan 

- Seeing some seemingly significant discrepancies between the same two vehicles depending on which one i select:

Pasted Graphic 2
Pasted Graphic 3

Other comments copied from slack:
image (4)
image (4)
- Seeing some diagram vs ladder discrepancies that seem significant.

arkadyan:
@ingridpierre Thanks, just pushed 2 bug fixes.

ingridpierre:
image
are we able to change these colors to grey also in this pr for headway routes?

katehedgpeth:
@ingridpierre we have a separate PR (#118) open for updating the icon color

ingridpierre:
image
Also seeing the line offset with its cap when there's no vehicle present at the ends of the diagram.

arkadyan:
@ingridpierre what browser is that in?

ingridpierre:
@arkadyan Chrome, on my Mac.

arkadyan:
@ingridpierre I changed how I handle the ends of the lines to avoid needing that circle, so that issue should be taken care of.

PR #396, opened 2020-01-02
single label field for LadderVehicle

skyqrose:
Followup to #379. This is what I meant by calling `vehicleLabel()` when making the `LadderVehicle` so that it doesn't need the `runId` and `endOfTripType` fields and so we don't need to do the type coercion to pass a `LadderVehicle` into `vehicleLabel()`. (The "label" field remains, but it's changing meaning from the bus number from the `VehiclePosition.label` to the string we show next to the triangle on the ladder.)


skyqrose:
Ah, yeah, I like the idea of passing a vehicle to the icon as a separate change, but there are a couple places where we want the behavior to depend on something other than the vehicle.

For most things, the direction of the icon depends on the LadderDirection.
In the incoming box, we don’t use a label, because we draw the text on the side instead.
And for vehicles incoming from another route, we should probably be showing the direction/variant of the upcoming trip, not of the vehicle ([task](https://app.asana.com/0/1152935389650390/1134321538456380)) which would require passing different arguments to the icon.

So we could probably make it work, but with a bit more difficulty than this, and it'd still be good to avoid the `ladderVehicle as Vehicle` coercion there, too.

PR #408, opened 2020-01-13
Refactor Busloc+Swiftly merging

skyqrose:
## What?

Simplifies a bunch of the merging / concentrate logic, creates specific types for the data we get from swiftly and busloc, and cuts out some middlemen in the data flow.

IMG_20200112_093724

Significant changes:
* Busloc and Swiftly each get their own structs to go with their parsers (`Concentrate.Busloc and Concentrate.Swiftly`)
* Removes `TripUpdate`, `StopTimeUpdate`, and `merge_group`. The Swiftly Api doesn't use either. Busloc doesn't make `StopTimeUpdate`. Busloc does make `TripUpdate`, but we don't use it, so now it lives on as a private `Concentrate.Busloc.TripDescriptor`.
* Replaces `Vehicle.from_vehicle_position(%VehiclePosition{})` with `Vehicle.from_sources(%{swiftly: %Swiftly{}, busloc: %Busloc{}})`. There's no more middle step where we make a combined `%VehiclePosition{}` before turning it into a `%Vehicle{}`.
* Moves the logic from `Consumer.VehiclePositions` into `Merge`, so Merge is the final consumer. Since we only have the one consumer and don't need backpressure there, this fits the GenStage best practices better.

Most of the new code is in `vehicle.ex` where the new logic for combining Swiftly+Busloc is, and in `merge.ex` which now keeps track of what comes from where and calls `Vehicle.from_sources`.

## Why?

* More specific types. The Busloc parser no longer returns a struct that has a `headway_secs` or `stop_id` field that it can never fill.
* More explicit control over where data comes from. Preferring the Swiftly trip assignment is now just `swiftly && swiftly.trip_id || busloc && busloc.trip_id`.
* Adding new fields (like block waivers) to just one source becomes easier.
* Net 1500 lines of code removed.


skyqrose:
With #442, we finally have a place that uses stop time updates and merge groups that this PR removed. But I think merge groups are still overkill for what we're doing. VPs are still totally separate from STUs. TUs are paired with STUs but we only use it to get a trip_id that the STUs already have.

I'll take a look at what #442 would look like in the world of this PR. I think that this PR will still lead to massive cleanup, even if block waivers require a more complicated merge with STUs added back in.

PR #416, opened 2020-01-15
rotate VPP icon to match orientation on ladder

skyqrose:
Asana Task: [Orient triangle in VPP to match the ladder](https://app.asana.com/0/1148853526253426/1126875696911113)

Screen Shot 2020-01-15 at 17 28 28
Screen Shot 2020-01-15 at 17 28 39

### DIFF
--- a/assets/src/components/propertiesPanel/header.tsx
+++ b/assets/src/components/propertiesPanel/header.tsx
@@ -4,8 +4,8 @@ import vehicleLabel from "../../helpers/vehicleLabel"
 import {
   directionOnLadder,
   getLadderDirectionForRoute,
-  LadderDirections,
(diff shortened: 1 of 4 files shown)


skyqrose:
I added one test for shuttles, but codecov is still failing because I don't test both a left-facing bus and a right-facing bus. Can you force merge?

arkadyan:
@skyqrose is there a reason not to just test both?

skyqrose:
It would be verbose without adding any additional safety.

arkadyan:
@skyqrose verbosity is what test code is all about.

I think if you wanted to write unit tests that would work too, but I agree with Codecov that this code should be completely tested.

PR #424, opened 2020-01-22
return default for failed api calls instead of crashing

skyqrose:
Deduplicates a lot of work writing and testing api calls.

Followup to https://github.com/mbta/skate/pull/415#discussion_r367615946

It's built on top of the branch for that PR, so until it merges, this PR will be a bit messy. The only new changes are in `api.ts` and `api.test.ts`.

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -33,15 +33,21 @@ const checkResponseStatus = (response: Response) => {
 
 const parseJson = (response: Response) => response.json()
 
-export const apiCall = <T>(url: string, parser: (data: any) => T): Promise<T> =>
+export const apiCall = <T>(url: string, parser: (data: any) => T, defaultResult?: T): Promise<T> =>
   fetch(url)
     .then(checkResponseStatus)
     .then(parseJson)
     .then(({ data: data }: { data: any }) => parser(data))
     .catch(error => {
-      // tslint:disable-next-line: no-console
-      console.error(error)
-      throw error
+      if (defaultResult === undefined) {
+        // tslint:disable-next-line: no-console
+        console.error(error)
+        throw error
+      } else {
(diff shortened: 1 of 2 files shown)


skyqrose:
Would calling it with named arguments in an object make it easier to understand the interface? like
```ts
export const fetchShapeForTrip = (tripId: TripId): Promise<Shape | null> =>
  apiCall({
    url: `/api/shapes/trip/${tripId}`,
    parser: (shape: Shape | null) => shape,
    defaultResult: null,
  })
```

arkadyan:
That does feel like an improvement to me. Do you agree?

assets/src/api.ts
```diff
@@ -45,6 +33,31 @@ const checkResponseStatus = (response: Response) => {
...
+    .then(parseJson)
+    .then(({ data: data }: { data: any }) => parser(data))
+    .catch(error => {
+      if (defaultResult === undefined) {
+        // tslint:disable-next-line: no-console
+        console.error(error)
+        throw error
+      } else {
+        // tslint:disable-next-line: no-console
+        console.warn(error)
```
arkadyan: We should listen to tslint and not call console here. Calling console is useless, and this is also doing what we want it to.

assets/tests/api.test.ts
```diff
@@ -23,6 +24,80 @@ const mockFetch = (status: number, json: any): void => {
...
+
+    apiCall({
+      url: "/",
+      parser: () => null,
+    })
+      .then(() => {
+        done("fetchRoutes did not throw an error")
+      })
+      .catch(error => {
+        expect(error).not.toBeUndefined()
```
arkadyan: There's also a [`.toBeDefined()`](https://jestjs.io/docs/en/expect#tobedefined)

PR #425, opened 2020-01-22
Pause channels if not on the relevant page

skyqrose:
Asana Task: [Pause channels if not on the relevant page](https://app.asana.com/0/1148853526253426/1137879685980033)

This moves the `useVehicles`/`useShuttleVehicles` calls into `LadderPage`/`ShuttleMapPage`, mirroring the way `useSearchResults` is in `SearchPage`.

An alternative way to do this would be to keep the hooks in `AppStateWrapper`, and then deactivate them somehow based on the route, but I think that'd be a lot more complicated, since `AppStateWrapper` doesn't have easy access to the routing.


arkadyan:
Needs a rebase

assets/src/components/ladderPage.tsx
```diff
@@ -60,14 +64,17 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
...
+        vehiclesByRouteId={vehiclesByRouteId}
         selectedVehicleId={selectedVehicleId}
       />
 
       {selectedVehicleOrGhost && (
-        <PropertiesPanel
-          selectedVehicleOrGhost={selectedVehicleOrGhost}
-          route={vehicleRoute(routes, selectedVehicleOrGhost)}
-        />
+        <VehiclesByRouteIdProvider vehiclesByRouteId={vehiclesByRouteId}>
```
arkadyan: Why not just wrap everything here with this and not have to change RouteLadders?

skyqrose: If it's just going one level deep, normal arguments are better than contexts. Contexts are good for avoiding prop drilling, but that's for the benefit of the intermediate components that would do nothing but forward the argument along. For the consumer, RouteLadders, getting the value from an argument is more direct and less magic than getting it from a context.

arkadyan: I can see how if you are only passing a piece of data into a single component why you wouldn’t need the overhead of a context, but once you already have a context it feels like consistency becomes much more important. For the consumer it’s much simpler to have one way of accessing this data than to think about different choices each time.

PR #426, opened 2020-01-22
persist selectedVehicleId. keep VPP open.

skyqrose:
Asana Task: [Save selected vehicle. Keep VPP open through refresh](https://app.asana.com/0/1148853526253426/1157713971731881)


skyqrose:
Same way it's currently handled when a selected vehicle disappears while the app is open: Anywhere that tries to use the selected vehicle treats not being able to find a selected vehicle the same as if the vehicle doesn't exist.

skyqrose:
And the workaround (deselect it) is so easy+familiar that people might not even realize something went wrong. I'm not worried at all, I don't think it's even worth making the followup task for it.

PR #436, opened 2020-01-30
show disconnected modal when socket disconnects

skyqrose:
Screenshot_20200130-115109

It's on dev right now.

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -1,7 +1,10 @@
 import React, { ReactElement, useContext } from "react"
 import { BrowserRouter, Route as BrowserRoute } from "react-router-dom"
+import { SocketContext } from "../contexts/socketContext"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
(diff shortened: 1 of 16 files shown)


assets/src/components/app.tsx
```diff
@@ -24,6 +28,9 @@ const App = (): ReactElement<HTMLDivElement> => {
         <BrowserRoute exact={true} path="/settings" component={SettingsPage} />
         <BrowserRoute path="/about" component={AboutPage} />
         <BrowserRoute exact={true} path="/search" component={SearchPage} />
+        {connectionStatus === ConnectionStatus.Disconnected ? (
```
arkadyan: It's also possible to do:

    {connectionStatus === ConnectionStatus.Disconnected && ( <DisconnectedModal /> )}

if you wanted to avoid the `null`.

skyqrose: eh, I don't mind the `null` since it's explicitly saying "don't draw anything".

PR #438, opened 2020-02-03
use static_path on dev for easier mobile testing

skyqrose:
`static_url` was saying assets were at `localhost`, which means they weren't loading when connecting over wifi from a mobile device.


skyqrose:
It broke the deploy so I changed the configuration a bit, but I just deployed it to dev and it's working.

PR #449, opened 2020-02-11
Update mix lock file after upgrading hex

arkadyan:
Hex version 0.20.5.


arkadyan:
@skyqrose I got this change running `mix deps.get` after updating hex like it recommended that I do (`mix local.hex`). It looks like hex added a piece of data to each entry in the file, but I don't know anything about the details.

PR #457, opened 2020-02-18
Display block waivers in vehicle and ghost properties panels

arkadyan:
Asana tickets:
[Vehicle states: ghost buses (type B) with block waiver](https://app.asana.com/0/1148853526253426/1128793574082088)
[Vehicle states: early/on-time/late with block waiver](https://app.asana.com/0/1148853526253426/1128793574082084)

Put behind a feature flag.

Import block waiver data for Vehicles and Ghosts.

Highlight vehicles on the route ladder that have block waivers. This is
quick & dirty styling not based on any design. We'll improve this in
future PRs.

~~Depends on https://github.com/mbta/skate/pull/442~~

Screen Shot 2020-02-18 at 15 19 28


assets/tests/components/propertiesPanel/blockWaivers.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  formatTimeOfDay,
+  hours12,
+  nowTimeOfDay,
+} = blockWaivers
+
+describe("BlockWaivers", () => {
+  test("renders", () => {
+    const blockWaiver: BlockWaiver = {
+      startTime: 18300,
+      endTime: 45480,
```
skyqrose: Without mocking `now`, will this snapshot change depending on when you run the test?

assets/tests/components/propertiesPanel/blockWaivers.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  CurrentFuturePastType,
+  currentFuturePastClass,
+  currentFuturePastTitle,
+  currentFuturePastType,
+  formatTimeOfDay,
+  hours12,
+  nowTimeOfDay,
+} = blockWaivers
+
+describe("BlockWaivers", () => {
```
skyqrose: I think snapshots for a past waiver and a future waiver would test the same functionality as some of the whitebox tests below without needing to expose and be sensitive to so much implementation detail.

assets/tests/components/ladder.test.tsx
```diff
@@ -538,9 +548,10 @@ describe("ladder", () => {
     const mockDispatch = jest.fn()
 
     const timepoints = ["t0", "t1", "t2"]
-    const incomingGhost: Ghost = {
+    const incomingGhost: Ghost = ({
       id: "ghost-incoming-y0622",
-    } as Ghost
+      blockWaivers: [],
+    } as unknown) as Ghost
```
skyqrose: Same question as in `routeLadder.test.tsx`. Why is the `as unknown` needed now?

assets/css/properties_panel/_block_waivers.scss
```diff
@@ -0,0 +1,55 @@
+.m-block-waiver {
```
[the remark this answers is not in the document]
arkadyan: Separated out BlockWaiverList and BlockWaiverBanner components for clarity.

assets/src/components/propertiesPanel/blockWaivers.tsx
```diff
@@ -0,0 +1,138 @@
...
+    {blockWaivers.map(blockWaiver => (
+      <BlockWaiver
+        blockWaiver={blockWaiver}
+        key={`${blockWaiver.startTime}-${blockWaiver.endTime}`}
+      />
+    ))}
+  </div>
+)
+
+export default BlockWaivers
```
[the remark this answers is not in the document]
arkadyan: Separated out BlockWaiverList and BlockWaiverBanner components for clarity.

assets/src/components/propertiesPanel/blockWaivers.tsx
```diff
@@ -0,0 +1,138 @@
...
+            {formatTimeOfDay(blockWaiver.endTime)}
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+)
+
+const BlockWaivers = ({ blockWaivers }: Props) => (
+  <div className="m-block-waivers">
```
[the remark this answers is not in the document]
arkadyan: I like this style to help guide me through the component structure.

assets/src/components/propertiesPanel/blockWaivers.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const minutesNum = Math.floor((timeOfDay % 3600) / 60)
+  const minutesStr = minutesNum < 10 ? `0${minutesNum}` : minutesNum
+
+  return `${hours}:${minutesStr}${ampm(hours24plus)}`
+}
+
+export const nowTimeOfDay = (now = new Date()): number =>
+  now.getHours() * 3600 + now.getMinutes() * 60 + now.getSeconds()
+
+export enum CurrentFuturePastType {
```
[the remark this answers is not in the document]
arkadyan: I'd be very happy for a better name for this, but personally, these don't feel like they make things more clear.

assets/css/properties_panel/_block_waivers.scss
```diff
@@ -0,0 +1,55 @@
...
+  margin-bottom: 0.5rem;
+  text-align: left;
+  width: 7rem;
+}
+
+.m-block-waiver__detail-value {
+  @include font-body;
+  margin-bottom: 1rem;
+
+  .highlighted {
```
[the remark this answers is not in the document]
arkadyan: Removed

assets/static/images/icon-alert-circle.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg"><path d="m24 0a24 24 0 1 0 17 7 23.93 23.93 0 0 0 -17-7z" fill="#fff" fill-rule="evenodd"/><circle cx="24" cy="24" fill="#606060" r="22.59"/><path d="m23.89 46.59a22.59 22.59 0 1 0 -22.48-22.7 22.59 22.59 0 0 0 22.48 22.7z" fill-rule="evenodd"/><g fill="#fff"><path d="m20.39 4.42h7.22a1.81 1.81 0 0 1 1.49 2.11l-1.49 23.22a1.73 1.73 0 0 1 -1.49 1.78h-4.24a1.73 1.73 0 0 1 -1.49-1.78l-1.49-23.22c-.07-1.13.61-2.11 1.49-2.11"/><path d="m21.57 34.54h4.87a1.88 1.88 0 0 1 1.82 1.93v5.17a1.89 1.89 0 0 1 -1.82 1.94h-4.87a1.89 1.89 0 0 1 -1.83-1.94v-5.17a1.88 1.88 0 0 1 1.83-1.93" fill-rule="evenodd"/></g></svg>
```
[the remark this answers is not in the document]
arkadyan: Do you think it's reasonable to think about this when implementing the on-ladder icons? I feel like the person implementing that will have a better ability to make these decisions because they will be working through whatever needs there are at that point.

assets/tests/components/routeLadder.test.tsx
```diff
@@ -496,7 +500,8 @@ describe("groupByPosition", () => {
           fractionUntilTimepoint: 0.2,
         },
       },
-    } as Vehicle
+      blockWaivers: [],
+    } as unknown) as Vehicle
```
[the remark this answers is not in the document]
arkadyan: As best as I can guess some upgrade made the linter more particular now and it now thinks this object is too far away from a Vehicle.

assets/tests/components/propertiesPanel/blockWaivers.test.tsx
arkadyan: No, if you look at the snapshot the time is the absolute start time and end time provided in the block waiver. (Also, it's easy to know if you need to mock now because running the tests a second time—including through Semaphore—will always fail.)

assets/tests/components/propertiesPanel/blockWaivers.test.tsx
arkadyan: I'm trying to make sure I'm understanding this properly. Are you referring to the tests of various functions, e.g. `formatTimeOfDay`? Those were very useful to me in developing those functions and I think would still serve an important purpose if we wanted to refactor. But I think more importantly, are you suggesting there are more rendering tests you'd like to see?

ingridpierre:
Even with the flag on, vehicles with known block waivers are missing the banner.

Am I missing something?

arkadyan:
@ingridpierre I forgot to mention, you'll also have to run `mix cache.clean`. Sorry I keep forgetting to say that!

assets/static/images/icon-alert-circle.svg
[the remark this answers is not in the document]
skyqrose: Good point. We'll need to keep in mind that now there are 2 places to update when setting those colors (in the SVG and in `_block_wavier.scss`) but that will be easier than trying to predict what styling we'll need in the future.

assets/tests/components/propertiesPanel/blockWaivers.test.tsx
skyqrose: The time formatting makes sense to have unit tests for, since there's a lot of detail there that you don't want to have to cover in the bigger `BlockWaiverBanner` tests.

I'm thinking about `currentFuturePastType()`, `currentFuturePastClass()`, and `currentFuturePastTitle()` tests. In all of those functions, all 3 branches would be covered by `BlockWaiverBanner` tests for the past, present, and future.

assets/src/components/propertiesPanel/blockWaivers.tsx
```diff
@@ -0,0 +1,138 @@
...
+  Current = 1,
+  Future,
+  Past,
+}
+
+export const currentFuturePastType = ({
+  startTime,
+  endTime,
+}: BlockWaiver): CurrentFuturePastType => {
+  const now = nowTimeOfDay()
```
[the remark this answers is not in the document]
arkadyan: Yep, that is better. Converted to epoch on the backend.

assets/tests/components/propertiesPanel/blockWaivers.test.tsx
arkadyan: Ah, got it. Added rendering tests for all three cases.

assets/tests/components/routeLadder.test.tsx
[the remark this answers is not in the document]
skyqrose: I looked at this and it's weird.
It sees `blockWaivers: []`, and not knowing that it's a `Vehicle` yet, gives the list type `never[]`.
Then it tries to convert `{..., blockWaivers: never[]}` to `Vehicle` and can't.

A safer way to nudge typescript in the right direction would be to do `blockWaivers: [] as BlockWaiver[]` and leave the `} as Vehicle` without the `unknown` step.

That way we stay protected from future errors like misspelling a new field and losing time chasing down why the test can't find that field.

ingridpierre:
@arkadyan Seeing this without styling by time period yet made me realize that we don't need the word "Notice" in "Future Notice" and "Past Notice". Removing it may help the scanability here I hope.

Otherwise looks good to my eyes.

I've also made that small change to the docs: https://zeroheight.com/3ccfe55fe/p/2132a0-overview/t/72522b

assets/tests/components/routeLadder.test.tsx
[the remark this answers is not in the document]
arkadyan: Oh great, updated.

arkadyan:
@ingridpierre Sounds good to me, removed the word "Notice".

assets/src/components/propertiesPanel/blockWaiverBanner.tsx
```diff
@@ -0,0 +1,128 @@
+import React from "react"
+import { alertCircleIcon } from "../../helpers/icon"
+import { BlockWaiver } from "../../realtime"
+
+interface Props {
+  blockWaiver: BlockWaiver
+}
+
+export const hours12 = (hours24plus: number): number => {
```
skyqrose: Since this is now using the output of `date.getHours()` instead of the GTFS time of day, I don't think this can be bigger than 23 anymore. (Same for `ampm()`)

It'll also need a 0 case now.

assets/src/components/propertiesPanel/blockWaiverBanner.tsx
```diff
@@ -0,0 +1,128 @@
...
+export const nowEpochSeconds = (now = Date.now()): number =>
+  Math.floor(now / 1_000)
+
+export enum CurrentFuturePastType {
+  Current = 1,
+  Future,
+  Past,
+}
+
+export const currentFuturePastType = ({
```
skyqrose: Some things in this file don't need to be exported anymore.

assets/src/components/propertiesPanel/blockWaiverBanner.tsx
arkadyan: Updated

assets/src/components/propertiesPanel/blockWaiverBanner.tsx
arkadyan: Removed

lib/realtime/block_waiver.ex
```diff
@@ -104,8 +104,8 @@ defmodule Realtime.BlockWaiver do
     {end_time, _} = List.last(trip_stop_time_waivers)
 
     %__MODULE__{
-      start_time: start_time,
-      end_time: end_time,
+      start_time: Util.Time.timestamp_for_time_of_day(start_time),
+      end_time: Util.Time.timestamp_for_time_of_day(end_time),
```
skyqrose: `date_of_timestamp` says in its `@doc` "This is the date that roles over at midnight, not the service date". So I think this won't handle the date rollover correctly.

Consider a trip from 02:30 to 03:00 and a trip from 26:30 to 27:00, both of which are running at the same time and both of which get a block waiver. This `timestamp_for_time_of_day` conversion needs different dates in order to turn those different times of days into the same timestamps. If you use the date that roles over at midnight, then the block waiver for 26:30 would get a timestamp 24 hours in the future.

The other place where `timestamp_for_time_of_day` is used (in `ghost.ex`) knows what date to use for the conversion by holding on to the date it found when digging through the calendar looking for active blocks. Unfortunately, since we're not starting from the calendar here, we don't already have a date, so that strategy won't work.

The `next_time_of_day_for_timestamp_after` conversion that's used in `timepoint_status.ex` converts the opposite direction, but uses a strategy that might be useful for this case. It tries doing the naive conversion, but if the time_of_day would be more than an hour before or more than 23 hours after the time we expect to be approximately correct, then we adjust the date so we get something that's in the range we expect.

Since this is the first time we're trying to do a conversion from time_of_day to timestamp that needs to handle the date rollover correctly, I think you'll need to write a new `next_timestamp_for_time_of_day_after` function, and choose a date that gives you a timestamp that's close to `Util.Time.now()`. For block waivers, maybe you'd aim for a timestamp that's between 4 hours ago and 20 hours in the future?

(Side note: there are a few uses of `date_of_timestamp()` in `time.ex` and `data.ex`, but the callers all check if the naive answer is correct and then adjust accordingly.)

assets/tests/components/ladder.test.tsx
skyqrose: I noticed you fixed the `as unknown` in `routeLadder.test.tsx`. Looks like there's one more here and one more in `routeLadder.test.tsx` (line 605).

lib/realtime/block_waiver.ex
arkadyan: We actually have the block and know it is active, so I changed it to behave the same way.

lib/realtime/block_waiver.ex
```diff
@@ -34,7 +34,7 @@ defmodule Realtime.BlockWaiver do
     block
     |> Enum.flat_map(&trip_stop_time_waivers/1)
     |> group_consecutive_sequences()
-    |> Enum.map(&from_trip_stop_time_waivers/1)
+    |> Enum.map(&from_trip_stop_time_waivers(&1, date_for_block(block)))
```
skyqrose: Yeah, this will correctly handle the day rollover 👍 

I'm a little worried about performance. For every block waiver, it will calculate all the active blocks, and then search through all of them. We could probably reduce some of that work with [Refactor: Make the BlockWaivers structs on write instead of read](https://app.asana.com/0/1148853526253426/1163344518291749) by fetching the active blocks only once and using it for all block waivers. I think it's probably fine to wait for that unless you've noticed it being a problem.

lib/realtime/block_waiver.ex
```diff
@@ -98,15 +98,35 @@ defmodule Realtime.BlockWaiver do
...
+  defp date_for_block(block) do
+    active_blocks_fn = Application.get_env(:realtime, :active_blocks_fn, &Gtfs.active_blocks/2)
+
+    now = Util.Time.now()
+    ten_minutes = 10 * 60
+    ten_minutes_ago = now - ten_minutes
+    in_ten_minutes = now + ten_minutes
+
+    {date, _blocks} =
+      ten_minutes_ago
```
skyqrose: Do we expect all block waivers to be for a block that's active within 10 minutes of now? Would a dispatcher ever enter a block waiver for a block that won't start for an hour? Would we want to see a block waiver on a bus that's so late it was supposed to finish its block an hour ago?

lib/realtime/block_waiver.ex
```diff
@@ -98,15 +98,35 @@ defmodule Realtime.BlockWaiver do
...
+      |> active_blocks_fn.(in_ten_minutes)
+      |> Enum.find(
+        {Util.Time.today(), []},
```
skyqrose: If we wanted to be stricter about mocking times, we could do `now = Application.get_env(:skate, :now_fn).()` above and then here could be `Util.Time.date_of_timestamp(now)`.

It looks like in this case the tests will still work since they also use `Util.Time.today()` in a matching way, so this is more a general feeling of not wanting to depend on using real times in tests than this causing a specific problem.

lib/realtime/block_waiver.ex
arkadyan: We don't show any vehicles that aren't active in the next 10 minutes. I expanded the reach into the past.

lib/realtime/block_waiver.ex
arkadyan: I'm not worried about that.

PR #459, opened 2020-02-19
Encapsulate window.location.reload call

arkadyan:
No Asana ticket, this is required (but not sufficient) for https://github.com/mbta/skate/pull/428 to pass.

This allows us to add mock implementation on a function we control.
This is required in order to upgrade to Jest version 25.1.0.

### DIFF
--- a/assets/tests/api.test.ts
+++ b/assets/tests/api.test.ts
@@ -26,6 +26,18 @@ const mockFetch = (status: number, json: any): void => {
 }
 
 describe("apiCall", () => {
+  let browserReloadSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    browserReloadSpy = jest
+      .spyOn(browser, "reload")
+      .mockImplementation(() => {})
+  })
+
+  afterAll(() => {
+    browserReloadSpy.mockRestore()
+  })
+
   test("returns parsed data", done => {
     mockFetch(200, { data: "raw" })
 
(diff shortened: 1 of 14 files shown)


assets/tests/api.test.ts
```diff
@@ -43,27 +44,27 @@ describe("apiCall", () => {
...
+      expect(browser.reload).toHaveBeenCalled()
       done()
     })
   })
 
   test("reloads the page if the response status is forbidden (403)", done => {
     mockFetch(403, { data: null })
 
-    window.location.reload = jest.fn()
+    jest.spyOn(browser, "reload").mockImplementation(() => {})
```
skyqrose: Could we just have a single spy for the whole file or describe, instead of one for each test? (We'd have to be careful that the mock call history doesn't pollute between the tests.)

Might it even be possible and worth doing to set up a mock in `tests/setup.ts` for all tests, assuming we never want the unmocked reload?

assets/tests/models/browser.test.ts
```diff
@@ -0,0 +1,33 @@
+import { reload } from "../../src/models/browser"
+
+describe("reload", () => {
+  let reloadSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    // Dirty: setting window.location as writable so we can spy on reload function.
+    // Doing this once here to avoid it in all other tests.
```
skyqrose: 👍

skyqrose:
The codecov failures are for code that was previously untested. I'm okay if you overrule codecov here.

PR #463, opened 2020-02-20
Style past and future block waivers

arkadyan:
Asana ticket: [Future & Past block waiver banner treatment](https://app.asana.com/0/1148853526253426/1162089955688355)

Make the background and border gray.

Make the start time bold for future waivers.

Depends on https://github.com/mbta/skate/pull/457

Screen Shot 2020-02-20 at 14 58 54

### DIFF
--- a/assets/css/properties_panel/_block_waiver_banner.scss
+++ b/assets/css/properties_panel/_block_waiver_banner.scss
@@ -26,6 +26,26 @@
   fill: $black;
   margin-right: 0.3rem;
   width: 15px;
+
+  .c-icon-alert-circle__outline {
+    fill: $white;
+  }
+
+  .c-icon-alert-circle__circle-fill {
+    fill: $color-component-dark;
+
+    .m-block-waiver-banner--past & {
+      fill: $color-component-medium;
+    }
+  }
+
+  .c-icon-alert-circle__exclamation-point {
+    fill: $white;
+
+    .m-block-waiver-banner--past & {
+      fill: $color-bg-medium;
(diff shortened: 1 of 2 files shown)


ingridpierre:
@arkadyan having trouble pulling down changes, but past waivers should have their end times in bold (future waivers have the opposite).

ingridpierre:
For past waivers we should also make sure the value text and the alert icon is grey as well. m-block-waiver__detail-value should be grey to futher make this block waiver seem "disabled" by comparison. I also want it to have the same icon in the banner as it would on the ladder.

**current:**
image

**desired:**
image

I have a grey icon in the skate icons folder here as well: https://drive.google.com/file/d/1Jbu3yZSb6JXdBaDBASg5hc9M26Za3Yd4/view?usp=sharing

skyqrose:
I guess we'll end up doing the icon coloring work described in https://github.com/mbta/skate/pull/457#discussion_r383444381 earlier than expected (instead of waiting until the on-ladder icons), but since this PR is so small, I think it's fine to do that here.

arkadyan:
@skyqrose I updated the icon to be colored via css rules. Ingrid gave a thumbs up, how does the code look to you?

assets/css/properties_panel/_block_waiver_banner.scss
```diff
@@ -26,6 +26,26 @@
   fill: $black;
```
skyqrose: Should the old `fill: $black;` still be here?

PR #472, opened 2020-02-26
Data Outage Banner

skyqrose:
Screen Shot 2020-02-26 at 10 25 11

Asana Task: [Let users know when Skate data is stale/we have lost live data connection](https://app.asana.com/0/1148853526253426/1156122231748753)
For simplicity, this is just the outage banner. The restored banner will be broken into a [separate task](https://app.asana.com/0/1148853526253420/1163528387235355).

---

A bus is considered stale if its timestamp is more than 2½ minutes old. We consider there to be an outage if more than 20% of buses are stale and there are at least 3 stale buses. There's a lot of wiggle room to adjust those thresholds.

In particular I'm a little unsure about what realtime data looks like overnight and how to make sure it doesn't say there's an outage when there isn't, hence the "at least 3 stale buses" check, but that also means that if we lose all realtime data and say there are no buses, it won't post the banner. I think that's okay since it'd be obvious if there are no buses, and I think avoiding false positives is probably more important, but want to get a second opinion on that.

---

Unfortunately, to display the banner, a lot of the places we use fixed positioning wouldn't work anymore (they would display over the banner), so I had to change a lot of the layout to use relative/absolute positioning instead. Most of the QAing for this task will be around making sure I didn't break anything by doing that.

The user-facing change that I know of from that is that the route ladder horizontal scrolling now happens on the route ladders instead of the whole page. (Notice the scrollbar in the screenshot.) This means:
* Swiping on the tab bar or route picker won't move the ladders anymore.
* The VPP modal prevents scrolling the ladders since they're behind the modal.

I'm pretty unsure about all of this since CSS is so messy. I'd be on the lookout for other regressions and better ways to organize things.

---

This is the 3rd place we're using channels like this, and I'm seeing a lot of duplication, inconsistencies in how we do things like close the channels, and some untested behavior (like getting the data from the initial result). I have a plan to unify and polish a bunch of that, but it'd make this PR a bit big, and the status quo is fine enough for this new work, so I'll do that as a followup.

---

I'm planning to deploy this to dev for easier testing on tablets. That will allow for checking most of the css changes. Unless we happen to have a data outage while this PR is open, you won't be able to see the banner, though. I have a modified version running locally that fakes an outage so we can see what it feels like with the banner open. Let me know if you want to take a look at that in person.


lib/realtime/server.ex
```diff
@@ -109,8 +126,11 @@ defmodule Realtime.Server do
         ) :: term()
   def update(update_term, server \\ __MODULE__)
 
-  def update({:vehicle_positions, vehicles_by_route_id, shuttles}, server) do
-    GenServer.cast(server, {:update, :vehicle_positions, vehicles_by_route_id, shuttles})
+  def update({:vehicle_positions, vehicles_by_route_id, shuttles, data_status}, server) do
```
paulswartz: This seems like an annoying place to include the data status, as demonstrated by all the tests you needed to update to include this data. 

What do you think about having a separate `Server` for the current status, which itself can listen for for updates to vehicle positions and do the appropriate calculations?

lib/realtime/data_status.ex
```diff
@@ -1,8 +1,51 @@
...
+  end
+
+  @spec vehicle_is_good(Vehicle.t(), Util.Time.timestamp()) :: boolean()
+  defp vehicle_is_good(vehicle, now) do
+    age_seconds = now - vehicle.timestamp
+    # 2 and a half minutes ago.
+    age_seconds < 150
+  end
+
+  @spec count_partition([element], (element -> bool())) :: {non_neg_integer(), non_neg_integer()}
```
paulswartz: You don't actually care about the number of good vehicles, so you could have this return `{predicate_count, total_count}` and change the predicate to `vehicle_is_bad`.

lib/realtime/data_status.ex
```diff
@@ -1,8 +1,51 @@
...
+    now = now_fn.()
+
+    {good_count, bad_count} =
+      vehicles
+      |> Enum.filter(&should_consider?/1)
+      |> count_partition(fn vehicle -> vehicle_is_good(vehicle, now) end)
+
+    total_count = good_count + bad_count
+
+    if bad_count >= 2 and bad_count / total_count > 0.20 do
```
paulswartz: What happens if all the vehicles are stale?

lib/realtime/data_status.ex
arkadyan: I would go further and say that I'm not sure you need to partition at all. You already know the total count: it is `vehicles |> Enum.filter(&should_consider?/1) |> length()`. You also know the threshold of bad results you have to cross, the max of `count * 0.2` and `3`, so you don't even actually care about the bad count. I think you could probably use `Stream.take_while/2` or something and lazily evaluate such that you can stop as soon as you've crossed the threshold.

lib/realtime/data_status.ex
skyqrose: If all of them are unassigned, then we'll show the status as good (but all the buses will be ghosts, so the problem will be obvious).

If all assigned buses are stale, then we'll show the outage banner.

lib/realtime/data_status.ex
skyqrose: `length()` is linear time, and we at least need that, so this needs to take linear time no matter what. Using `take_while` for the second step doesn't save us from that.

As for getting the total count, I chose this way to stay the same as [`Enum.partition()`](https://hexdocs.pm/elixir/1.0.5/Enum.html#partition/2), which returns `{true_results, false_results}`.

arkadyan:
What's the plan for testing this and should Ingrid be a reviewer?

lib/realtime/server.ex
skyqrose: It is annoying. I gave it a shot in `Realtime.DataStatusRegistry` (though I left the calculations in the same place, and the new genserver just passes that result through)

lib/realtime/supervisor.ex
```diff
@@ -5,11 +5,15 @@ defmodule Realtime.Supervisor do
...
   end
 
+  @spec registry_name() :: Registry.registry()
+  def registry_name(), do: Realtime.Registry
+
   @impl true
   def init(:ok) do
     children = [
-      {Registry, keys: :duplicate, name: Realtime.Server.registry_name()},
+      {Registry, keys: :duplicate, name: registry_name()},
```
paulswartz: Do you need two separate registries? Maybe a pattern like [events.ex](https://github.com/mbta/api/blob/master/apps/events/lib/events.ex) where clients provide the key they're registering for would work, with keys like `:data_status` and `{:route, ID}`.

skyqrose: It is just one registry. The key is the pid of the pubsub genserver (`server.ex` or `data_status_registry.ex`).

paulswartz: Ah, can we give it a different name then? Maybe `DataStatusServer`?

skyqrose:
Akira tested it and said it was good, but if you could run it locally too during your review for another set of eyes, that'd help me feel more confident in it.

To force the banner to show up locally, change `Realtime.DataStatus.data_status` to always return `:outage`.

Ingrid just had one comment about font styling that I still need to fix but shouldn't get in the way of your review.

lib/realtime/supervisor.ex
skyqrose: `DataStatusPubsub`?

skyqrose:
Yeah, I tested it on the tablets.

I'm looking at the codecov failures.
One is due to not testing data initially returned from joining the channel. We don't currently have a way to test that, but the followup PR I mentioned will be able to test that case (for this channel and the others).
One is from not testing Consumer.VehiclePositions.start_link, but that's unrelated to this PR
One is for not having a test for Realtime.Supervisor, but that also seems outside the scope of this PR.

Can you force squash?

PR #476, opened 2020-02-27
fix intermittent test failure

skyqrose:
when the :active_blocks_fn was unassigned, it got the value nil
which means Application.get_env returns nil instead of the default

this caused intermittent test failures depending on the order of tests
example of a failing seed beforehand: commit ceb6330, mix test --seed 145576

deleting the value instead of assigning nil means that future calls to
get_env will use their default values

Fixes the failing build in #475 and others.


skyqrose:
I think codecov got confused by a forcepush and isn't going to respond. Can you force squash?

PR #477, opened 2020-02-27
useChannel

skyqrose:
Follow up to #472 

While doing that PR, I noticed some inconsistencies and problems with how we do channels from the frontend. We have 3 places that use data from a single channel now. This PR unifies and polishes those implementations.

This PR:
* Removes a ton of duplication
* Fixes how we close channels. In particular, the useShuttleVehicles channel wasn’t getting closed properly, so we were still loading shuttle vehicles after clicking away from the shuttle map.
* Adds a missing test for the data returned from the initial channel join.
* Lets the channel users focus on the parts that are special for that channel instead of the channel lifecycle.

It doesn’t change the useVehicles channel, since that needs to handle and remember multiple channels, and is significantly different from the other 3 places we use channels.

### DIFF
--- a/assets/src/hooks/useChannel.ts
+++ b/assets/src/hooks/useChannel.ts
@@ -1,0 +1,64 @@
+import { Channel, Socket } from "phoenix"
+import { useEffect, useState } from "react"
+import { reload } from "../models/browser"
+
+/** Opens a channel for the given topic
+ *  and returns the latest data that's been pushed to it
+ *
+ *  Only listens for a single event
+ *  Assumes the data returned on join and the data from subsequent pushes are the same format
+ *  If topic is null, does not open a channel.
+ *  Returns offState if the channel is not open yet
+ *  Returns loadingState after the channel has been opened but before receiving the first response
+ */
+export const useChannel = <T>({
+  socket,
+  topic,
+  event,
+  parser,
+  loadingState,
+  offState,
+}: {
+  socket: Socket | undefined
+  topic: string | null
+  event: string
+  parser: (data: any) => T
+  loadingState: T
+  offState: T
+}): T => {
+  const [result, setResult] = useState<T>(offState)
+
+  useEffect(() => {
+    let channel: Channel | undefined
+
+    if (socket !== undefined && topic !== null) {
+      setResult(loadingState)
+      channel = socket.channel(topic)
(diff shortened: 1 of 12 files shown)


assets/src/hooks/useChannel.ts
```diff
@@ -0,0 +1,64 @@
...
+  parser,
+  loadingState,
+  offState,
+}: {
+  socket: Socket | undefined
+  topic: string | null
+  event: string
+  parser: (data: any) => T
+  loadingState: T
+  offState: T
```
arkadyan: I'm confused by what `offState` means. Can we just use `loadingState`?

assets/src/hooks/useChannel.ts
```diff
@@ -0,0 +1,64 @@
...
+}): T => {
+  const [result, setResult] = useState<T>(offState)
```
arkadyan: Nit: To me, `result` makes me think of a single response. Would something like `state` convey what this is better?

assets/tests/hooks/useChannel.test.ts
```diff
@@ -0,0 +1,242 @@
...
+        parser: jest.fn(),
+        loadingState: "loading",
+        offState: "off",
+      })
+    )
+    expect(result.current).toEqual("off")
+    expect(mockSocket.channel).not.toHaveBeenCalled()
+  })
+
+  test("giving a topic subscribes to that topic", () => {
```
arkadyan: s/giving/given/ ?

assets/src/hooks/useChannel.ts
skyqrose: This is a single response. There's a simple `return result` at the end of the hook.

(The channel isn't saved by a `setState`.)

arkadyan: Sorry, I phrased that poorly. What I meant was more like a single value and the interaction is over, as opposed to the current value of some data that will be updated over time.

assets/src/hooks/useChannel.ts
skyqrose: `useSearchResults` returns a different value to distinguish between "we're waiting for a response (`loadingState`,`null`) and "we haven't initialized the socket yet"/"we're not searching (`offState`,`undefined`), but then we don't do anything interesting with those values. I'll consolidate them.

PR #479, opened 2020-02-28
Display the operator's last login time in search results

arkadyan:
Asana ticket: [Add "Last Login" to search results cards](https://app.asana.com/0/1148853526253426/1160813278503801)

Screen Shot 2020-02-28 at 14 33 37


assets/src/components/propertiesPanel/blockWaiverBanner.tsx
```diff
@@ -1,34 +1,12 @@
...
-
-export const dateFromEpochSeconds = (time: number): Date =>
-  new Date(time * 1_000)
-
-export const formattedTime = (date: Date): string => {
-  const hours24 = date.getHours()
-  return `${hours12(hours24)}:${zeroPad(date.getMinutes())}${ampm(hours24)}`
-}
-
 export const formatEpochSeconds = (epochSeconds: number): string => {
```
skyqrose: If you're pulling out time helpers to another file, this and `nowEpochSeconds` could be pulled out, too.

assets/src/util/dateTime.ts
```diff
@@ -0,0 +1,34 @@
+export const dateFromEpochSeconds = (time: number): Date =>
+  new Date(time * 1_000)
+
+export const now = (): Date => new Date()
```
skyqrose: I wonder if it would be possible for this file's interface to work entirely with seconds instead of working with a mix of `Date` and seconds. The fact that this file uses `Date` under the hood would then be an opaque implementation detail.

I think you could get by with only exporting 3 functions: `nowEpochSeconds()`, `formattedTime(epochSeconds: number)` and `formattedTimeDiff(epochSeconds1, epochSeconds2)`.

assets/tests/components/__snapshots__/propertiesPanel.test.tsx.snap
```diff
@@ -307,6 +315,20 @@ Array [
...
+            >
+              <td
+                className="m-properties-list__property-label"
+              >
+                Last Login
+              </td>
+              <td
+                className="m-properties-list__property-value"
+              >
+                4h 3m; 1:38pm
```
skyqrose: Adding the logon time to the properties list also adds it to the VPP. I just want to double check that we want that.

assets/src/components/propertiesList.tsx
```diff
@@ -90,14 +109,17 @@ const highlightRegex = (highlightText: string): RegExp => {
   return new RegExp(allowNonAlphanumeric, "i")
 }
 
+const labelName = (label: string): string =>
+  label.toLowerCase().replace(" ", "-")
```
skyqrose: I'm a little hesitant about using human-facing copy as css property names. Could we instead have a separate (maybe optional) field in `Property` for the css modifier?

assets/tests/components/__snapshots__/propertiesPanel.test.tsx.snap
arkadyan: Yes, good question. I discussed with Ingrid and we are doing that intentionally.

assets/src/util/dateTime.ts
arkadyan: I hesitate to use epoch seconds more since that isn't a "normal" unit in Javascript. This actually kind of makes me wonder: maybe we should be converting from epoch seconds to Date when we ingest the block waiver data? Then we could get rid of the redundant feeling `formatEpochSeconds` function and not have to worry about a different time standard anywhere else in JS land. How would you feel about that change?

arkadyan: I made that change.

PR #485, opened 2020-03-03
LadderVehicle uses whole vehicle instead of taking a few fields

skyqrose:
[Display block waiver markers on ghost bus ladder icons](https://app.asana.com/0/1148853526253426/1163145000751641) will require `ladder.tsx` to check the block waivers to decide what kind of icon to draw. Rather than adding all the block waivers to the `LadderVehicle`, or checking the styling when creating the `LadderVehicle`, it's probably easier pass the whole vehicle through.

Old `LadderVehicle`:
```ts
export interface LadderVehicle {
  vehicleId: VehicleId
  label: string
  viaVariant: ViaVariant | null
  status: DrawnStatus
  hasBlockWaivers: boolean
  headwaySpacing: HeadwaySpacing | null
  x: number
  y: number
  vehicleDirection: VehicleDirection
  scheduledY?: number
  scheduledVehicleDirection?: VehicleDirection
  lane: number
} 
```
new `LadderVehicle`:
```ts
export interface LadderVehicle {
  vehicle: VehicleOrGhost
  x: number
  y: number
  vehicleDirection: VehicleDirection
  scheduledY?: number
  scheduledVehicleDirection?: VehicleDirection
}
```

### DIFF
--- a/assets/src/components/headwayLines.tsx
+++ b/assets/src/components/headwayLines.tsx
@@ -3,6 +3,8 @@ import { VehicleDirection } from "../models/ladderDirection"
 import { LadderVehicle } from "../models/ladderVehicle"
 import { HeadwaySpacing, headwaySpacingToString } from "../models/vehicleStatus"
 import { CENTER_TO_LINE } from "./ladder"
+import { VehicleOrGhost } from "../realtime"
+import { isVehicle } from "../models/vehicle"
 
 interface Props {
   height: number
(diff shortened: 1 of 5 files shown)


assets/src/models/ladderVehicle.ts
```diff
@@ -1,40 +1,23 @@
...
-  status: DrawnStatus
-  hasBlockWaivers: boolean
-  headwaySpacing: HeadwaySpacing | null
+  vehicle: VehicleOrGhost
   x: number
   y: number
   vehicleDirection: VehicleDirection
   scheduledY?: number
   scheduledVehicleDirection?: VehicleDirection
-  lane: number
```
arkadyan: Why does this go away?

assets/src/components/headwayLines.tsx
```diff
@@ -21,32 +23,35 @@ const drawHeadwayLine = (
     return acc
   }
 
-  const [currentVehicle, ...rest] = ladderVehicles
+  const [currentLadderVehicle, ...rest] = ladderVehicles
+  const vehicle: VehicleOrGhost = currentLadderVehicle.vehicle
+  const headwaySpacing =
```
arkadyan: Has this lost the feature flag check?

assets/src/models/ladderVehicle.ts
skyqrose: Yeah, I wasn't quite sure what to do about this. It's never used outside of this file, so I figured it shouldn't be part of the public interface, but the way the LadderVehicle object is created with ` return {...vehicleInLane, x}`, technically the lane still in the object when it's returned.

assets/src/components/headwayLines.tsx
skyqrose: I did remove the check that `LadderVehicle` used to do before passing `headwaySpacing` through. But `<Ladder>` checks the feature flag before drawing `<HeadwayLines>`, so that check was redundant.

arkadyan: Ah, got it, ok.

assets/src/models/ladderVehicle.ts
arkadyan: Yeah, personally I'd just describe the data completely and let other things worry about what they do or do not want to use, but I don't feel strongly.

skyqrose: I looked at adding it back, but it requires adding unused fields to a bunch of test data, too. That's not a big deal, but I don't think it's worth it to make the interface reflect the implementation when ideally it should probably be the implementation that changes to reflect the interface we'd use.

skyqrose:
Ready for a squash merge, but I need you to do that because of codecov's weird accounting.

PR #486, opened 2020-03-03
Sort search results by most recent logon time

arkadyan:
Asana ticket: [Sort search results by most recent vehicle login time](https://app.asana.com/0/1148853526253426/1160813278503793)

Screen Shot 2020-03-03 at 11 00 56

### DIFF
--- a/assets/tests/components/searchResults.test.tsx
+++ b/assets/tests/components/searchResults.test.tsx
@@ -1,7 +1,9 @@
 import { mount } from "enzyme"
 import React from "react"
 import renderer from "react-test-renderer"
-import SearchResults from "../../src/components/searchResults"
+import SearchResults, {
+  byOperatorLogonTime,
+} from "../../src/components/searchResults"
 import { StateDispatchProvider } from "../../src/contexts/stateDispatchContext"
 import { HeadwaySpacing } from "../../src/models/vehicleStatus"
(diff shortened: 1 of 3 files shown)


assets/tests/components/searchResults.test.tsx
```diff
@@ -285,3 +423,34 @@ describe("SearchResults", () => {
...
+      id: "1",
+      operatorLogonTime: new Date("2018-08-15T13:30:00.000Z"),
+    } as Vehicle
+    const ghost: Vehicle = {
+      id: "ghost-2",
+    } as Vehicle
+
+    expect(byOperatorLogonTime(vehicle, ghost)).toEqual(1)
+    expect(byOperatorLogonTime(ghost, vehicle)).toEqual(-1)
+    expect(byOperatorLogonTime(ghost, ghost)).toEqual(0)
```
skyqrose: I know we had a bug a while ago where the `-1` and `+1` were backwards. It's hard to know by looking at these whether the test is correct.

What if these tests were instead something like
```ts
expect([vehicle, ghost].sort(byOperatorLogonTime)).toEqual([ghost, vehicle])
```
where the test is based on easy to understand list ordering?

arkadyan: Yeah, I like that much better.

PR #492, opened 2020-03-05
fix map on safari

skyqrose:
Asana Task: [🐞 Layout issues in Safari](https://app.asana.com/0/1148853526253426/1164221537676417)

`height: 100%` didn't work on children with `flex-basis: auto` in safari

also:
* remove width: 100% that caused the page to be slightly too wide (because of margins)
* hide route ladders scroll bar if the page is narrow (overflow-x: auto)

### DIFF
--- a/assets/css/_ladder_page.scss
+++ b/assets/css/_ladder_page.scss
@@ -1,5 +1,6 @@
 .m-ladder-page {
   box-sizing: border-box;
+  height: 100%;
   position: relative;
   transition: $transition-slide;
 
--- a/assets/css/_page.scss
+++ b/assets/css/_page.scss
@@ -2,7 +2,6 @@
   height: 100%;
   margin-left: $tab-bar-width;
   overflow: scroll;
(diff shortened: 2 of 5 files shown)


arkadyan:
Desktop Safari looks better, but I'm still seeing some issues in mobile Safari. The shuttle map and search views are extending taller than the device. And just to make things more annoying I'm not seeing that in the "responsive" mode in desktop Safari…

IMG_0652

IMG_0653

skyqrose:
The iphone viewport height bug is [separate](https://medium.com/@susiekim9/how-to-compensate-for-the-ios-viewport-unit-bug-46e78d54af0d). I think I have a fix, but I'd rather make a new PR for that to review it separately.

skyqrose:
With the height issue fixed separately, is this PR ready to go?

arkadyan:
@skyqrose I think it would be good to test this combined with the height changes to make sure that everything is covered between them.

PR #493, opened 2020-03-05
Refactor: Move StopTimeUpdate consumer into a new pipeline

arkadyan:
Asana ticket: [Refactor: Move StopTimeUpdate consumer into a new pipeline](https://app.asana.com/0/1148853526253426/1163344518291746)


lib/concentrate/pipeline.ex
```diff
@@ -0,0 +1,21 @@
...
+      },
+      id: source
+    )
+  end
+
+  @spec consumer(module(), term(), term()) :: Supervisor.child_spec()
+  def consumer(module, id, provider) do
+    Supervisor.child_spec(
+      {module, subscribe_to: [{provider, [max_demand: 1]}]},
+      id: id
```
paulswartz: You only need an ID if the `module` isn't unique: I think that's the case here?

lib/concentrate/pipeline/stop_time_updates_pipeline.ex
```diff
@@ -0,0 +1,28 @@
...
+    [source(opts), consumer()]
+  end
+
+  def source(opts) do
+    Pipeline.source(
+      :trip_updates_enhanced,
+      opts[:trip_updates_url],
+      Concentrate.Parser.GTFSRealtimeEnhanced
+    )
+  end
```
skyqrose: Will this behave well if the `trip_updates_url` isn't set?

lib/concentrate/supervisor.ex
```diff
@@ -27,108 +20,9 @@ defmodule Concentrate.Supervisor do
...
-  end
-
-  def outputs_with_options(outputs, options) do
-    for name <- outputs do
-      {name, options}
-    end
+    Enum.concat([
+      Concentrate.Pipeline.VehiclePositionsPipeline.pipeline(opts),
+      Concentrate.Pipeline.StopTimeUpdatesPipeline.pipeline(opts)
+    ])
```
skyqrose: I'm thinking of the stop times pipeline as being pretty independent from the concentrate pipeline. What do you think about giving them separate supervisors to reflect that, and so if one pipeline dies, the other one doesn't have to get restarted with it. (I'm imagining both pipeline-supervisors would be started by Realtime.Supervisor)

lib/concentrate/pipeline/stop_time_updates_pipeline.ex
arkadyan: No. I was thinking that isn't something we have a need to worry about—is there something I'm missing?

lib/concentrate/pipeline/stop_time_updates_pipeline.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  def consumer() do
```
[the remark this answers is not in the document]
arkadyan: I actually went with both, tell me what you think about my changes. I made Pipeline a behaviour, where it requires an `init` function that returns a list of Supervisor children. Then those two existing functions I put into a Helpers module.

lib/concentrate/supervisor.ex
arkadyan: Added individual supervisors for each pipeline.

lib/concentrate/pipeline/stop_time_updates_pipeline.ex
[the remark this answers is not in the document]
paulswartz: You don't really need the helpers module in that case; the functions can live inside `Pipeline` directly. But it's fine this way too.

lib/concentrate/pipeline/stop_time_updates_pipeline_supervisor.ex
```diff
@@ -0,0 +1,16 @@
...
+  use Supervisor
+
+  @spec start_link(keyword()) :: Supervisor.on_start()
+  def start_link(opts) do
+    Supervisor.start_link(__MODULE__, opts)
+  end
+
+  @impl true
+  def init(opts) do
+    Supervisor.init(Concentrate.Pipeline.StopTimeUpdatesPipeline.init(opts), strategy: :rest_for_one)
```
paulswartz: This is also something like you could include in `Pipeline`, since you can pass a list of children to `Supervisor.start_link`. The two separate pipelines would be configured like this:
```
{Concentrate.Pipeline, module: Concentrate.Pipeline.StopTimeUpdates, ..other_opts..},
{Concentrate.Pipeline, module: Concentrate.Pipeline.VehiclePositions, ..other opts..}
```
The `init/1` callback for `Pipeline` behaviors is the same, and then you implement `start_link/1` and `child_spec/1` on the `Pipeline` module directly. This is also a case where you'll want an `id` on the child spec: you can probably use the module for the ID.

lib/concentrate/supervisor.ex
skyqrose: 👍 They can be restarted separately, but I think the structure is more complex than necessary.
It's now
```
Realtime.Supervisor
├ ...
└ Concentrate.Supervisor
  ├ Concentrate.Pipeline.VehiclePositionsPipelineSupervisor
  │ └ Concentrate.Pipeline.VehiclePositionsPipeline
  │   ├ Concentrate.Producer.HTTP
  │   ├ Concentrate.Producer.HTTP
  │   ├ Concentrate.Merge
  │   └ Concentrate.Consumer.VehiclePositions
  └ Concentrate.Pipeline.StopTimeUpdatesPipelineSupervisor
    └ Concentrate.Pipeline.StopTimeUpdatesPipeline
      ├ Concentrate.Producer.HTTP
      └ Concentrate.Consumer.StopTimeUpdates
```
but we can knock out a couple layers like this:
```
Realtime.Supervisor
├ ...
├ Concentrate.VehiclePositionsPipeline
│ ├ Concentrate.Producer.HTTP
│ ├ Concentrate.Producer.HTTP
│ ├ Concentrate.Merge
│ └ Concentrate.Consumer.VehiclePositions
└ Concentrate.Pipeline.StopTimeUpdatesPipeline
  ├ Concentrate.Producer.HTTP
  └ Concentrate.Consumer.StopTimeUpdates
```

lib/concentrate/pipeline/stop_time_updates_pipeline.ex
skyqrose: It might cause problems to any non-MBTA users since they won't be able to give a trip_updates_url that has the format we expect, but wouldn't make a difference to us.

lib/concentrate/pipeline/stop_time_updates_pipeline_supervisor.ex
arkadyan: I feel like I might not be fully grasping what you are saying here, perhaps starting with the bit about "you can pass a list of children to Supervisor.start_link". Are you suggesting something counter to the suggestion Sky made [here](https://github.com/mbta/skate/pull/493#discussion_r389722614) of separating out the pipelines from a single supervisor? Or is what you are saying different?

lib/concentrate/supervisor.ex
arkadyan: That was intentional to encapsulate all the pipeline configuration knowledge. It didn't seem like something the Realtime supervisor should be concerned with.

skyqrose: `Realtime.Supervisor` wouldn't know anything about how the pipelines are set up. It would just know that there's 2 separate pipelines, one for VPs and one for STUs.

All the knowledge about which children to start and which env vars to use would be written in `VehiclePositionsPipeline` or `StopTimeUpdatesPipeline`.

lib/concentrate/pipeline/stop_time_updates_pipeline_supervisor.ex
paulswartz: I'll write it up more thoroughly tomorrow morning.

arkadyan: Thanks! We can also talk through it in person if that's easier.

paulswartz:
```elixir
defmodule Pipeline do
  @moduledoc """Create a pipeline supervisor for a given module."""
  
  @callback init(Keyword.t) :: [Supervisor.child_spec]

  def start_link(mod, opts) do
    children = mod.init(opts)
    Supervisor.start_link(children, restart: :rest_for_one)
  end

  def child_spec(opts) do
   mod = Keyword.fetch(opts, :module)
    %{
      type: :supervisor, 
      id: mod,
      start: {__MODULE__, :start_link, [mod, opts]}
    }
  end
end

# in Concentrate.Supervisor
def start_link(...) do
  Supervisor.start_link([
    {Pipeline, module: StopTimePipeline, ...},
    {Pipeline, module: VehiclePositionSupervisor, ...}
  ], restart: :one_for_one)
end

Comment (arkadyan) [reply lib/concentrate/pipeline/stop_time_updates_pipeline_supervisor.ex]: Thanks, that's helpful. Let me know if you have any suggestions on my adaptation.

File: lib/concentrate/supervisor.ex (hunk 3 above)
Comment (arkadyan) [reply lib/concentrate/supervisor.ex]: Yes, knowing about what pipelines exist is what I wanted to encapsulate.

File: lib/concentrate/pipeline/stop_time_updates_pipeline.ex [hunk 6]
```diff
@@ -0,0 +1,26 @@
+defmodule Concentrate.Pipeline.StopTimeUpdatesPipeline do
+  @behaviour Concentrate.Pipeline
+  alias Concentrate.Pipeline
+
+  @type opts :: [
+          trip_updates_url: String.t()
+        ]
+
+  @impl Concentrate.Pipeline
+  @spec init(opts()) :: [Supervisor.child_spec()]
```
paulswartz: - You alias `Concentrate.Pipeline` but then keep using `Concentrate.Pipeline` in most places (in both modules).
- You don't need the `@spec` here, as it's implicitly the same as the `@callback` you're `@impl`ing.

lib/concentrate/pipeline.ex
```diff
@@ -0,0 +1,41 @@
...
+  @callback init(keyword()) :: [Supervisor.child_spec()]
+
+  @spec start_link(module(), keyword()) :: Supervisor.on_start()
+  def start_link(module, opts) do
+    children = module.init(opts)
+
+    Supervisor.start_link(children, strategy: :rest_for_one)
+  end
+
+  def child_spec(opts) do
```
skyqrose: Does this need a `@impl true` to say it implements the `Supervisor.child_spec` callback? (Or would that get in the way of this being a behavior itself?)

(Same for `Concentrate.Supervisor.child_spec` which doesn't currently have an `@impl true`.)

lib/realtime/supervisor.ex
```diff
@@ -17,11 +17,13 @@ defmodule Realtime.Supervisor do
...
-         swiftly_realtime_vehicles_url:
-           Application.get_env(:skate, :swiftly_realtime_vehicles_url),
-         trip_updates_url: Application.get_env(:skate, :trip_updates_url)
+         [
+           busloc_url: Application.get_env(:skate, :busloc_url),
+           swiftly_authorization_key: Application.get_env(:skate, :swiftly_authorization_key),
+           swiftly_realtime_vehicles_url:
+             Application.get_env(:skate, :swiftly_realtime_vehicles_url),
+           trip_updates_url: Application.get_env(:skate, :trip_updates_url)
+         ]
```
skyqrose: Why did this extra nesting come up?

lib/concentrate/pipeline.ex
arkadyan: I don't believe so because these aren't declaring a behaviour. I added a typespec here though.

lib/realtime/supervisor.ex
arkadyan: Oops, moved that to the child_spec where it should actually be.

arkadyan:
@paulswartz does this PR look good to you at this point?

PR #498, opened 2020-03-09
fix iphone height bug

skyqrose:
No Asana Task. Fixes an issue raised in #492 

on iphone, 100vh includes height for the address bar,
which causes a small vertical scroll.

using height: 100% instead gives the behavior we want,
where we the address bar doesn't overlap the app,
but it has to be specified all the way up the heirarchy.

Before:
IMG_0023

After:
IMG_0024

This was a pre-existing issue, even before the safari bug that #492 addresses was introduced in #472. It's currently visible on prod: The route ladder in this screenshot has that small amount of scroll (though before #472, including what's on prod now, the tab bar and route picker avoided the problem by being `fixed`, which is why it became more noticable in #472.)
IMG_0025

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -39,6 +39,14 @@ $picker-width-wide: $picker-width-narrow + 3.1;
 @import "vehicle_icon";
 @import "vehicle_map";
 
+html,
+body,
+main,
+#app,
+.m-app {
+  height: 100%;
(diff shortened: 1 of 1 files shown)


arkadyan:
@skyqrose When you get a chance could you please put this onto dev if you haven't already?

PR #499, opened 2020-03-09
Add a banner for ghosts with no block waivers

arkadyan:
Asana ticket: [Enhanced "Ghost bus with no block waiver" display](https://app.asana.com/0/1148853526253426/1153879901566649)

Screen Shot 2020-03-09 at 17 37 31


assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
```diff
@@ -20,6 +38,12 @@ const GhostPropertiesPanel = ({ selectedGhost, route }: Props) => (
       <BlockWaiverList blockWaivers={selectedGhost.blockWaivers} />
     )}
 
+    {featureIsEnabled("block_waivers") &&
+      !hasBlockWaiver(selectedGhost) &&
+      !isLateVehicleIndicator(selectedGhost) && (
```
skyqrose: We don't show the Ghost Properties Panel for virtual ghosts. (It shows the VPP for the corresponding real vehicle.) So I don't think you need to check this.

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
```diff
@@ -6,12 +6,30 @@ import { Route } from "../../schedule"
...
+    <div className="m-ghost-properties-panel__no-waiver-banner-header">
+      <span className="m-ghost-properties-panel__no-waiver-banner-alert-icon">
+        <IconAlertCircle />
+      </span>
+      <div className="m-ghost-properties-panel__no-waiver-banner-title">
+        Unknown Ghost Bus - No Dispatcher Note
+      </div>
+    </div>
+    A ghost bus or dropped trip has been automatically detected on this route.
+    Please follow up with dispatch about Run {runId || "Not Available"} as
```
skyqrose: This will read funny if the run is missing:
"Please follow up with dispatch about Run Not Available." Should we rephrase it in that case?

assets/css/properties_panel/_ghost_properties_panel.scss
```diff
@@ -7,3 +7,37 @@
     flex: 1 0 auto;
   }
 }
+
+.m-ghost-properties-panel__no-waiver-banner {
+  @include font-body;
+  background-color: transparentize($color-component-highlight, 0.6);
+  border: 1px solid $color-component-highlight;
+  font-weight: 400;
+  padding: 1rem 1rem 1rem 2rem;
```
skyqrose: Is there some of this that's redundant with `<BlockWaiverBanner>` that would mean we should combine some of the banner functionality?

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
arkadyan: Good question. What do you think, @ingridpierre?

assets/css/properties_panel/_ghost_properties_panel.scss
arkadyan: There are similarities, but `BlockWaiverBanner` is specific to a certain format and `NoWaiverBanner` is very static and straightforward. My thought is that trying to combine them would make things more complicated in a way that would outweigh any benefits of combining functionality.

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
arkadyan: Ah, good point. Removed the check.

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
arkadyan: I spoke with Ingrid and we decided to remove the run number in all cases.

PR #501, opened 2020-03-11
Give "Reason" a reference tooltip

arkadyan:
Add a question mark icon.

Screen Shot 2020-03-10 at 14 51 49

Screen Shot 2020-03-11 at 15 02 23


arkadyan:
This is currently on dev for testing (the tooltip will not show up running in your local development server).

ingridpierre:
@arkadyan can we make "Reason" display as "REASON" to match the other little labels below?

arkadyan:
@ingridpierre Good catch, I fixed the uppercase thing.

ingridpierre:
One more lil tweak, current waivers should use green in the label. like so:

Current:
image
Desired:
image

arkadyan:
@ingridpierre @skyqrose Updated to use a local tooltip instead of Appcues.

ingridpierre:
"Operator Error" should be "Operator error" for consistency.

assets/css/properties_panel/_block_waiver_banner.scss
```diff
@@ -101,3 +101,45 @@
...
+
+  .m-block-waiver-banner--future &,
+  .m-block-waiver-banner--past & {
+    fill: $color-font-grey;
+  }
+}
+
+.m-block-waiver-banner__reason-tooltip.__react_component_tooltip.show {
+  opacity: 1;
+  text-align: left;
```
skyqrose: With the ladder styling moved back to be specifically about the ladder, this property isn't needed anymore.

assets/src/skate.d.ts
```diff
@@ -5,6 +5,7 @@ declare global {
     Appcues?: {
       identify: (userId: string) => void
       page: () => void
+      show: (id: string) => void
```
skyqrose: I don't see this function called anywhere. What happens to the tests if it's not mocked?

arkadyan:
@ingridpierre thanks, made error lowercase.

assets/css/properties_panel/_block_waiver_banner.scss
arkadyan: Yep, removed.

assets/src/skate.d.ts
arkadyan: Yep, that is no longer needed.

ingridpierre:
@arkadyan lgtm 👍

PR #502, opened 2020-03-11
Block waiver alert icon on laying over and incoming buses.

skyqrose:
Asana Tasks:
[Display markers on vehicle laying over icons for "has current waiver" and "has future or past waiver"](https://app.asana.com/0/1148853526253426/1163145000751649)
[Display markers on ghost bus laying over icons for "has waiver" and "does not have waiver"](https://app.asana.com/0/1148853526253426/1163145000751643)
[Display markers on vehicle incoming icons for "has current waiver" and "has future or past waiver"](https://app.asana.com/0/1148853526253426/1163145000751651)
[Display markers on ghost bus incoming icons for "has waiver" and "does not have waiver"](https://app.asana.com/0/1148853526253426/1163145000751645)

Screen Shot 2020-03-11 at 14 53 09
Screen Shot 2020-03-11 at 15 27 45
Screen Shot 2020-03-11 at 15 28 35

The actual change is like 6 lines in `layoverBox.tsx` and `incomingBox.tsx`, but there's some stuff that needed to change to make that happen:
* Writing the same classes and css to color the icon in like 4 places was getting a little ridiculous, so I moved the coloring to be an argument passed into the `<IconAlertCircle>` component.
* The viewbox for the vehicle icon needed to be expanded to make room for the extra icon.
* New classes wrap the vehicle icon in the incoming box and around the alert icon. Sorry, these really mess up the snapshot diffs.
* The way the ghosts highlighted the label moved from `ladder.scss` to `vehicle_icon.scss`.

### DIFF
--- a/assets/src/components/iconAlertCircle.tsx
+++ b/assets/src/components/iconAlertCircle.tsx
@@ -3,6 +3,7 @@ import React from "react"
 export enum AlertIconStyle {
   Black = 1,
   Grey,
+  GreyOnGrey,
   Highlighted,
 }
 
@@ -12,6 +13,8 @@ const styleClass = (style: AlertIconStyle): string => {
       return "--black"
     case AlertIconStyle.Grey:
       return "--grey"
+    case AlertIconStyle.GreyOnGrey:
+      return "--grey-on-grey"
     case AlertIconStyle.Highlighted:
       return "--highlighted"
   }
--- a/assets/src/components/iconAlertCircle.tsx
+++ b/assets/src/components/iconAlertCircle.tsx
@@ -1,20 +1,41 @@
 import React from "react"
 
-const IconAlertCircle = () => (
+export enum AlertIconStyle {
+  Black = 1,
+  Grey,
+  Highlighted,
+}
+
+const styleClass = (style: AlertIconStyle): string => {
+  switch (style) {
+    case AlertIconStyle.Black:
+      return "--black"
+    case AlertIconStyle.Grey:
+      return "--grey"
+    case AlertIconStyle.Highlighted:
+      return "--highlighted"
+  }
+}
+
+export interface Props {
+  style: AlertIconStyle
+}
+
+const IconAlertCircle = (props: Props) => (
   <svg viewBox="-3 -3 54 54" className="c-icon-alert-circle">
-    <IconAlertCircleSvgNode />
+    <IconAlertCircleSvgNode {...props} />
   </svg>
 )
 
-export const IconAlertCircleSvgNode = () => (
-  <>
+export const IconAlertCircleSvgNode = ({ style }: Props) => (
+  <g className={`c-icon-alert-circle${styleClass(style)}`}>
     <circle cx="24" cy="24" className="c-icon-alert-circle__outline" r="27" />
     <circle cx="24" cy="24" className="c-icon-alert-circle__fill" r="22.59" />
     <g className="c-icon-alert-circle__exclamation-point">
       <path d="m20.39 4.42h7.22a1.81 1.81 0 0 1 1.49 2.11l-1.49 23.22a1.73 1.73 0 0 1 -1.49 1.78h-4.24a1.73 1.73 0 0 1 -1.49-1.78l-1.49-23.22c-.07-1.13.61-2.11 1.49-2.11" />
       <path d="m21.57 34.54h4.87a1.88 1.88 0 0 1 1.82 1.93v5.17a1.89 1.89 0 0 1 -1.82 1.94h-4.87a1.89 1.89 0 0 1 -1.83-1.94v-5.17a1.88 1.88 0 0 1 1.83-1.93" />
     </g>
-  </>
+  </g>
 )
 
 export default IconAlertCircle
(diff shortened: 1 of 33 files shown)


arkadyan:
Haven't had a chance to look yet, but just a heads up that this needs a rebase.

assets/src/components/iconAlertCircle.tsx
```diff
@@ -1,20 +1,41 @@
...
+const IconAlertCircle = (props: Props) => (
   <svg viewBox="-3 -3 54 54" className="c-icon-alert-circle">
-    <IconAlertCircleSvgNode />
+    <IconAlertCircleSvgNode {...props} />
   </svg>
 )
 
-export const IconAlertCircleSvgNode = () => (
-  <>
+export const IconAlertCircleSvgNode = ({ style }: Props) => (
```
arkadyan: Is there a reason to consider having a default style?

assets/src/components/vehicleIcon.tsx
```diff
@@ -21,7 +21,7 @@ export interface Props {
   label?: string
   variant?: string | null
   status?: DrawnStatus
-  alertIcon?: boolean
+  alertIcon?: AlertIconStyle
```
arkadyan: To me, this sounds like you are passing in an icon. Would `alertIconStyle` be more descriptive?

assets/src/components/iconAlertCircle.tsx
```diff
@@ -1,20 +1,41 @@
...
+    case AlertIconStyle.Highlighted:
+      return "--highlighted"
+  }
+}
+
+export interface Props {
+  style: AlertIconStyle
+}
+
+const IconAlertCircle = (props: Props) => (
```
arkadyan: This is being used by `GhostPropertiesPanel`, does that need to be updated as well?

assets/src/models/blockWaiver.ts
```diff
@@ -51,18 +45,16 @@ export enum BlockWaiverDecoratorStyle {
  */
 export const blockWaiverDecoratorStyle = (
```
arkadyan: It feels to me like this function name deserves a change. It feels like it no longer accurately describes what it is providing.

arkadyan:
@skyqrose Can you please assign this to Ingrid to review as well?

assets/src/components/incomingBox.tsx
```diff
@@ -42,6 +45,7 @@ const IncomingBoxVehicle = ({
           status={drawnStatus(vehicleOrGhost)}
         />
       </div>
+      {alertIconStyle ? <IconAlertCircle style={alertIconStyle} /> : null}
```
arkadyan: I'm not sure the best words to describe it, but this feels like "logic as a side effect" to me in a way that feels problematic. That is, I think there are two questions here: a) should I display an alert icon?, and b) if so, what style icon should I use? But they are both encoded in `alertIconStyle` in a way that doesn't seem intuitive to me.

assets/src/components/iconAlertCircle.tsx
skyqrose: I don't think there is yet.

assets/src/components/iconAlertCircle.tsx
skyqrose: Yeah, it caused a merge conflict, so that fix ended up snuck into 090a2a6 "alert icon controls its own styling"

assets/src/components/incomingBox.tsx
skyqrose: The idea is that if `alertIconStyle` is undefined, that means "don't draw the icon at all". I can make that clearer by doing a `alertIconStyle === undefined` check and by declaring the type `const alertIconStyle : AlertIcontyle | undefined`.

skyqrose:
Another screenshot because it covers a case the other screenshots don't (a laying over vehicle on the bottom).

Screen Shot 2020-03-12 at 16 23 16

arkadyan:
Looks good to me pending a 👍 from Ingrid.

ingridpierre:
A couple visual things:

- For incoming vehicles, the vertical alignment should be even through the center of the ghost, icon and number. The ones I've seen have had the (!) icon a little high or a little low. Desired:
image


- For laying over vehicles, can we move the icon to be closer to the thin "point" and maybe shrink it by 2px for all laying over vehicles including ghosts? As-is, I feel the icon is overwhelming the arrow making it hard to see the status/vehicle icon underneath.

**Current:**
image

**Desired:**
image

skyqrose:
I bumped the icon up a couple pixels
Screen Shot 2020-03-13 at 11 54 42

I don't see any real world examples of laying over vehicles with block waivers, but here's the icons I was testing with, showing the adjusted and shrunk icon on the small vehicles.
Screen Shot 2020-03-13 at 12 09 43

ingridpierre:
@skyqrose that's looking better in most places.

-Let's have the (!) icon left or right aligned with the tip of sideways/laying over arrows. In the image above they sometimes look centered.

skyqrose:
Screen Shot 2020-03-13 at 14 02 38

ingridpierre:
image

I don't see the change I was expecting, maybe I'm mistaken. Are these (leftmost and rightmost examples) used for laying-over vehicles? Those icons still look centered.

skyqrose:
The laying over vehicles are all the small size that I shifted. The centered ones are the bigger size that's used for the up and down vehicles on the ladder. I left those in the same position as they currently are. Do you want those moved forward too?

PR #505, opened 2020-03-12
Bump react-tooltip from 4.1.0 to 4.1.1 in /assets

arkadyan:
Combines https://github.com/mbta/skate/pull/504 with updates to our snapshots.


arkadyan:
Ugh, it looks like these class names chang every single time they are run. That's annoying, we'll have to think harder about dealing with this change…

skyqrose:
Hmm... some options:
* Mock this so we don't see the library-generated content in tests.
* Don't keep this package up to date (or do so more rarely, with major versions / security fixes / if we need a new feature).
* Just keep going with these upgrades.

I think I'd go for the 2nd option. This is such a small part of our app that any upgrades are unlikely to be important to us.

arkadyan:
@skyqrose I added the mocking. I'm reluctant to give up on updates at this point in case something comes down that fixes the bugs we have with the library.

arkadyan:
@skyqrose good point, changed the mocking to happen globally.

PR #506, opened 2020-03-13
Enable block waivers for all users

arkadyan:
Remove the feature flag.

Waiting until we want this to go live in production before making a full PR.

~~Depends on https://github.com/mbta/skate/pull/502 (just the last commit is new).~~


arkadyan:
@skyqrose I removed a bunch more. Let me know if you spot any I got overly enthusiastic about.

PR #507, opened 2020-03-13
Refactor: Make the BlockWaivers structs on write instead of read

arkadyan:
Asana ticket: [Refactor: Make the BlockWaivers structs on write instead of read](https://app.asana.com/0/1148853526253426/1163344518291749)


lib/realtime/stop_time_updates_by_trip.ex
```diff
@@ -0,0 +1,16 @@
...
+  @type t :: %{Trip.id() => [StopTimeUpdate.t()]}
+
+  @spec stop_time_updates_for_trip(t(), Trip.id()) :: [StopTimeUpdate.t()]
+  def stop_time_updates_for_trip(stop_time_updates_by_trip, trip_id) do
+    Map.get(stop_time_updates_by_trip, trip_id, [])
+  end
+
+  @spec trip_ids(t()) :: [Trip.id()]
+  def trip_ids(stop_time_updates_by_trip) do
+    Map.keys(stop_time_updates_by_trip)
```
skyqrose: This `trip_ids()` function is only used in `stop_time_updates.ex`, and the `stop_time_updates_for_trips()` function is only used in `block_waiver.ex`. What do you think about moving these functions to those files, or inlining them since they're simple `Map` calls, so this module can go away?

test/concentrate/consumer/stop_time_updates_test.exs
```diff
@@ -39,6 +55,9 @@ defmodule Concentrate.Supervisor.StopTimeUpdatesTest do
 
   describe "handle_events/3" do
     setup do
+      reassign_env(:realtime, :trip_fn, fn _trip_id -> @trip end)
+      reassign_env(:realtime, :block_fn, fn _block_id, _service_id -> nil end)
```
skyqrose: If the mocked block function doesn't return a block, what's this testing? Should it return `[@trip]`?

test/realtime/block_waiver_store_test.exs
```diff
@@ -0,0 +1,62 @@
...
+    test "starts up and lives" do
+      {:ok, server} = BlockWaiverStore.start_link(name: :start_link)
+
+      Process.sleep(10)
+
+      assert Process.alive?(server)
+    end
+  end
+
+  describe "block_waivers_for_block_and_service/2" do
```
skyqrose: I think this could also use a test for the case where we don't have a block waiver for the block.

lib/concentrate/consumer/stop_time_updates.ex
```diff
@@ -29,19 +27,70 @@ defmodule Concentrate.Consumer.StopTimeUpdates do
...
   @spec stop_time_updates_by_trip([TripUpdate.t() | StopTimeUpdate.t()]) ::
-          stop_time_updates_by_trip()
+          StopTimeUpdatesByTrip.t()
   def stop_time_updates_by_trip(all_updates) do
     all_updates
     |> Enum.filter(&match?(%StopTimeUpdate{}, &1))
     |> Enum.group_by(& &1.trip_id)
   end
+
+  @type block_and_service_id :: {Block.id(), Service.id()}
```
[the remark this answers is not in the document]
arkadyan: Oh yeah, much better. I missed that, thanks.

lib/realtime/stop_time_updates_by_trip.ex
arkadyan: To me this is a collection of data (kind of a DTO) that we have during a part of our flow and I find it helpful to name it and encapsulate actions on it. Yes it happens to use a map and live in memory right now, but other objects that interact with it don't need to worry about those details.

arkadyan:
@skyqrose Thanks, I took in all your feedback.

arkadyan:
@skyqrose Oh yeah, I didn't think about those names but that's a good idea.

PR #514, opened 2020-03-16
Don't display a vehicle as invalid if it is a block overload

arkadyan:
Asana ticket: [Don't display a vehicle as invalid if it is a block overload](https://app.asana.com/0/1148853526253426/1166770269732563)


skyqrose:
Oh hold on reread the asana task:

> Swifly won't currently provide predictions for these vehicles, so we believe they will appear as inactive. 

We don't care about predictions. Will it also not provide trip assignments?

If so, this might not be enough to show these vehicles correctly. Busloc provides a trip assignment but not a stop assignment, so we wouldn't be able to show them on the ladder.

arkadyan:
@skyqrose I've updated this substantially now that we are seeing this data come through the pipeline. There are several new commits before the ones that were there. Swiftly is giving us assignments, so we are able to draw the vehicles on the ladder.

lib/concentrate/vehicle_position.ex
```diff
@@ -181,6 +181,23 @@ defmodule Concentrate.VehiclePosition do
...
+    defp overload_priority(block_id1, block_id2) do
+      cond do
+        String.contains?(block_id1, "-OL") ->
+          block_id1
+
+        String.contains?(block_id2, "-OL") ->
+          block_id2
+
+        true ->
+          first_value(block_id1, block_id2)
```
skyqrose:
```suggestion
          block_id1
```
`first_value()` checks for `nil`, but the function cases already check for `nil`, so this can just use the value directly.

lib/concentrate/vehicle_position.ex
```diff
@@ -181,6 +181,23 @@ defmodule Concentrate.VehiclePosition do
...
+    defp overload_priority(block_id1, nil), do: block_id1
+
+    defp overload_priority(nil, block_id2), do: block_id2
+
+    defp overload_priority(block_id1, block_id2) do
+      cond do
+        String.contains?(block_id1, "-OL") ->
+          block_id1
+
+        String.contains?(block_id2, "-OL") ->
```
skyqrose: Can use `Block.overload?()`

lib/realtime/vehicle.ex
```diff
@@ -189,6 +192,7 @@ defmodule Realtime.Vehicle do
       via_variant: via_variant,
       bearing: VehiclePosition.bearing(vehicle_position),
       block_id: block_id,
+      block_id_with_overload: block_id_with_overload,
```
skyqrose: It looks aside from the off course check, this value isn't used anymore, so we could take the field out of the struct.

arkadyan:
@skyqrose Swiftly is giving us the stop assignment, but not necessarily the right trip or other info, but we can get those from Busloc.

lib/realtime/vehicle.ex
arkadyan: At least for the moment I want to leave it on the client so we can debug the data in the browser without having to add breakpoints in the server or whatever.

arkadyan:
@skyqrose Thanks, made a couple updates.

lib/realtime/vehicle.ex
skyqrose: I don't like the idea of adding a field to the vehicle struct just for debugging. What about making a `data_discrepancy` for `block_id`? (Then it'd also show up when the debug flag is on.)

arkadyan: Ok, changed to that.

PR #517, opened 2020-03-17
tolerate slightly changed stop_ids

skyqrose:
Followup to #516


skyqrose:
Screen Shot 2020-03-17 at 11 10 16

skyqrose:
This problem was fixed upstream, and this fix is no longer needed.

PR #523, opened 2020-03-18
Add "New" indicator to search result cards

arkadyan:
For buses that have logged in within the last 30 minutes.

Asana ticket: [Add "New" indicator to search result cards for buses that have logged in within last 30 min](https://app.asana.com/0/1148853526253426/1160813278503797)

Screen Shot 2020-03-18 at 19 09 00

### DIFF
--- a/assets/css/_properties_list.scss
+++ b/assets/css/_properties_list.scss
@@ -28,5 +28,9 @@
   .m-properties-list__property--last-login & {
     color: $color-font-grey;
     font-size: 0.75rem;
+
+    .m-search-results__card--new & {
+      color: $color-secondary-dark;
+    }
   }
 }
--- a/assets/css/_properties_list.scss
+++ b/assets/css/_properties_list.scss
@@ -28,9 +28,5 @@
   .m-properties-list__property--last-login & {
     color: $color-font-grey;
     font-size: 0.75rem;
-
-    .m-search-results__card--new & {
-      color: $color-secondary-dark;
-    }
   }
 }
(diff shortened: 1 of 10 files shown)


ingridpierre:
Looks great, some vehicle info looked incorrect due to the data issues we're having (no trip headsigns displayed on active vehicles), but that seems separate from this pr.

One subtle detail to add. Can we change the dropshadow to the new cards to this:
```
.m-search-results__card--new {
box-shadow: 0 2px 4px 0 rgba(77, 182, 172, 0.3);
}
```

assets/tests/components/searchResults.test.tsx
```diff
@@ -120,6 +120,72 @@ describe("SearchResults", () => {
     expect(tree).toMatchSnapshot()
   })
 
+  test("renders vehicles that have logged in within the past 30 minutes", () => {
```
skyqrose: A name like "shows the new badge for vehicle that have logged in within the past 30 minutes" would better indicate why we need the test.

assets/css/_properties_list.scss
```diff
@@ -28,5 +28,9 @@
   .m-properties-list__property--last-login & {
     color: $color-font-grey;
     font-size: 0.75rem;
+
+    .m-search-results__card--new & {
```
skyqrose: Either `_properties_list.scss` needs a `.m-search-results` selector, or `_search_results.scss` needs a `.m-properties-list` selector. I think I'd rather this be in `_search_results.scss` and have the parent know the details about its child, than to have it here where the child would need to know details about its parent.

arkadyan:
@ingridpierre thanks, updated that box shadow.

assets/css/_properties_list.scss
arkadyan: Yeah, I went back and forth, but I agree with you, that's better.

ahakuta:
Just wanted to let you know @arkadyan I took a look and this ticket looks great!

PR #524, opened 2020-03-19
Don't make a ghost indicator if another vehicle is on the run

arkadyan:
Asana ticket: [🐞 Showing both a ghost and a regular bus on the route](https://app.asana.com/0/1148853526253426/1161461620801441)

### DIFF
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -184,19 +184,16 @@ const runNotSharedByAnotherVehicle = (vehiclesAndGhosts: VehicleOrGhost[]) => (
     return false
   }
 
-  const otherVehicles = onlyVehicles(vehiclesAndGhosts).filter(
-    ({ id }) => id !== vehicle.id
-  )
+  const otherVehicles = vehiclesAndGhosts.filter(({ id }) => id !== vehicle.id)
   const otherRunIds = runIds(otherVehicles)
 
   return !otherRunIds.includes(vehicle.runId)
 }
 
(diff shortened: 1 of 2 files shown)


assets/src/components/routeLadder.tsx
```diff
@@ -177,6 +177,27 @@ export const groupByPosition = (
...
+
+  const otherVehicles = onlyVehicles(vehiclesAndGhosts).filter(
+    ({ id }) => id !== vehicle.id
+  )
+  const otherRunIds = runIds(otherVehicles)
+
+  return !otherRunIds.includes(vehicle.runId)
+}
+
+const onlyVehicles = (vehiclesAndGhosts: VehicleOrGhost[]): Vehicle[] =>
```
skyqrose: Is this needed? Ghosts wouldn't be made with a matching run, and even if there was one, you probably wouldn't want to draw the virtual ghost.

assets/src/components/routeLadder.tsx
```diff
@@ -177,6 +177,27 @@ export const groupByPosition = (
   }
 }
 
+const runNotSharedByAnotherVehicle = (vehiclesAndGhosts: VehicleOrGhost[]) => (
+  vehicle: Vehicle
+): boolean => {
+  if (vehicle.runId === null) {
```
skyqrose: When do we ever have a null run id?

arkadyan: I don't have a great sense, but it seems like they come in occasionally, but this is also just about dealing with the type definition.

assets/src/components/routeLadder.tsx
arkadyan: Yeah, I guess that's true.

PR #545, opened 2020-03-30
Only create block waivers for stop time updates with a “remark”

arkadyan:
This field indicates a block waiver, as opposed to a prediction or other
types of stop time update.

Require block waivers to include a remark.

Asana ticket: [Only create block waivers for stop time updates with a “remark”](https://app.asana.com/0/1148853526253426/1168716458967945)


arkadyan:
@paulswartz if you have a chance, could you please take a look over this?

PR #546, opened 2020-03-30
Don't consider a vehicle "invalid" when on a shuttle run

arkadyan:
Asana ticket: [Shuttle vehicles should not be labeled invalid](https://app.asana.com/0/1148853526253426/1168749847411529)


arkadyan:
@paulswartz I think the big reason is that the "clients" that want to know about this are all in Javascript on the client-side. So we need to pass data across, we can't rely on an Elixir function.

arkadyan:
@paulswartz makes sense, added a `is_shuttle` property to Vehicle.

ingridpierre:
@arkadyan This looked good to me 👍, invalid vehicles showed the message everywhere that they were found and all the shuttles did not. For better or worse there were a lot of specials online right now to test with :-/

PR #549, opened 2020-03-31
Add train icons to shuttle map

arkadyan:
Asana ticket: [Add train icons to shuttle map](https://app.asana.com/0/1148853526253426/1140898276160665)

~~**NB:** Requires 2 new env vars, `API_URL` and `API_KEY`.~~ Added

Note that the location data we get from the trains doesn't always place them directly over the route lines.

Screen Shot 2020-03-31 at 15 10 43


ingridpierre:
@arkadyan Let's update the subway shape stroke width to be 4. Also, it looks like the stroke on the vehicles is on the thin side. Should I provide a *thicker* svg or should we change that programmatically to look more like the stroke on the stop icons and the other bus vehicles.

assets/src/realtime.d.ts
```diff
@@ -49,6 +49,16 @@ export type RouteStatus = "on_route" | "laying_over" | "pulling_out"
 
 export type RunId = string
 
+export type TrainVehicleId = string
+
+export interface TrainVehicle {
```
skyqrose, line 54: "TrainVehicle" sounds verbose. What do you think about a global find/replace "TrainVehicle" to just "Train"?

assets/src/components/map.tsx
```diff
@@ -158,6 +160,74 @@ export const updateMarkers = (
...
+          <path fill="#fff" d="m42.88 45.83a2.1 2.1 0 0 1 -.87-.19l-15.92-7.17a5.23 5.23 0 0 0 -2.09-.47 5.14 5.14 0 0 0 -2.08.44l-15.92 7.2a2.1 2.1 0 0 1 -.87.19 2.14 2.14 0 0 1 -1.76-1 2 2 0 0 1 -.12-2l18.86-40.83a2.08 2.08 0 0 1 3.78 0l18.87 40.87a2 2 0 0 1 -.12 2 2.14 2.14 0 0 1 -1.76.96z"/>
+        </g>
+    </svg>`,
+    iconAnchor: [centerX, centerY],
+    className: "m-vehicle-map__train-icon",
+  })
+}
+
+export const updateTrainVehicleMarkers = (
+  newTrainVehicles: { [id: string]: TrainVehicle },
```
skyqrose, line 178:
```suggestion
  newTrainVehicles: TrainVehiclesById,
```

since that type alias is used elsewhere.

assets/src/realtime.d.ts
```diff
@@ -49,6 +49,16 @@ export type RouteStatus = "on_route" | "laying_over" | "pulling_out"
...
+  id: TrainVehicleId
+  directionId: DirectionId
```
skyqrose, line 56: Is this field used? Can we remove it, and just skip getting it from the api?

assets/src/components/shuttleMapPage.tsx
```diff
@@ -23,6 +25,14 @@ const filterShuttles = (
   )
 }
 
+export const allTrainVehicles = (
+  trainVehiclesByRouteId: ByRouteId<TrainVehicle[]>
+): TrainVehicle[] =>
+  Object.values(trainVehiclesByRouteId).reduce(
+    (acc, trains) => acc.concat(trains),
+    []
+  )
```
skyqrose, line 34: I think we have this as `helpers/array.ts` `flatten()`.

assets/src/components/shuttleMapPage.tsx
```diff
@@ -39,6 +49,14 @@ const ShuttleMapPage = ({}): ReactElement<HTMLDivElement> => {
   const { socket }: { socket: Socket | undefined } = useContext(SocketContext)
   const shuttles: Vehicle[] | null = useShuttleVehicles(socket)
   const shapes: Shape[] = useRouteShapes(selectedShuttleRouteIds)
+
+  const selectedSubwayRouteIds: RouteId[] = selectedShuttleRouteIds.filter(
+    isASubwayRoute
+  )
+  const trainVehicles: TrainVehicle[] = allTrainVehicles(
+    useTrainVehicles(socket, selectedSubwayRouteIds)
```
skyqrose, line 57: This does meet the rules for calling hooks, since it's called in the same way every time, but it still surprised me to see a hook call inside a function argument instead of as a top level variable assignment.

assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -116,3 +119,20 @@ describe("Shuttle Map Page", () => {
     expect(tree).toMatchSnapshot()
   })
 })
+
+describe("allTrainVehicles", () => {
```
skyqrose, line 123: I expected to see a test in here that's something to do with the `useTrainVehicles` hook. Maybe something like checking that the hook is called with the selected subway route, or that results from the hook get passed into the map. I don't quite know what's testable, but it feels like this one new unit test doesn't really cover all the new functionality that `shuttleMapPage` got.

arkadyan:
@ingridpierre increased both stroke widths.

arkadyan:
@skyqrose The major pieces moving up the chain are: `api`, which is the base level of streaming any data from the API; `train_vehicles`, which streams specifically subway vehicles and publishes diffs over `Phoenix.PubSub`; and finally `TrainVehiclesPubSub` sits in our Realtime app, subscribes to that pub-sub, as a data source, maintains a complete data store and a registry for client channels like our other realtime modules do. This is all comparable to—though simpler than—how dotcom does things.

assets/src/realtime.d.ts
arkadyan: I went back and forth on this. The big sticking point for me was that everywhere else in the world these are simply vehicles. Since we default to vehicle meaning specifically a bus I wanted to differentiate them, but I was afraid of adding confusion in the long run by divorcing them completely from the vehicle terminology.

lib/realtime/supervisor.ex
```diff
@@ -14,6 +14,7 @@ defmodule Realtime.Supervisor do
       {Registry, keys: :duplicate, name: registry_name()},
       {Realtime.BlockWaiverStore, name: Realtime.BlockWaiverStore.default_name()},
       {Realtime.Server, name: Realtime.Server.default_name()},
+      {Realtime.TrainVehiclesPubSub, name: Realtime.TrainVehiclesPubSub.default_name()},
```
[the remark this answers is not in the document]
arkadyan: The data is coming from a separate part of the tree, but it feels to me like what this module is doing is very specific to the realtime functionality supporting client channels. It manages a data store and a registry for channels—it's very equivalent to `Vehicles` or `DataStatusPubSub`.

assets/tests/components/shuttleMapPage.test.tsx
arkadyan: Added a test for rendering ShuttleMapPage with train vehicles.

lib/api/stream.ex
```diff
@@ -0,0 +1,108 @@
...
+  defmodule Event do
+    @moduledoc """
+    Struct representing a parsed Api server-sent event.
+    """
+    defstruct data: nil, event: :unknown
+    @type event :: :reset | :add | :update | :remove
+    @type t :: %__MODULE__{
+            event: event | :unknown,
+            data: nil | JsonApi.t() | {:error, any}
+          }
```
skyqrose, line 33: Nothing makes the `data: nil` case or the `event: :unknown` case. They could be removed, and the fields marked by `@enforce_keys`.

lib/api/json_api.ex
```diff
@@ -0,0 +1,191 @@
+defmodule JsonApi.Item do
+  defstruct [:type, :id, :attributes, :relationships]
+
+  @type t :: %JsonApi.Item{
+          type: String.t(),
+          id: String.t(),
+          attributes: %{String.t() => any},
+          relationships: %{String.t() => list(JsonApi.Item.t())}
```
skyqrose, line 8: I see this is hydrating relationships by pulling their values from the included data. I have some vague memories of hearing other teams regret doing that, and I don't think it'd work for streaming anyway since included resources are only sent once the first time they're seen, so we probably shouldn't.

lib/api/stream.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  defmodule Event do
+    @moduledoc """
+    Struct representing a parsed Api server-sent event.
+    """
+    defstruct data: nil, event: :unknown
+    @type event :: :reset | :add | :update | :remove
+    @type t :: %__MODULE__{
+            event: event | :unknown,
+            data: nil | JsonApi.t() | {:error, any}
```
skyqrose, line 32: `:error` will always be `JsonApi.error()`, so we can be more specific than `any`.

lib/api/json_api.ex
```diff
@@ -0,0 +1,191 @@
...
+  defp parse_data(%{"id" => _} = data) do
+    # V3Api.Stream receives :add, :update, and :remove data as single items
+    parse_data(%{"data" => data})
+  end
+
+  defp parse_data(%{}) do
+    {:error, :invalid}
+  end
+
+  def parse_data_item(%{"type" => type, "id" => id, "attributes" => attributes} = item, included) do
```
skyqrose, line 100: A bunch of functions in this file are missing `@spec`s.

lib/api/json_api.ex
```diff
@@ -0,0 +1,191 @@
...
+end
+
+defmodule JsonApi do
+  defstruct links: %{}, data: []
+  @type t :: %JsonApi{links: %{String.t() => String.t()}, data: list(JsonApi.Item.t())}
+
+  @spec empty() :: JsonApi.t()
+  def empty do
+    %JsonApi{
+      links: %{},
```
skyqrose, line 30: I don't expect we'll use links at all. Could we remove it?

lib/train_vehicles/supervisor.ex
```diff
@@ -0,0 +1,39 @@
+defmodule TrainVehicles.Supervisor do
```
skyqrose, line 1: Right now we have data going from the GenState pipeline, into a pubsub library, and then into our own homegrown pubsub, which seems a bit redundant. Could we remove one of those steps? A few ideas for ways we might do that:
* Call a `GenServer.cast` directly, like `DataStatusPubSub.update()`, instead of using the pubsub library.
* Make `TrainVehiclesPubSub` a GenStage consumer so it gets the data via the genstage instead of as a separate pubsub step. I think `Realtime` isn't set up like this because the broadcasting to channels predates the genstage pipeline.
* Use the pubsub library for TrainVehiclesPubSub instead of making our own pubsub with registry.

lib/train_vehicles/stream.ex
```diff
@@ -0,0 +1,88 @@
...
+  @spec broadcast([TrainVehicle.t() | String.t()], event_type, broadcast_fn) :: :ok
+  defp broadcast([], _type, _broadcast_fn), do: :ok
+
+  defp broadcast(data, type, broadcast_fn) do
+    TrainVehicles.PubSub
+    |> broadcast_fn.("train_vehicles", {type, data})
+    |> do_broadcast()
+  end
+
+  @spec do_broadcast(:ok | {:error, any}) :: :ok
```
skyqrose, line 80: This isn't doing any broadcasting. Maybe `log_errors` would be a better name.

lib/train_vehicles/stream.ex
```diff
@@ -0,0 +1,88 @@
...
+    |> broadcast(type, broadcast_fn)
+  end
+
+  @typep broadcast_fn :: (atom, String.t(), any -> :ok | {:error, any})
+  @spec broadcast([TrainVehicle.t() | String.t()], event_type, broadcast_fn) :: :ok
+  defp broadcast([], _type, _broadcast_fn), do: :ok
+
+  defp broadcast(data, type, broadcast_fn) do
+    TrainVehicles.PubSub
+    |> broadcast_fn.("train_vehicles", {type, data})
```
skyqrose, line 76: This "train_vehicles` argument is always the same anywhere. Since we control the `broadcast_fn` type, could we just remove that argument from the function? (I know you'd still need to specify it somewhere for the pubsub topic, but it could be specified in the function definition instead of here.)

lib/realtime/train_vehicles_pub_sub.ex
```diff
@@ -0,0 +1,177 @@
...
+
+    new_state = Map.put(state, :train_vehicles_by_route_id, train_vehicles_by_route_id)
+
+    _ = broadcast(new_state)
+
+    {:noreply, new_state}
+  end
+
+  def handle_info(
+        {:add, new_train_vehicles},
```
skyqrose, line 81: This module has two major responsibilities:
* broadcasting new data to channels.
* updating the list of known vehicles as streaming data trickles in

Both are pretty complex tasks. Could we pull the "updating with new stream data" into a separate module (that's just functions, not a gen server). It'd have an interface like `@spec update_data(train_vehicles_by_route_id(), event(), new_data()) :: train_vehicles_by_route_id())`

lib/api/stream.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  Required options:
+  `:path` (e.g. "/vehicles")
+  `:name` -- name of module
+  `:subscribe_to` -- pid or name of a ServerSentEventStage
+  for the Api.Stream to subscribe to. This should be
+  started as part of a supervision tree.
+
+  Other options are made available for tests, and can include:
+  - :name (name of the GenStage process)
```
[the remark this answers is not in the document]
arkadyan: Removed that, it's the same.

lib/api/json_api.ex
arkadyan: This is all copied directly from dotcom so we can consider future improvements, but it works for streaming.

lib/api/stream.ex
arkadyan: I don't think that's true from what JsonApi says.

lib/train_vehicles/supervisor.ex
arkadyan: The benefit of this approach is that streaming from the API and sending data to a web client over a channel are decoupled. The former doesn't need to know anything about the latter and we have flexibility to do things like have something new subscribe to train data without having to make any changes.

lib/train_vehicles/stream.ex
arkadyan: I'm not sure I'm following your suggestion. The `broadcast_fn` is `PubSub.broadcast/3` which takes the topic.

lib/realtime/supervisor.ex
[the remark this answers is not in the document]
arkadyan: If I'm following right I think this would be another benefit of the decoupling - this has no dependence on the train_vehicle/api streaming stuff, so there's no need to restart this if that restarts.

lib/realtime/train_vehicles_pub_sub.ex
arkadyan: Good idea, I refactored out a `TrainVehiclesStore`.

lib/train_vehicles/stream.ex
skyqrose: Ah, I missed that you were literally using `&Pubsub.broadcast/3`. I thought you were defining your own function somewhere. Nevermind.

lib/realtime/supervisor.ex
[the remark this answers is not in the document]
skyqrose: `TrainVehiclesPubSub` has a lot more dependence on the train_vehicle/api streaming than it does on `Server` or `DataStatusPubSub`.

lib/realtime/train_vehicles_pub_sub.ex
```diff
@@ -0,0 +1,121 @@
...
+  def handle_info({:reset, train_vehicles}, state) do
+    new_state = TrainVehiclesStore.reset(state, train_vehicles)
+
+    _ = broadcast(new_state)
+
+    {:noreply, new_state}
+  end
+
+  def handle_info({:add, new_train_vehicles}, state) do
+    new_state = TrainVehiclesStore.add(state, new_train_vehicles)
```
skyqrose, line 70: I like `TrainVehiclesStore`. Can move one more layer of logic into it, by having `TVPubSub` get a generic "new train vehicles" message, and let `TVStore` take care of casing over whether it's a reset/add/update/remove?

skyqrose:
It looks like the codecov misses are all logging, impossible defaults in exhaustive switch statements, and stuff like that. I'm fine with ignore them.

PR #565, opened 2020-04-06
Separate GTFS and HASTUS files

skyqrose:
Old file structure:
```
├ gtfs.ex
└ gtfs/
  ├ supervisor.ex
  ├ cache_file.ex
  ├ csv.ex
  ├ data.ex
  ├ health_server.ex
  ├ helpers.ex
  │
  ├ block.ex
  ├ calendar.ex
  ├ direction.ex
  ├ route.ex
  ├ route_pattern.ex
  ├ run.ex
  ├ service.ex
  ├ shape.ex
  ├ stop.ex
  ├ stop_time.ex
  ├ timepoint.ex
  ├ trip.ex
  │
  └ hastus/
    ├ activity.ex
    ├ place.ex
    ├ schedule.ex
    └ trip.ex
```

New file structure
```
├ static.ex
└ static/
  ├ supervisor.ex
  ├ cache_file.ex
  ├ csv.ex
  ├ data.ex
  ├ health_server.ex
  ├ helpers.ex
  │ 
  ├ block.ex
  ├ route.ex
  ├ trip.ex
  │
  ├ gtfs/
  │ ├ calendar.ex
  │ ├ direction.ex
  │ ├ route.ex
  │ ├ route_pattern.ex
  │ ├ service.ex
  │ ├ shape.ex
  │ ├ stop.ex
  │ ├ stop_time.ex
  │ ├ timepoint.ex
  │ └ trip.ex
  │
  └ hastus/
    ├ activity.ex
    ├ place.ex
    ├ run.ex
    ├ schedule.ex
    └ trip.ex
```

When the only static data we had was from GTFS, having it all in a folder named `gtfs/` made sense. But now that HASTUS data is more fleshed out, HASTUS and GTFS can be treated more symmetrically, with some shared resources (like csv parsing and `data.ex`) not belonging to one or the other.

Everything in `gtfs/` is pure GTFS, directly mirrors the GTFS format, and doesn't reference `hastus/`.
Everything in `hastus/` is pure HASTUS, directly mirrors the csv files, and doesn't reference `gtfs/`.
Shared resourced and skate-specific data (like our merged trip) live directly in `static/`.

Having a separation between the GTFS trip type and our own merged trip type will make things easier as we start doing more complicated merging, with nonrevenue trips and new fields, in order to support minischedules.

Almost everything in this PR is just file and module name changes. The two actual code changes:
* Splits the old gtfs trip into a `static/gtfs/trip.ex` that mirrors the GTFS format, and a `static/trip.ex` that's our representation with run ids and stop times.
* As a followup from #562, deduplicates hastus run id formatting into `static/hastus/run.ex`

~Depends on #562. Wait til that merges to review this.~


arkadyan:
How would you feel about a name like "schedule" or something along those lines instead of "static"? I bump on static for two reasons, 1) it makes me think static assets, and 2) it's not that this data never changes, it just changes much slower than the realtime data. Currently we only load it once at load, but we have back-burnered thoughts about possibly reloading fresh versions, so that doesn't out of the realm of possibility. To me, I think when we talk about our data we generally refer to schedule data vs realtime.

skyqrose:
#562 is merged, so this is ready for a review now.

I've done the static/schedule rename. I like that.

PR #569, opened 2020-04-08
Make Minischedules data from HASTUS data

skyqrose:
Asana Task: [Import the new mini-schedule-related HASTUS data into Skate](https://app.asana.com/0/1148853526253426/1166438708199679)

Preprocesses hastus data into the format we'll need for minischedules, and stores it in Schedule.Data

This does not include a way to get it out of Data yet.

Has a placeholder implementation for the details of turning trips into pieces and understanding Operator and Sign-on activities.

The big picture of separating trips and activities into their blocks and runs is done, and will let us start working on fetching this data into the frontend, though the placeholder means the data won't be quite right.

### DIFF
--- a/lib/schedule/minischedule/block.ex
+++ b/lib/schedule/minischedule/block.ex
@@ -1,0 +1,27 @@
+defmodule Schedule.Minischedule.Block do
+  alias Schedule.Block
+  alias Schedule.Minischedule.Piece
+  alias Schedule.Hastus.Schedule
+
+  @type key :: {Schedule.id(), Block.id()}
+
+  @type by_id :: %{key() => t()}
+
+  @type t :: %__MODULE__{
+          schedule_id: Schedule.id(),
+          id: Block.id(),
+          pieces: [Piece.t()]
+        }
+
+  @enforce_keys [
+    :schedule_id,
+    :id,
+    :pieces
+  ]
+
+  defstruct [
+    :schedule_id,
+    :id,
+    :pieces
+  ]
+end
--- a/lib/schedule/minischedule/load.ex
+++ b/lib/schedule/minischedule/load.ex
@@ -6,10 +6,7 @@ defmodule Schedule.Minischedule.Load do
   alias Schedule.Helpers
   alias Schedule.Hastus.Activity
   alias Schedule.Hastus.Trip
-  alias Schedule.Minischedule.Block
-  alias Schedule.Minischedule.Break
-  alias Schedule.Minischedule.Piece
-  alias Schedule.Minischedule.Run
+  alias Schedule.Minischedule.{Block, Break, Piece, Run}
 
   @spec from_hastus([Activity.t()], [Trip.t()]) ::
           %{runs: Run.by_id(), blocks: Block.by_id()}
--- a/lib/schedule/minischedule/load.ex
+++ b/lib/schedule/minischedule/load.ex
@@ -11,10 +11,12 @@ defmodule Schedule.Minischedule.Load do
   @spec from_hastus([Activity.t()], [Trip.t()]) ::
           %{runs: Run.by_id(), blocks: Block.by_id()}
   def from_hastus(activities, trips) do
-    activities_by_run = group_activities_by_run(activities)
-    trips_by_run = group_trips_by_run(trips)
-    run_groups = pair_activities_and_trips(activities_by_run, trips_by_run)
-    runs_and_pieces = Enum.map(run_groups, &run_and_pieces_from_run_group/1)
+    activities_by_run = Enum.group_by(activities, fn a -> {a.schedule_id, a.run_id} end)
+    trips_by_run = Enum.group_by(trips, fn t -> {t.schedule_id, t.run_id} end)
+    activities_and_trips_by_run = Helpers.pair_maps(activities_by_run, trips_by_run)
+    runs_and_pieces = Enum.map(activities_and_trips_by_run, fn {run_key, {activities, trips}} ->
+      run_and_pieces(run_key, activities, trips)
+    end)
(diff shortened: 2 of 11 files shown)


lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+
+  @spec run_key_for_trip(Trip.t()) :: Run.key()
+  defp run_key_for_trip(trip) do
+    {trip.schedule_id, trip.run_id}
+  end
+
+  @typep run_group :: {Run.key(), [Activity.t()], [Trip.t()]}
+
+  # Assumes inputs are both sorted by run_key.
+  # Assumes no runs are in trips.csv but not in activities.csv
```
arkadyan, line 60: Trying to be sure I understand this. Is this rephrasing equivalent? : "Assumes that any run in trips.csv is also in activities.csv"

lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+  @spec pair_activities_and_trips([activity_group()], [trip_group()]) :: [run_group()]
+  defp pair_activities_and_trips([], _trip_groups) do
```
arkadyan, line 62: I think some unit tests might be helpful for this function, personally I don't find it intuitively obvious what it's doing.

lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+    []
+  end
+
+  defp pair_activities_and_trips([activity_group | other_activity_groups], trip_groups) do
+    {run_key, activities} = activity_group
+
+    case trip_groups do
+      [{^run_key, trips} | other_trip_groups] ->
```
arkadyan, line 70: If `activity_groups` and `trip_groups` were maps instead of lists of tuples would a lot of this complexity and recursion go away? I.e. could you just iterate over activities if you could reference trips by `run_key`?

(If so I might retract the unit test comment.)

lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+    # Currently, turns the trips directly into pieces, and makes every activity a break.
+    # The real way to do it would be to integrate Sign-on and Operator activities into pieces.
+    {schedule_id, run_id} = run_key
+
+    breaks = Enum.map(activities, &break_from_activity/1)
+
+    pieces =
+      trips
+      |> Helpers.split_by(fn trip -> trip.block_id end)
+      |> Enum.map(fn {block_id, trips} ->
```
arkadyan, line 96: This feels like a candidate for breaking out as a named function.

lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+      schedule_id: schedule_id,
+      id: run_id,
+      activities: breaks ++ pieces
+    }
+
+    {run_key, run, pieces}
+  end
+
+  @spec break_from_activity(Activity.t()) :: Break.t()
+  defp break_from_activity(activity) do
```
arkadyan, line 128: Would it make sense for this to live on `Break`?

lib/schedule/minischedule/load.ex
```diff
@@ -0,0 +1,157 @@
...
+      start_place: activity.start_place,
+      end_place: activity.end_place
+    }
+  end
+
+  @spec blocks_from_pieces([Piece.t()]) :: Block.by_id()
+  defp blocks_from_pieces(pieces) do
+    pieces
+    |> Enum.group_by(&block_key_for_piece/1)
+    |> Enum.map(fn {{schedule_id, block_id} = block_key, pieces} ->
```
arkadyan, line 142: Can this be combined into the `Map.new` line below?

lib/schedule/minischedule/load.ex
skyqrose: Yes, that's a better way to put it.

skyqrose: Ended up being obsolete after changing to using maps instead of lists of tuples.

lib/schedule/minischedule/load.ex
skyqrose: After changing to using maps instead of lists of tuples, this became `Helpers.pair_maps()`, and got easier to understand.

lib/schedule/minischedule/load.ex
skyqrose: Yeah, it all became a lot better. 👍

lib/schedule/minischedule/load.ex
skyqrose: If it were permanent, yes, but this is the placeholder implementation. The real implementation for making pieces will need to depend on activities in some way (so the piece can start with a sign-on). It will be easier to do that transition if all of the placeholder is contained in this one function.

skyqrose: I did end up making a change (just returning the run, instead of the run and pieces) that makes this big function a bit shorter.

lib/schedule/minischedule/block.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Schedule.Minischedule.Block do
+  alias Schedule.Block
```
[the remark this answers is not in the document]
skyqrose: I like that idea and I tried it, but then the way `{Schedule.id(), Schedule.Block.id()}` has two different `Schedule`s in it was a little weird, and then the aliasing started expanding `Schedule.Block` into `Schedule.Hastus.Schedule.Block`.... it just got messy so I gave up.

skyqrose:
Ready for another look. Using maps instead of lists ended up cleaning a lot of things up.

PR #581, opened 2020-04-14
Show "ADDED" for overloaded vehicles

arkadyan:
Asana ticket: [Treatment for overloaded vehicles (route ladder, Search, VPP, maps)](https://app.asana.com/0/1148853526253426/1169592928381123)

Screen Shot 2020-04-14 at 11 01 52
Screen Shot 2020-04-14 at 11 02 10
Screen Shot 2020-04-14 at 11 02 55
Screen Shot 2020-04-14 at 11 03 23
Screen Shot 2020-04-14 at 11 06 56

### DIFF
--- a/assets/src/components/propertiesList.tsx
+++ b/assets/src/components/propertiesList.tsx
@@ -29,10 +29,11 @@ const vehicleProperties = (vehicle: Vehicle): Property[] => {
   return [
     {
       label: "Run",
-      value:
-        isVehicle(vehicle) && vehicle.isShuttle
-          ? formattedRunNumber(vehicle)
-          : runId || "Not Available",
+      value: vehicle.isShuttle
+        ? formattedRunNumber(vehicle)
+        : vehicle.isOverload && !!vehicle.runId
+        ? `ADDED ${runId}`
+        : runId || "Not Available",
     },
     {
       label: "Vehicle",
--- a/assets/src/helpers/vehicleLabel.ts
+++ b/assets/src/helpers/vehicleLabel.ts
@@ -21,7 +21,11 @@ const vehicleLabel = (
 
   switch (vehicleLabelSetting(settings, vehicleOrGhost)) {
     case VehicleLabelSetting.RunNumber:
-      return runIdToLabel(vehicleOrGhost.runId)
+      if (isVehicle(vehicleOrGhost) && vehicleOrGhost.isOverload) {
+        return "ADDED"
+      } else {
+        return runIdToLabel(vehicleOrGhost.runId)
+      }
     case VehicleLabelSetting.VehicleNumber:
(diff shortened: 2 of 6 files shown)


ingridpierre:
For the most part, labels looked good where they were applied. I'm still trying to observe more overload examples to provide feedback on, as they've been a little scant lately. The only added vehicle I was able to observe was incoming and not on the ladder.

-----
*Null route name*
It looks like the route # is coming up as null when the vehicle is inbound. Since the incoming vehicle is associated with the route somehow to belong to the incoming box, is it possible to update the "null" here with the route number?
image

-------------
**Missing invalid + ghost overload?**
On the same route last night there were two overloads on the route (one was invalid, then became a ghost), only the OL 2 which was in the incoming box appeared in skate:
image
image

arkadyan:
@ingridpierre Do you know if maybe the inbound vehicle was pulling out from a garage? The properties panel shows the work the vehicle is currently doing, which means for an incoming vehicle it is not the route that you are currently looking at.

ahakuta:
I'm looking at an overload right now on rte 350 which is making me wonder the following: **should PULL-B and SW-OFF labels override the ADDED label on overloads?** See the image below, where the second bus on the left side of the rte 350 ladder with the PULL-B label is the overloaded vehicle (it's a green arrow).

Screenshot 2020-04-15 15 10 37

After speaking with @ingridpierre about this, we think the answer is that the PULL-B and SW-OFF labels shouldn't override the ADDED label. The ADDED label is important to see, and the overload duration may well exceed the duration of the run end time that is generating the PULL-B label. It also seems unlikely that there would be a SW-OFF for any block overloads (I haven't observed this personally, but I'm guessing an overloaded bus can also inherit that label from the schedule data of the original run).

So our proposed solution is that the ADDED label should override both PULL-B and SW-OFF labels on the route ladder, in VPP, and on the map views. 

Maybe in some future state it might be worthwhile to show the end time of the overload, but that is obviously out of scope here. @arkadyan out of curiosity, do we have access to the block overload start/end time?

arkadyan:
@ahakuta We do not have access to the block overload start/end times.

ingridpierre:
@arkadyan Sort of? It's pulling out, but not from a garage (this person was picking up a vehicle on the street from someone who had already ended their block several hours prior). These screenshots were taken just a couple minutes after they had logged on and before they had started their first trip. 

I think all overloads look like pull-outs because their runs don't connect to any previous work.

I wonder if we should make a separate ticket to say incoming or something when we don't have the current (or previous) route assignment & it's in the incoming box. I felt "Null" was just confusing.

arkadyan:
@ahakuta Updated so that the ADDED label takes precedence over PULL-B or SW-OFF.

PR #582, opened 2020-04-14
filter by block_is_active on the backend

skyqrose:
No Asana task.

The ladder was filtering by `blockIsActive`, but nothing else was. As I move the layover box into the ladder, rather than try to figure out how to make that filtering consistent between both the layover box and onRoute vehicles, I figured that those vehicles shouldn't even be making it to the frontend at all.

(I only removed them when subscribed to a route, not when searching, so they will still show up in search if you're specifically looking for an inactive vehicle.)


arkadyan:
What about for a bus pulling out for the start of the day: would it’s block not yet be active, and this change would mean it no longer shows up in the incoming box when it used to?

skyqrose:
This would newly hide any laying over or incoming vehicles with `block_is_active: false`. But that field is only set to false if the vehicle is already off course, and I don't think we ever have incoming/laying over vehicles that are off course.

Also, only the block end time is checked, not the start. So pulling out vehicles whose block hasn't started yet wouldn't be removed anyway.

So I think it's fine.

arkadyan:
Got it, sound good.

PR #587, opened 2020-04-15
Display a page-specific Appcue when you click help

arkadyan:
Remove About page and move the "?" button to the bottom to be the new
help button.

Asana ticket: [Remove the /about page link, add a '?' icon to the sidebar that launches a unique Appcue flow within each Skate screen](https://app.asana.com/0/1148853526253426/1156635492523553)

Currently on Dev for testing. (The Appcues won't do anything locally.)


assets/css/_tab_bar.scss
```diff
@@ -53,9 +53,14 @@
...
   bottom: 0;
-  padding: 0.5rem;
+  display: flex;
+  flex-direction: column;
   position: absolute;
   width: 100%;
+
+  button {
+    padding: 0.5rem;
+  }
```
skyqrose, line 65: Does this need `flex: 1 1 auto`?

skyqrose:
The codecov failures are because you didn't test the help button on every page. I'm fine with ignoring that.

assets/css/_tab_bar.scss
arkadyan: I don't think so, I'm not seeing any effect from that.

arkadyan:
@skyqrose I tested the bad ID manually—you get an error in the console but there's nothing user-visible.

ahakuta:
@arkadyan, @ingridpierre found an interesting bug on the shuttle map that we're guessing is related to the code in this ticket (although it may not be). The moment the shuttle map is scrolled, the nav bar on the left disappears. This only happens on dev, not on prod. 

Screenshot 2020-04-16 10 21 25

ahakuta:
Thanks for the map fix.

This PR is ready to ship as far as I'm concerned!

PR #590, opened 2020-04-16
Disable map recenter button when auto centering is on

arkadyan:
Asana ticket: [Disable map recenter button if autoCenter is on](https://app.asana.com/0/1148853526253426/1145445573523887)

New version of https://github.com/mbta/skate/pull/585 that was reverted by https://github.com/mbta/skate/pull/589. This version fixes the bug where our tab bar would disappear that was discovered here: https://github.com/mbta/skate/pull/587#issuecomment-614684877.

Background: Leaflet doesn't want its div changing, that caused the bug. I moved the class name that represents the current state of things to an adjacent div.

### DIFF
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -55,6 +55,7 @@
   .m-vehicle-map-state--auto-centering + .m-vehicle-map & a {
     background-color: $color-button-disabled;
     color: $color-font-grey;
+    cursor: default;
 
     path {
       stroke: $color-font-grey;
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -48,7 +48,7 @@
 .m-vehicle-map__recenter-button {
   path {
     fill: none;
-    stroke: $black;
+    stroke: $color-button-secondary;
     stroke-width: 2;
   }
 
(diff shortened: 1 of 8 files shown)


assets/css/_vehicle_map.scss
```diff
@@ -51,4 +51,13 @@
     stroke: $black;
     stroke-width: 1.8;
   }
+
+  .m-vehicle-map-state--auto-centering + .m-vehicle-map & a {
```
skyqrose, line 55: I noticed that the hover styles are still happening. You might need to actively undo them when they button is disabled.

arkadyan: The hover styles seem fixed, but I also changed the cursor to an arrow when the button is disabled.

ingridpierre:
@arkadyan Akira and I noticed a few small visual things things, a couple tweaks to request. Some of these are visual polishes to make it tidier while we're here, and others address Akira's concern that the on state is not emphatic enough.
 
**1. Match disabled grey colors to the zoom control disabled state**
The original ticket specified we use our disabled button colors, but it looks like these greys are subtly different from the greys on the disabled state for the map zoom controls. See the slight difference here:
image

**2. Add Primary Green color to active "on" recenter button state** 
**3. Adjust icon to be more visually centered**
Example : viewBox="-5 -2 36 36"
**4. Increase stroke width slightly to match zoom controls**
use stroke width of 2
image

arkadyan:
@ingridpierre Made the style tweaks. Deploying to dev right now for testing.

PR #598, opened 2020-04-21
Add 3 tabs to the properties panel

arkadyan:
Only when not a shuttle. Shuttles do not have run or block views, so
should not have any tabs.

Asana ticket: [Mini-schedule | Frontend | Add 3 tabs to the properties panel](https://app.asana.com/0/1148853526253426/1153848861993988)

You’ll need to enable the “Mini-schedule” fag in order to see this.

Screen Shot 2020-04-21 at 15 15 42


assets/src/components/propertiesPanel/status_run_block_tabs.tsx
```diff
@@ -0,0 +1,65 @@
...
+      cx="7"
+      cy="7"
+      r="4"
+    />
+  </svg>
+)
+
+const StatusRunBlockTabs = ({ statusContent }: Props) => (
+  <div className="m-tabs">
+    <Tabs className="m-tabs__tabs">
```
skyqrose, line 27: The wrapping div is redundant. This could be a single `<Tabs className="m-tabs">`.

assets/css/properties_panel/_status_run_block_tabs.scss
```diff
@@ -0,0 +1,75 @@
...
+
+.m-tabs__tab {
+  @include button-base;
+  background-color: $color-button-simple;
+  border-left: 1px solid $white;
+  border-radius: 0;
+  color: $color-font-grey;
+  display: inline-block;
+  flex: 1 0 auto;
+  font-weight: 900;
```
skyqrose, line 27: We only load weights `400`, `500`, and `700`, so this is showing up as `700`. I think that's fine. `button-base` sets the weight to `700`, so this line doesn't end up doing anything, and could be removed.

assets/tests/components/__snapshots__/propertiesPanel.test.tsx.snap
```diff
@@ -154,28 +154,150 @@ Array [
...
+          <div
+            aria-labelledby="react-tabs-8"
+            className="m-tabs__tab-panel"
+            id="react-tabs-9"
+            role="tabpanel"
+          />
+          <div
+            aria-labelledby="react-tabs-10"
+            className="m-tabs__tab-panel"
+            id="react-tabs-11"
```
skyqrose, line 297: I'm a little worried about global state making these snapshots fragile. When doing a single `StatusRunBlockTabs`, the 6 components get ids 0-5. But this file has 2 tests, one with ids 0-5 and one with 6-11. Is there risk of these ids changing from a distance and causing intermittent test failures?

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
```diff
@@ -38,7 +39,9 @@ const GhostPropertiesPanel = ({ selectedGhost, route }: Props) => (
       <NoWaiverBanner />
     )}
 
-    <PropertiesList vehicleOrGhost={selectedGhost} />
+    <StatusRunBlockTabs
+      statusContent={<PropertiesList vehicleOrGhost={selectedGhost} />}
```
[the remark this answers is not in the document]
arkadyan: One thing that might help is that I was only imagining passing in content for the status tab. I only did that because this content is different for vehicles and ghosts. But I think the run and block content is all shared, so I was expecting the run and block content to live in / be nested in StatusRunBlockTabs. I think that would make it more straightforward to think about the data lifecycle.

assets/css/properties_panel/_status_run_block_tabs.scss
arkadyan: @ingridpierre Do you want the type weight on the tabs to be 900, or is 700 ok with you?

assets/tests/components/__snapshots__/propertiesPanel.test.tsx.snap
arkadyan: I've run the tests many times and haven't seen a problem at this point.

assets/css/properties_panel/_status_run_block_tabs.scss
ingridpierre: 700 looks alright in your screenshot. I was more concerned about emphasis when I was thinking about it as a brand new feature, less so now.

ingridpierre:
@arkadyan Small tweak. Can we make this vertical divider line the same light grey color as the overall background for vehicle properties panel when it's separating two inactive tabs? (I like the divider line being white when it borders the active tab though, I'm finding it adds subtle dimension to the tab in that situation.)

image

If so let me know so I can update the component styles in the documentation, I like it enough that I'd like to tweak what's there (what I have feels a bit less elegant) to match.

arkadyan:
@ingridpierre Sounds good, made those adjustments to the vertical dividers.

arkadyan:
@skyqrose I think that everything is addressed at this point and ready for another look, please.

PR #604, opened 2020-04-22
Save nonrevenue trips

skyqrose:
Asana Task: [Mini-schedule | Store nonrevenue trips and trip schedule_ids](https://app.asana.com/0/1148853526253426/1171782113097825)

### DIFF
--- a/lib/schedule/gtfs/stop_time.ex
+++ b/lib/schedule/gtfs/stop_time.ex
@@ -31,6 +31,13 @@ defmodule Schedule.Gtfs.StopTime do
 
   @type timepoint_id :: String.t()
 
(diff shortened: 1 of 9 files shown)


lib/schedule/gtfs/stop_time.ex
```diff
@@ -8,6 +8,8 @@ defmodule Schedule.Gtfs.StopTime do
   alias Schedule.Gtfs.Stop
   alias Schedule.Gtfs.Trip
 
+  @type by_id :: %{Trip.id() => [t()]}
```
arkadyan, line 11: What do you think about calling this `by_trip_id`? When I read it I think it's going to be by stop time ID.

PR #610, opened 2020-04-23
Upgrade Elixir and Erlang

arkadyan:
elixir v1.9.4-otp-22
erlang 22.3.2

No Asana ticket. Erlang 22.3.1+ is required for Catalina.


paulswartz:
Any reason you're not updating to Elixir 1.10.2?

arkadyan:
@paulswartz Because I just copied the versions from screens and didn't look any closer? (No there's not). I'll update to that, good call.

PR #620, opened 2020-04-28
bugfix: read hastus activities

skyqrose:
Oops, this is important. Without this, we weren't actually loading any activities.

### DIFF
--- a/lib/schedule.ex
+++ b/lib/schedule.ex
@@ -369,6 +369,7 @@ defmodule Schedule do
     ]
 
     hastus_file_names = [
+      "activities.csv",
       "trips.csv"
     ]
 


skyqrose:
Right now it intentionally treats missing files as empty so that we don't have to specify a bunch of empty files when mocking data for tests. It probably wouldn't be too hard to make sure we only access expected files though.

arkadyan:
@skyqrose Yeah, but at the same time these don't change very often and it's not like it takes us that long to figure out if something's wrong, so maybe it isn't worth even that small effort.

skyqrose:
Codecov is confused. Can you squash it?

PR #621, opened 2020-04-28
Use native releases instead of distillery

arkadyan:
No ticket, preliminary work to make it easier to [Support AWS Secrets Manager](https://app.asana.com/0/1148853526253426/1169250151322401).

I was able to deploy this to dev where it appears to be running properly.


rel/config.exs
```diff
@@ -6,40 +6,3 @@
...
-# For a full list of config options for both releases
-# and environments, visit https://hexdocs.pm/distillery/config/distillery.html
-
-# You may define one or more environments in this file,
-# an environment's settings will override those of a release
-# when building in that environment, this combination of release
-# and environment configuration is called a profile
-
-environment :prod do
-  set(include_erts: true)
```
arkadyan, line 25: @paulswartz I tried moving all these `set()` lines to the `releases` file at first, but it didn't seem happy about that. Screens doesn't seem to have any of these. Is it ok for these to go away, or is there something we need to carry forward in a new way?

arkadyan:
@skyqrose I had a specific question for Paul, but pinging you in case you are interested in looking this over as well.

rel/config.exs
paulswartz: `include_erts: true` is the default, `include_src` isn't a feature of the native releases, and the cookie is fetched from `RUNTIME_COOKIE` by default. So the only thing you need to do here is switch the environment variable with the cooking from ERLANG_COOKIE to RUNTIME_COOKIE.

Full documentation: https://hexdocs.pm/mix/1.9.0/Mix.Tasks.Release.html#module-runtime-configuration

paulswartz: There's also some stuff in `rel/vm.args` that you can either port over or delete the file.

arkadyan:
@paulswartz Looking at the docs, is the variable name actually `RELEASE_COOKIE`?

paulswartz:
> @paulswartz Looking at the docs, is the variable name actually `RELEASE_COOKIE`?

yep!

PR #625, opened 2020-04-30
Fetch minischedules data in frontend 

skyqrose:
Asana Task: [Mini-schedule | Frontend | Fetch data](https://app.asana.com/0/1148853526253426/1170680075334329)

Screen Shot 2020-04-29 at 14 09 38
Screen Shot 2020-04-29 at 14 09 47

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -1,0 +1,21 @@
+import React from "react"
+import {
+  useMinischeduleBlock,
+  useMinischeduleRun,
+} from "../../hooks/useMinischedule"
+import { Block, Run } from "../../minischedule"
+import { TripId } from "../../schedule"
+import Loading from "../loading"
+
+export interface Props {
+  activeTripId: TripId
+}
+export const MinischeduleRun = ({ activeTripId }: Props) => {
+  const run: Run | null = useMinischeduleRun(activeTripId)
+  return run === null ? <Loading /> : <div>{JSON.stringify(run)}</div>
+}
+
+export const MinischeduleBlock = ({ activeTripId }: Props) => {
+  const block: Block | null = useMinischeduleBlock(activeTripId)
+  return block === null ? <Loading /> : <div>{JSON.stringify(block)}</div>
+}
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -11,11 +11,23 @@ export interface Props {
   activeTripId: TripId
 }
 export const MinischeduleRun = ({ activeTripId }: Props) => {
-  const run: Run | null = useMinischeduleRun(activeTripId)
-  return run === null ? <Loading /> : <div>{JSON.stringify(run)}</div>
+  const run: Run | null | undefined = useMinischeduleRun(activeTripId)
+  if (run === undefined) {
+    return <Loading />
+  } else if (run === null) {
+    return <>No run found</>
+  } else {
+    return <div>{JSON.stringify(run)}</div>
+  }
 }
 
 export const MinischeduleBlock = ({ activeTripId }: Props) => {
-  const block: Block | null = useMinischeduleBlock(activeTripId)
-  return block === null ? <Loading /> : <div>{JSON.stringify(block)}</div>
+  const block: Block | null | undefined = useMinischeduleBlock(activeTripId)
+  if (block === undefined) {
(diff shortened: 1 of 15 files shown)


assets/src/models/minischeduleData.ts
```diff
@@ -0,0 +1,97 @@
...
+      : pieceFromData(activityData)
+  ),
+})
+
+export const blockFromData = (blockData: BlockData): Block => ({
+  id: blockData.id,
+  pieces: blockData.pieces.map(pieceFromData),
+})
+
+export const breakFromData = (breakData: BreakData): Break => ({
```
arkadyan, line 70: Not a big deal, but several of these functions are exported but not used outside of this file.

assets/src/components/propertiesPanel/status_run_block_tabs.tsx
```diff
@@ -49,13 +52,13 @@ const StatusRunBlockTabs = ({ statusContent }: Props) => (
       className="m-tabs__tab-panel"
       selectedClassName="m-tabs__tab-panel--selected"
     >
-      <h2>Run Content</h2>
+      {activeTripId && <MinischeduleRun activeTripId={activeTripId} />}
```
arkadyan, line 55: This is likely a question for Ingrid, but if there is no active trip, do we want to display some message, or not show tabs in the first place or something?

assets/src/models/minischeduleData.ts
```diff
@@ -0,0 +1,97 @@
...
+const isBreakData = (
+  activityData: BreakData | PieceData
+): activityData is BreakData => activityData.hasOwnProperty("break_type")
+
+export const runFromData = (runData: RunData): Run => ({
+  id: runData.id,
+  activities: runData.activities.map((activityData) =>
+    isBreakData(activityData)
+      ? breakFromData(activityData)
+      : pieceFromData(activityData)
```
arkadyan, line 61: It looks like this branch isn't being tested yet?

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -0,0 +1,21 @@
...
+} from "../../hooks/useMinischedule"
+import { Block, Run } from "../../minischedule"
+import { TripId } from "../../schedule"
+import Loading from "../loading"
+
+export interface Props {
+  activeTripId: TripId
+}
+export const MinischeduleRun = ({ activeTripId }: Props) => {
+  const run: Run | null = useMinischeduleRun(activeTripId)
```
arkadyan, line 14: Waiting to test these two until we have actual content there seems fine by me.

assets/src/hooks/useMinischedule.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from "react"
+import { fetchMinischeduleBlock, fetchMinischeduleRun } from "../api"
+import { Block, Run } from "../minischedule"
+import { TripId } from "../schedule"
+
+export const useMinischeduleRun = (tripId: TripId): Run | null => {
+  const [run, setRun] = useState<Run | null>(null)
+  useEffect(() => {
+    fetchMinischeduleRun(tripId).then(setRun)
```
arkadyan, line 9: These API calls can return `null` as a legitimate response, right? Do we want to set the state to be `undefined` initially to distinguish between loading and empty results?

assets/src/hooks/useMinischedule.ts
```diff
@@ -0,0 +1,20 @@
...
+  }, [])
```
arkadyan, line 10: Should this `useEffect` (and below) pass `tripId` as a dependency? If the vehicle's trip changes while this panel is open we'd want to load fresh data, right?

skyqrose: Good catch!

assets/src/components/propertiesPanel/status_run_block_tabs.tsx
skyqrose: @ingridpierre ?

assets/src/components/propertiesPanel/minischedule.tsx
skyqrose: I ended up adding tests, since the loading/null/data conditionals are reasonably permanent, even if the content of the snapshot will change soon.

assets/src/models/minischeduleData.ts
skyqrose: `pieceFromData` is tested by the the piece in the block test.

PR #626, opened 2020-04-30
Move the "last GPS ping" status into the header

arkadyan:
Only for mini-schedule version. The stable version remains unchanged.

Asana ticket: [Mini-schedule | Frontend | Move the "last GPS ping" status into the header](https://app.asana.com/0/1148853526253426/1153848861993994)

Without flag:

without flag

With flag:

with flag

### DIFF
--- a/assets/css/properties_panel/_header.scss
+++ b/assets/css/properties_panel/_header.scss
@@ -4,11 +4,6 @@
   .m-route-variant-name {
     @include font-header;
   }
-
-  .m-close-button {
-    display: flex;
-    flex-direction: column;
-  }
 }
 
 .m-properties-panel__label {
@@ -80,7 +75,15 @@
   }
 }
 
-.m-properties-panel__header > button {
-  flex: 0 0 auto;
+.m-properties-panel__close-ping {
+  align-items: flex-end;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
   padding: 1rem;
+}
+
+.m-properties-panel__last-gps-ping {
+  @include font-small;
+  color: $color-font-grey;
 }
--- a/assets/css/properties_panel/_header.scss
+++ b/assets/css/properties_panel/_header.scss
@@ -80,7 +80,11 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  padding: 1rem;
+
+  .m-close-button,
+  .m-properties-panel__last-gps-ping {
+    padding: 1rem;
+  }
 }
 
 .m-properties-panel__last-gps-ping {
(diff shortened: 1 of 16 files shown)


assets/src/components/propertiesPanel/header.tsx
```diff
@@ -90,10 +119,17 @@ const directionName = (
...
 ): string => (route ? route.directionNames[directionId] : "")
 
+const nowInSeconds = (): number => Math.floor(Date.now() / 1000)
+
 const Header = ({ vehicle, route }: Props) => {
   const [{ ladderDirections, settings }, dispatch] = useContext(
     StateDispatchContext
   )
+  const [epocNowInSeconds, setEpocNowInSeconds] = useState(nowInSeconds())
+  useInterval(() => setEpocNowInSeconds(nowInSeconds()), 1000)
```
skyqrose, line 129: This can use `now()` from `dateTime.ts` instead of defining `nowInSeconds`.

assets/css/properties_panel/_header.scss
```diff
@@ -80,7 +75,15 @@
   }
 }
 
-.m-properties-panel__header > button {
-  flex: 0 0 auto;
+.m-properties-panel__close-ping {
```
skyqrose, line 78: Currently, the click target for the x is big, including the padding and taking up most of the right side of the header. I think this wrapper might make the click target much smaller, so it's just the x. If that's the case, is there a way to rearrange things so that the click target stays as large as possible?

assets/src/components/propertiesPanel/header.tsx
arkadyan: Well, we need to compare now with the vehicle timestamp we are getting which is in epoch seconds, so we could start with dates, but we'd have to convert the timestamp instead, and then convert the result into seconds, so this feels a little simpler to me.

assets/css/properties_panel/_header.scss
arkadyan: That's interesting. I can add an `onclick` to the entire div. So you'd be able to click on the entire right side to close panel—does that seem like a good approach?

arkadyan: Made that update.

skyqrose: That'd make the clicking work, but the highlighting+pointer behaviors of the button wouldn't work unless you're actually over the button element.

Could you instead put padding on both the button and the `last-gps-ping` elements separately, instead of the wrapping div?

assets/src/components/propertiesPanel/header.tsx
skyqrose: Ah, right 👍

arkadyan:
Got a 👍 from @ahakuta

PR #631, opened 2020-05-04
Use AWS SecretsManager for secret environment variables

arkadyan:
Asana ticket: [Support AWS Secrets Manager](https://app.asana.com/0/1148853526253426/1169250151322401)

~~N.B.: Requires [IAM roles](https://app.asana.com/0/1113179098808463/1169714476500353/f) before we deploy to production.~~

Currently running on dev.


arkadyan:
@ianwestcott @paulswartz happy to hear any feedback from either of you if you have any.

lib/skate/secrets_manager.ex
```diff
@@ -0,0 +1,32 @@
...
+  def env do
+    System.get_env("ENVIRONMENT_NAME")
+  end
+
+  @spec secret_key_name(String.t()) :: String.t()
+  def secret_key_name(env_name) do
+    env_name
+    |> String.downcase()
+    |> String.replace("_", "-")
+  end
```
skyqrose, line 31: Could we put the corrected name in the config instead of having to process it here? So the config would have `{:secret, "api-key"}` instead of `{:secret, "API_KEY"}`?

lib/skate_web/endpoint.ex
```diff
@@ -46,14 +48,14 @@ defmodule SkateWeb.Endpoint do
 
   # callback for runtime configuration
   def init(:supervisor, config) do
-    secret_key_base = System.get_env("SECRET_KEY_BASE")
-
     config =
-      if secret_key_base do
+      try do
+        secret_key_base = SecretsManager.fetch!("SECRET_KEY_BASE")
```
skyqrose, line 53: Shouldn't this be loaded into the application config by `Application.runtime_config()`? Can we replace this whole `try`/`rescue` and `SecretsManager.fetch` with an `Application.get_env`?

config/releases.exs
```diff
@@ -1 +1,16 @@
...
+Application.ensure_all_started(:ex_aws)
+Application.ensure_all_started(:ex_aws_secretsmanager)
+
+config :skate,
+  secret_key_base: SecretsManager.fetch!("SECRET_KEY_BASE")
+
+config :ueberauth, Ueberauth.Strategy.Cognito,
+  client_secret: SecretsManager.fetch!("COGNITO_CLIENT_SECRET")
+
+config :skate, SkateWeb.AuthManager, secret_key: SecretsManager.fetch!("GUARDIAN_SECRET_KEY")
```
skyqrose, line 16: Why this as opposed to `{:secret, "GUARDIAN_SECRET_KEY"}` and let `Application.runtime_config` take care of it? 

Does this config file get run when devloping locally, and would that be a problem?

If we can load config at runtime in this file, should we replace the whole `runtime_config` system with `System.get_env` calls in this file?

README.md
```diff
@@ -34,10 +34,20 @@ There are a number of configuration details defined in environment variables. Th
...
-- **ERLANG_COOKIE** Used by Erlang (only required in production)
+- **API_URL**: URL of the API for retrieving live train positions
+- **API_KEY**: Access key for the API (for dev only, see below for prod)
+- **ERLANG_COOKIE**: Used by Erlang (only required in production)
 - **COGNITO_DOMAIN**, **COGNITO_CLIENT_ID**, **COGNITO_CLIENT_SECRET**, **COGNITO_USER_POOL_ID**, **COGNITO_AWS_REGION**, and **GUARDIAN_SECRET_KEY**: Authentication/authorization details (only required in production)
 - **STATIC_SCHEME**, **STATIC_HOST**, **STATIC_PATH**, and **STATIC_PORT**: CDN details (only required in production)
+
+Additionally, there a number of secret variables (only required in production) defined in [AWS SecretsManager](https://console.aws.amazon.com/secretsmanager):
+
+- **API_KEY**: Access key for the API
```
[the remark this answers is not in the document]
arkadyan: Yes, that is intentional. We want `API_KEY` to come from SecretsManager in prod, but that doesn't work for dev. So the env var lets you continue setting the value locally for development. I tried to indicate this with the "for dev only" comment above, but that's probably doesn't explain very well. Do you have suggestions for a better way to describe it? Maybe I even just copy the core part of this comment?

lib/skate/secrets_manager.ex
arkadyan: Following up on my comment above, `API_KEY` is actually a good example to show why I kept it this way. The idea was to be consistent between env vars and secret names, which allows you to use different modes per environment (e.g. env vars in dev vs secrets in prod).

README.md
```diff
@@ -34,10 +34,20 @@ There are a number of configuration details defined in environment variables. Th
...
-- **SWIFTLY_REALTIME_VEHICLES_URL** and **SWIFTLY_AUTHORIZATION_KEY**: Source of Swiftly vehicle data
+- **SWIFTLY_REALTIME_VEHICLES_URL**: Source of Swiftly vehicle data
+- **SWIFTLY_AUTHORIZATION_KEY**: for dev only, see below for prod
 - **SKATE_HASTUS_URL**: Source of extended schedule data
 - **TRIP_UPDATES_URL**: Source of GTFS-realtime enhanced TripUpdates json data file (optional)
-- **SECRET_KEY_BASE** Used for writing encrypted cookies. Generate a value using `mix phx.gen.secret` (only required in production)
-- **ERLANG_COOKIE** Used by Erlang (only required in production)
+- **API_URL**: URL of the API for retrieving live train positions
+- **API_KEY**: Access key for the API (for dev only, see below for prod)
+- **ERLANG_COOKIE**: Used by Erlang (only required in production)
```
paulswartz, line 43: I thought this changed to `RELEASE_COOKIE` or similar?

arkadyan: It sure did, good catch.

lib/skate/secrets_manager.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  @spec env_specific_key_name(String.t()) :: String.t()
+  def env_specific_key_name(key_name) do
+    env() <> "-" <> secret_key_name(key_name)
+  end
+
+  @spec env() :: String.t()
+  def env do
+    System.get_env("ENVIRONMENT_NAME")
```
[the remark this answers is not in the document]
arkadyan: It comes from the server config, and good catch, it absolutely should be documented with the others.

config/releases.exs
arkadyan: This config file does not get run when developing locally, and that's the reason for these to be here—they are only needed in prod and not in dev.

lib/skate_web/endpoint.ex
arkadyan: We are intentionally not using `Application.runtime_config()` for this. For dev and test we don't care about the values, so we have them defined in those environment config files so it's one less thing you have to worry about in your `.env` file.

skyqrose: I think you could accomplish that with
`dev.exs`:
```
secret_key_base: "local_secret_key_base_at_least_64_bytes_________________________________"
```
`prod.exs`:
```
secret_key_base: {:secret, "SECRET_KEY_BASE"}
```

config/releases.exs
skyqrose: I think it'd be possible to fetch it from the secrets manager in prod, and not in dev, and have a consistent way of specifying all the config/secrets if it was like this:
```exs
#config.exs
secret_key: nil
```
```exs
#prod.exs
secret_key: {:secret, "GUARDIAN_SECRET_KEY"}
```
```exs
#releases.exs
#no runtime configuration here, it's all prod.exs and Application.runtime_config()
```

arkadyan: Personally, that feels more confusing to me—the key being defined in prod but undefined elsewhere seems less clear.

skyqrose: `config.exs` already has that line.

The goal with this suggestion would be consolidate all the places where prod checks the secrets manager (currently `releases.exs`, `prod.exs/Application.runtime_config`, and `endpoint.ex`) into just one place (`prod.exs/Application.runtime_config`).

Dev configuration wouldn't change.

arkadyan: Oh sorry, I was confused because that setting isn't on `:skate`, it's on `SkateWeb.AuthManager`. The application `load_runtime_config` is just specific to the skate app.

skyqrose: Ah, I didn't realize that. Should we make `load_runtime_config` able to work on all apps? We could write the list of keys as `[{:skate, :api_url}, {SkateWeb.AuthManager, :secret_key}]`

lib/skate_web/endpoint.ex
arkadyan: Updated this.

config/releases.exs
arkadyan: Yeah, I could see that, but I think that's outside the scope of this ticket. This change isn't really making things any more or less consolidated. That said as I think about it the amount of variation is pretty limited and the SecretsManager code is all encapsulated, so I'm not sure how much extra effort on `load_runtime_config` would be worth it…

PR #632, opened 2020-05-04
react-leaflet

skyqrose:
Asana Task: [Dev cleanup: Use react-leaflet package](https://app.asana.com/0/1152340551558956/1153855588392657)

As part of looking into [🐞 Vehicle doesn't stay centered in the properties panel map](https://app.asana.com/0/1148853526253426/1171784929698249) , I thought I'd check whether react-leaflet would handle the state changes better, so I updated and finished the old investigations into it.

It didn't end up fixing the bug, but react-leaflet works now, and it's so much nicer than writing custom state updating.


assets/src/components/map.tsx
```diff
@@ -330,156 +184,56 @@ export const autoCenter = (
...
+  turnOnAutoCenter: () => void
+}) => (
+  <Control position="topright">
+    <div className="leaflet-bar m-vehicle-map__recenter-button">
+      <a
+        href="#"
+        title="Recenter map"
+        role="button"
+        aria-label="Recenter map"
+        onClick={(e) => {
```
arkadyan, line 199: Codecov is dinging this. I don't think it's the most crucial thing, but would it be relatively straightforward that clicking this does something, like e.g. the `autoCenteringClass` changes (even though that's kind of a second-order effect).

assets/tests/components/map.test.tsx
```diff
@@ -78,272 +69,55 @@ const vehicle: Vehicle = {
...
 }
-const trainVehicle: TrainVehicle = {
-  id: "red1",
-  latitude: 42.24615,
-  longitude: -71.00369,
-  bearing: 15,
-}
 
 describe("map", () => {
-  test("renders", () => {
```
arkadyan, line 89: Is there a reason not to keep a basic render test?

skyqrose: Yeah. We need the test file to use the real Leaflet implementation for other tests,`react-test-renderer` can't handle the real implementation and needs a mock, and I couldn't figure out how to unmock Leaflet for only part of the test file, it had to be all or nothing.

If we wanted to keep the snapshots, one thing we could try is having a separate test file that only does the snapshots, so that file could use the mocked Leaflet while this file doesn't. But the snapshots were so shallow, and the `draws x` tests do render everything, so I figured it wasn't worth it.

assets/__mocks__/leaflet.js
```diff
@@ -0,0 +1,166 @@
+// from https://github.com/PaulLeCam/react-leaflet/blob/master/__mocks__/leaflet.js
```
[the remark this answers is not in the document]
skyqrose: This is the implementation of leaflet that all tests except `map.test.tsx` use. I think it would be better to have them in the tests directory, but unfortunately, that's not where they're supposed to go. https://jestjs.io/docs/en/manual-mocks

assets/src/components/map.tsx
```diff
@@ -330,156 +184,56 @@ export const autoCenter = (
...
-  })
+  const [appState] = useContext(StateDispatchContext)
+  const mapRef: MutableRefObject<ReactLeafletMap | null> =
+    // this prop is only for tests, and is consistent between renders, so the hook call is consistent
+    // tslint:disable-next-line: react-hooks-nesting
+    props.reactLeafletRef || useRef(null)
   const [shouldAutoCenter, setShouldAutoCenter] = useState<boolean>(true)
   const isAutoCentering: MutableRefObject<boolean> = useRef(false)
 
   useEffect(() => {
```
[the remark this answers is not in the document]
skyqrose: Hmm... It'd cause some merge conflicts with #634. I'll wait until after that PR and then see if it would help.

assets/tests/components/map.test.tsx
arkadyan: Yeah, makes sense.

assets/src/components/map.tsx
skyqrose: 👍 I just pushed a test that checks the class and the map center.

PR #634, opened 2020-05-05
Bugfix: Autocentering loses focus

skyqrose:
Asana Task: [🐞 Vehicle doesn't stay centered in the properties panel map](https://app.asana.com/0/1148853526253426/1171784929698249)

Built on top of #632 

This was the result of two overlapping problems:

1. `useEffect` dependencies use a shallow `===` comparison, and `[] !== []`, so it thought that the list of vehicles was moving every render, and it called `autoCenter()` every render.

That should be fine, autoCentering should be idempotent, but

2. If `autoCenter()` is called twice in quick succession, there's a chance that events will happen in this order:
- 1st `autoCenter` starts, setting `isAutoCentering` to `true`
- 1st `movestart` triggers. `isAutoCentering` is `true`.
- 2nd `autoCenter` call happens. `isAutoCentering` is set to `true` again.
- 1st `moveend` triggers. `isAutoCentering` is set to `false`.
- 2nd `movestart` triggers. `isAutoCentering` is `false`, so it thinks it's a manual move, and turns `shouldAutoCenter` off.

This PR fixes the 1st problem by changing the list of vehicles to a form that `useEffect` can compare, so `autoCenter` is only called if the vehicles move. (Once every few seconds, instead of several times per second.)

Fixing the 2nd bug would require bigger changes to how we tell the difference between autocentering and manual moves. I guess it's possible that if there were two real moves in quick succession it could still happen, but in practice, updates are far enough apart that one move can finish before the next starts. I left the VPP open for 15 minutes and it kept focus on the vehicle the whole time.

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -238,7 +238,11 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
         appState.pickerContainerIsVisible
       )
     }
-  }, [shouldAutoCenter, latLngs, appState.pickerContainerIsVisible])
+  }, [
+    shouldAutoCenter,
+    JSON.stringify(latLngs),
+    appState.pickerContainerIsVisible,
+  ])
 
   const autoCenteringClass = shouldAutoCenter
     ? "m-vehicle-map-state--auto-centering"
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -24,7 +24,7 @@ import { drawnStatus, statusClass } from "../models/vehicleStatus"
 import { TrainVehicle, Vehicle, VehicleId } from "../realtime.d"
 import { Shape } from "../schedule"
 import { Settings } from "../settings"
-import { selectVehicle, State as AppState } from "../state"
+import { selectVehicle } from "../state"
 
 interface Props {
   vehicles: Vehicle[]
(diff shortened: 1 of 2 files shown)


ahakuta:
Took a look at this on my local machine, love this bug fix. Thanks Sky!

assets/src/components/map.tsx
```diff
@@ -164,13 +164,10 @@ const Shape = ({ shape }: { shape: Shape }) => {
 
 export const autoCenter = (
   map: LeafletMap,
-  vehicles: Vehicle[],
+  latLngs: LatLngExpression[],
```
arkadyan, line 167: Maybe `vehicleLatLngs` would be a little more descriptive?

assets/src/components/map.tsx
```diff
@@ -238,7 +238,11 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
         appState.pickerContainerIsVisible
       )
     }
-  }, [shouldAutoCenter, latLngs, appState.pickerContainerIsVisible])
+  }, [
+    shouldAutoCenter,
+    JSON.stringify(latLngs),
```
arkadyan, line 243: So how does this end up fixing the bug? Is it possible to write a test case to prevent a regression? Or maybe just a description here if not—I can imagine this line looking strange/unnecessary out of context.

assets/src/components/map.tsx
skyqrose: At this point, you don't really care what the latlngs are of. Maybe someday we'd decide that any stops should be in view, too, and you could passing `autoCenter(vehicleLatLngs.concat(stopLatLngs))`. This function wouldn't change.

arkadyan: Ok, sure makes sense.

assets/src/components/map.tsx
skyqrose: Yeah, it is pretty strange. Added a comment.

PR #639, opened 2020-05-06
Add an architecture document

arkadyan:
Asana ticket: [Add an ARCHITECTURE.md file describing the app's architecture](https://app.asana.com/0/1152340551558956/1130855786258908/f)


ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    - Driver information
+    - TransitMaster route assignments
+  - Trip updates
+    - Block waivers
+- Swiftly: GTFS-rt data
+  - Vehicle updates
+    - Inferred location information (what stop is the bus approaching)
+    - Schedule adherence (On-Time Performance)
+    - Headway
+    - "Predicted" route assignments based on location
```
skyqrose, line 33: I don't think we do anything with predictions from Swiftly. We get trip assignments, but just for the current time. (It may be worth noting that swiftly's trip assignments are generally better than TransitMaster's).

arkadyan:
@skyqrose Thanks, updated the diagram.

PR #644, opened 2020-05-07
Minischedules trip list

skyqrose:
Asana Task: [Mini-schedule | Trip List Component | Base component](https://app.asana.com/0/1148853526253426/1153870789408601)

Screen Shot 2020-05-07 at 11 38 27
Screen Shot 2020-05-07 at 11 38 54

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -55,7 +55,7 @@ export const MinischeduleBlock = ({ activeTripId }: Props): ReactElement => {
 
 const Header = ({ label, value }: { label: string; value: string }) => (
   <div className="m-minischedule__header">
-    <span className="m-minischedule__label">{label}</span>
+    <span className="m-minischedule__header-label">{label}</span>
     {value}
   </div>
 )
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -1,33 +1,68 @@
-import React from "react"
+import React, { ReactElement } from "react"
 import {
   useMinischeduleBlock,
   useMinischeduleRun,
 } from "../../hooks/useMinischedule"
-import { Block, Run } from "../../minischedule"
+import { Block, Break, Piece, Run, Trip } from "../../minischedule"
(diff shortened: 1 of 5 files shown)


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -1,33 +1,68 @@
...
+      <>
+        {block.pieces.map((piece) => (
+          <Piece piece={piece} key={piece.start.time} />
+        ))}
+      </>
+    )
   }
 }
+
+const Break = ({ break: breakk }: { break: Break }) => (
```
arkadyan, line 63: I wouldn't be surprised if these break out into their own files as they get fleshed out, but this is fine for now. (Though as I think about it you might want to think about whether breaking them out sooner than later will help avoid conflicts as you start working on things in parallel…)

arkadyan, line 63: Typo: `breakk`

assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -36,7 +59,8 @@ describe("MinischeduleRun", () => {
 
   test("renders a run", () => {
     ;(useMinischeduleRun as jest.Mock).mockImplementationOnce(() => ({
-      run: "run",
+      id: "run",
+      activities: [breakk, piece],
```
arkadyan, line 63: Typo: the "k" key is being extra eager again.

assets/src/components/propertiesPanel/minischedule.tsx
skyqrose: That's actually intentional. `break` is a keyword in js, so you can't name a variable that. Since "Break" is such a specific concept, I figured modifying the name was better than trying to find a synonym, but if you have other ideas on how to handle this, I'm open to it.

arkadyan: Oh, ha! Makes sense.

skyqrose: Yeah. They might eventually want their own unit/snapshot tests too if we start doing interesting logic on them, but not yet.

PR #646, opened 2020-05-11
Add start_time and end_time to minischedules trip

phildarnowsky:
Minischedules trips now contain a start_time and an end_time. If a GTFS
trip is present the minischedule uses the times of the first and last
stops. Otherwise the start_time and end_time from the HASTUS trip are
used.

### DIFF
--- a/lib/schedule/trip.ex
+++ b/lib/schedule/trip.ex
@@ -21,7 +21,9 @@ defmodule Schedule.Trip do
           shape_id: Shape.id() | nil,
           schedule_id: Hastus.Schedule.id() | nil,
           run_id: Run.id() | nil,
-          stop_times: [StopTime.t()]
+          stop_times: [StopTime.t()],
+          start_time: Util.Time.time_of_day(),
+          end_time: Util.Time.time_of_day()
         }
 
   @enforce_keys [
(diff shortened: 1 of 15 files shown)


lib/schedule/trip.ex
```diff
@@ -21,7 +21,9 @@ defmodule Schedule.Trip do
           shape_id: Shape.id() | nil,
           schedule_id: Hastus.Schedule.id() | nil,
           run_id: Run.id() | nil,
-          stop_times: [StopTime.t()]
+          stop_times: [StopTime.t()],
+          start_time: Util.Time.time_of_day() | nil,
+          end_time: Util.Time.time_of_day() | nil
```
skyqrose, line 26: Since we always have either GTFS or HASTUS, these won't be nil.

skyqrose:
Looks like you need to run `npm run format`.

The test failure is an unrelated intermittent failure that was recently fixed, and if you rebase onto master you'll get the fix and it shouldn't fail again.

phildarnowsky:
@skyqrose this version should address all your feedback. Where commits are tagged `WIP`, that indicates that I mean to squash them down into the preceding "regular" commit before merging.

And speaking of merging, I see that you use a squash-merge strategy that ends with one commit per PR. Are you married to that? I've found it to be advantageous to use a strategy with multiple commits per PR, with commits organized and neatened so that each one is a single self-contained change that can be sensibly considered in isolation. This makes review easier, and leaving the commits unsquashed means you get more fine-grained results when you find bugs with `git bisect`.

skyqrose:
I agree that organized commits are really valuable during the PR, but we've never had to use git bisect in a way where it'd be useful to have them broken apart after merging to master. I like to polish and organize the squash commit message so the features are clear in case we ever need to read that, but if I ever need to see them broken apart by commit, I usually go back to the github PR (where I can see any PR comments and context, too).

So no, not married, but happily dating. I could be convinced to change.

PR #653, opened 2020-05-12
Minischedule.Trip type with via_variant

skyqrose:
A step towards [Mini-schedule | Style Revenue Trips](https://app.asana.com/0/1148853526253426/1170698291365302)

The designs for that task call for showing the via variant, but the via variant isn't in the trip type we sent to the frontend.

This makes a new type on the backend, `Minischedule.Trip`, that includes the via variant field (and omits some other fields we don't use). The conversion from `Schedule.Trip` to `Minischedule.Trip` happens at the same time as we hydrate the trip ids into trip objects.


lib/schedule/minischedule/piece.ex
```diff
@@ -43,15 +43,14 @@ defmodule Schedule.Minischedule.Piece do
     :end
   ]
 
-  @spec hydrate(t(), Trip.by_id()) :: t()
+  @spec hydrate(t(), Schedule.Trip.by_id()) :: t()
```
phildarnowsky, line 46: @skyqrose "hydrate" is a term I'm hearing a lot these days but I'm not clear on what it means.

skyqrose: When we store the minischedules, we store them "dehydrated" with the trip ids. Then, when reading them out of storage, we "hydrate" them by replacing the ids with the trip data. Analogous to dehydrating food for storage.

There are 2 reasons we do this:
1. Because trips, with all their stop times, are a huge stress on our memory. We already have to have ridiculously large AWS servers to hold them. Storing the minischedules data dehydrated avoids making that problem worse.
2. When we store them with just the id, we work with just HASTUS trips. When we hydrate them, we do that with merged HASTUS+GTFS trips. This means we can get data from GTFS (like `headsign`) without having to think about GTFS while parsing+loading minischedules.

PR #661, opened 2020-05-18
Use proper icons in minischedule

phildarnowsky:
We've been using the same question-mark icon for all rows of the
minischedule. This adds proper icons for the following:

* swing on/off
* deadhead
* pull out/back
* trip (both directions)

In addition, we refactor existing icon helper tests and give icon
helpers consistent types.

Card: https://app.asana.com/0/1148853526253426/1175607249486042


assets/tests/helpers/icon.test.tsx
```diff
@@ -11,536 +17,82 @@ import {
...
+  plusIcon,
+  questionMarkIcon,
+  redLineIcon,
+  reverseIcon,
+  searchIcon,
+  triangleDownIcon,
+  triangleUpIcon,
+}
+
+for (const key in testMap) {
```
skyqrose, line 60: nice!

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -126,28 +142,51 @@ const DeadheadTrip = ({
...
+  const directionIcon =
+    trip.directionId !== null
+      ? // Safe to assume routeId is not null, since if it were, we'd be
+        // rendering a deadhead row instead.
+        iconForDirectionOnLadder(
+          trip.directionId,
+          ladderDirections,
+          trip.routeId!
+        )
+      : questionMarkIcon()
```
skyqrose, line 185: If `iconForDirectionOnLadder` takes `DirectionId | null`, then all the decision about which icon to show can happen in the one function, instead of being split between that function and this `?:`.

skyqrose:
Oh right, codecov. Looks like you'll need to add a test that uses
`{...revenueTrip, directionId: 0, id: "otherDirection"}`
and
`{...revenueTrip, directionId: null, headsign: null, viaVariant: null, id: "withoutGtfsData"}`
trips to one of the tests.

PR #662, opened 2020-05-18
Show breaks in minischedule

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1170698291365291

Breaks are now shown in the minischedule in a formatted fashion, rather than a blob of JSON.


phildarnowsky:
@skyqrose implemented your suggestion, CI seems extremely slow right now.

PR #667, opened 2020-05-19
Show cause_description for block waivers

skyqrose:
Asana Task: [Pull new TM Block Waiver fields into Skate](https://app.asana.com/0/1148853526253426/1173835630372489)

Screen Shot 2020-05-19 at 14 22 53
Screen Shot 2020-05-19 at 14 23 08

### DIFF
--- a/lib/realtime/block_waiver.ex
+++ b/lib/realtime/block_waiver.ex
@@ -11,13 +11,16 @@ defmodule Realtime.BlockWaiver do
   @type t :: %__MODULE__{
           start_time: Util.Time.time_of_day(),
           end_time: Util.Time.time_of_day(),
-          remark: String.t()
+          cause_id: integer(),
+          cause_description: String.t(),
+          remark: String.t() | nil
         }
 
   @enforce_keys [
     :start_time,
     :end_time,
-    :remark
+    :cause_id,
+    :cause_description
   ]
 
   @derive Jason.Encoder
@@ -25,6 +28,8 @@ defmodule Realtime.BlockWaiver do
   defstruct [
     :start_time,
     :end_time,
+    :cause_id,
+    :cause_description,
     :remark
   ]
 
@@ -48,8 +53,10 @@ defmodule Realtime.BlockWaiver do
       {stop_time.time,
        Enum.find(
          stop_time_updates,
-         &(StopTimeUpdate.stop_id(&1) == stop_time.stop_id && StopTimeUpdate.remark(&1) != nil &&
-             StopTimeUpdate.remark(&1) != "")
+         fn stu ->
+           StopTimeUpdate.stop_id(stu) == stop_time.stop_id &&
+             StopTimeUpdate.cause_id(stu) != nil
+         end
        )}
     end)
   end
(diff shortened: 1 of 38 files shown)


lib/realtime/block_waiver.ex
```diff
@@ -94,22 +101,25 @@ defmodule Realtime.BlockWaiver do
   defp bank_working([], working), do: [working]
   defp bank_working(acc, working), do: acc ++ [working]
 
-  @spec same_remark?(map(), map()) :: boolean
-  defp same_remark?(%StopTimeUpdate{remark: remark1}, %StopTimeUpdate{remark: remark2}),
-    do: remark1 == remark2
+  @spec same_cause?(StopTimeUpdate.t(), StopTimeUpdate.t()) :: boolean()
```
skyqrose, line 104: Dialyzer is failing here because it expects this to be `same_cause?(atom() | STU.t(), atom() | STU.t())  :: boolean()`. Any idea why it'd expect that?

phildarnowsky: I don't, but I don't know much about Dialyzer's type inference algorithm and wasn't able to readily find out.

lib/realtime/block_waiver.ex
```diff
@@ -94,22 +101,25 @@ defmodule Realtime.BlockWaiver do
   defp bank_working([], working), do: [working]
   defp bank_working(acc, working), do: acc ++ [working]
 
-  @spec same_remark?(map(), map()) :: boolean
-  defp same_remark?(%StopTimeUpdate{remark: remark1}, %StopTimeUpdate{remark: remark2}),
-    do: remark1 == remark2
+  @spec same_cause?(StopTimeUpdate.t(), StopTimeUpdate.t()) :: boolean
+  defp same_cause?(stu1, stu2) do
+    stu1.cause_id == stu2.cause_id and stu1.remark == stu2.remark
```
phildarnowsky, line 106: @skyqrose is `remark` a free-text field here, since if so checking for equality can be problematic

lib/realtime/block_waiver.ex
skyqrose: It's because concentrates `defstruct_accessors` macro expects you get the fields with accessor functions instead of accessing the fields directly.

Discussion: https://mbta.slack.com/archives/C10EPAU6Q/p1590005263127900

lib/realtime/block_waiver.ex
skyqrose: Yes, it is free text. How can it be problematic?

phildarnowsky: People, even if they're trying to be, are not 100% consistent about how they enter data, with regards to capitalization, whitespace, punctuation, etc. So you have issues where things that "should" be identical like `"Foo Bar"` and `"Foo bar"` are not so.

But actually, I think I heard something about this remark value ultimately originating in a dropdown so perhaps it's OK after all?

skyqrose: cause_id/_description are from the dropdown. `remark` is from the accompanying free text.

It's entered once, and then duplicated by TransitMaster software across all the stops, so I don't think small human errors would really be a problem with this equality check. If it was entered by a human twice with a small difference (maybe for two halves of the run), I'd be okay with it being grouped into two separate block waivers.

PR #668, opened 2020-05-19
Layovers in minischedules

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1153870789408619

We now calculate and show layovers between trips in the minischedule.

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -84,12 +84,16 @@ const Break = ({ break: breakk }: { break: Break }) => {
   )
 }
 
-const Layover = ({ trips, index }: { trips: Trip[]; index: number }) => {
-  const nextTrip = trips[index + 1]
+const Layover = ({
+  currentTrip,
+  nextTrip,
+}: {
+  currentTrip: Trip
+  nextTrip?: Trip
+}) => {
   if (!nextTrip) {
     return null
   }
-  const currentTrip = trips[index]
   const layoverDuration = nextTrip.startTime - currentTrip.endTime
   if (layoverDuration === 0) {
     return null
@@ -126,8 +130,8 @@ const Piece = ({ piece, view }: { piece: Piece; view: "run" | "block" }) => (
             <Trip trip={trip} sequence={sequence} />
             {view === "run" ? (
               <Layover
-                trips={piece.trips}
-                index={index}
+                currentTrip={trip}
+                nextTrip={piece.trips[index + 1]}
                 key={`layover-${trip.endTime}`}
               />
             ) : null}
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -84,6 +84,20 @@ const Break = ({ break: breakk }: { break: Break }) => {
   )
 }
 
+const Layover = ({ trips, index }: { trips: Trip[]; index: number }) => {
+  const nextTrip = trips[index + 1]
+  if (!nextTrip) {
+    return null
+  }
+  const currentTrip = trips[index]
+  const layoverDuration = nextTrip.startTime - currentTrip.endTime
+  if (layoverDuration === 0) {
+    return null
+  }
+
+  return <Row text="Layover" rightText={formattedDuration(layoverDuration)} />
+}
+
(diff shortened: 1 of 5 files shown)


phildarnowsky:
@skyqrose interesting point, I will meditate on this a little and see what implementation seems best.

phildarnowsky:
@skyqrose so what I'm thinking, is first to leave the `<Layover>` as a separate, sibling element to `<Trip>`. As you point out here it's probably the easiest implementation for a human to follow. When it comes time to do the highlighting on an active trip's layover, we're fine as-is, because when we create the `<Layover>` here we know what the current trip is, hence can determine if it's active or not.

As for the special styling, I think my inclination is to add the ability to customize the classes on a `<Row>` when you instantiate it, to do so for the `<Row>`s that correspond to layovers, and then work out the actual styling difference in CSS.

phildarnowsky:
@skyqrose this includes CSS fixes, I see I still have to cover one code path to satisfy Codecov.

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -205,12 +240,14 @@ const Row = ({
...
   rightText,
+  extraClasses,
 }: {
   icon?: ReactElement
   text: string | ReactElement
   rightText?: string
+  extraClasses?: string
 }) => (
-  <div className="m-minischedule__row">
+  <div className={`m-minischedule__row ${extraClasses ? extraClasses : ""}`}>
```
skyqrose, line 250: we have a helper `className()` in `dom.ts` that will avoid adding the trailing whitespace. If it makes the call simpler, you could change it to take `(string | undefined)[]` instead of `string[]`.

phildarnowsky:
@skyqrose ...and that's codecov, ready for another look

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -107,7 +131,18 @@ const Piece = ({ piece, view }: { piece: Piece; view: "run" | "block" }) => (
...
+          <React.Fragment key={trip.id}>
+            <Trip trip={trip} sequence={sequence} />
+            {view === "run" ? (
+              <Layover
+                currentTrip={trip}
+                nextTrip={piece.trips[index + 1]}
+                key={`layover-${trip.endTime}`}
+              />
+            ) : null}
+          </React.Fragment>
```
skyqrose, line 144: Laying over vehicles are assigned to their upcoming trip in the realtime data. I think that the active trips ticket might be easier if the layover component here was drawn before the trip with the same id, instead of after the previous trip, so the two components in the fragment are checking for the same `trip.id`, instead of one checking for `trip.id` and one for `piece.trips[index+1]?.id`.

```jsx
<>
  <Layover
    previousTrip={piece.trips[index-1]}
    nextTrip={trip}
  />
  <Trip/>
</>
```

(But you're the one doing that work so you know better than me if this would help.)

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -107,7 +131,18 @@ const Piece = ({ piece, view }: { piece: Piece; view: "run" | "block" }) => (
...
             : "middle"
-        return <Trip trip={trip} sequence={sequence} key={trip.id} />
+        return (
+          <React.Fragment key={trip.id}>
+            <Trip trip={trip} sequence={sequence} />
+            {view === "run" ? (
+              <Layover
+                currentTrip={trip}
+                nextTrip={piece.trips[index + 1]}
+                key={`layover-${trip.endTime}`}
```
skyqrose, line 141: With the key on the parent fragment, this key isn't necessary.

assets/src/components/propertiesPanel/minischedule.tsx
phildarnowsky: @skyqrose hmm yes I think that might be a bit more elegant, I'll try it

phildarnowsky: @skyqrose actually I'm looking at the active-trip-row card and it says:

> If vehicle is laying over, highlight the smaller "Layover [duration in min]" row *below* the completed trip as well.

[emphasis mine], so I think the current implementation is preferable after all.

phildarnowsky:
@skyqrose this should address all your feedback

PR #672, opened 2020-05-21
rad/wad in minischedules

skyqrose:
Asana Task: [Mini-schedule | RADs/WADs](https://app.asana.com/0/1148853526253426/1173661402437253)

Screen Shot 2020-05-21 at 17 17 48

It shows as "Work as directed" if the data has "wad".

Will require running `mix cache.clean` locally to see it, but they're pretty rare.

Examples of what rads/wads appear like in the HASTUS data are in the asana tasks. Looking at that will help understand the parsing step.

### DIFF
--- a/lib/schedule/minischedule/load.ex
+++ b/lib/schedule/minischedule/load.ex
@@ -43,7 +43,7 @@ defmodule Schedule.Minischedule.Load do
 
     activities =
       activities
-      |> Enum.map(fn activity -> operator_to_piece(activity, trips) end)
+      |> operator_activities_to_pieces(trips)
       |> add_deadheads_to_pieces()
       |> add_sign_ons_to_pieces()
       |> activities_to_breaks()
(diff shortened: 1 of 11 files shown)


lib/schedule/minischedule/load.ex
```diff
@@ -95,17 +121,93 @@ defmodule Schedule.Minischedule.Load do
...
   @spec trip_in_operator(Activity.t(), Trip.t()) :: boolean()
   defp trip_in_operator(%Activity{activity_type: "Operator"} = activity, trip) do
     trip.start_time >= activity.start_time and
       trip.end_time <= activity.end_time and
       String.contains?(trip.block_id, activity.partial_block_id)
   end
 
+  @spec operator_is_as_directed?(Activity.t()) :: boolean()
+  defp operator_is_as_directed?(%Activity{activity_type: "Operator"} = activity) do
+    String.contains?(activity.partial_block_id, "ad")
```
phildarnowsky, line 133: @skyqrose this makes me a little leery but I'm assuming that if you did it this way there's probably not a better way to distinguish as-directed activities.

skyqrose: Here's an activity for scheduled trips:
```
aba20011;123;    1504;05:39;09:37;albny;kenbs;Operator; 60 - 52
```
and here's one for wad
```
abc20011;123;    1151;20:30;23:30;cabot;cabot;Operator;wad -342
```
so no, not really, that's the only difference.

PR #682, opened 2020-05-28
show whether breaks are paid or not

skyqrose:
Asana Task: [Mini-schedule | Show if breaks are paid](https://app.asana.com/0/1148853526253426/1170732391370740)

Screen Shot 2020-05-27 at 17 01 54
Screen Shot 2020-05-27 at 16 57 49

### DIFF
--- a/assets/tests/components/propertiesPanel/minischedule.test.tsx
+++ b/assets/tests/components/propertiesPanel/minischedule.test.tsx
@@ -1,6 +1,8 @@
+import { mount } from "enzyme"
 import React from "react"
 import renderer from "react-test-renderer"
 import {
+  BreakRow,
   MinischeduleBlock,
   MinischeduleRun,
 } from "../../../src/components/propertiesPanel/minischedule"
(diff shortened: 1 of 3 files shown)


ingridpierre:
@skyqrose this lgtm! 👍

skyqrose:
@phildarnowsky codecov is failing because I'm not testing every activity_type in the switch (but I am testing every unique path, I'm just missing cases that fall through into tested cases). Are you okay if I override codecov and merge?

phildarnowsky:
@skyqrose my preference, if it's practical, rather than override codecov, would be to programatically generate a number of similar tests, like I did a couple weeks ago when cleaning up those icon tests. let me look at the diff again and see if I can make a concrete suggestion in this line.

assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -404,3 +405,51 @@ describe("MinischeduleBlock", () => {
...
+      breakType: "Split break",
+      startTime: 0,
+      endTime: 0,
+    }
+
+    const wrapper = mount(<BreakRow break={breakk} />)
+    expect(wrapper.html()).toContain("Break (Unpaid)")
+  })
+
+  test("Paid breaks show as paid", () => {
```
phildarnowsky, line 421: @skyqrose would this satisfy codecov, if we did this instead here:

```
constPaidBreakTypes = ["Paid meal before", "Paid meal after", "Joinup", "Technical break"]

for(breakType in paidBreakTypes) {
  test(`Breaks of type ${breakType} are paid`, () => {
    const breakk: Break = {
      breakType: breakType,
      startTime: 0,
      endTime: 0
    }

    const wrapper = mount(<BreakRow break={breakk} />)
    expect(wrapper.html()).toContain("Break (Paid)")
  }
}
```

skyqrose: When the source needs a hardcoded list or constant, I like to avoid duplicating it in the tests. But I think I found a way to abuse fall through order so that codecov doesn't care about getting all of the strings as long as each unique path is covered.

PR #686, opened 2020-06-03
 style past and future trips and pieces 

skyqrose:
Asana Task: [Mini-schedule | Trip List Component | Past Trip or Event row styling](https://app.asana.com/0/1148853526253426/1153870789408625)

Screen Shot 2020-06-02 at 14 41 38
Screen Shot 2020-06-02 at 14 42 03
Screen Shot 2020-06-02 at 14 42 10

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -22,7 +22,7 @@ import {
   LadderDirections,
   VehicleDirection,
 } from "../../models/ladderDirection"
-import { drawnStatus } from "../../models/vehicleStatus"
+import { drawnStatus, DrawnStatus } from "../../models/vehicleStatus"
 import { VehicleOrGhost } from "../../realtime"
 import { DirectionId, RouteId, TripId } from "../../schedule"
 import { formattedDuration, formattedScheduledTime } from "../../util/dateTime"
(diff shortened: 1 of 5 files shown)


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -189,31 +188,63 @@ const Piece = ({
...
-                  nextTrip={piece.trips[index + 1]}
-                  vehicleOrGhost={vehicleOrGhost}
+                <AsDirected
+                  asDirected={trip}
+                  extraClasses={[
+                    `m-minischedule__row--${onRouteTimeBasedStyle}`,
+                  ]}
                 />
-              ) : null}
+              )}
```
skyqrose, line 247: This function is getting really big, but to pull out the function in the map would require a function with like 7 arguments. Any idea for a good way to break it into smaller pieces?

phildarnowsky: Maybe a function with 7 arguments is the lesser evil here, this really is getting big.

skyqrose: Got it down to 6 arguments. It's not perfect, but it is a little easier to see the big picture now.

ingridpierre:
Non-revenue events like starts/breaks/done should all also appear in grey text like past trips. None of the text in the rows above the active trips should be black.

skyqrose:
Okay, I've made the past breaks and start/done rows gray.

@phildarnowsky there's some significant code changes, so I'd like another review

Screen Shot 2020-06-11 at 14 05 47
Screen Shot 2020-06-11 at 14 06 03

ingridpierre:
@skyqrose changes look great! 👍

PR #689, opened 2020-06-05
Polish active trip

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1177853123505321

A few presentational polishes to the active-trip row:

* Background color should be lighter. 
* Text should be bold.
* Icon should get fill and stroke set
* Dark border on the left


ingridpierre:
Currently, the trip arrows in the active row are using a light background, but they should use the regular fully saturated adherence color:

**CURRENT:**
image
**DESIRED:**
image

Relative size of active trip arrow should be larger than the other rows. For example:
image

phildarnowsky:
@ingridpierre ready for another look

ingridpierre:
@phildarnowsky lgtm! 👍

PR #692, opened 2020-06-08
Convert leading/following deadhead activities

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1170698291365294

Lynn has some "Deadhead to" (pull back) and "Deadhead from" (pull out)
activities. Everywhere else, deadheads are represented as a trip with no
route_id. Here we convert those deadheads into dummy trips that get
tacked onto the correct end of the adjacent piece.

### DIFF
--- a/lib/schedule/minischedule/trip.ex
+++ b/lib/schedule/minischedule/trip.ex
@@ -2,7 +2,6 @@ defmodule Schedule.Minischedule.Trip do
   alias Schedule.Block
   alias Schedule.Gtfs.{Direction, Route, RoutePattern}
   alias Schedule.Hastus.{Activity, Run}
-  alias Schedule.Minischedule.Piece
 
   @type id :: Schedule.Trip.id()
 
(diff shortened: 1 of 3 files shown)


lib/schedule/minischedule/trip.ex
```diff
@@ -50,4 +51,28 @@ defmodule Schedule.Minischedule.Trip do
       end_time: trip.end_time
     }
   end
+
+  @spec from_following_deadhead(Activity.t(), Piece.t()) :: t()
+  def from_following_deadhead(deadhead, piece) do
+    %__MODULE__{
+      id: "following_deadhead_#{piece.block_id}_#{deadhead.start_time}",
+      block_id: piece.block_id,
```
skyqrose, line 59: There's a small type error here, in that the piece's `block_id` can potentially be nil if there are no trips in that piece. (The piece doesn't have its own block_id, and only figures out its block_id from its trips' block_id.)

Trips, meanwhile expect a non-nil block_id.

Some (non necessarily mutually exclusive) ways to handle this:
1. Leave it, it's a small thing and it's fine. I don't think we'd realistically ever have to make a fake trip on a piece that doesn't have any real trips.
2. Make trip allow a nil block_id (could be a lot of work maybe?)
3. Don't make a dummy trip if we can't determine a non-nil block_id for it.
4. In the trip_id, use the run_id instead of the block_id in the trip id to prevent duplicate ids between two trips with unknown block_ids.

(I'd lean towards just 4. If you're feeling particularly strict with the types, then 3 could be simple enough to do.)

phildarnowsky: #4 it is, this is not worth overthinking.

PR #696, opened 2020-06-10
Show duty details of run in minischedules

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1153848861994002

We show a summary of the "duty details" of the given run in the run view
in the minischedule: the "spread" (total length of the working day),
number of hours actually spent working, and number of hours on paid
break.

### DIFF
--- a/assets/tests/components/propertiesPanel/minischedule.test.tsx
+++ b/assets/tests/components/propertiesPanel/minischedule.test.tsx
@@ -419,6 +419,97 @@ describe("MinischeduleRun", () => {
 
     expect(tree).toMatchSnapshot()
   })
+
+  test("renders duty details of run", () => {
+    const piece1: Piece = {
+      runId: "run",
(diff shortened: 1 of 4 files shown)


assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -346,6 +346,113 @@ describe("MinischeduleRun", () => {
 
     expect(tree).toMatchSnapshot()
   })
+
+  test("renders duty details of run", () => {
```
skyqrose, line 350: This shows up in other snapshots, so unless you're testing a specific case here, I don't think having another test for it is necessary.

phildarnowsky: I did specifically want to test a complex case here, to make sure I was exercising all the different parts of the logic.

PR #697, opened 2020-06-10
Make header of VPP sticky

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1179728644442403

This makes the header of the VPP sticky, both with and without minischedules enabled.


ahakuta:
This looks good to me, thanks @phildarnowsky!

PR #701, opened 2020-06-15
Show alert icons on route ladders for current block waivers only

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1176421268419395

For vehicles, we now only show the alert icon on the route ladder if there's a current block waiver, as opposed to a past or future one. We also extend the definition of "current" in this case to cover future block waivers that start no more than 240 minutes in the future.

### DIFF
--- a/assets/src/models/blockWaiver.ts
+++ b/assets/src/models/blockWaiver.ts
@@ -9,13 +9,16 @@ export enum CurrentFuturePastType {
   Past,
 }
 
-export const currentFuturePastType = ({
-  startTime,
-  endTime,
-}: BlockWaiver): CurrentFuturePastType => {
+export const currentFuturePastType = (
+  { startTime, endTime }: BlockWaiver,
+  startThresholdInMinutes: number = 0
+): CurrentFuturePastType => {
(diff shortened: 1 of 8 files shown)


assets/src/models/blockWaiver.ts
```diff
@@ -32,25 +35,26 @@ export const hasCurrentBlockWaiver = ({
...
   vehicleOrGhost: VehicleOrGhost
 ): AlertIconStyle | undefined => {
   if (hasCurrentBlockWaiver(vehicleOrGhost)) {
     return AlertIconStyle.Black
   }
+
   if (isGhost(vehicleOrGhost) && !isLateVehicleIndicator(vehicleOrGhost)) {
     return AlertIconStyle.Highlighted
-  } else {
-    return hasBlockWaiver(vehicleOrGhost) ? AlertIconStyle.Grey : undefined
```
skyqrose, line 54: Since we no longer have grey icons anymore, that value can be removed from the enum in `iconAlertCircle.tsx` and from the scss in `_icon_alert_circle.scss`

PR #705, opened 2020-06-16
Rework implementation of VPP/GPP tab controls

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1176968901192802

Several commits ago, we made the vehicle-status header at the top of the
VPP/GPP sticky via a simple CSS tweak. We would also like the tabs
themselves to stick, so that a user can scroll up and down the full
length of a run/block and not have to scroll back to the top to get the
controls back.

Unfortunately, the react-tabs library that we were using locked us into
a certain hierarchy of components, which was not compatible with how
we'd have to alter the page's structure to get the tabs in the sticky
container too. Here we get rid of the react-tabs library and roll our
own implementation of tabs.

sticky_tabs


assets/src/components/propertiesPanel/header.tsx
```diff
@@ -121,7 +141,30 @@ const directionName = (
 
 const nowInSeconds = (): number => Math.floor(Date.now() / 1000)
 
-const Header = ({ vehicle, route }: Props) => {
+const Tab = ({
```
skyqrose, line 144: Could a lot of this live in `tabPanels.tsx` instead of here? The content half has a nicely abstracted component, and I think the button half would benefit from the same.

assets/src/components/propertiesPanel/header.tsx
```diff
@@ -134,39 +177,47 @@ const Header = ({ vehicle, route }: Props) => {
...
+        <div className="m-properties-panel__close-ping">
+          <CloseButton onClick={hideMe} />
+
+          {isVehicle(vehicle) && (
+            <div className="m-properties-panel__last-gps-ping">
+              {secondsAgo(vehicle.timestamp)}
+            </div>
+          )}
+        </div>
+      </div>{" "}
```
skyqrose, line 215: Why are these spaces needed?

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
```diff
@@ -29,21 +29,31 @@ const NoWaiverBanner = () => (
...
-    />
-  </div>
-)
+      {hasBlockWaiver(selectedGhost) ? (
+        <BlockWaiverList blockWaivers={selectedGhost.blockWaivers} />
+      ) : (
+        <NoWaiverBanner />
+      )}
+
+      <TabPanels
```
skyqrose, line 50: By moving the tabs to the header, any block waivers will now show between the tabs and the content. Is that what we want?

I realize that it might be tricky to do it the other way if we want the block waivers to scroll but the tabs to be sticky. Maybe the block waivers should live as part of the statusContent and not show in the run/block tabs?

ahakuta:
@phildarnowsky this looks good to me- signing off on this. Thanks!

assets/src/components/propertiesPanel/header.tsx
phildarnowsky: I think my editor added those on its own initiative.

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
phildarnowsky: Right, doing it where the block waivers are above but can scroll out of sight while the tabs stick gets us back to the complex and hackish-feeling JS implementation that we were trying to avoid--or possibly back to trying those flexbox tricks from the other day.

Considering that block waivers currently do show in all three tabs, I think in the absence of a compelling reason to show them on just the status tab, that we keep them on all three tabs.

assets/src/components/propertiesPanel/header.tsx
phildarnowsky: Yeah I think the thing to do here would be to extract a third file, `tab.tsx`

assets/src/components/propertiesPanel/ghostPropertiesPanel.tsx
skyqrose: If we want the block waivers above the tabs, and don't mind them being sticky too, we could do:
```
<div class=panel>
  <div class=sticky-wrapper>
    <Header/>
    <BlockWaivers/>
    <Tabs/>
  </div>
  <PanelContent (status or minischedule)/>
</div>
```
(moving the tabs and the sticky behavior out of the header and into the VPP/GPP)

phildarnowsky: @skyqrose I think at this point it'd be best to ask @ingridpierre which she'd prefer:

* Keep block waivers above tabs, make them sticky.
* Block waivers go below tabs, and scroll

I'd figure the former, but perhaps not?

phildarnowsky: @ingridpierre @skyqrose or third option, if we wanted, put the block waivers below the tabs but make them sticky.

skyqrose: Another option would be to move them to the status tab only and make them not sticky.

ahakuta: @skyqrose @phildarnowsky thanks for flagging this question about what to do with block waivers.

I just spoke with @ingridpierre and we agreed the following would be best:

- Put the block waivers below the tabs (specifically in the STATUS tab, the way it already is)

- Block waiver information should scroll like the rest of VPP content (making it sticky would lead to too much vertical height being consumed by sticky elements on mobile)

@skyqrose @phildarnowsky I know we already support scrolling of the RUN and BLOCK views in mini schedules, but do we need a ticket to support scrolling of the STATUS tab to support the requirements outlined above for block waivers?

ahakuta:
@skyqrose @phildarnowsky I spoke with @ingridpierre and we all agree that block waiver information should only be visible in the STATUS tab. 

As noted above, it should not be sticky either- thanks!

phildarnowsky:
@ahakuta @skyqrose @ingridpierre block waivers tweaked as discussed, ready for another look

assets/src/components/propertiesPanel/tabPanels.tsx
```diff
@@ -0,0 +1,36 @@
...
+        vehicleOrGhost.tripId && (
+          <MinischeduleBlock vehicleOrGhost={vehicleOrGhost} />
+        )
+      )
+  }
+}
+
+const TabPanels = (props: Props) => (
+  <div className="m-tabs">
+    <div className="m-tabs__tab-panel m-tabs__tab-panel--selected">
```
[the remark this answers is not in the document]
phildarnowsky: I got this markup by aping what `react-tabs` did so that I could re-use our existing CSS; perhaps we could do without the nesting here, I'll try.

phildarnowsky:
@ahakuta and/or @ingridpierre this could use one more look to verify that I didn't introduce any visual regressions in the process of carrying out Sky's suggestions, otherwise ready to merge.

ingridpierre:
@phildarnowsky I was having difficulty telling whether the status tab will properly scroll because currently the map resizes so as to rarely need a vertical scroll. To find this bug I had to shrink my window down quite a lot. Setting a min-height on the map was created as a separate ticket, but if it's easy maybe it's worth doing in this ticket since it will help us troubleshoot here?

Looks like the map doesn't go "behind" the sticky elements as intended:
image

phildarnowsky:
@ingridpierre this latest should both fix the z-index issue with the map, and give the map a decent min-height.

ingridpierre:
@phildarnowsky Those changes look great.  👍 
-----
I'm going to make a separate ticket to address the scroll/panning trap issue I'm seeing here. It's not bad enough to block this ticket, but we should have a discussion about what to do when maps that take up most of the panel prevent your from scrolling up/down.

PR #706, opened 2020-06-16
Mid Route Swings

skyqrose:
Asana Task: [Mini-schedule | Mid-Route Swings](https://app.asana.com/0/1148853526253426/1170698291365289)

~Depends on #702~

Doesn't highlight the trip if it's active. I've pushed that off to [a separate task](https://app.asana.com/0/1148853526253420/1180597825747907) because this PR is already big.

Requires running `mix cache.clean` if you want to run it locally.

Run view swing on:
Screen Shot 2020-06-16 at 10 21 25

Run view swing off:
Screen Shot 2020-06-16 at 09 59 29

Block view swing off and swing on:
Screen Shot 2020-06-16 at 10 21 46


lib/schedule/minischedule/load.ex
```diff
@@ -119,9 +119,9 @@ defmodule Schedule.Minischedule.Load do
 
   @spec trip_in_operator(Activity.t(), Trip.t()) :: boolean()
   defp trip_in_operator(%Activity{activity_type: "Operator"} = activity, trip) do
-    trip.start_time >= activity.start_time and
-      trip.end_time <= activity.end_time and
-      String.contains?(trip.block_id, activity.partial_block_id)
+    String.contains?(trip.block_id, activity.partial_block_id) and
```
phildarnowsky, line 155: this is something I noticed before and meant to ask about, what's the deal with sometimes having partial and sometimes what I assume are complete block IDs?

phildarnowsky:
@skyqrose aside from my concerns about `operator_activity_to_piece` this looks good.

lib/schedule/minischedule/load.ex
skyqrose: `trips.csv` lists its block ids like `"A57-11"`, which is what GTFS uses, but `activities.csv` only gives us the `"57-11"` part. So we have to do this `String.contains?` matching instead of `==`.

An alternative might be to try to reconstruct the full block id. The missing letter, which represents the garage, shows up (in lowercase) as the 3rd character of the schedule_id, but using the trip's full block id seemed easier.

lib/schedule/minischedule/load.ex
```diff
@@ -103,6 +112,30 @@ defmodule Schedule.Minischedule.Load do
           nil
       end
 
+    start_mid_route? =
```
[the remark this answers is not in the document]
skyqrose: 👍  Done.

lib/schedule/minischedule/load.ex
phildarnowsky: No doubt, just wondering what that was about.

PR #710, opened 2020-06-22
Fullscreen button on VPP map

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1173065850148378

...pretty much what it says on the label.

full_screen_map

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -18,6 +18,9 @@ import {
 } from "react-leaflet"
 // @ts-ignore
 import Control from "react-leaflet-control"
+// @ts-ignore
+import FullscreenControl from "react-leaflet-fullscreen"
+import "react-leaflet-fullscreen/dist/styles.css"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import vehicleLabelString from "../helpers/vehicleLabel"
 import { drawnStatus, statusClass } from "../models/vehicleStatus"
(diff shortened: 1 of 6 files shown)


ingridpierre:
@phildarnowsky Looking good so far, are we able to use the SVGs attached to the asana ticket (one for full screen, one for collapse) instead of the one that's standard with this library?

phildarnowsky:
@ingridpierre I figured out how to use your SVGs here. Also noticed that the controls were in the wrong order so I corrected that.

assets/src/components/map.tsx
```diff
@@ -18,6 +18,9 @@ import {
 } from "react-leaflet"
 // @ts-ignore
 import Control from "react-leaflet-control"
+// @ts-ignore
+import FullscreenControl from "react-leaflet-fullscreen"
+import "react-leaflet-fullscreen/dist/styles.css"
```
skyqrose, line 23: Why is this (and the moduleNameMapper/cssStubber) necessary to import the css? Is there something different about this package compared to like the react-tooltip package that means we need to import the css explicitly? And if there is, is this import better here or in webpack/app.scss configuration?

phildarnowsky: @skyqrose hmmmm, it didn't work without an explicit import before...and now it does! I guess I fixed it.

ahakuta:
Finally got my local machine working and was able to checkout this branch and run it- I'm seeing some slight weirdness with the SVG icons that @ingridpierre will investigate.

ingridpierre:
@phildarnowsky I've added two new svgs "collapse" and "expand" to the asana ticket. These use polygons instead of paths, wondering if that will help avoid the pixel distortion Akira was observing.

@ahakuta when/if it's ready for another look can you retest it on the same screen on which you took the examples you showed me?

phildarnowsky:
@ingridpierre @ahakuta give this a whirl, what kind of distortion are you seeing exactly?

ahakuta:
Screenshot 2020-06-25 12 01 49
Screenshot 2020-06-25 12 01 41

@ingridpierre just re-tested this, here's what I'm seeing. It looks fine to my eyes, with maybe a little weirdness with the EXIT FULLSCREEN button, but not so bad that I think we should hold off on this longer...

lmk what you think!

ingridpierre:
@ahakuta it's still a bit wonky/weird, but mostly in Chrome. Safari and FF render it perfectly fine, think this is just a browser rendering issue.

let's call this one 👍 gtg.

PR #711, opened 2020-06-22
Show past trips toggle

skyqrose:
Asana Task: [Mini-schedule | “Show past trips” button](https://app.asana.com/0/1148853526253426/1153890233353385)

The CSS and snapshot have some indentation changes and are easier to read with the whitespace changes disabled.


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -42,22 +42,53 @@ export interface Props {
 
 export const MinischeduleRun = ({ vehicleOrGhost }: Props): ReactElement => {
   const run: Run | null | undefined = useMinischeduleRun(vehicleOrGhost.tripId!)
+  return (
+    <Minischedule runOrBlock={run} vehicleOrGhost={vehicleOrGhost} view="run" />
```
phildarnowsky, line 53: @skyqrose rather than pass a separate and redundant `view` argument, could we instead have a [type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) to determine if `runOrBlock` is a run or if it's a block?

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -42,22 +42,53 @@ export interface Props {
...
+          label={view === "run" ? "Run" : "Block"}
+          value={runOrBlock.id}
+        />
+        {view === "run" ? <DutyDetails run={runOrBlock as Run} /> : null}
+        {activities.map((activity, index) =>
           isPiece(activity) ? (
             <Piece
               piece={activity}
-              view="run"
+              view={view}
```
phildarnowsky, line 105: overall very nice refactoring in this commit!

assets/src/components/propertiesPanel/minischedule.tsx
skyqrose: I tried that, and it did make a bunch of things nicer, but we wouldn't be able to say something different for `No run found`/`No block found`, since we wouldn't have an object to call the type predicate on.

phildarnowsky: Oh right. We could probably do something clever with the type system to avoid the redundancy, but that's going a little overboard.

ahakuta:
This PR looks good to me- signing off on this!

PR #718, opened 2020-06-25
Prevent wrapping of time since last gps ping

skyqrose:
Asana Task: [🐞 Last GPS ping in VPP can cause weird wrapping issues](https://app.asana.com/0/1148853526253426/1180518784070169)

Headsign will wrap without the time wrapping:
Screen Shot 2020-06-25 at 09 49 50

Removed the padding between headsign and last gps ping so the headsign can be longer without wrapping:
Screen Shot 2020-06-25 at 09 45 51

and make headsign less likely to wrap and unwrap by preventing width changes until the data is 100 seconds old

### DIFF
--- a/assets/css/properties_panel/_header.scss
+++ b/assets/css/properties_panel/_header.scss
@@ -16,7 +16,9 @@
 .m-properties-panel__variant {
   border-left: 1px solid $color-rule;
   flex: 1 1 auto;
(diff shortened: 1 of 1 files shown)


ahakuta:
Signing off on this PR- thanks @skyqrose!

PR #719, opened 2020-06-25
Add crowding display to VPP

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1181069452654991

Screen Shot 2020-06-25 at 2 38 08 PM


phildarnowsky:
@skyqrose @ingridpierre I need to raise test coverage, but this is complete enough to start reviewing.

lib/concentrate/vehicle_position.ex
```diff
@@ -41,6 +41,11 @@ defmodule Concentrate.VehiclePosition do
     :scheduled_headway_secs,
     :sources,
     :data_discrepancies,
+    :load,
+    :route_has_reliable_crowding_data,
+    :capacity,
+    :occupancy_status,
+    :occupancy_percentage,
```
skyqrose, line 48: Did you consider whether these would be cleaner as a struct/map in a single nilable field, so they could all be `nil` together instead of each field being `nil` one-by-one?

It could reduce the amount of boilerplate in tests, and could make things like this typesafe:
```ts
  // If vehicle.load isn't null, vehicle.occupancyStatus shouldn't be either.
  return vehicle.occupancyStatus! 
```

but I don't know if the data is consistent enough or if our vehicle_position processing code could handle it cleanly.

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+const GenericCrowdingDiagram = ({
+  vehicle,
+  statusDescription,
+}: {
+  vehicle: Vehicle
+  statusDescription: string
+}) => {
+  const classModifier = statusDescription.toLowerCase().replace(" ", "-")
```
skyqrose, line 30: I generally don't like taking human readable strings ("Some crowding") and turning them into machine-readable strings ("some-crowding") because at the string literal it's not clear that some other code (the css selector) is going to match against it. In the future we might decide that "Medium crowding" is better microcopy and break the css.

What if you did (roughly, I know I'm misusing js switch statements here)
```
    case "FEW_SEATS_AVAILABLE":
      return <GenericCrowdingDiagram vehicle={vehicle} status="FEW_SEATS_AVAILABLE"/>
```
and then in this component
```
statusDescription = switch status {
  case "FEW_SEATS_AVAILABLE": "Some crowding"
}
classModifier = switch status {
  case "FEW_SEATS_AVAILABLE": "some-crowding"
}
```

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -0,0 +1,110 @@
+import React from "react"
+import { crowdingIcon } from "../../helpers/icon"
+import { Vehicle } from "../../realtime"
+
+const extendedOccupancyStatus = (vehicle: Vehicle): string => {
```
skyqrose, line 5: Is the list of potential occupancy statuses defined in a type anywhere? If it were, it'd make it easier to make sure that any case over all the options doesn't miss a status.

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <div className="m-crowding-diagram__crowding-icon-wrapper">
+        {crowdingIcon(
+          `m-crowding-diagram__crowding-icon m-crowding-diagram__crowding-icon--${classModifier}`
+        )}
+      </div>
+    </div>
+  )
+}
+
+const NoDataCrowdingDiagram = () => (
```
skyqrose, line 56: This is pretty similar to the generic component. Would it be reasonable to have this drawn by the generic instead of being separate? The css class and text is already dynamic there. It'd just need to put a couple of pieces in a conditional.

assets/css/_crowding_diagram.scss
```diff
@@ -0,0 +1,82 @@
+$color-empty-text: $color-component-medium;
+$color-empty-icon: $color-secondary-light;
+$color-not-crowded: #5cbf95;
+$color-some-crowding: #fca400;
+$color-crowded: #e90d0d;
```
skyqrose, line 5: Since these are global variables, perhaps they should be namespaced, like `$color-crowding-crowded`/`$color-crowding-empty-icon` or something?

assets/css/_crowding_diagram.scss
```diff
@@ -0,0 +1,82 @@
...
+  position: absolute;
+  right: 1rem;
+
+  svg {
+    height: 2.5rem;
+    width: 2.5rem;
+
+    circle,
+    path {
+      fill: $color-bg-medium;
```
skyqrose, line 46: This could get an alias with the others at the top of the file.

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const SomeCrowdingCrowdingDiagram = ({ vehicle }: { vehicle: Vehicle }) => (
+  <GenericCrowdingDiagram vehicle={vehicle} statusDescription="Some crowding" />
+)
+
+const CrowdedCrowdingDiagram = ({ vehicle }: { vehicle: Vehicle }) => (
+  <GenericCrowdingDiagram vehicle={vehicle} statusDescription="Crowded" />
+)
+
+const CrowdingDiagram = ({ vehicle }: { vehicle: Vehicle }) => {
```
skyqrose, line 88: It could be easier to test and reuse this if instead of taking a whole `vehicle`, it just took the little bit of information related to crowding.

(That could be even easier if the crowding data was all in one object instead of as 5 different fields in the vehicle.)

ingridpierre:
@phildarnowsky Some little things:

- .m-crowding-diagram__status-description should get font-weight: 500
- "No data available" doesn't need a period at the end
- The "No data available" icon uses a really subtle grey-green color, rather than the grey it has here currently. Should be: #D4E7E7. I think it's called $secondary-lightest or something like that.

lib/concentrate/vehicle_position.ex
phildarnowsky: @skyqrose yes, I think you're right, we have a data clump here.

phildarnowsky:
note to self: you forgot to actually _check_ the feature flag.

lib/concentrate/vehicle_position.ex
phildarnowsky: Although we do still have to do that slightly inelegant `!` there.

ahakuta:
Hey @phildarnowsky this is looking good! Two things I noticed for tweaking:

- We say "1 riders", would be cleaner to say "rider" if the load = 1

- We don't have a tooltip which we had intended to add to the crowding panel in VPP. This tooltip would explain that the data is an estimate based on the onboard APC units. We can make this a separate task if you'd prefer that @Phil Darnowsky but I think @Ingrid Pierre, @Maeg Keane and I would agree that that tooltip should be there before we ship this feature

phildarnowsky:
@skyqrose @ingridpierre @ahakuta Ready for another look, although I do once again have to raise the test coverage.

phildarnowsky:
@skyqrose @ingridpierre @ahakuta note that you'll actually have to turn on the feature flag now to see crowding--I was forgetting to check it before.

lib/concentrate/vehicle_position.ex
```diff
@@ -152,7 +154,8 @@ defmodule Concentrate.VehiclePosition do
               second.layover_departure_time,
               first.sources,
               first.layover_departure_time
-            )
+            ),
+          crowding: merge_crowding(first.crowding, second.crowding)
```
skyqrose, line 158: Crowding data will all come from the same source and we'd never need to take half the object from one source and half from another, so I think `first_value(first.crowding, second.crowding)` would be fine here, rather than diving into the struct.

lib/realtime/crowding.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Realtime.Crowding do
+  @type t :: %__MODULE__{
+          load: non_neg_integer(),
+          capacity: non_neg_integer(),
+          occupancy_status: String.t(),
+          occupancy_percentage: float()
```
skyqrose, line 6: These should all be `| nil`, right, for the case where the route has crowding data but the vehicle doesn't?

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -0,0 +1,103 @@
+import React from "react"
+import ReactTooltip from "react-tooltip"
+import { crowdingIcon } from "../../helpers/icon"
+import { Crowding, OccupancyStatus } from "../../realtime"
+
+const extendedOccupancyStatus = (crowding: Crowding): OccupancyStatus => {
```
skyqrose, line 6: Should this transformation be handled instead on the backend when we're parsing the busloc data? (If so, we could make `occupancyStatus` nonnullable.)

phildarnowsky:
@skyqrose this is ready for another look, /cc @ahakuta @ingridpierre

ahakuta:
Taking a look at this today @phildarnowsky, apologies for the delay.

lib/concentrate/parser/gtfs_realtime_enhanced.ex
```diff
@@ -119,13 +119,32 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
         load: Map.get(vp, "load"),
         capacity: Map.get(vp, "capacity"),
         occupancy_status: Map.get(vp, "occupancy_status"),
-        occupancy_percentage: Map.get(vp, "occupancy_percentage")
+        occupancy_percentage: decode_occupancy_status(vp)
```
skyqrose, line 122: Are `_status` and `_percentage` mixed up here?

lib/concentrate/parser/gtfs_realtime_enhanced.ex
```diff
@@ -119,13 +119,32 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
...
+  defp decode_occupancy_status(%{load: nil}) do
+    "NO_DATA"
+  end
+
+  defp decode_occupancy_status(%{load: 0}) do
+    "EMPTY"
+  end
+
+  defp decode_occupancy_status(vp) do
+    Map.get(vp, "occupancy_percentage")
```
skyqrose, line 145: and here?

lib/concentrate/parser/gtfs_realtime_enhanced.ex
phildarnowsky: whoops!

ahakuta:
Screenshot 2020-06-29 11 41 46

@phildarnowsky I think the CROWDED state seems to be missing color/fill for the people icons.

ingridpierre:
Good catch from Akira on the crowded state. Looks like we're still missing the secondary-lightest color on the "no data available" state:

**Current**
image

**Desired**
image

-----
I also hadn't seen it mentioned here explicitly, but let's take out the tooltip behavior here and make that a separate ticket (ux and wording TBD, but coming soon).

phildarnowsky:
@ahakuta @ingridpierre regarding the tooltip, what if we launched crowding without it, but just don't remove the crowding feature flag yet. The advantage here is that we then unblock https://app.asana.com/0/1148853526253426/1181069452654993, which is going to re-use a good deal of the code from this PR.

phildarnowsky:
@ingridpierre @ahakuta This removes the tooltip, and should fix the icon colors.

ahakuta:
Icon colors look good to me- thanks @phildarnowsky!

phildarnowsky:
@ahakuta Behind a flag is what I had in mind, I'll wait to merge this until you get in just so I can confirm that we're on the same page.

phildarnowsky:
Merging per @ahakuta

PR #720, opened 2020-06-29
Fix location of subway vehicles on shuttle map

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1179942155819579

Screen Shot 2020-06-29 at 9 36 53 AM

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -104,11 +104,11 @@ const Vehicle = ({ vehicle }: { vehicle: Vehicle }) => {
 }
 
 const makeTrainVehicleIcon = ({ bearing }: TrainVehicle): Leaflet.DivIcon => {
-  const centerX = 12
-  const centerY = 12
+  const centerX = 21
+  const centerY = 22.5
   return Leaflet.divIcon({
     html: `<svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 36 36">
-        <g transform="rotate(${bearing}, ${centerX}, ${centerY})">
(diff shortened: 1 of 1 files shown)


skyqrose:
Here's the old center:
Screen Shot 2020-06-29 at 10 18 29

And here's the new center:
Screen Shot 2020-06-29 at 10 17 00

So definitely an improvement, but it's still a bit off. I think `centerX` should be `24`. Ingrid probably designed the icon at 48x48, like most of our other icons.

(The red box is the viewbox, but it looks like leaflet ignores it, using `iconAnchor` instead.)

Also, I don't quite know how `iconAnchor` works, but I think the new `translate(36, 36)` probably doesn't do anything, and maybe could cause problems. Here's the viewbox and `iconAnchor` after the translation:
Screen Shot 2020-06-29 at 10 25 10

Did you look into whether it was possible to use the existing bus svg to dodge all of these problems?

phildarnowsky:
@skyqrose I did try using the bus icon's path, and didn't get results, but that was also very early in this process so perhaps having experimented with this some, it's worth another look.

phildarnowsky:
@skyqrose TBH I don't remember by what logic I added the `translate`--but when I remove it, the icons jump noticeably away from the line again. On the one hand, I'm not wild about having that modifier in there without a good theory of why it fixes things. On the other hand, it seems to work consistently (comparing against dotcom).

My day is close to over so let me come back to this one with fresh eyes in the morning, either to think of a better fix or to account for why that translate helps.

phildarnowsky:
Closing in favor of https://github.com/mbta/skate/pull/745

PR #721, opened 2020-06-30
Add tooltip to crowding diagram

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1182434887083497

Tooltip activated by clicking on the "?" icon

Screen Shot 2020-06-30 at 8 19 01 AM

### DIFF
--- a/assets/tests/components/propertiesPanel.test.tsx
+++ b/assets/tests/components/propertiesPanel.test.tsx
@@ -3,6 +3,7 @@ import React from "react"
 import renderer from "react-test-renderer"
 import PropertiesPanel, {
   handleSwipe,
+  hideMeIfNoCrowdingTooltip,
 } from "../../src/components/propertiesPanel"
 import { StateDispatchProvider } from "../../src/contexts/stateDispatchContext"
 import { HeadwaySpacing } from "../../src/models/vehicleStatus"
@@ -154,5 +155,24 @@ describe("handleSwipe", () => {
     const map = document.createElement("div")
     map.id = "id-vehicle-map"
     handleSwipe(hidePanelCB)("Right", map)
+  })
+})
+
+describe("hideMeIfNoCrowdingTooltip", () => {
+  test("hides panel if no tooltip open", () => {
+    const hidePanelCB = jest.fn()
+    hideMeIfNoCrowdingTooltip(hidePanelCB)
+
+    expect(hidePanelCB).toHaveBeenCalled()
+  })
+
+  test("does not hide panel if a tooltip is open", () => {
+    const hidePanelCB = jest.fn()
+    const newDiv = document.createElement("div")
+    // @ts-ignore
+    document.getElementsByClassName = () => [newDiv]
+    hideMeIfNoCrowdingTooltip(hidePanelCB)
+
+    expect(hidePanelCB).not.toHaveBeenCalled()
   })
 })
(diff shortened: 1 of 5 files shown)


ahakuta:
Screenshot 2020-06-30 08 35 51

@phildarnowsky thanks for getting this PR up so quickly. I tested this on a small device (see screenshot above) and the tooltip text isn't fully visible. Is there any feature in the tooltip code that allows it to auto-wrap? 

CC: @ingridpierre

phildarnowsky:
@ahakuta that's nasty, at the very worst I should be able to put in some manual line breaks.

phildarnowsky:
@ahakuta @ingridpierre see if that latest fixes up the over-wide tooltips on narrow devices

ahakuta:
Screenshot 2020-06-30 08 50 18

Looks way better! Thanks @phildarnowsky.

Can we put this onto dev for additional device-level testing? I have an inspector's old tablet on hand, so I'd like to test on it as well.

phildarnowsky:
@ahakuta I think I can get it onto dev, let me try.

ahakuta:
@skyqrose thanks for flagging that, I hadn't noticed that! @phildarnowsky is it possible to allow the tooltip to close on click anywhere else? It doesn't feel like the user should have to click exactly on the tooltip again to hide it.

phildarnowsky:
@ahakuta @skyqrose despite the fact that that functionality's built into `react-tooltip`, it's a bit tricky, because clicking elsewhere also makes the VPP want to close. I'm looking at an approach to keep that from happening.

phildarnowsky:
@ahakuta @skyqrose the tooltip should now behave more like you would expect.

ahakuta:
Will take a look, thanks @phildarnowsky!

ahakuta:
@phildarnowsky this looks good to me, thanks for chipping away at this one to get it to where we wanted to!

One small suggestion on a copy tweak that I ran by @ingridpierre: can we change the tooltip copy to the following, to make it a little shorter and concise?

**Riders are estimated using Automated Passenger Counters (APCs).**

I'm not sure what the right line breaks should be for this shorter copy, but I like the current width of the tooltip and think we should use line break to try and keep it where it is now width-wise.

phildarnowsky:
@ahakuta sure, see how this latest grabs you.

ahakuta:
@phildarnowsky this PR looks good to me, thanks for the copy tweak!

phildarnowsky:
@skyqrose do you have any feedback on the new commits, particularly https://github.com/mbta/skate/pull/721/commits/fcbdd7abf78e5eaede330b196c967cbac49fba3c?

assets/tests/components/propertiesPanel.test.tsx
```diff
@@ -156,3 +157,22 @@ describe("handleSwipe", () => {
...
+    hideMeIfNoCrowdingTooltip(hidePanelCB)
+
+    expect(hidePanelCB).toHaveBeenCalled()
+  })
+
+  test("does not hide panel if a tooltip is open", () => {
+    const hidePanelCB = jest.fn()
+    const newDiv = document.createElement("div")
+    // @ts-ignore
+    document.getElementsByClassName = () => [newDiv]
```
skyqrose, line 173: This will need to be cleaned up with `after()` to avoid polluting the global test environment. `routeLadder.test.tsx`/`routeLadder` has an example.

PR #723, opened 2020-06-30
Add Sentry to Skate

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1177675442396473

This adds error monitoring via Sentry to both the frontend and backend.

Getting Sentry working in prod will require setting an environment variable. Since this reboots the server, and thus flushes the cache, we'll incur a couple of minutes of downtime when we set that variable. However, we can deploy this anytime and set the variable at our convenience--I verified that the worst thing that happens if that env var is not set is that a warning is logged locally, and nothing is sent to Sentry.


assets/src/app.tsx
```diff
@@ -20,6 +21,11 @@ import ReactDOM from "react-dom"
 import ResizeObserver from "resize-observer-polyfill"
 import AppStateWrapper from "./components/appStateWrapper"
 
+Sentry.init({
+  dsn:
+    "https://e4ef550df0f644d5916bef033772db15@o89189.ingest.sentry.io/5303927",
```
skyqrose, line 26: Does this mean that local development will send to sentry? If so, could we set this to only work in prod in the same way we configure fullstory and appcues?

.env.example
```diff
@@ -2,6 +2,7 @@ API_URL=
 API_KEY=
 GTFS_URL=
 BUSLOC_URL=
+SENTRY_DSN=
```
skyqrose, line 5: This file is meant for local setup. Assuming we only want prod to send to sentry, this should go in the README instead.

config/config.exs
```diff
@@ -112,6 +112,12 @@ config :ueberauth, Ueberauth,
     cognito: nil
   ]
 
+# Sentry for error tracking
+config :sentry,
+  dsn: {:system, "SENTRY_DSN"},
```
skyqrose, line 117: for `application.ex` to load this, you'll need to add it to the list in `load_runtime_config()`.

config/config.exs
```diff
@@ -112,6 +112,12 @@ config :ueberauth, Ueberauth,
...
+  included_environments: [:prod],
+  environment_name: Mix.env()
```
skyqrose, line 119: Would putting `config :sentry, dsn: {:system, "SENTRY_DSN"}` only in `prod.exs` accomplish the same thing?

assets/src/app.tsx
phildarnowsky: It does not, that's what the `included_environments` you noticed below will do.

config/config.exs
phildarnowsky: It would. The Sentry docs present several ways of getting to this same end, and that's another way. I did it like this because it was convenient for local testing: I was able to add `:dev` as an `included_environment` and generate errors locally (see e.g. https://sentry.io/organizations/mbtace/issues/1755323107/?project=5303878&query=&statsPeriod=14d).

If you have a strong preference for putting this in `prod.exs`, that's fine by me. I would also be fine leaving this as-is.

assets/src/app.tsx
phildarnowsky: I'll look into it.

phildarnowsky:
@skyqrose Also I'm going to move the frontend DSN to an env var.

phildarnowsky:
@skyqrose ready for another look

PR #724, opened 2020-06-30
show other buses on the same route in VPP

skyqrose:
Asana Task: [Include other vehicles on the VPP map](https://app.asana.com/0/1148853526253426/1134108215297994)

Screen Shot 2020-06-30 at 15 36 20


ingridpierre:
**I was hoping the "other buses" labels in VPP would be about the same size as the ones on the ladders.** Here you can see the labels are a touch larger in VPP map:

image

------

There is some additional functionality that I'm observing that I'd like to remove. **It looks like you can change the vehicle with the heavier/larger treatment by clicking.** The only vehicle with that size/weight should be the one to which the panel belongs.

------

Also noticing that **clicking vehicles makes them hop around a bit**:



-----

Let's hold off on adding extra interactivity to vehicles in the VPP, I don't think we've thought it through entirely.

assets/src/components/map.tsx
```diff
@@ -40,7 +44,10 @@ export const defaultCenter: LatLngExpression = {
   lng: -71.05891,
 }
 
-const makeVehicleIcon = (vehicle: Vehicle): Leaflet.DivIcon => {
+const makeVehicleIcon = (
+  vehicle: Vehicle,
+  primary: boolean
```
phildarnowsky, line 49: @skyqrose small thing, but let's call this `isPrimary`

assets/src/components/map.tsx
```diff
@@ -53,24 +60,31 @@ const makeVehicleIcon = (vehicle: Vehicle): Leaflet.DivIcon => {
         <path
           class="${statusClass(drawnStatus(vehicle))}"
           d="m10 2.7-6.21 16.94a2.33 2.33 0 0 0 1.38 3 2.36 2.36 0 0 0 1.93-.14l4.9-2.67 4.89 2.71a2.34 2.34 0 0 0 3.34-2.8l-5.81-17a2.34 2.34 0 0 0 -4.4 0z"
-          transform="rotate(${vehicle.bearing}, ${centerX}, ${centerY})"
+          transform="scale(${primary ? 1.0 : 0.8}) rotate(${
```
phildarnowsky, line 63: I assume no trouble here with icons being in the wrong place like I was grappling with?

skyqrose: Theoretically no, but I never actually checked that. I will do so.

skyqrose:
Okay, made some changes:

Previously, the location of the other buses were loaded separately from the focused bus,  which is what caused the jumping. Now they all come from the same data loaded at the same time (the same data as is being used for the ladders), so it should be consistent as you switch between buses.

When you click another bus, it doesn't just change the map focus. It changes the whole VPP focus, just the same as if you had clicked that bus on the ladder. I think being able to select a new bus helps answer questions like "This bus is very close to the bus in front of it. Is that bus crowded?". With the non-jumpy locations, it's behaving a lot better. Do you still want me to take it out?

Screen-Recording-2020-07-01-at-1

As part of making the locations more consistent, I ended up removing the ability to see other buses on routes that you don't have open on the ladder right now. That happens if you click an incoming bus that's coming from a route you don't have open, or if you're on the search page. I figured in those cases inspectors are less likely to care about other buses on the route, and it was a little messy to implement, and it re-introduced the inconsistencies/jumping (if you don't already have the route's ladder open, we need to make a separate request to get the other buses, and that separate request was what caused the inconsistencies). But I can bring the feature back in those situations if you think it's important there.

Other small changes:
* adjusted the label size
* verified that the icons are still centered
* added a test

skyqrose:
After discussing with ingrid and others, I've removed the feature for clicking on another bus because it can still be disorienting (we'll reconsider it with the whole-route map idea ingrid is desiging now), and added back the ability to fetch the buses on a route if we don't already have it loaded.

PR #728, opened 2020-07-02
Add action that deliberately fails, to test Sentry

phildarnowsky:
No card. Added this to deliberately generate a harmless error in production, so that we can make sure that Sentry is working properly in prod.


phildarnowsky:
@skyqrose I should mention maybe, the idea is to revert  this commit entirely after we've verified Sentry; I don't want to e.g. keep ignoring `router.ex` for Codecov purposes, just to expedite this one thing.

PR #733, opened 2020-07-02
highlight pullout and deadhead minischedule rows

skyqrose:
Asana Task: [Minischedules | Show active status on more than just revenue trips](https://app.asana.com/0/1148853526253426/1180597825747907)

One problem: The 🚍 icon doesn't play as nicely with highlighting the stroke as the simple triangles do. Any ideas what to do about that?

Screen Shot 2020-07-02 at 15 04 10
Screen Shot 2020-07-02 at 15 20 55

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -511,7 +511,7 @@ const Trip = ({
           />
         )
       ) : (
-        <AsDirected asDirected={trip} timeBasedStyle={tripTimeBasedStyle} />
+        <AsDirected asDirected={trip} timeBasedStyle={onRouteTimeBasedStyle} />
       )}
     </>
   )
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -467,20 +467,23 @@ const Trip = ({
         ? "current"
         : "past"
       : tripTimeBasedStyle
-  tripTimeBasedStyle =
+  const onRouteTimeBasedStyle =
     tripTimeBasedStyle === "current"
       ? vehicleOrGhost.routeStatus === "laying_over"
         ? "future"
         : "current"
       : tripTimeBasedStyle
+  const deadheadTimeBasedStyle = tripTimeBasedStyle
   const layoverActiveStatus: DrawnStatus | null =
     layoverTimeBasedStyle === "current" ? drawnStatus(vehicleOrGhost) : null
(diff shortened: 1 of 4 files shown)


skyqrose:
And also mid-block deadheads:
(I've removed the layover between the deadhead and the next trip since there's no way to tell the deadhead and layover apart, and the concept of a layover there isn't really that meaningful.)

Screen Shot 2020-07-02 at 17 10 49

Here's what it looks like on prod at the same time:

Screen Shot 2020-07-02 at 17 11 01

skyqrose:
phil, you're welcome to take over ownership of this PR while I'm on vacation and merge it / figure out what to do with the icons if you think it doesn't need any back and forth review process, but more realistically, I think it's fine to just leave it sitting here for a week. This isn't urgent.

phildarnowsky:
@skyqrose I took a look and didn't see anything noteworthy in the PR so what I'm thinking is, let's figure out what to do about the icons, then I'll take a second pass through the diff.

skyqrose:
@ingridpierre thoughts on what should happen to these icons when they're highlighted?

ingridpierre:
@skyqrose slight size increase is desired, but let's avoid color/style changes to the icons otherwise. Or in other words, let's just lose the white stroke on the non-revenue trips here.

I can take another look directly in a bit, right now just giving feedback based on screenshots here.

skyqrose:
Okay, I've removed the stroke from the pullout and deadhead and left it on the triangles:

Screen Shot 2020-07-14 at 09 10 36
Screen Shot 2020-07-14 at 09 11 09
Screen Shot 2020-07-14 at 09 12 13

@ingridpierre @phildarnowsky ready for another look from both of you.

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -546,16 +546,16 @@ const iconForDirectionOnLadder: (
   routeId: RouteId
 ) => ReactElement = (directionId, ladderDirections, routeId) => {
   if (directionId === null) {
-    return questionMarkIcon()
+    return questionMarkIcon("m-minischedule__svg--revenue")
```
phildarnowsky, line 549: @skyqrose since we're using this string literal three times, perhaps it's time to extract it to a named constant.

ingridpierre:
I know we can't really show pullbacks in this ticket, but it seems like they are always coming through as invalids/or still doing their last trip and being very late. Could we maybe detect that a bus should have finished its last trip and use the invalid status to highlight a pull back?

Otherwise for this ticket the changes do look good to me. 👍

PR #745, opened 2020-07-10
Fix positioning of train vehicles on shuttle map

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1179942155819579

Screen Shot 2020-07-10 at 1 40 13 PM

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -141,15 +141,14 @@ const Vehicle = ({
 }
 
 const makeTrainVehicleIcon = ({ bearing }: TrainVehicle): Leaflet.DivIcon => {
-  const centerX = 12
-  const centerY = 12
+  const centerX = 24
+  const centerY = 24
   return Leaflet.divIcon({
     html: `<svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 36 36">
         <g transform="rotate(${bearing}, ${centerX}, ${centerY})">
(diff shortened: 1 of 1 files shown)


assets/src/components/map.tsx
```diff
@@ -141,15 +141,14 @@ const Vehicle = ({
...
-  const centerY = 12
+  const centerX = 24
+  const centerY = 24
   return Leaflet.divIcon({
     html: `<svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 36 36">
         <g transform="rotate(${bearing}, ${centerX}, ${centerY})">
           <path fill="#fff" d="m42.88 45.83a2.1 2.1 0 0 1 -.87-.19l-15.92-7.17a5.23 5.23 0 0 0 -2.09-.47 5.14 5.14 0 0 0 -2.08.44l-15.92 7.2a2.1 2.1 0 0 1 -.87.19 2.14 2.14 0 0 1 -1.76-1 2 2 0 0 1 -.12-2l18.86-40.83a2.08 2.08 0 0 1 3.78 0l18.87 40.87a2 2 0 0 1 -.12 2 2.14 2.14 0 0 1 -1.76.96z"/>
         </g>
     </svg>`,
-    iconAnchor: [centerX, centerY],
```
skyqrose, line 152: `iconAnchor` is [centered by default](https://leafletjs.com/reference-1.6.0.html#icon-iconanchor), so this should work, but I think it'd still be good to be explicit and leave it in. Like, I'm not sure if this is center of the viewbox (which it otherwise ignores but would be 18) or the drawing (which would be 24).

PR #750, opened 2020-07-15
Use human readable route name

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1135983810723213

Screen Shot 2020-07-15 at 1 56 46 PM
Screen Shot 2020-07-15 at 1 57 17 PM
Screen Shot 2020-07-15 at 1 57 28 PM

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -5,6 +5,7 @@ import React, {
   useContext,
   useState,
 } from "react"
+import RoutesContext from "../../contexts/routesContext"
 import { StateDispatchContext } from "../../contexts/stateDispatchContext"
 import { className } from "../../helpers/dom"
 import {
@@ -22,7 +23,6 @@ import {
   useMinischeduleBlock,
   useMinischeduleRun,
 } from "../../hooks/useMinischedule"
-import useRoutes from "../../hooks/useRoutes"
 import {
   AsDirected,
   Block,
@@ -79,7 +79,7 @@ export const Minischedule = ({
   view: "run" | "block"
 }) => {
   const [showPast, setShowPast] = useState<boolean>(false)
-  const routes = useRoutes()
+  const routes = useContext(RoutesContext)
 
   if (runOrBlock === undefined) {
     return <Loading />
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -22,6 +22,7 @@ import {
   useMinischeduleBlock,
   useMinischeduleRun,
 } from "../../hooks/useMinischedule"
+import useRoutes from "../../hooks/useRoutes"
 import {
   AsDirected,
   Block,
@@ -39,8 +40,9 @@ import {
 } from "../../models/ladderDirection"
 import { drawnStatus, DrawnStatus } from "../../models/vehicleStatus"
 import { VehicleOrGhost } from "../../realtime"
-import { DirectionId, RouteId, TripId } from "../../schedule"
+import { DirectionId, Route, RouteId, TripId } from "../../schedule"
 import { formattedDuration, formattedScheduledTime } from "../../util/dateTime"
+import { routeNameOrId } from "../../util/route"
 import Loading from "../loading"
 
 export interface Props {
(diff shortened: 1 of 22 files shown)


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -77,6 +79,8 @@ export const Minischedule = ({
   view: "run" | "block"
 }) => {
   const [showPast, setShowPast] = useState<boolean>(false)
+  const routes = useRoutes()
```
skyqrose, line 82: This will make a new api call. Since the routes are already in the properties panel, could they be passed in as an argument?

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -428,13 +450,15 @@ const Trip = ({
...
   view,
+  routes,
 }: {
   trip: Trip | AsDirected
   previousEndTime: Time | undefined
   sequence: "first" | "middle" | "last"
   tripTimeBasedStyle: TimeBasedStyle
   vehicleOrGhost: VehicleOrGhost
   view: "run" | "block"
+  routes: Route[] | null
```
skyqrose, line 461: There's a lot of prop drilling to do this feature. Did you consider whether a context would fit well?

assets/src/state.ts
```diff
@@ -48,7 +48,7 @@ export const selectRoute = (routeId: RouteId): SelectRouteAction => ({
   payload: { routeId },
 })
 
-interface DeselectRouteAction {
+export interface DeselectRouteAction {
```
skyqrose, line 51: what else uses this?

assets/tests/components/routePicker.test.tsx
```diff
@@ -18,6 +18,11 @@ describe("RoutePicker", () => {
         directionNames: { 0: "Outbound", 1: "Inbound" },
         name: "111",
       },
+      {
+        id: "741",
+        directionNames: { 0: "Outbound", 1: "Inbound" },
+        name: "SL1",
```
skyqrose, line 24: If you added this to test the name being used in the selected list, you'll also need to add it to the `selectedRouteIds` below.

assets/tests/components/routeVariantName.test.tsx
```diff
@@ -96,8 +101,23 @@ describe("RouteVariantName", () => {
...
 
-    const wrapper = shallow(<RouteVariantName vehicle={testVehicle} />)
+    const wrapper = shallow(
+      <RouteVariantName vehicle={testVehicle} routes={[]} />
+    )
 
     expect(wrapper.text()).toEqual("Shuttle")
   })
+
+  test("uses route name if available", () => {
```
skyqrose, line 111: 👍

ahakuta:
This looks good to me! Thanks @phildarnowsky.

CC @ingridpierre

assets/src/components/propertiesPanel/minischedule.tsx
phildarnowsky: I didn't, that sounds like a good idea.

phildarnowsky: also, filing away term "prop drilling" for future use.

assets/src/state.ts
phildarnowsky: We use it to construct one type in https://github.com/mbta/skate/pull/750/files#diff-92d87afe2518867c84b521efaa759c23. That said, I find it kind of inelegant exporting just this one interface, if you have a suggestion for not having to do that I'd be interested.

phildarnowsky:
@skyqrose overhauled this to use a context, which was a fair bit of work but simplified things greatly. Good call there.

assets/tests/components/searchResults.test.tsx
```diff
@@ -492,7 +495,7 @@ describe("SearchResults", () => {
     const testDispatch = jest.fn()
     const wrapper = mount(
       <StateDispatchProvider state={state} dispatch={testDispatch}>
-        <SearchResults vehicles={[]} />
+        <SearchResults vehicles={[]} routes={[]} />
```
skyqrose, line 498: I think these props in the tests are obsolete now.

assets/src/state.ts
skyqrose: You could make it the type `Dispatch` (from `state.ts`, not React's `Dispatch` type).

assets/tests/components/searchResults.test.tsx
phildarnowsky: Do you know a way to force recompilation of all tests? Webpack doesn't seem to consistently catch breaking changes in tests--meaning, really, I don't understand the exact conditions. But I'd rather just be able to start fresh when I want to.

PR #752, opened 2020-07-15
Show nearest intersection

skyqrose:
Asana Task: [Add nearest intersection to VPP](https://app.asana.com/0/1148853526253426/1120980093841759)

Screen Shot 2020-07-15 at 15 54 12

If it's still loading / failed to load:
Screen Shot 2020-07-15 at 15 54 32

Old:
Screen Shot 2020-07-15 at 15 44 17

It still needs some QA for edge cases (bridges, tunnels, etc).

We're waiting for a response from geonames to see if we can add a token to the request. Right now it's completely open, all you need is the username to make the request.

Codecov is failing because we don't (and shouldn't) test the call to Geonames. Everything else is covered.

### DIFF
--- a/assets/css/properties_panel/_vehicle_properties_panel.scss
+++ b/assets/css/properties_panel/_vehicle_properties_panel.scss
@@ -24,12 +24,12 @@
   padding: 1rem;
 }
 
-// Not in use temporarily - will be used once we have a current location and
-// can move the "seconds ago" timestamp back up to the top-right. --MSS 2019-08-16
-.m-vehicle-properties-panel__timestamp {
-  @include font-small;
-  color: $color-font-grey;
-  float: right;
+.m-vehicle-properties-panel__latlng {
+  flex: 0 1 4.25rem;
+}
+
+.m-vehicle-properties-panel__next-stop {
+  flex: 0 0 auto;
 }
 
 .m-vehicle-properties-panel__label {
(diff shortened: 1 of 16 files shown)


ahakuta:
@skyqrose this is looking good! A quick question: if I open a VPP, do we make a service request right away, or is there some set heartbeat for making the requests running in the background? If the latter, is there a way to ensure that opening a VPP results in a new call? 

I've seen some instances of VPP being opened and no intersection data appearing for 5-7 seconds which can feel a bit odd, but that may be due to Geonames also failing to return a result other than NULL?

ingridpierre:
Looks like we have a funny state  "&" when there's not a proper intersection:
image

ingridpierre:
When the current location flashes in when the VPP loads, the elements underneath all jump up/down slightly. Can we force a height to current location or add a placeholder so the height of things doesn't change so much?

skyqrose:
We only make the request when somebody opens the panel, and then each time the bus moves while the panel is open.

5-7 seconds is long enough that I'd guess something went wrong with the first request, and then when the bus moved, it worked for the next latlng.

skyqrose:
Added a space so that it doesn't jump when the intersection loads:
Kapture 2020-07-16 at 10 24 10

And I added a check for the case where there's no intersection, so nothing will show instead.

ahakuta:
This look good to me- thanks @skyqrose

PR #758, opened 2020-07-21
Prevent mobile devices from sleeping with NoSleep

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1182803855145267

Using NoSleep.JS, prevent mobile screen from locking. Confirmed to work on my Android phone.


assets/package.json
```diff
@@ -17,9 +17,11 @@
   "dependencies": {
     "@rehooks/component-size": "^1.0.3",
     "@sentry/react": "^5.19.2",
+    "@types/nosleep.js": "^0.9.0",
```
skyqrose, line 20: `@types/` packages should be in dev dependencies.

phildarnowsky:
@skyqrose re the timeout, @ahakuta asked me the same thing. It doesn't look like there's a way built into NoSleep but perhaps it would be possible to roll our own, I'll think about it.

phildarnowsky:
Per discussion with @ahakuta, and as @skyqrose asks about above, will rework to include a 30-minute timeout from the last user interaction.

phildarnowsky:
@ahakuta @skyqrose one thing I came across is: https://github.com/richtr/NoSleep.js/blob/master/src/index.js#L88, which might be problematic if any users are on very old versions of iOS (version < 10). Those versions have < 1% market share worldwide now but wanted to flag that just in case.

skyqrose:
Does "interrupting request in progress" include stopping open sockets? That could cause problems. Otherwise, I think it's probably fine.

We might be able to check on fullstory how many people are using those devices for skate.

phildarnowsky:
@skyqrose I'm not clear on that; worldwide the problematic browsers are <1% of market share, but yeah we should check our own userbase.

ingridpierre:
> iOS (version < 10

Eavesdropping, we're not able to use FullStory to get info about OS version, just the overall OS type.

phildarnowsky:
@ingridpierre don't suppose we have GA or anything else that would tell us the OS version?

phildarnowsky:
@skyqrose this is ready for another look, with an extended but not indefinite sleep. @ahakuta I'll push it to dev.

Still haven't tested on iOS, at this point it's a race between XCode and the new iPhone to see which one can be ready for testing first.

phildarnowsky:
@skyqrose @ahakuta confirmed to work on this new iPhone.

skyqrose:
Sentry reports OS version with every crash report, and the oldest iOS I saw was iOS 11, so I think we're good on that front.

skyqrose:
> after some initial setup and testing, this didn't seem to work more than one time (ie timeout was only delayed once, then the system-defined timeout time took over)

PR #761, opened 2020-07-22
Upload sourcemaps to Sentry

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1184712585012913

Sentry has the ability to accept an uploaded sourcemap for our frontend code, so that instead of having stacktraces that look like this:

Screen Shot 2020-07-22 at 9 18 24 AM

we can have stacktraces that look like this:

Screen Shot 2020-07-22 at 9 18 30 AM

This PR adds such an upload to our npm `deploy` script.

One thing to note is that this does leave a public-facing `app.js.map` that one could download if they knew to look at, e.g., https://skate-dev.mbtace.com/js/app.js.map. Since our code is all open-source that seems OK to me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,10 @@ RUN mix local.hex --force && \
 
 RUN apk add --update nodejs nodejs-npm
 
+ARG SENTRY_ORG=$SENTRY_ORG
+ARG SENTRY_PROJECT=$SENTRY_PROJECT
+ARG SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN
+
 ENV MIX_ENV=prod
 
 ADD . .
--- a/assets/package.json
+++ b/assets/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@rehooks/component-size": "^1.0.3",
     "@sentry/react": "^5.19.2",
+    "@sentry/webpack-plugin": "^1.12.0",
     "core-js": "^3.6.4",
     "leaflet": "^1.6.0",
     "leaflet-defaulticon-compatibility": "^0.1.1",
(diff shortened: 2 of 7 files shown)


assets/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@rehooks/component-size": "^1.0.3",
     "@sentry/react": "^5.19.2",
+    "@sentry/webpack-plugin": "^1.12.0",
```
skyqrose, line 20: If this package is just used during webpack building, and doesn't need to be shipped to the client, it should go in `devDependencies`.

Dockerfile
```diff
@@ -15,6 +15,10 @@ RUN mix local.hex --force && \
 
 RUN apk add --update nodejs nodejs-npm
 
+ARG SENTRY_ORG=$SENTRY_ORG
+ARG SENTRY_PROJECT=$SENTRY_PROJECT
+ARG SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN
```
skyqrose, line 20: Where is this defined, what's the path it takes to get here, and does it need to be secret?

I'm not familiar with `ARG` (or with docker at all really), but in trying to figure out what this does, I saw this:
> ARG values can be easily inspected after an image is built, by viewing the docker history of an image. Thus they are a poor choice for sensitive data.
https://vsupalov.com/docker-arg-env-variable-guide/

Is that an issue for us?

phildarnowsky: These start off as the corresponding `SENTRY_*` env vars in dev. The `--build-arg` options added to `build.sh` expose those for use inside the `Dockerfile`, and calling `ARG` makes those environment variables available during the build, but not to the resulting image itself, which is exactly what we want here.

I don't think we have to worry about anyone viewing the image history, because IIUC if someone has access to that, they probably have wide-ranging access to our server in general, perhaps root access, so we already have bigger problems.

PR #764, opened 2020-07-23
Add hack to keep iPhones from zooming in on inputs

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1152520728603475

By setting the `font-size` for `input`s to 24px (rather than the 16px that's theoretically supposed to work), the e.g. search input no longer zooms in when selected.

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -72,3 +72,12 @@ body {
   flex: 1 1 0;
   min-height: 0;
 }
+
+// Hack to prevent iPhones from automatically zooming in when an input is
+// selected. A font size of 16px is supposed to do the job, but for some
+// reason this doesn't work at less than 24 px.
+@media screen and (max-width: 767px) {
+  input {
+    font-size: 24px !important;
+  }
+}
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -72,12 +72,3 @@ body {
   flex: 1 1 0;
   min-height: 0;
 }
-
-// Hack to prevent iPhones from automatically zooming in when an input is
-// selected. A font size of 16px is supposed to do the job, but for some
-// reason this doesn't work at less than 24 px.
-@media screen and (max-width: 767px) {
-  input {
-    font-size: 24px !important;
-  }
-}
--- a/assets/src/app.tsx
+++ b/assets/src/app.tsx
@@ -36,16 +36,6 @@ if (!("ResizeObserver" in global)) {
   window.ResizeObserver = ResizeObserver
 }
 
-// Show more content on small screens by changing the meta viewport scale.
-if (screen.width < 1000 && screen.height < 1000) {
-  document
-    .getElementsByName("viewport")[0]
-    .setAttribute(
-      "content",
-      "width=device-width, initial-scale=.8, minimum-scale=.8"
-    )
-}
-
 // Import local files
 //
 // Local files can be imported directly using relative paths, for example:


assets/css/app.scss
```diff
@@ -72,3 +72,12 @@ body {
   flex: 1 1 0;
   min-height: 0;
 }
+
+// Hack to prevent iPhones from automatically zooming in when an input is
+// selected. A font size of 16px is supposed to do the job, but for some
+// reason this doesn't work at less than 24 px.
+@media screen and (max-width: 767px) {
```
skyqrose, line 79: There's a `mobile-max-width` variable with this value.
```suggestion
@media screen and (max-width: $mobile-max-width) {
```

phildarnowsky:
@skyqrose it is extremely terrible. But on the bright side, I have spent the past 24 hours immersed in Apple ecosystem minutiae, and as a result, I've come to appreciate Linux even more.

ingridpierre:
Overall, having the font be 24px makes it obnoxiously large and seems to break/cut-off/stretch other neighboring elements.

At the moment the hack looks and feels pretty undesirable. Can't approve it.

I tried to track down the pr where skate added some responsive sizing (which is what I suspect is driving the need for the hack to cancel out some other sizing effects), but had no luck. Did you happen to come across any possible media size rules in your investigation Phil? I'm truly confused, but it feels like somewhere we need to set the body to 100% again.

Also seems related or affects this ticket, demonstrating extremely tiny scaled font sizing in VPP: https://app.asana.com/0/1148853526253426/1185567580329333/f

phildarnowsky:
@ingridpierre I found some code that, on narrow screens, overrides the `initial-scale` and `minimum-scale` dynamically to 0.8 (see https://github.com/mbta/skate/blob/master/assets/src/app.tsx#L39). That's the culprit.

We have two possible fixes that I can see here. 

One is to remove that chunk of code and not display as much on phones. The before-and-after of this change is below. As you can see, that's problematic because the right side of the first ladder is no longer in view at page load. The effect is similar, though more exaggerated, on my Android phone.
Screen Shot 2020-08-03 at 4 34 30 PM
Screen Shot 2020-08-03 at 4 34 44 PM

The other possibility is to set a `maximum-scale` of 0.8 as well. That prevents the automatic zoom nicely. On Android devices, however, it also prevents user-initiated zoom. That seems like a non-starter from an accessibility point of view, especially given the prevalence of Android in our userbase.

I think the least of all evils here is to leave the code as it originally was, mark the ticket "won't fix", and accept the auto-zoom behavior. If you'd prefer a different alternative though please let me know.

ingridpierre:
@phildarnowsky  Can we try the first option: Remove the .8 scaling, set the input size back to 16px. And can we have that put up on dev so that Akira can also test on his devices?

ahakuta:
@ingridpierre @phildarnowsky I just tested on a test iPhone 6, and on one of the tablets that our bus officials use. 
Screenshots of what I saw/found are below. My takeaway though is that this is a nice improvement that makes it easier to see everything on the ladders at a glance. Some users on small screens may need to learn to hide the left hand route picker, but Ingrid already pulled some data on this and many already are, so it's just a matter of getting the stragglers to use it. My recommendation as of right now is that we roll this out.

### iPhone 6 with no scaling, with left hand panel open
IMG_0064

### iPhone 6 with no scaling, with left hand panel minimized
IMG_0065

### Compare the two above to prod, with scaling on the iPhone 6:
IMG_0066
IMG_0067

### Tablet

### No scaling, portrait mode
Screenshot_20200804-155849_Chrome

### Prod with scaling, portrait mode
Screenshot_20200804-155759_Chrome

### No scaling, landscape
Screenshot_20200804-162722_Chrome

### Scaling, landscape
Screenshot_20200804-162741_Chrome

ingridpierre:
I agree with Akira that this appears to be an overall improvement. With more recent research on the ways different folks have personal device zoom/customize their viewports, we aren't too worried to miss that right edge.

I think this one is gtg.

-----
Note with this change, I've added a separate ticket for addressing some smal issues I observed with the settings pages without scaling:

https://app.asana.com/0/1148853526253420/1187547356580673/f

PR #773, opened 2020-07-29
Sort incoming vehicles by departure time

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1173324653606105

We now sort vehicles in the incoming box, according to the departure time of their first trip on the new route.

### DIFF
--- a/lib/realtime/vehicles.ex
+++ b/lib/realtime/vehicles.ex
@@ -18,11 +18,19 @@ defmodule Realtime.Vehicles do
     # Includes blocks that are scheduled to be pulling out
     active_blocks_by_date = Schedule.active_blocks(now, now)
 
+    date_by_block_id =
+      Schedule.active_blocks(now, in_fifteen_minutes)
+      |> Enum.flat_map(fn {date, blocks} ->
+        Enum.map(blocks, &{&1.id, date})
+      end)
+      |> Map.new()
+
     result =
       group_by_route_with_blocks(
         ungrouped_vehicles,
         incoming_trips,
         active_blocks_by_date,
+        date_by_block_id,
         now
       )
 
@@ -44,6 +52,7 @@ defmodule Realtime.Vehicles do
           [Vehicle.t()],
           [Trip.t()],
           %{Date.t() => [Block.t()]},
+          %{Block.id() => Date.t()},
           Util.Time.timestamp()
         ) ::
           Route.by_id([VehicleOrGhost.t()])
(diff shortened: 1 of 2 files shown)


phildarnowsky:
@ahakuta /cc @skyqrose  Regarding testing: one thing I found helpful was to edit the file `lib/realtime/vehicles.ex`, and on line 14, where it says: `in_fifteen_minutes = now + 15 * 60`, change that `15` to `150`. If you do that and restart the server, Skate will consider vehicles incoming if they're starting a new route within 150 minutes rather than the usual 15.

The advantage of doing this is that it artificially inflates the number of vehicles Skate considers incoming, so where without that tweak there might be 1 or 2 vehicles in an incoming box, now there might be 4 or 5. That makes checking the start times of those vehicles on the route, and verifying that they are in the correct order, is a much more meaningful test.

lib/realtime/vehicles.ex
```diff
@@ -57,6 +72,22 @@ defmodule Realtime.Vehicles do
...
+  @spec sort_incoming_vehicles_and_ghosts(Route.id(), [Vehicle.t() | Ghost.t()], [Trip.t()]) :: [
+          Vehicle.t() | Ghost.t()
+        ]
+  defp sort_incoming_vehicles_and_ghosts(route_id, vehicles_and_ghosts, incoming_trips) do
+    vehicles_and_ghosts
+    |> Enum.sort_by(fn vehicle_or_ghost ->
+      incoming_trip =
+        incoming_trips
+        |> Enum.find(fn trip ->
+          trip.block_id == vehicle_or_ghost.block_id && trip.route_id == route_id
```
skyqrose, line 84: What if a block is `[route 71, route 73, route 71]`, and a bus on the 2nd trip incoming into the 3rd trip. This could select the 1st trip instead of the 3rd, and give the wrong start time.

lib/realtime/vehicles.ex
```diff
@@ -15,39 +15,54 @@ defmodule Realtime.Vehicles do
...
         ungrouped_vehicles,
-        incoming_blocks_by_route,
+        incoming_trips,
         active_and_incoming_blocks_by_date,
         now
       ) do
     ghosts = Ghost.ghosts(active_and_incoming_blocks_by_date, ungrouped_vehicles, now)
     vehicles_and_ghosts = ghosts ++ ungrouped_vehicles
 
     incoming_from_another_route =
```
skyqrose, line 59: The `incoming_from_another_route` also needs to be sorted with the laying over for which one will leave first.

lib/realtime/vehicles.ex
phildarnowsky: Conceptually it would be simple to further filter trips by time, rejecting those that started in the past. I've been thinking about it though and here's a problem: our concept of "now" here is a Unix timestamp. The trip start/end times are given as GTFS times-of-day. There are multiple representations of any given timestamp in time-of-day format, to disambiguate you need a service date.

What I'm thinking about doing this is assuming that if the current wall-clock time is between midnight and, say, 5 AM, the service date is calendar yesterday. Otherwise it's calendar today.

Do you see anything wrong with that approach?

phildarnowsky: Ah, I'm thinking of subway service, where there is a definite period every day without any service at all.

lib/realtime/vehicles.ex
phildarnowsky: @skyqrose I'm not clear on how laying-over vehicles are involved here: they aren't displayed in the incoming box among  incoming vehicles, so their order wouldn't affect anything about how the incoming vehicles are displayed.

lib/realtime/vehicles.ex
```diff
@@ -11,43 +11,59 @@ defmodule Realtime.Vehicles do
   def group_by_route(ungrouped_vehicles) do
     now = Util.Time.now()
     in_ten_minutes = now + 10 * 60
-    in_fifteen_minutes = now + 15 * 60
+    # TODO: don't merge this, even though it would be sort of funny.
+    in_fifteen_minutes = now + 150 * 60
```
skyqrose, line 15: Adding a comment here so it's easier to remember to remove this.

lib/realtime/vehicles.ex
skyqrose: Oops. `s/laying over/pulling out`. The incoming box shows two kinds of buses. Buses that are incoming from another route, and buses that are pulling out on that route. Both those categories should be sorted together.

(Sorry, I managed to have important factual errors in both my substantive comments 🤦 )

ahakuta:
This looks good to me @phildarnowsky! 👍

phildarnowsky:
@skyqrose @ahakuta This is ready for another look: it should sort correctly in the case when we have both interlining and pulling-out buses on a route.

lib/realtime/vehicles.ex
```diff
@@ -44,48 +52,87 @@ defmodule Realtime.Vehicles do
...
+          incoming_trip_start_timestamp =
+            Util.Time.timestamp_for_time_of_day(incoming_trip.start_time, block_date)
+
+          {vehicle_or_ghost, incoming_trip_start_timestamp}
+        end
+      )
+
+    pulling_out_start_timestamps =
+      Map.new(pulling_out_vehicles_and_ghosts, fn vehicle_or_ghost ->
+        {vehicle_or_ghost, vehicle_or_ghost.layover_departure_time}
```
skyqrose, line 126: It feels a little weird to have this big object as a map key. 3 options:
1. It's not a big deal, leave it.
2. Use the vehicle id as a key instead.
3. Instead of a map, make it a list with elements [{vehicle, time}], then sort the list, then map over it to drop the times.

phildarnowsky: I'm inclined to leave it, I don't think it's a big deal to use a whole object as a hash key when it's convenient.

phildarnowsky:
@skyqrose I extended the one test to check pullouts as well.

ahakuta:
Signing off on this, thanks @phildarnowsky!

PR #774, opened 2020-07-29
Don't close VPP on swipe right

phildarnowsky:
Ticket: https://app.asana.com/0/1148853526253426/1185856327874259

There are other ways to close the VPP and this is believed to be the
cause of apparently-random closures of the VPP.


ahakuta:
lgtm on dev, thanks @phildarnowsky

PR #779, opened 2020-07-30
make blocks a struct instead of a list

skyqrose:
Refactor in preparation for [Detect pullouts dynamically based on the duration of the pullout, instead of a fixed threshold before the trip](https://app.asana.com/0/1148853526253426/1182075650453425)
To do that task, I'm planning on setting `block.start_time` to include the nonrevenue time, but that means that block needs to be a struct with a `start_time` field instead of calculating it from revenue trips.

This PR should have no user-facing changes. The `start_time` is still calculated as the start of the first revenue trip. I'll mix in the nonrevenue trips later.

### DIFF
--- a/lib/realtime/timepoint_status.ex
+++ b/lib/realtime/timepoint_status.ex
@@ -144,16 +144,12 @@ defmodule Realtime.TimepointStatus do
     nil
   end
 
-  def scheduled_location([], _now) do
-    nil
-  end
-
   def scheduled_location(block, now) do
     now_time_of_day =
       Util.Time.next_time_of_day_for_timestamp_after(
         now,
         # Allow a little wiggle room in case a bus appears just before its block starts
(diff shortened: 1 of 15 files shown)


lib/realtime/timepoint_status.ex
```diff
@@ -179,7 +175,8 @@ defmodule Realtime.TimepointStatus do
               run_id: trip.run_id,
               time_since_trip_start_time: now_time_of_day - trip.start_time,
               headsign: trip.headsign,
-              via_variant: RoutePattern.via_variant(trip.route_pattern_id),
+              via_variant:
+                trip.route_pattern_id && RoutePattern.via_variant(trip.route_pattern_id),
```
phildarnowsky, line 179: @skyqrose this would appear to be a separate fix, is that correct?

skyqrose: Yeah. It just popped up with the changed test data.

PR #783, opened 2020-08-03
Show pulling out ghosts buses based on scheduled pullout times instead of 10min threshhold

skyqrose:
Asana Task: [Detect pullouts dynamically based on the duration of the pullout, instead of a fixed threshold before the trip](https://app.asana.com/0/1148853526253426/1182075650453425)

~Builds on top of #779. Wait until that merges to review this.~

### Preventing false negatives:
Screen Shot 2020-08-03 at 16 36 38
This screenshot was taken at 16:36:38.
The route 23 pullout is scheduled from 16:35 to 16:48. The route 77 pullout is scheduled from 16:36 to 16:54. In both cases, revenue service was more than 10m away, so prod showed nothing in either incoming box.

### Preventing false positives:
run 123-1150 is scheduled to pull out from 17:01 to 17:05
At 16:55, prod showed a ghost bus for it:
Screen Shot 2020-08-03 at 16 55 10
This PR didn't show a ghost bus:
Screen Shot 2020-08-03 at 16 55 18

### DIFF
--- a/lib/schedule/block.ex
+++ b/lib/schedule/block.ex
@@ -1,5 +1,6 @@
 defmodule Schedule.Block do
   alias Schedule.Gtfs.Service
+  alias Schedule.Hastus
   alias Schedule.Trip
 
   @type id :: String.t()
@@ -13,6 +14,7 @@ defmodule Schedule.Block do
   @type t :: %__MODULE__{
           id: id(),
           service_id: Service.id(),
+          schedule_id: Hastus.Schedule.id() | nil,
           start_time: Util.Time.time_of_day(),
           end_time: Util.Time.time_of_day(),
           # only revenue trips. always nonempty
@@ -30,6 +32,7 @@ defmodule Schedule.Block do
   defstruct [
     :id,
     :service_id,
+    :schedule_id,
     :start_time,
     :end_time,
     :trips
@@ -55,6 +58,7 @@ defmodule Schedule.Block do
     %__MODULE__{
       id: first_trip.block_id,
       service_id: first_trip.service_id,
+      schedule_id: first_trip.schedule_id,
       start_time: first_trip.start_time,
       end_time: List.last(trips).end_time,
       trips: trips
--- a/lib/schedule/block.ex
+++ b/lib/schedule/block.ex
@@ -52,7 +52,7 @@ defmodule Schedule.Block do
     |> Helpers.map_values(fn trips ->
       block_id = List.first(trips).block_id
       schedule_id = List.first(trips).schedule_id
-      deadheads = nonrevenue_trips[{block_id, schedule_id}] || []
+      deadheads = Map.get(nonrevenue_trips, {block_id, schedule_id}, [])
       block_from_trips(trips, deadheads)
     end)
   end
@@ -61,7 +61,7 @@ defmodule Schedule.Block do
   def block_from_trips(revenue_trips, nonrevenue_trips \\ []) do
     revenue_trips = Enum.sort_by(revenue_trips, & &1.start_time)
     nonrevenue_trips = Enum.sort_by(nonrevenue_trips, & &1.start_time)
-    [first_trip | _] = revenue_trips
+    first_trip = List.first(revenue_trips)
 
     start_time =
       if nonrevenue_trips == [] do
--- a/lib/schedule/block.ex
+++ b/lib/schedule/block.ex
@@ -40,28 +40,56 @@ defmodule Schedule.Block do
 
   @spec blocks_from_trips([Trip.t()]) :: by_id()
   def blocks_from_trips(trips) do
-    trips
-    |> group_trips_by_block()
-    |> Helpers.map_values(&block_from_trips/1)
-  end
+    {revenue_trips, nonrevenue_trips} =
+      Enum.split_with(trips, fn trip -> trip.route_id != nil end)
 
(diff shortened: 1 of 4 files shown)


ahakuta:
I love this improvement, and I really appreciate the clear documentation of testing @skyqrose 👏  🙌

ahakuta:
@skyqrose I did some additional digging into the scenarios you captured in your screenshots looking at Swiftly playback, and one thing I noticed that I think we should keep in mind, not as additional scope for this ticket, but as a potential follow-on improvement:

For the **Preventing false positive** scenario you captured, the new logic works well, certainly better than production. The bus on run 1150 (vehicle 1899) didn't actually pull out/get assigned (according to Swiftly anyways) to run 1150 until 5:07PM, two minutes after its scheduled first trip, so we would have showed a ghost bus for that bus for 6 minutes, between 5:01PM (start of pullout time) and 5:07PM (when it actually started run 1150). On prod, we showed that bus as a ghost between 4:55PM and 5:07PM, or 12 minutes, so twice as long. 

For the **Preventing false negatives** scenario you captured, the new logic works, but in one scenario (for the route 77) has an inverse effect of arguably unnecessarily actually showing a ghost in the incoming box for _longer_, while in the other scenario for route 23, it worked well. 

- For the route 77 example the bus (vehicle 1981) pulled out right on time at 4:53PM, but it didn't appear in Swiftly to be assigned to run 1159 until 4:53PM because it had been sitting at the Mt. Auburn pullout, which is right at the start of the route 77. In other words, it was sitting (logged out) right next to the start point of its route.  

- For the route 23 pullout captured at this same time, the new logic in this PR worked well- the bus left the yard a few minutes late, and it didn't end up starting its run until 4:53PM, 5 minutes late. So this new PR  would have shown that bus in the inbound box was a ghost for 3 more minutes than we did on prod, and given the inspector managing this route an early heads up that this bus was likely to be a bit late. 

I think this PR is a solid improvement- we are working with schedule data now rather than the 10 minute arbitrary threshold. I'm left wondering though, how might we improve the underlying logic even more in order to minimize the noise introduced in the scenario with route 77? Would we need crew schedules in order to know that the operator on bus 1981 was sitting by the start of route 77 and ready to go? Are there any other ways we might know that in advance?

skyqrose:
@phildarnowsky #779 is merged, so this is ready to be reviewed now.

skyqrose:
@ahakuta if I understand correctly, you're saying the 77 showed up as a ghost because nobody was pulling out on it, but the inspector had already handled the situation by giving that run to a bus that was already at the terminal (and so didn't need to pull out)? If the bus isn't logged in, we have no way in the data to distinguish between a missing pullout that's unhandled and a missing pullout where the inspector has a plan, so I think showing the ghost bus is still the right thing to do.

lib/schedule/block.ex
```diff
@@ -40,28 +40,56 @@ defmodule Schedule.Block do
...
+    end)
   end
 
-  @spec block_from_trips([Trip.t()]) :: t()
-  def block_from_trips([first_trip | _] = trips) do
+  @spec block_from_trips([Trip.t()], [Trip.t()]) :: t()
+  def block_from_trips(revenue_trips, nonrevenue_trips \\ []) do
+    revenue_trips = Enum.sort_by(revenue_trips, & &1.start_time)
+    nonrevenue_trips = Enum.sort_by(nonrevenue_trips, & &1.start_time)
+    [first_trip | _] = revenue_trips
```
phildarnowsky, line 64: This is a line that I think is clearer _not_ using pattern matching: `first_trip = List.first(revenue_trips)`.

skyqrose:
I guess we could solve the 77 problem if we had a separate system (not bus logins) to know who's doing what. Maybe HASTUS Daily (if it records which bus will do a run, or even if it just says "the inspector has confirmed that this run will happen"), or Inspector input into Glides-like trainsheets. But both of those would be such huge lifts, and this is such a small reason to do them, that I don't think it's worth worrying about.

ahakuta:
> @ahakuta if I understand correctly, you're saying the 77 showed up as a ghost because nobody was pulling out on it, but the inspector had already handled the situation by giving that run to a bus that was already at the terminal (and so didn't need to pull out)? If the bus isn't logged in, we have no way in the data to distinguish between a missing pullout that's unhandled and a missing pullout where the inspector has a plan, so I think showing the ghost bus is still the right thing to do.

Sorry for the slow response on this @skyqrose. What I was noting above was that with the route 77 example, with this new PR we would show a ghost bus for longer than we (used to) production because we know the pullout time for that bus is longer than our old default of 10 minutes. So, as soon as the pullout time matched the current time, we started to show a ghost because nobody was logged into the appropriate incoming run. In reality, the bus that was going to make that run was sitting right at the start of the route, ready to go, it just wasn't logged into the right run until the very last minute. 

I think the problem I'm bringing up ultimately isn't related to pullout durations, although it does mean that in some circumstances like this one with the 77, it may be slightly worse in the sense that we may be showing ghosts in incoming for slightly longer than we used to. The overarching problem I think is, **how might we avoid showing ghosts in the incoming box for buses that are in fact incoming to a given route?**

This is now making me remember that @ldnash  recently mentioned to me that there is such a thing as a run number for a pullout for a bus to get to the start of a route, and that these run numbers are actually used by operators. @ldnash is that correct? If so, that seems like it could be a potential way to solve this?

PR #806, opened 2020-08-17
Phoenix channel/socket for notifications

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1187690663845801

This adds infrastructure for the Phoenix socket and channel over which notifications will flow. The intent is to merge this with https://github.com/mbta/skate/pull/803 when both are ready, which will cause new block waivers to be forwarded from busloc clear through to the browser. The browser won't do anything with notifications yet except log them, but that's coming.


phildarnowsky:
@skyqrose You may observe that, rather than have a separate topic per set of routes (e.g. `notifications:1,77,83`) I went with a single `notifications` topic for everything, meaning every client will receive every notification.

Notifications seem as though they'll be sparse, perhaps several dozen per day once we add the new types, so filtering at the client end is feasible. Doing so greatly simplifies the issue I've been discussing with Akira and Ingrid, where we want to ensure that if you the inspector have routes X and Y open, and a notification comes through that affects both X and Y, you only see one notification, not two.

lib/skate_web/channels/notifications_channel.ex
```diff
@@ -0,0 +1,23 @@
...
+      :ok = push(socket, "notification", %{data: message})
+      {:noreply, socket}
+    else
+      :ok = push(socket, "auth_expired", %{})
+      {:stop, :normal, socket}
+    end
+  end
+
+  @impl true
+  def join("notifications", _message, socket) do
```
skyqrose, line 19: This doesn't have a way to specify which routes to subscribe to. If the intent is to push that off to keep this PR smaller, let's add another ticket ahead of [Create front end socket](https://app.asana.com/0/1148853526253426/1187690663845805), so that ticket can stay focused on the frontend half.

assets/src/components/routeLadders.tsx
```diff
@@ -15,6 +18,14 @@ const RouteLadders = ({
   timepointsByRouteId,
   selectedVehicleId,
 }: Props) => {
+  const { socket } = useContext(SocketContext)
+  useChannel<void | null>({
+    socket,
+    topic: "notifications",
+    event: "notification",
+    parser: console.log,
```
skyqrose, line 26: Are we okay with logging on clients' consoles like this without a flag? I'm thinking yes it's fine, but want to double check.

lib/skate_web/channels/notifications_channel.ex
phildarnowsky: @skyqrose see my comment on the PR above--essentially, while you were out, we clarified part of the UX related to waivers that affect multiple routes, and as a consequence I decided it would make more sense to have all notifications go through a common channel, such that all clients get all notifications, and it's up to the frontend to decide whether to display a notification or not.

assets/src/components/routeLadders.tsx
phildarnowsky: I think it would be OK, @ahakuta do you have any objection? Clients wouldn't see this unless they open the developer tools, and my thinking is that if you open that, you're probably fine with the occasional look behind the scenes.

phildarnowsky:
@skyqrose regarding order, I'm thinking yes, as you say, #803 merged first and then this PR merged on top of it. But they're decoupled to the point that it should be possible to do whichever order is more convenient.

lib/skate_web/channels/notifications_channel.ex
skyqrose: Ah, I missed that comment. 👍

phildarnowsky:
@skyqrose I feel like the changes I made in the last few commits are substantial enough to merit more review, would you mind looking at them?

lib/notifications/notification_server.ex
```diff
@@ -42,6 +49,8 @@ defmodule Notifications.NotificationServer do
           inspect(new_notifications)
```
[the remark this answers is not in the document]
phildarnowsky: I'd like to keep them around a little while at least until this code has had a little time to break in.

PR #809, opened 2020-08-21
Parse and log notifications on the frontend

skyqrose:
Asana Task: [Create front end socket](https://app.asana.com/0/1148853526253426/1187690663845805)

There's nothing to demo or screenshot here. It logs the data to the client's console in basically the same way as it did before this PR.

### DIFF
--- a/assets/src/hooks/useNotifications.ts
+++ b/assets/src/hooks/useNotifications.ts
@@ -1,0 +1,52 @@
+import { Channel, Socket } from "phoenix"
+import { useContext, useEffect } from "react"
+import { SocketContext } from "../contexts/socketContext"
+import { StateDispatchContext } from "../contexts/stateDispatchContext"
+import { reload } from "../models/browser"
+import { notificationsFromData } from "../models/notificationData"
+import { Notification } from "../realtime.d"
+
+export const useNotifications = (
+  handleNewNotification: (notification: Notification) => void
+): void => {
+  const { socket }: { socket: Socket | undefined } = useContext(SocketContext)
+  const topic: string = "notifications"
+  const event: string = "notifications"
+  const [{ selectedRouteIds }] = useContext(StateDispatchContext)
+
+  useEffect(() => {
+    let channel: Channel | undefined
+
+    if (socket !== undefined) {
+      channel = socket.channel(topic)
+      channel.on(event, ({ data: data }) => {
+        const notifications: Notification[] = notificationsFromData(data)
+        notifications.forEach((notification) => {
+          if (
+            notification.routeIds.some((routeId) =>
+              selectedRouteIds.includes(routeId)
+            )
+          ) {
+            handleNewNotification(notification)
+          }
+        })
+      })
+      channel
+        .join()
+        .receive("error", ({ reason }) =>
+          // tslint:disable-next-line: no-console
+          console.error(`joining topic ${topic} failed`, reason)
+        )
+        .receive("timeout", () => {
+          reload(true)
+        })
+    }
+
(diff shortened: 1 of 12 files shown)


assets/src/models/notificationData.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const notificationsFromData = (
+  notificationsData: NotificationData[]
+): Notification[] => notificationsData.map(notificationFromData)
+
+const notificationFromData = (
+  notificationData: NotificationData,
+  index: number
+): Notification => ({
+  id: Date.now() + index,
```
phildarnowsky, line 21: @skyqrose I've thought this through and it's a safe way to generate these indices, but I did have to think it through, so let's maybe document the reasoning: that busloc pushes new block waivers to Skate only every few seconds, meaning that a collision of IDs here could only happen if thousands of block waivers are generated per second for multiple seconds running.

assets/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,52 @@
...
+): void => {
+  const { socket }: { socket: Socket | undefined } = useContext(SocketContext)
+  const topic: string = "notifications"
+  const event: string = "notifications"
+  const [{ selectedRouteIds }] = useContext(StateDispatchContext)
+
+  useEffect(() => {
+    let channel: Channel | undefined
+
+    if (socket !== undefined) {
```
phildarnowsky, line 20: @skyqrose I'm assuming that you're rolling a fresh implementation of this because the existing `useChannel` isn't suitable for some reason?

assets/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,52 @@
...
+      channel = socket.channel(topic)
+      channel.on(event, ({ data: data }) => {
+        const notifications: Notification[] = notificationsFromData(data)
```
phildarnowsky, line 23: @skyqrose in the old implementation it was possible for the data to be `undefined` (i.e. when a client first joined). Is that accounted for here?

assets/src/hooks/useNotifications.ts
skyqrose: Yeah. `useChannel` is more meant for "I'll return the most up-to-date state of the world". It keeps its own state under the hood. For notifications, the caller keeps track of the state, and the channel pushes the new notifications incrementally through a callback. (The notifications component needs to keep the state so it can remove read cards.)

Also because of this difference, the notifications channel doesn't return anything on join which would break `useChannel` when it tries to parse the response.

If you think it's worth it, we could have a single channel implementation like this:
```ts
useChannelCallback = (socket, topic, event, callback) => {
  useEffect(() => {
    // only implementation of channel.join
    channel.on(event, callback)
  })
}

useChannel = (socket, topic, event, parser) => {
  [state, setState] = useState()
  useChannelCallback(socket, topic, event, parse and setState)
  return state
}

useNotifications = (callback) => {
  useChannelCallback(socket, "notifications", "notifications", parse and callback)
}
```
but we'd have to either return an empty list (instead of returning nothing) when joining the notification channel, or let `useChannelCallback` decide whether to try to parse the join response.

assets/src/hooks/useNotifications.ts
skyqrose: When the backend notifications channel doesn't send a response to joining the chanel, the default response is `{}`. The old version logged undefined, because it was doing `console.log({}.data)`. `useNotifications()` avoids that by not giving a `channel.receive("ok",)` handler to the join response like `useChannel()` did.

PR #819, opened 2020-08-26
Set up Postgres database

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1184959932880359

This includes basic setup and documentation for dev, test, and prod
environments.


config/config.exs
```diff
@@ -88,6 +90,14 @@ config :skate, SkateWeb.AuthManager,
   issuer: "skate",
   secret_key: nil
 
+config :skate, Skate.Repo,
+  database: "skate_dev",
+  username: System.get_env("POSTGRES_USERNAME", System.get_env("USER")),
+  password: System.get_env("POSTGRES_PASSWORD", ""),
+  hostname: System.get_env("POSTGRES_HOSTNAME", "localhost"),
+  port: System.get_env("POSTGRES_PORT", "5432") |> String.to_integer(),
```
skyqrose, line 98: Will these be read at run time or build time? I thought that this would need to do the `{:system, "POSTGRES_USERNAME"}` thing like the other env variables in this file, but I see Glides' database config calls `System.get_env()`, so now I'm not sure. If it was `{:system,}`, would that even work? Would it be able to remove the need for the `add_prod_credentials` function? Should prod get these from secrets manager instead of ENV variables?

test/skate/repo_test.exs
```diff
@@ -0,0 +1,48 @@
...
+      original_username = System.get_env("POSTGRES_USERNAME", System.get_env("USER"))
+
+      System.put_env("POSTGRES_HOSTNAME", "db_server_hostname")
+      System.put_env("POSTGRES_USERNAME", "my_username")
+      System.put_env("POSTGRES_PORT", "6789")
+
+      on_exit(fn ->
+        System.put_env("POSTGRES_HOSTNAME", original_hostname)
+        System.put_env("POSTGRES_USERNAME", original_username)
+        System.put_env("POSTGRES_PORT", original_port)
```
skyqrose, line 17: If the defaults above get used, this doesn't perfectly restore the environment. You'd have to detect that it was unset and call `System.delete_env()` instead.

skyqrose:
I just ran it locally and it works, though I also had to run `env MIX_ENV=test mix ecto.create` (and I'd probably also have to run the migrations too if we had any yet).

It seems like the ``env `cat .env``` part isn't necessary if you're not setting any of the variables locally. Everything else in `.env` is only needed when starting the application.

config/config.exs
phildarnowsky: I think we can actually use `:system` now, it was causing errors in an earlier iteration of the code but it looks to be OK now. Will push update assuming CI is good. One thing about that is that henceforth we'll have to do `env \`cat .env\` mix test`.

We do still need the `add_prod_credentials` though, and (most of) the prod credentials do belong in environment variables. As I discussed with Ian, and feel free to correct me if I'm wrong @ianwestcott, of the credentials the only secret is the password. But that password's not a secret that lives in Secrets Manager: it's a temporary password that we fetch through the RDS API (that's what that `auth_token_fn` is about in `add_prod_credentials`), and fetching that is really `add_prod_credentials`' main job.

phildarnowsky: @skyqrose actually, `:system` breaks things in CI:

```
** (Protocol.UndefinedError) protocol List.Chars not implemented for {:system, "POSTGRES_HOSTNAME"} of type Tuple
```

(e.g. https://semaphoreci.com/mbta/skate/branches/pwd-postgres-docs/builds/39)

skyqrose (approved):

PR #828, opened 2020-08-31
Notification Cards

skyqrose:
Asana Task: [Notifications cards](https://app.asana.com/0/1148853526253426/1187690663845807)

Screen Shot 2020-08-31 at 11 04 30

And here it is behind the VPP:

Screen Shot 2020-08-31 at 11 04 56

The hover states work. The close button removes the notification, but clicking the body of the notification does nothing.

Opening the VPP, showing the specific route that an accident happened on, and showing the operator will depend on getting that information out of Transit Master

Animation, showing the modal if we can't find the vehicle, and fullstory tracking are all separate tickets and not part of this.

### DIFF
--- a/assets/src/components/notifications.tsx
+++ b/assets/src/components/notifications.tsx
@@ -44,7 +44,7 @@ const NotificationCard = ({
   currentTime: Date
 }) => (
   <div className="m-notifications__card">
-    <button className="m-notifications__card-info" onClick={() => {}}>
+    <button className="m-notifications__card-info">
       <div className="m-notification__title-row">
         <div className="m-notification__title">
           {title(notification.reason)}
--- a/assets/src/components/notifications.tsx
+++ b/assets/src/components/notifications.tsx
@@ -7,9 +7,6 @@ import { formattedTimeDiff, now } from "../util/dateTime"
 import PropertiesList from "./propertiesList"
 
 export const Notifications = () => {
-  const [currentTime, setCurrentTime] = useState(now())
-  useInterval(() => setCurrentTime(now()), 1000)
-
   const [notifications, setNotifications] = useState<Notification[]>([])
   const addNotification = (notification: Notification): void => {
     if (featureIsEnabled("notifications")) {
@@ -20,6 +17,10 @@ export const Notifications = () => {
     setNotifications((previous) => previous.filter((n) => n.id !== id))
   }
   useNotifications(addNotification)
+
+  const [currentTime, setCurrentTime] = useState(now())
+  useInterval(() => setCurrentTime(now()), 1000)
+
   return (
     <div className="m-notifications">
       {notifications.map((notification) => (
@@ -34,7 +35,7 @@ export const Notifications = () => {
   )
 }
 
-const NotificationCard = ({
+export const NotificationCard = ({
   notification,
   remove,
   currentTime,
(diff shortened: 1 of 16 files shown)


assets/src/components/notifications.tsx
```diff
@@ -1,15 +1,109 @@
...
+  notification,
+  remove,
+  currentTime,
+}: {
+  notification: Notification
+  remove: (id: number) => void
+  currentTime: Date
+}) => (
+  <div className="m-notifications__card">
+    <button className="m-notifications__card-info" onClick={() => {}}>
```
phildarnowsky, line 47: @skyqrose may as well just remove the `onClick` handler then, unless I'm missing something?

assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
```diff
@@ -13,13 +14,12 @@ import {
...
 import { Route, RouteId, Shape } from "../../schedule"
 import Map from "../map"
-import PropertiesList from "../propertiesList"
+import PropertiesList, { vehicleProperties } from "../propertiesList"
 import BlockWaiverList from "./blockWaiverList"
 import CrowdingDiagram from "./crowdingDiagram"
 import Header from "./header"
 import HeadwayDiagram from "./headwayDiagram"
 import TabPanels, { TabMode } from "./tabPanels"
-import { useNearestIntersection } from "../../hooks/useNearestIntersection"
```
phildarnowsky, line 22: @skyqrose what's up with this coming out?

phildarnowsky:
Approval subject to CI passing of course. Some of what's not passing in Semaphore looks like you probably forgot to update some snaps, but there are some other tests that look like they may expose something worth investigating.

assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
skyqrose: The imports were out of order. It was inserted above. Not sure why CI didn't catch it before.

skyqrose:
It's not just forgetting to update the snapshots. I totally forgot to update the tests for the new functionality. There's a bunch of new test code, so I'd like another 👀 on that.

Codecov doesn't like that I don't test the string for every reason's description, but I got all the interesting cases. Is it worth writing tests that go through all those, or should I overrule codecov?

ahakuta:
@skyqrose the cards look great on desktop! I'm attaching a screenshot of how they look though when I simulated a Moto G4 using the Chrome user agent switcher- is there a way to ensure that these cards fit on all mobile screen sizes?
Screen Shot 2020-09-01 at 10 20 06 PM

skyqrose:
The cards now shrink to fit small screens:
Screen Shot 2020-09-03 at 14 54 01

phildarnowsky:
@skyqrose regarding covering the various statuses, my preference is to write test code similar to the bit starting https://github.com/mbta/skate/blob/0f83883324bef79dd4099977a7fee6fcb723aa65/test/notifications/notification_server_test.exs#L96: that is, it iterates over a collection to test each case, without having to write a lot of repetitive tests by hand.

assets/src/realtime.d.ts
```diff
@@ -65,6 +65,7 @@ export type NotificationReason =
   | "adjusted"
   | "operator_error"
   | "traffic"
+  | string
```
phildarnowsky, line 68: @skyqrose I'm not clear on why we're adding support for arbitrary reason strings: we control both ends of the pipe, so we can ensure that only recognized reasons are sent.

phildarnowsky (commented):
New stuff mostly looks good, barring my one suggestion and one question.

PR #829, opened 2020-08-31
Notification Animation

skyqrose:
Asana Task: [Notification animation](https://app.asana.com/0/1148853526253426/1191003352789117)

Kapture 2020-08-31 at 16 54 04

Animating closing the notifications would be a little harder, so I skipped that. (You'd have to keep the component around until the animation finishes, and you might also want to animate the other notifications sliding into position.)

~Built on top of #828~ Ready for review. There was an indentation change, so view it with whitespace turned off.

### DIFF
--- a/assets/src/components/notifications.tsx
+++ b/assets/src/components/notifications.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useEffect, useState } from "react"
 import useInterval from "../hooks/useInterval"
 import { useNotifications } from "../hooks/useNotifications"
 import featureIsEnabled from "../laboratoryFeatures"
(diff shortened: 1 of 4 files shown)


assets/src/components/notifications.tsx
```diff
@@ -43,37 +43,49 @@ export const NotificationCard = ({
...
   currentTime: Date
-}) => (
-  <div className="m-notifications__card">
-    <button className="m-notifications__card-info">
-      <div className="m-notification__title-row">
-        <div className="m-notification__title">
-          {title(notification.reason)}
+}) => {
+  const [isNew, setIsNew] = useState<boolean>(true)
+  useEffect(() => {
```
phildarnowsky, line 48: @skyqrose perhaps `useLayoutEffect` (https://reactjs.org/docs/hooks-reference.html#uselayouteffect) would be more appropriate here. From the documentation:

> However, not all effects can be deferred. For example, a DOM mutation that is visible to the user must fire synchronously before the next paint so that the user does not perceive a visual inconsistency.

skyqrose: That certainly would make sense, but I tried it, and it made it consistently not work. I think it's backwards from what we want, in that it forces DOM>effect>paint, rather than DOM>paint>effect. (We want the paint to be first so that the css transition sees the starting state.)

(Without the timeout, `useEffect` works about half the time.)

phildarnowsky (approved):
LGTM pending coverage.

PR #835, opened 2020-09-08
fullstory event for delivering a notification

skyqrose:
Asana Task:
[Test Fullstory custom event tracking using dummy cards](https://app.asana.com/0/1148853526253426/1191912668336954)
part of [Notifications tracking?](https://app.asana.com/0/1148853526253426/1191223285175310)

This sends "Notification delivered" on every notification (but doesn't do the "Notification stacked" or "Notification VPP unavailable" that the second task asks for.)

I verified with https://github.com/mbta/skate/commit/fc0bcaccf604534511d410d50ed759929c8e799f (which is currently deployed to dev) that `FS.event()` is getting called, and I think that it's getting sent to fullstory, but it's a little opaque how fullstory does its stuff, so @ahakuta , can you check that the event made it to the fullstory dashboard? (It'd be from me on firefox this afternoon)

### DIFF
--- a/assets/src/components/notifications.tsx
+++ b/assets/src/components/notifications.tsx
@@ -6,7 +6,7 @@ import { Notification, NotificationReason } from "../realtime.d"
 import { formattedTimeDiff, now } from "../util/dateTime"
 import PropertiesList from "./propertiesList"
 
-const fullstoryEvent = (): void => {
+const deliveryFullstoryEvent = (): void => {
   if (window.FS && window.username) {
     window.FS.event("Notification delivered")
   }
@@ -16,7 +16,7 @@ export const Notifications = () => {
   const [notifications, setNotifications] = useState<Notification[]>([])
   const addNotification = (notification: Notification): void => {
     if (featureIsEnabled("notifications")) {
-      fullstoryEvent()
+      deliveryFullstoryEvent()
       setNotifications((previous) => [...previous, notification])
     }
   }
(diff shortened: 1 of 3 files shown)


assets/src/components/notifications.tsx
```diff
@@ -6,10 +6,17 @@ import { Notification, NotificationReason } from "../realtime.d"
 import { formattedTimeDiff, now } from "../util/dateTime"
 import PropertiesList from "./propertiesList"
 
+const fullstoryEvent = (): void => {
```
phildarnowsky, line 9: @skyqrose I would suggest naming this `deliveryFullstoryEvent`. That makes it clear at a glance what circumstances trigger this event (notification delivered, i.e. successfully rendered).

assets/src/components/notifications.tsx
```diff
@@ -6,10 +6,17 @@ import { Notification, NotificationReason } from "../realtime.d"
...
+  if (window.FS && window.username) {
```
phildarnowsky, line 10: I expect we're also going to be doing a lot of this `if (window.FS && ...` in the near future, do you think we should extract a helper now or is that maybe premature?

phildarnowsky (changes requested):
Minor changes noted.

assets/src/components/notifications.tsx
skyqrose: Let's wait to see how the next use of fullstory works to see what would be a way to DRY this.

phildarnowsky (approved):

PR #843, opened 2020-09-11
Force geonames to use TLS v1.2

skyqrose:
This is a guess as to what might fix [🐛Investigate Geonames access failures](https://app.asana.com/0/1148853526253426/1193058913335769)

The situation is most succinctly described here: https://campezzi.ghost.io/httpoison-ssl-connection-closed/

~I'll deploy to dev, and if that is the issue, then this should fix it.~
Dev only sees this error every few days, and without a specific way to reproduce it, I think we won't be able to see if this works until it's on prod where we'd be able to notice the errors stopping. We can revert it if it doesn't solve the problem on prod.

### DIFF
--- a/lib/geonames.ex
+++ b/lib/geonames.ex
@@ -30,11 +30,13 @@ defmodule Geonames do
         |> Jason.decode!(strings: :copy)
 
       response ->
-        sanitized_url = if geonames_token do
-          String.replace(url, geonames_token, "SECRET")
-        else
-          url
-        end
+        sanitized_url =
+          if geonames_token do
+            String.replace(url, geonames_token, "SECRET")
+          else
+            url
+          end
+
         Logger.warn(fn -> "Unexpected response from #{sanitized_url} : #{inspect(response)}" end)
         nil
     end
--- a/lib/geonames.ex
+++ b/lib/geonames.ex
@@ -29,7 +29,12 @@ defmodule Geonames do
         |> Jason.decode!(strings: :copy)
 
       response ->
-        Logger.warn(fn -> "Unexpected response from #{url} : #{inspect(response)}" end)
+        sanitized_url = if geonames_token do
+          String.replace(url, geonames_token, "SECRET")
+        else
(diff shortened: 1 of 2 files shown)


lib/geonames.ex
```diff
@@ -23,13 +23,21 @@ defmodule Geonames do
...
+    case HTTPoison.get(url, [], ssl: [versions: [:"tlsv1.2"]]) do
       {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
         body
         |> Jason.decode!(strings: :copy)
 
       response ->
-        Logger.warn(fn -> "Unexpected response from #{url} : #{inspect(response)}" end)
+        sanitized_url =
+          if geonames_token do
+            String.replace(url, geonames_token, "SECRET")
```
phildarnowsky, line 35: not strictly within the scope of this PR but how secret is our Geonames token anyhow? is it having been logged sufficient grounds to roll it?

lib/geonames.ex
```diff
@@ -23,7 +23,8 @@ defmodule Geonames do
         token_param
       }"
 
-    case HTTPoison.get(url) do
+    # force TLS v1.2, see https://campezzi.ghost.io/httpoison-ssl-connection-closed/
+    case HTTPoison.get(url, [], ssl: [versions: [:"tlsv1.2"]]) do
```
phildarnowsky, line 27: Should we set this as a global option on `ssl` perhaps? I wouldn't expect trouble with other places that we might be using `HTTPoison` to access third-party services, since the problematic protocol versions are deprecated and nobody should be using them. Nevertheless it would be best, if we do set this globally, to let it run on dev a while and make sure there's no uptick in errors.

lib/geonames.ex
skyqrose: I think it's probably fine.

skyqrose:
I put it on dev, and everything is still working

phildarnowsky (approved):

PR #845, opened 2020-09-14
Realtime server tracks vehicles by trip ID

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1191003352789105

The realtime server now tracks vehicles by trip ID, and allows querying on one or more trip IDs. The notification server uses this feature to augment the `Notification`s that it creates by adding data from the active vehicle, if any.

### DIFF
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -8,7 +8,7 @@ defmodule Realtime.Server do
 
   use GenServer
 
-  alias Schedule.Route
+  alias Schedule.{Route, Trip}
 
   alias Realtime.{
     Ghost,
@@ -21,7 +21,8 @@ defmodule Realtime.Server do
   @enforce_keys [:ets]
 
   defstruct ets: nil,
-            active_route_ids: []
+            active_route_ids: [],
+            active_trip_ids: []
 
   @type subscription_key ::
           {:route_id, Route.id()} | :all_shuttles | :all_vehicles | {:search, search_params()}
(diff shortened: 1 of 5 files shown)


lib/realtime/server.ex
```diff
@@ -149,32 +171,67 @@ defmodule Realtime.Server do
...
     for {route_id, vehicles_and_ghosts} <- vehicles_by_route_id do
       active_vehicles_and_ghosts = Enum.filter(vehicles_and_ghosts, &block_is_active?/1)
       _ = :ets.insert(ets, {{:route_id, route_id}, active_vehicles_and_ghosts})
     end
 
-    :ets.insert(ets, {:all_vehicles, Enum.uniq(all_vehicles(vehicles_by_route_id) ++ shuttles)})
+    all_vehicles = Enum.uniq(all_vehicles(vehicles_by_route_id) ++ shuttles)
+
+    for vehicle <- all_vehicles do
+      _ = :ets.insert(ets, {{:trip_id, vehicle.trip_id}, vehicle})
```
skyqrose, line 231: What happens if two vehicles have the same trip id? (It shouldn't happen, since overloads get an `ADDED` trip id, but I just want to make sure nothing will explode.)

skyqrose (approved):
You might need a `notification_server_test` for if a ghost is active or no vehicle is active.

lib/realtime/server.ex
phildarnowsky: `ets` has type `set`, so the last `vehicle` iterated over in that case is the one recorded. But as you say, shouldn't happen.

phildarnowsky:
@skyqrose the recent commits merit review, especially the refactoring of the notification server test.

skyqrose (approved):

PR #860, opened 2020-09-17
Clicking notification opens VPP

phildarnowsky:
When a notification is clicked, the server tries to determine what
vehicle/ghost, if any, is related to the current trip out of the trips
affected by the notification. If there is such a vehicle or ghost, we
open the VPP with it, even if the ladder for the route in question is
not currently displayed.


phildarnowsky:
@skyqrose once the tests pass, of course, Codecov is still going to be extremely angry about this PR; bringing coverage up is next on my agenda but this should be feature-complete and it Works On My Machine™.

phildarnowsky:
note to self: you forgot the Fullstory events

lib/skate_web/router.ex
```diff
@@ -70,6 +70,7 @@ defmodule SkateWeb.Router do
     get "/minischedule/run/:trip_id", MinischeduleController, :run
     get "/minischedule/block/:trip_id", MinischeduleController, :block
     get "/intersection", IntersectionController, :intersection
+    get "/vehicle_for_trips", VehicleForTripController, :show
```
skyqrose, line 73: If we want the VPP to either get updates if the vehicle on those trips changes, or get realtime updates if the bus is on a non-selected route, this would need to be a channel instead of a GET.

assets/src/state.ts
```diff
@@ -29,6 +30,7 @@ export interface State {
   selectedShuttleRunIds: RunId[] | "all"
   selectedVehicleId?: VehicleId
   settings: Settings
+  vehicleAndRouteForNotification?: VehicleOrGhostAndRoute
```
skyqrose, line 33: We've been using `state.ts` just for interface stuff, and the data lives in the state of the hook that gets it. So this would be like `selectedTripIdsForNotification`, and somewhere you would have `const vehicleAndRoute = useVehicleAndRouteForNotification(selectedTripIdsForNotification)`, where there's a `setState` or `useChannel` to save the data inside that hook.

assets/src/components/ladderPage.tsx
```diff
@@ -78,6 +82,14 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
                 route={vehicleRoute(routes, selectedVehicleOrGhost)}
               />
             )}
+            {vehicleAndRouteForNotification && (
+              <PropertiesPanel
```
skyqrose, line 86: Is it a problem that this could draw two PropertiesPanels? Should we have something like
```jsx
vehicleForTheVPP = selectedVehicleOrGhost || vehicleAndRouteForNotification
vehicleForTheVPP && <PropertiesPanel/>
```
?

assets/tests/components/notifications.test.tsx
```diff
@@ -155,6 +159,7 @@ describe("NotificationCard", () => {
         notification={notification}
         remove={jest.fn()}
         currentTime={now()}
+        fetchCurrentVehicle={() => null}
```
skyqrose, line 162: `jest.fn()` is the standard way to make empty functions for this (and then you can have a test that asserts that it's called on click).

lib/skate_web/controllers/vehicle_for_trip_controller.ex
```diff
@@ -0,0 +1,28 @@
...
+      case Server.peek_at_vehicles(trip_ids) do
+        [vehicle_or_ghost] -> vehicle_or_ghost
+        _ -> nil
+      end
+
+    route = vehicle_or_ghost && Schedule.route_by_id(vehicle_or_ghost.route_id)
+
+    payload = %{
+      data:
+        case vehicle_or_ghost do
```
phildarnowsky, line 20: note to self: must also handle null here

lib/skate_web/router.ex
phildarnowsky: I haven't heard any talk of updating if the vehicle on trip changes (nor against it), but yeah presumably we do want realtime updates. Will rework.

assets/src/components/ladderPage.tsx
phildarnowsky: It shouldn't be able to render two PPs, since `selectedVehicleOrGhost` and `vehicleAndRouteForNotification` are only mutated through reducers that ensure that they're not both set at once. But that said, never hurts to check.

Actually here's an idea: what if we go with what you're suggesting here, but we also add something like:

```
if(selectedVehicleOrGhost && vehicleAndRouteForNotification) {
  sendWarningToSentry("supposedly impossible state happening");
}
```

phildarnowsky:
@skyqrose this should address all your feedback, I verified that live update is working (both when the vehicle is on a visible ladder, and not). proceeding to embiggen test coverage.

lib/skate_web/channels/vehicle_channel.ex
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  @impl Phoenix.Channel
+  def join("vehicle:trip_ids:" <> trip_ids, _message, socket) do
+    trip_ids = String.split(trip_ids, ",")
+    vehicle_or_ghost = Realtime.Server.peek_at_vehicles(trip_ids) |> List.first()
+    payload = make_payload(vehicle_or_ghost)
+
+    if vehicle_or_ghost do
+      _ = Duration.log_duration(Server, :subscribe_to_vehicle, [vehicle_or_ghost.id])
```
skyqrose, line 22: The duration logging was to investigate a specific issue a while ago, and doesn't need to be copied over. (We could probably remove the old duration logging too.)

lib/skate_web/channels/vehicle_channel.ex
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    {:ok, payload, socket}
+  end
+
+  defp make_payload([]) do
+    %{data: %{}}
+  end
+
+  defp make_payload([vehicle_or_ghost]) do
```
skyqrose, line 32: I think the `List.first()` call above makes these list cases unnecessary.

lib/skate_web/channels/vehicle_channel.ex
```diff
@@ -0,0 +1,41 @@
...
+    make_payload(vehicle_or_ghost)
+  end
+
+  defp make_payload(vehicle_or_ghost) do
+    route = Schedule.route_by_id(vehicle_or_ghost.route_id)
```
skyqrose, line 37: Can `vehicle_or_ghost` be `nil` here? If so this might crash.

(Also, this function needs a `@spec`)

assets/src/hooks/useVehicleAndRouteForNotification.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (isVehicle(newVehicleOrGhostAndRoute.vehicleOrGhost)) {
+          window.FS.event("Notification linked to VPP")
+        } else {
+          window.FS.event("Notification linked to ghost")
+        }
+      } else if (selectedTripIdsForNotification) {
+        window.FS.event("Notification link failed")
+      }
+    }
+  }, [newVehicleOrGhostAndRoute])
```
skyqrose, line 70: This will trigger every time the vehicle updates. I'm not sure how to prevent that without keeping some kind of "reported for this vehicle already" state though.

lib/skate_web/channels/vehicle_channel.ex
```diff
@@ -0,0 +1,41 @@
+defmodule SkateWeb.VehicleChannel do
```
skyqrose, line 1: This modules needs tests.

skyqrose (commented):
Let's put off the loading state into a [separate task](https://app.asana.com/0/1148853526253420/1195672895821953)

phildarnowsky:
@skyqrose once more, with feeling! Still waiting to hear from codecov but I think this should pass coverage.

skyqrose (approved):

PR #867, opened 2020-09-21
Show notifications only on ladder page 

skyqrose:
Asana Task: [Restrict notification cards to only appear on route ladder view](https://app.asana.com/0/1148853526253426/1194621236689393)

### DIFF
--- a/assets/tests/contexts/notificationsContext.test.tsx
+++ b/assets/tests/contexts/notificationsContext.test.tsx
@@ -1,0 +1,92 @@
+import { act as hooksAct, renderHook } from "@testing-library/react-hooks"
+import { mount } from "enzyme"
+import React, { useContext } from "react"
+import { act as testUtilsAct } from "react-dom/test-utils"
+import {
+  NotificationsContext,
+  NotificationsProvider,
+} from "../../src/contexts/notificationsContext"
+import { useNotifications } from "../../src/hooks/useNotifications"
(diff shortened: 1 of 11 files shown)


assets/tests/contexts/notificationsContext.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  test("starts empty", () => {
+    const { result } = renderHook(() => useContext(NotificationsContext), {
+      wrapper: NotificationsProvider,
+    })
+    expect(result.current.notifications).toEqual([])
+  })
+
+  test("receives incoming notifications", () => {
+    let handler: (notification: Notification) => void
+    ;(useNotifications as jest.Mock).mockImplementationOnce((h) => {
```
phildarnowsky, line 48: @skyqrose I had been meaning to ask, what's up with this odd `;(useNotifications as jest.Mock)` idiom?

phildarnowsky (approved):

assets/tests/contexts/notificationsContext.test.tsx
skyqrose: The parens are necessary because otherwise it looks like `useNotifciations as (jest.Mock.mockImplementation)`.
The `;` is necessary because otherwise the parens make it look like a function call continued from the previous line: `void(useNotifications)`

(JavaScript is such a disaster that even its syntax has surprises.)

PR #873, opened 2020-09-29
Open modal on clickthrough of notification when no vehicle found

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1191003352789108

If no vehicle is found, we open a modal with some pertinent copy, and give the user an affordance to delete the notification.


assets/src/components/inactiveNotificationModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  setInactiveNotification,
+}: {
+  notification: Notification
+  setInactiveNotification: (notification: Notification | null) => void
+}) => {
+  const [, dispatch] = useContext(StateDispatchContext)
+  const { removeNotification } = useContext(NotificationsContext)
+
+  const closeModal = () => {
+    setInactiveNotification(null)
```
phildarnowsky, line 20: note to self: this is gross, try to figure out a way to get rid of the redundancy

ahakuta:
Screenshot_20200929-190513

Hey @phildarnowsky, thanks again for whipping this together.

A few pieces of feedback on this based on what @ingridpierre's mockup shows:

- The **Keep** and **Remove** colors differ from the mockup

- Can you reduce the width of the modal a bit? I tested on my phone and on Chrome simulating a Moto G4 (smaller screen) and on both, the modal spans the full width of the screen. I'm not sure how much to reduce the width by, but I think enough  so that the edges of the modal don't reach the edges of the screen. 

- Center the Keep and Remove buttons so that they're not so far apart

That's all I got- thanks again @phildarnowsky!

phildarnowsky:
@ahakuta the colors do differ from the mockup, but the ones I've used are the standard primary and secondary button colors from the styleguide, which is correct?

ahakuta:
@phildarnowsky ah that makes sense.

I'm assuming @ingridpierre deviated from those with purpose, so let's stick to the ones in the mockup. Thanks for flagging that.

assets/src/components/inactiveNotificationModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { title } from "./notifications"
+
+const InactiveNotificationModal = ({
+  notification,
+  removeNotification,
+  setInactiveNotification,
+}: {
+  notification: Notification
+  removeNotification: (id: number) => void
+  setInactiveNotification: (notification: Notification | null) => void
```
skyqrose, line 15: Some places (like here) this is `inactiveNotification` and other places (like css) it's `invalidNotification`. Let's standardize on one.

assets/src/components/inactiveNotificationModal.tsx
skyqrose: Right now if `useVehicle....()` fails, it sets the result elsewhere, which is why this information is in two places. What if we could detect failure from the same place we detect success, and roughly do:
```
case useVehicle(selectedNotification) do
  undefined -> <Loading/>
  null -> <FailureModal/>
  vehicle -> <VPP/>
end
```
and we could remove the `inactiveNotificationContext` entirely.

skyqrose (commented):
I'm not done reviewing, but I have some comments that I figured I would get committed before standup.

assets/src/components/inactiveNotificationModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+  )
+}
+
+const runIdPhrase = (notification: Notification): string => {
+  if (notification.runIds.length === 0) {
+    return "No runs associated with this notification are"
+  }
+
+  if (notification.runIds.length === 1) {
+    return `Run ${notification.runIds[0]} is not`
```
skyqrose, line 72: Design question about a weird edge case that (if we decide to act on it) would be a separate PR after the release: @ingridpierre 

It's mostly but not entirely true that the run isn't active. The block waiver might not be applied to the whole run, and if a bus is on the right run, but doing a trip that the block waiver wasn't applied to, we wouldn't match it. Should we:
1. Change the wording to something to the effect of "Nobody's doing the trips this notification applies to"
2. Change the matching so we can show a bus that's doing any part of this run, not just the trips on the block waiver
3. Leave it as is

?

assets/css/_invalid_notification_modal.scss
```diff
@@ -0,0 +1,42 @@
...
+  margin-top: 1rem;
+}
+
+.m-invalid-notification-modal__buttons {
+  display: flex;
+  height: 1rem;
+  margin-top: 1.5rem;
+}
+
+.m-invalid-notification-modal__button-wrapper {
```
skyqrose, line 25: To get the buttons centered, instead of `position: absolute`ing them to the sides of a wrapper, I think you can remove the wrappers with:
```
.buttons {
  display: flex;
  justify-content: center
}

.keep-button {
  flex: 0 1 auto; // the 0 prevents it from growing above its size
  width: 7rem;
}

.discard-button {
  // same
}

File: assets/src/hooks/useVehicleAndRouteForNotification.ts [hunk 5]
```diff
@@ -33,15 +33,18 @@ const parseVehicleOrGhostAndRouteData = ({
 }
 
 const useVehicleAndRouteForNotification = (
-  selectedTripIdsForNotification?: TripId[]
+  notification?: Notification
 ): VehicleOrGhostAndRoute | undefined => {
```
skyqrose, line 37: This would need to change to `: VehicleOrGhostAndRoute | null | undefined` to be able to use it to detect when it fails instead of using the separate context for that.

skyqrose (commented):
No big comments except for the one from this morning.

phildarnowsky:
@skyqrose this should address all your feedback

assets/src/components/ladderPage.tsx
```diff
@@ -71,8 +68,13 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
   )
 
   const vehicleAndRouteForNotification = useVehicleAndRouteForNotification(
-    selectedTripIdsForNotification
+    selectedNotification
   )
+  useEffect(() => {
+    if (vehicleAndRouteForNotification === null) {
+      dispatch(setNotificationIsInactive())
```
skyqrose, line 75: Not necessary if you want to get this out a bit faster, but what if we showed the modal right from `ladderPage` instead of setting the state and drawing from `App`? It's all absolutely/fixedly positioned, so I think it'd work. (If you do this, you don't need to wait for another review on it)

assets/src/components/inactiveNotificationModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  notification,
+  removeNotification,
+}: {
+  notification: Notification
+  removeNotification: (id: number) => void
+}) => {
+  const [, dispatch] = useContext(StateDispatchContext)
+
+  const closeModal = () => {
+    dispatch(setNotification(undefined))
```
skyqrose, line 18: I think you also need `dispatch(setSelectedNotificationIsInactive(false))` so the next time you click a notification it doesn't show the modal right away

skyqrose (approved):
See comments, but ✅ 🎉

ahakuta:
Thank you @skyqrose!

assets/src/components/ladderPage.tsx
phildarnowsky: Personally I like having all the modals in one place; this is something about which reasonable people might disagree. But in my experience modals are one of those common elements that are surprisingly tricky to get right, so having them all co-located might make them easier to reason about.

assets/src/components/inactiveNotificationModal.tsx
phildarnowsky: No we're covered there, check out the `selectedNotificationIsInactiveReducer`. Any time a `SET_NOTIFICATION` event is dispatched--including when the notification is `undefined`--we clear the is-inactive flag.

PR #880, opened 2020-10-07
Index vehicles by run ID for notification purposes

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1197215845308509

We had been indexing vehicles by trip ID in the realtime server, so that
when a user clicked a notification, we could use the trip IDs associated
with the notification to look up the pertinent vehicle.

This led to a counterintuitive behavior where, if a vehicle was on a run
affected by a block waiver, but not on one of the affected trips, it
would not be considered associated with the waiver.

Here we change to indexing and retrieving vehicles by run ID instead, to
avoid that problem.


skyqrose (commented):
`notification_server.ex` is still calling `peek_at_vehicles(trip_ids)` when we match at the time of creation.

phildarnowsky:
@skyqrose ready for another look

skyqrose (approved):

PR #890, opened 2020-10-14
Loading state for notification clickthrough

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1195459399895689


assets/src/components/ladderPage.tsx
```diff
@@ -73,8 +73,15 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
...
+    } else if (
+      vehicleAndRouteForNotification === undefined &&
+      selectedNotification
+    ) {
+      dispatch(setNotificationIsLoading(true))
+    } else if (vehicleAndRouteForNotification) {
+      dispatch(setNotificationIsLoading(false))
     }
-  }, [vehicleAndRouteForNotification])
+  }, [vehicleAndRouteForNotification, selectedNotification])
```
skyqrose, line 84: I'm worried about missing edge cases where this won't keep `selectedNotificationIsLoading/Inactive` up to date. What happens if you switch away from the ladder page while it's loading or inactive? What should happen? (If there is a problem there, which I'm not sure there is, returning a cleanup function from `useEffect` is probably the way to handle it.)

Are there any other times when vehicleAndRouteForNotification might change unexpectedly and we need to do extra work to keep everything in sync?

skyqrose (approved):

assets/src/components/ladderPage.tsx
phildarnowsky: @skyqrose On a semi-related note, I'm thinking the best place for this `useEffect` would be within `useVehicleAndRouteForNotification` rather than here. Does that make sense to you? I don't think it would alter the functionality any.

phildarnowsky: @skyqrose after some experimentation and thought, I've decided that the best thing to do would be to clear the `selectedNotification` when unmounting, via the `useEffect` cleanup function.

phildarnowsky: actually that doesn't seem to be the right thing--continuing to experiment

phildarnowsky: @skyqrose OK! I had to try a great many things but there turns out to be a simple fix. Ready for another look.

skyqrose (approved):

PR #896, opened 2020-10-20
Add more timepoint order hints

skyqrose:
Followup to #895 

I looked for all the routes that ran into [this kind of conflict](https://github.com/mbta/skate/pull/895#discussion_r508555899) and added hints (manual overrides) to straighten them out.

Screen Shot 2020-10-20 at 15 04 09

There are still conflicts that can't be straightened into a line with timepoint order hints in:
* 708 and 60 (there's a loop with multiple layover points)
* 26 (a loop that sometimes goes clockwise and sometimes counterclockwise)
* 23 (a [bug in GTFS](https://app.asana.com/0/881264583703207/1198563159603145) means one of the variants is listed in the wrong direction)
* 435 (a variant loops to the same timepoint twice, which we can't show)
but I've added hints to be sure those cases are being drawn in the best way possible.

I did not search for places where a branch or short turn (like route 28's malcx school short turn) had an ambiguous resolution to the merge, only places with a conflict with a clearly incorrect resolution, so it's still possible for inspectors to report timepoints being out of order. But this should get the worst of them.

### DIFF
--- a/data/timepoint_order.json
+++ b/data/timepoint_order.json
@@ -1,18 +1,55 @@
 {
+  "11": {
+    "1": ["brway", "waher", "tumed", "sosta", "bdfch"]
+  },
   "15": {
-    "0": ["roxbs", "malcx", "nubn", "kane", "peter", "genbo"]
+    "comment": "malcx and peter are timepoints when they're a short turn terminal",
+    "0": [
+      "roxbs",
+      "malcx",
+      "nubn",
+      "upham",
+      "kane",
+      "peter",
+      "genbo",
+      "fldcr",
+      "dotav",
+      "ashmt"
+    ]
   },
   "19": {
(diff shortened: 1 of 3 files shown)


data/timepoint_order.json
```diff
@@ -1,18 +1,55 @@
 {
+  "11": {
+    "1": ["brway", "waher", "tumed", "sosta", "bdfch"]
+  },
   "15": {
-    "0": ["roxbs", "malcx", "nubn", "kane", "peter", "genbo"]
+    "comment": "malcx and peter are timepoints when they're a short turn terminal",
```
phildarnowsky, line 6: Ah, embedded comments, excellent idea.

phildarnowsky (approved):

PR #899, opened 2020-10-22
Refactor Right Panel

skyqrose:
Pulls the logic for the right panel out of `LadderPage` so it will be easier to add the notifications drawer to and use on every page.
Supports [Bell icon and opening an empty notification drawer](https://app.asana.com/0/1148853526253426/1192202587402575)

Also:
* Moves the `useRoutes()` fetch to `appStateWrapper` so we only make it once and it's accessible everywhere.
* Adds a `useRoute()` wrapper around that context for the common case of getting one route by id.
* Changes `useVehicleAndRouteForNotification` to `useVehicleForNotification` now that we can get the route from anywhere.
* Moves `useVehicleForNotification` to `appStateWrapper` with a context so `modal.tsx` can see the result, removing the need for setting `state.notificationIsLoading` and `state.notificationIsInvalid`.

Right now, the `RightPanel` is only called by `LadderPage`, but in the future, every page will have its own `RightPanel`.

This PR should have no user-visible changes.


assets/src/components/appStateWrapper.tsx
```diff
@@ -1,21 +1,26 @@
...
   const socketStatus = useSocket()
+  const routes = useRoutes()
 
   return (
     <StateDispatchProvider state={state} dispatch={dispatch}>
       <SocketProvider socketStatus={socketStatus}>
-        <NotificationsProvider>
-          <App />
-        </NotificationsProvider>
+        <RoutesContext.Provider value={routes}>
```
phildarnowsky, line 25: @skyqrose for consistency let's define a `RoutesProvider` component in `routes_context.tsx`

assets/tests/components/__snapshots__/ladderPage.test.tsx.snap
```diff
@@ -509,15 +526,15 @@ exports[`LadderPage renders with timepoints 1`] = `
       >
         <li>
           <button
-            className="m-route-picker__route-list-button m-route-picker__route-list-button--unselected"
+            className="m-route-picker__route-list-button m-route-picker__route-list-button--selected"
```
phildarnowsky, line 529: does this change and the similar one below from `unselected` to `selected` represent a possible regression?

assets/src/components/rightPanel.tsx
```diff
@@ -0,0 +1,90 @@
...
+      dispatch(setNotificationIsLoading(false))
+    }
+  }, [vehicleAndRouteForNotification, selectedNotification])
+  const history = useHistory()
+  if (history) {
+    /* istanbul ignore next */
+    history.listen(() => dispatch(setNotification(undefined)))
+  }
+
+  const vehicleOrGhostForVPP:
```
phildarnowsky: not directly related to the PR but while we're here maybe let's add a comment about the `null`/`undefined` convention for loading vehicles

assets/src/components/appStateWrapper.tsx
```diff
@@ -3,22 +3,32 @@ import { NotificationsProvider } from "../contexts/notificationsContext"
...
+    state.selectedNotification
+  )
+
   return (
     <StateDispatchProvider state={state} dispatch={dispatch}>
       <SocketProvider socketStatus={socketStatus}>
         <RoutesContext.Provider value={routes}>
           <NotificationsProvider>
-            <App />
+            <VehicleAndRouteForNotificationContext.Provider
```
phildarnowsky: likewise let's have a `VehicleAndRouteForNotificationProvider`

phildarnowsky (approved):
@skyqrose nice work simplifying this stuff, I have only minor feedback.

assets/tests/components/__snapshots__/ladderPage.test.tsx.snap
skyqrose: No.
The test `renders with timepoints` didn't select the routes, so it didn't actually render with timepoints.  [Fixing that test to be as intended](https://github.com/mbta/skate/pull/899/files#diff-41b776228e88467a5a40b857932c4e403705be68b76270e8b50f440ec6abcb3cR65) is the cause of all the changes in this ladderPage snapshot.

PR #915, opened 2020-10-29
Store route settings in database rather than localStorage

phildarnowsky:
Here we start persisting users' route preferences (which routes are visible, which directions they see ladders in, and crowding toggles for ladders) on the backend rather than in the browser's localStorage.

@skyqrose you'll notice that the schema I went with differs from what we discussed the other day; I decided it would be wiser to have the backend data's structure mirror that of the frontend. This is the correction of an "impedance mismatch" that I mention in two commit messages.


lib/skate/settings/ladder_direction.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Skate.Settings.LadderDirection do
+  use Ecto.Type
```
skyqrose, line 2: This is as much related to the DB as the schemas are, so it seems inconsistent that the schemas are in `db/` while this isn't.
I see 3 options:
1. Move the `Ecto.Type`s into `db/`
2. rename `db/` to `schema/`
3. Leave it, it's fine.

skyqrose (commented):
The backend is good. Still needs `route_settings_controller_test.exs` and `route_settings_test.exs`, though.

I'll review the frontend after standup.

assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -27,40 +26,96 @@ const PERSISTED_KEYS: Key[] = [
...
+
+  const [firstLoadDone, setFirstLoadDone] = useState(false)
+
+  useEffect(() => {
+    if (firstLoadDone) {
+      putRouteSettings(state)
+    } else {
+      setFirstLoadDone(true)
+    }
+  }, [selectedRouteIds, ladderDirections, ladderCrowdingToggles])
```
skyqrose, line 45: Ooh, I like this way of updating the backend more than the approach I did for usersettings of calling the api call in every place the settings change.
Do you think it'd be possible/easy to split this into like a `useUpdateBackendAboutRouteSettings()` hook and also refactor the user settings updating into a `useUpdateBackendAboutUserSettings()` hook?
(This PR is big and it might be better to do those in a followup PR)

assets/tests/hooks/usePersistedStateReducer.test.ts
```diff
@@ -86,30 +90,28 @@ describe("usePersistedStateReducer", () => {
...
     const { result } = renderHook(() => usePersistedStateReducer())
     const [state] = result.current
-    expect(state.settings.shuttleVehicleLabel).toEqual(
+    expect(state.userSettings.shuttleVehicleLabel).toEqual(
       VehicleLabelSetting.RunNumber
     )
   })
 
-  test("if settings are in localstorage, copies them to the backend and uses them", () => {
+  test("if user settings are in localstorage, copies them to the backend and uses them", () => {
```
skyqrose, line 114: This test should also test that the route settings are copied from local storage to the backend.

skyqrose (approved):

assets/src/hooks/usePersistedStateReducer.ts
phildarnowsky: I do think that would be possible and worth doing, but I do also think that a followup is indicated, since as you say this has turned out to be a big PR.

skyqrose (approved):

PR #925, opened 2020-11-05
Filter notifications on backend rather than frontend

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402603

### DIFF
--- a/lib/notifications/notification_server.ex
+++ b/lib/notifications/notification_server.ex
@@ -152,7 +152,7 @@ defmodule Notifications.NotificationServer do
 
         Enum.each(entries, fn {pid, username} ->
           if Enum.member?(usernames, username) do
-            send(pid, {:notifications, notifications})
+            send(pid, {:notification, notification})
           end
         end)
       end)
--- a/lib/notifications/notification_server.ex
+++ b/lib/notifications/notification_server.ex
@@ -4,6 +4,7 @@ defmodule Notifications.NotificationServer do
   alias Notifications.Notification
   alias Realtime.{BlockWaiver, Ghost, Vehicle}
   alias Schedule.Block
+  alias Skate.Settings.User
 
   require Logger
 
@@ -23,10 +24,10 @@ defmodule Notifications.NotificationServer do
     GenServer.cast(server, {:new_block_waivers, new_waivers_by_block_key})
   end
 
-  def subscribe(server \\ default_name()) do
+  def subscribe(username, server \\ default_name()) do
     registry_key = GenServer.call(server, :subscribe)
 
-    Registry.register(Notifications.Supervisor.registry_name(), registry_key, :dummy_value)
+    Registry.register(Notifications.Supervisor.registry_name(), registry_key, username)
     :ok
   end
 
@@ -146,8 +147,14 @@ defmodule Notifications.NotificationServer do
 
   defp broadcast(notifications, registry_key) do
     Registry.dispatch(Notifications.Supervisor.registry_name(), registry_key, fn entries ->
-      Enum.each(entries, fn {pid, _} ->
-        send(pid, {:notifications, notifications})
+      Enum.each(notifications, fn notification ->
+        usernames = User.usernames_for_route_ids(notification.route_ids)
+
+        Enum.each(entries, fn {pid, username} ->
+          if Enum.member?(usernames, username) do
+            send(pid, {:notifications, notifications})
+          end
+        end)
       end)
     end)
   end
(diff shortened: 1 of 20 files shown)


lib/notifications/notification_server.ex
```diff
@@ -146,8 +147,14 @@ defmodule Notifications.NotificationServer do
...
   defp broadcast(notifications, registry_key) do
     Registry.dispatch(Notifications.Supervisor.registry_name(), registry_key, fn entries ->
-      Enum.each(entries, fn {pid, _} ->
-        send(pid, {:notifications, notifications})
+      Enum.each(notifications, fn notification ->
+        usernames = User.usernames_for_route_ids(notification.route_ids)
+
+        Enum.each(entries, fn {pid, username} ->
+          if Enum.member?(usernames, username) do
+            send(pid, {:notifications, notifications})
```
skyqrose, line 155:
```suggestion
            send(pid, {:notification, notification})
```
Now that you're looping over the notifications one by one, I think this needs to be singular.

test/notifications/notification_server_test.exs
```diff
@@ -1,13 +1,17 @@
 defmodule Notifications.NotificationServerTest do
-  use ExUnit.Case, async: false
+  use ExUnit.Case, async: true
```
skyqrose, line 2: Is this safe to do with tests that access the database?

test/notifications/notification_server_test.exs
```diff
@@ -214,7 +216,7 @@ defmodule Notifications.NotificationServerTest do
       assert log == ""
     end
 
-    test "broadcasts and logs new notifications for waivers with recognized reason for vehicles" do
+    test "broadcasts and logs new notifications for waivers with recognized reason for vehicles on selected routes" do
```
skyqrose, line 219: With the log assertions removed from `assert_notification`, the test names should also remove "and logs". And if you don't want to test the logging (which is fine with me), you might also consider if there's a different way to do the "does nothing if no new block waivers are received" test above. (It's possible testing the log is the best way to do that.)

test/notifications/notification_server_test.exs
```diff
@@ -244,7 +246,7 @@ defmodule Notifications.NotificationServerTest do
       end
     end
 
-    test "broadcasts and logs new notifications for waivers with recognized reason when no vehicle or ghost is associated" do
+    test "broadcasts and logs new notifications for waivers with recognized reason when no vehicle or ghost is associated on selected routes" do
```
skyqrose, line 249: A good complement to this "on selected routes" addition would be a test that it doesn't send a notification to a user who's not looking at that route. Otherwise, the old code that sent every notification to everybody would still pass all these tests.

test/notifications/notification_server_test.exs
phildarnowsky: probably not...checked this in by mistake.

phildarnowsky: scratch that, I apparently did that for a reason. The reason is that, with `async: false`, I get the following quite verbose error:

```
     ** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.8354.0>.

     When using ownership, you must manage connections in one
     of the four ways:

     * By explicitly checking out a connection
     * By explicitly allowing a spawned process
     * By running the pool in shared mode
     * By using :caller option with allowed process

     The first two options require every new process to explicitly
     check a connection out or be allowed by calling checkout or
     allow respectively.

     The third option requires a {:shared, pid} mode to be set.
     If using shared mode in tests, make sure your tests are not
     async.

     The fourth option requires [caller: pid] to be used when
     checking out a connection from the pool. The caller process
     should already be allowed on a connection.

     If you are reading this error, it means you have not done one
     of the steps above or that the owner process has crashed.

     See Ecto.Adapters.SQL.Sandbox docs for more information.
```

So I will investigate further what this means and how to resolve it.

test/notifications/notification_server_test.exs
phildarnowsky: I actually put the log assertions back in. With all the chaos of the last week I don't actually remember why I took them out, but I assume the tests weren't working properly at one point and I suspected those assertions.

test/notifications/notification_server_test.exs
phildarnowsky: Looks like I just had to use `Skate.DataCase` here instead of `ExUnit.Case`, and a few other corresponding changes.

phildarnowsky:
@skyqrose this latest version should address all your feedback to this point, I do need to bring coverage up.

skyqrose (approved):

PR #929, opened 2020-11-10
Show at most one floating notification

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402579

Now we only show the most recent notification, which we can now hide but no longer remove. The distinction will become important when we start populating the drawer.

### DIFF
--- a/assets/src/components/notifications.tsx
+++ b/assets/src/components/notifications.tsx
@@ -2,12 +2,14 @@ import React, { useContext, useEffect, useState } from "react"
 import { NotificationsContext } from "../contexts/notificationsContext"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import useCurrentTime from "../hooks/useCurrentTime"
-import { Notification, NotificationId } from "../realtime.d"
+import { Notification } from "../realtime.d"
 import { setNotification } from "../state"
 import { NotificationContent } from "./notificationContent"
 
 export const Notifications = () => {
-  const { notifications, removeNotification } = useContext(NotificationsContext)
+  const { notifications, showLatestNotification, hide } = useContext(
+    NotificationsContext
+  )
   const currentTime = useCurrentTime()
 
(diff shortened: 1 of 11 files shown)


assets/src/components/notifications.tsx
```diff
@@ -16,29 +18,31 @@ export const Notifications = () => {
...
+
   return (
     <div className="m-notifications">
-      {notifications.map((notification) => (
+      {showLatestNotification && latestNotificationIndex >= 0 && (
         <NotificationCard
-          key={notification.id}
-          notification={notification}
-          remove={removeNotification}
+          key={notifications[latestNotificationIndex].id}
```
skyqrose, line 27: `key` is no longer needed since we're not rendering a list of them anymore.

assets/tests/components/notificationDrawer.test.tsx
```diff
@@ -30,13 +30,14 @@ describe("NotificationDrawer", () => {
     expect(dispatch).toHaveBeenCalledWith(closeNotificationDrawer())
   })
 
-  test("renders notifications", () => {
+  test("renders latest notification", () => {
```
skyqrose, line 33: The notification drawer should show all the notifications, so I don't think this name should change.

assets/tests/components/inactiveStateModal.test.tsx
```diff
@@ -85,45 +82,37 @@ describe("InactiveNotificationModal", () => {
...
-      />
+      <InactiveNotificationModal notification={notification} hide={hide} />
     )
 
     wrapper
       .find(".m-inactive-notification-modal__discard-button")
       .first()
       .simulate("click")
-    expect(removeNotification).toHaveBeenCalledWith(notification.id)
+    expect(hide).toHaveBeenCalled()
```
skyqrose, line 104: I think this test is a duplicate of the one above, and could be removed.

skyqrose (approved):

PR #937, opened 2020-11-18
Store notifications in the database

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402599

We now store notifications in the database when they come in. There is some extra logic that prevents identical notifications from being logged or written by multiple nodes at once.

### DIFF
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -57,7 +57,7 @@ defmodule Notifications.Notification do
     :end_time
   ]
 
-  @spec create([t()]) :: [t()]
+  @spec create(t(), (DbNotification.t() -> t())) :: t()
   def create(notification_without_id, insertion_callback \\ fn _ -> nil end) do
     changeset =
       DbNotification.changeset(
@@ -76,6 +76,6 @@ defmodule Notifications.Notification do
           nil
       end
 
-    %{notification_without_id | id: id}
+    %__MODULE__{notification_without_id | id: id}
   end
 end
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -60,8 +60,8 @@ defmodule Notifications.Notification do
     :end_time
   ]
 
-  @spec get_or_create(t(), (DbNotification.t() -> t())) :: t()
-  def get_or_create(notification_without_id, insertion_callback \\ fn _ -> nil end) do
+  @spec get_or_create(t()) :: t()
+  def get_or_create(notification_without_id) do
     identifying_fields =
       Map.take(notification_without_id, [:start_time, :end_time, :block_id, :service_id, :reason])
       |> Map.to_list()
@@ -82,7 +82,6 @@ defmodule Notifications.Notification do
           db_notification = insert!(changeset)
           notification_with_id = %{notification_without_id | id: db_notification.id}
           log_creation(notification_with_id)
-          insertion_callback.(notification_with_id)
           db_notification
         end
       end)
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -1,6 +1,7 @@
 defmodule Notifications.Notification do
   alias Schedule.Route
   alias Schedule.Trip
+  alias Schedule.Hastus.Run
 
   @type notification_reason :: :manpower | :disabled | :diverted | :accident
 
@@ -8,7 +9,7 @@ defmodule Notifications.Notification do
           created_at: Util.Time.timestamp(),
           reason: notification_reason(),
           route_ids: [Route.id()],
-          run_ids: [String.t()],
+          run_ids: [Run.id()],
           trip_ids: [Trip.id()],
           operator_id: String.t() | nil,
           operator_name: String.t() | nil,
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -1,13 +1,18 @@
 defmodule Notifications.Notification do
+  import Skate.Repo
+
   alias Schedule.Route
   alias Schedule.Trip
   alias Schedule.Hastus.Run
+  alias Notifications.NotificationReason
+  alias Notifications.Db.Notification, as: DbNotification
 
-  @type notification_reason :: :manpower | :disabled | :diverted | :accident
+  @type id :: integer()
 
   @type t() :: %__MODULE__{
+          id: id() | nil,
           created_at: Util.Time.timestamp(),
-          reason: notification_reason(),
+          reason: NotificationReason.t(),
           route_ids: [Route.id()],
           run_ids: [Run.id()],
           trip_ids: [Trip.id()],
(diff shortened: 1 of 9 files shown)


priv/repo/migrations/20201110152910_create_notifications.exs
```diff
@@ -0,0 +1,34 @@
...
+      index(
+        :notifications,
+        [:start_time, :end_time, :block_id, :service_id, :reason],
+        unique: true,
+        name: "notifications_unique_index"
+      )
+    )
+  end
+
+  def down do
```
skyqrose, line 31: Was there a reason to use separate `up`/`down` functions instead of a single `change` function?

lib/notifications/notification.ex
```diff
@@ -27,6 +53,29 @@ defmodule Notifications.Notification do
     :operator_id,
     :operator_name,
     :route_id_at_creation,
-    :start_time
+    :start_time,
+    :end_time
   ]
+
+  @spec create(t(), (DbNotification.t() -> t())) :: t()
+  def create(notification_without_id, insertion_callback \\ fn _ -> nil end) do
```
skyqrose, line 61: This callback seems not very elixir-y to me. What about either doing the log call inside this function, or having the caller case on whether they got an id back or not?

lib/notifications/notification_server.ex
```diff
@@ -145,16 +152,14 @@ defmodule Notifications.NotificationServer do
     end
   end
 
-  defp broadcast(notifications, registry_key) do
+  defp broadcast(notification, registry_key) do
```
skyqrose, line 155: 👍

skyqrose (approved):

priv/repo/migrations/20201110152910_create_notifications.exs
phildarnowsky: There is--the index. If you have something like:

```
def change do
  create table :foo do
    # some columns defined
  end
  create index :foo, [:some_column, :other_column]
end
```

and you roll that back, it wants to drop the index before dropping the table. But dropping the index separately isn't necessary, since when you drop a table all the indices on it disappear too. 

It also doesn't really hurt to drop the index separately, except that during development, when I was tweaking the index, the separate-drop scheme led to some annoyances that were breaking my flow.

lib/notifications/notification.ex
phildarnowsky: Logging within this function does make sense to me, because we've established that it's useful to log the creation of every single notification. But I do like the approach of passing a callback better than examining the result in the caller. It's more explicit IMHO to say "here is a way to request some processing be done if we successfully insert", than to have to know that a nil ID means insert failed in the caller.

phildarnowsky:
@skyqrose I realized late yesterday that there's a problem with this code, which I alluded to at grooming. Basically, in trying to avoid duplicate notifications, I've gone a bit too far such that each user will now receive only about 1/N of the notifications, where N is the number of nodes currently running.

Imagine that we have the code as in this PR running on two or more separate instances in dev/prod. A block waiver comes in from Busloc. Each of our N instances grabs that waiver, constructs a notification out of it, and tries to save the notification to the DB. Thanks to the constraint, only one of them actually writes to the DB. The writing instance then logs and broadcasts the notification, but it only broadcasts _to the clients connected to that instance_. That is, about 1/N of them.

Fix to come.

phildarnowsky:
@lemald I only really need review on the last commit, but the rest of the PR is probably necessary context. And of course if you have any feedback on the other parts of the PR I'd be glad to hear it.

lib/notifications/notification.ex
```diff
@@ -59,27 +60,34 @@ defmodule Notifications.Notification do
...
+    identifying_fields =
+      Map.take(notification_without_id, [:start_time, :end_time, :block_id, :service_id, :reason])
+      |> Map.to_list()
 
-    id =
-      case insert(changeset) do
-        {:ok, db_notification} ->
+    {:ok, db_record} =
+      Skate.Repo.transaction(fn ->
+        existing_record = Skate.Repo.one(from(DbNotification, where: ^identifying_fields))
```
lemald, line 71: Isn't there still a race condition here? It's inside a transaction but I don't think there's any guarantee that another client won't insert something between this `from` query and the `insert` below.

phildarnowsky: I have just spent some time with the documentation and it looks like you're correct, will fix.

lib/notifications/notification.ex
```diff
@@ -27,6 +56,41 @@ defmodule Notifications.Notification do
...
+          insertion_callback.(notification_with_id)
+          db_notification
+        end
+      end)
+
+    %__MODULE__{notification_without_id | id: db_record.id}
+  end
+
+  defp log_creation(notification) do
+    Logger.warn("Notification created new notification new_notification=#{inspect(notification)}")
```
lemald, line 94: `Notification created new notification` sort of repeats itself, can you just say `Notification created` and drop the `new norification`? But not the `new_notification=`.

lib/notifications/notification_reason.ex
```diff
@@ -0,0 +1,56 @@
...
+    if reason in @valid_reasons do
+      {:ok, reason}
+    else
+      :error
+    end
+  end
+
+  @impl true
+  def load(reason) do
+    reason_as_atom = String.to_atom(reason)
```
lemald, line 39: `String.to_existing_atom` might be a bit safer, since `String.to_atom` has a potential memory leak due to the fact that atoms are apparently not garbage collected. OTOH this data is coming from our own database, not a random piece of user input or something, so it's probably not that huge an issue.

priv/repo/migrations/20201110152910_create_notifications.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Skate.Repo.Migrations.StoreNotifications do
+  use Ecto.Migration
+
+  def up do
+    create table(:notifications) do
+      add(:created_at, :bigint, null: false)
+      add(:reason, :string, null: false)
```
lemald, line 7: Any particular reason not to implement this as an `enum` type in Postgres?

lib/notifications/notification_reason.ex
phildarnowsky: Right, we control both sides here so it should be fine, but it would also be good to get in the habit of using `to_existing_atom` when possible.

priv/repo/migrations/20201110152910_create_notifications.exs
phildarnowsky: No, that makes sense.

phildarnowsky:
@lemald these should address all your feedback

priv/repo/migrations/20201110152910_create_notifications.exs
```diff
@@ -29,6 +33,7 @@ defmodule Skate.Repo.Migrations.StoreNotifications do
   end
 
   def down do
+    execute("DROP TYPE notification_reason")
```
lemald, line 36: Should this maybe be in the opposite order (drop the table using the type, _then_ the type)? Or does Ecto already automagically reverse the order of the steps in `down`?

lemald (approved):
Looks good to me!

PR #938, opened 2020-11-20
feat: send FullStory notification when bus goes invalid

lemald:
Card: [Add FS custom event tracking when a bus goes invalid on a ladder](https://app.asana.com/0/1148853526253426/1199159279046397/f)

This is my first PR, and it has side-effects in a system that I don't yet have access to, so it should receive extra-careful product review. 😅 

I got a little tangled up with the intricacies of testing channels since I haven't done much with that in the past. I'd _like_ to additionally test that a FullStory event _doesn't_ get triggered when a bus stays valid between two different messages, but I was having a lot of trouble crafting a test where the channel receives one message and then another.

### DIFF
--- a/assets/src/hooks/useVehicles.ts
+++ b/assets/src/hooks/useVehicles.ts
@@ -1,12 +1,19 @@
 import { Channel, Socket } from "phoenix"
-import { Dispatch as ReactDispatch, useEffect, useReducer } from "react"
+import {
+  Dispatch as ReactDispatch,
+  useEffect,
+  useReducer,
+  useState,
+} from "react"
 import { reload } from "../models/browser"
 import {
   VehicleOrGhostData,
   vehicleOrGhostFromData,
 } from "../models/vehicleData"
-import { VehicleOrGhost } from "../realtime.d"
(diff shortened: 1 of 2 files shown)


ahakuta:
@lemald do you think there's anything that might cause duplicate events to be fired? I'm looking at events logged to FS today during my testing, and all of the events that are in this screenshot fired at 4:12pm. 

I imagine it may well be the case that the invalid state can "flicker" on and off quickly which would cause this (or of course, multiple buses turning invalid at the same time), but I'm bringing this to your attention just in case there's some investigation that is worth doing here. I'm gonna keep poking to see if I can get a better sense of what is happening here- I have so many routes open that Fullstory can't show me the ladders that these invalids were on when this tracking was fired.

Screen Shot 2020-11-23 at 11 50 25 AM

ahakuta:
As commented in Slack, I think we're good to go ahead with this- I watched some more sessions in Fullstory and I didn't see anything too weird. I'll keep an eye once this hits prod.

phildarnowsky (approved):

PR #944, opened 2020-11-30
Username in Sentry (frontend)

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1196917567568801

### DIFF
--- a/assets/src/app.tsx
+++ b/assets/src/app.tsx
@@ -25,6 +25,10 @@ if (window.sentry) {
   Sentry.init({
     dsn: window.sentry.dsn,
   })
+
+  if (window.username) {
+    Sentry.setUser({ username: window.username })
+  }
 }
 
 if (window.FS && window.username) {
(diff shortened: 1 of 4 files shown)


assets/src/app.tsx
```diff
@@ -25,6 +25,10 @@ if (window.sentry) {
   Sentry.init({
     dsn: window.sentry.dsn,
   })
+
+  if (window.username) {
+    Sentry.setUser({ username: window.username })
```
lemald, line 30: For what it's worth, it looks like this is the username of the form `mbta-active-directory_emaldonado`, do we think that's good to send to Sentry? It's easy enough to parse out as a human reading it but I'm not sure whether or not we might want to join this later on to some data that has the `emaldonado` or `emaldonado@mbta.com` username format.

lemald:
One minor thing - it might be good to just drop commits `3361d93` and `7971f46` before merging, just to keep `master` clean.

assets/src/app.tsx
phildarnowsky: I thought about that some, and decided that for now, the best course of action was to leave the usernames as raw as possible--we can always translate and/or parse them later if we need to.

phildarnowsky:
@lemald we do squash-merges in Skate, so the questionable commits will disappear anyway--I'll remove their commit messages when I squash.

lemald:
> @lemald we do squash-merges in Skate, so the questionable commits will disappear anyway--I'll remove their commit messages when I squash.

Ah, good to know, given that I'm so used to doing rebase and merge on the TRC team.

assets/src/app.tsx
lemald: Makes sense to me!

lemald (approved):
👍

PR #946, opened 2020-12-01
Join notifications to users on notification create

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402601

When a notification is created, it's now linked via a new join table to users who are subscribed to any affected routes.

### DIFF
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -8,8 +8,10 @@ defmodule Notifications.Notification do
   alias Schedule.Route
   alias Schedule.Trip
   alias Schedule.Hastus.Run
+  alias Skate.Settings.User
   alias Notifications.NotificationReason
   alias Notifications.Db.Notification, as: DbNotification
+  alias Notifications.Db.NotificationUser, as: DbNotificationUser
 
   require Logger
 
@@ -72,6 +74,7 @@ defmodule Notifications.Notification do
       if db_record.id do
         notification_with_id = %__MODULE__{notification_without_id | id: db_record.id}
         log_creation(notification_with_id)
+        link_notification_to_users(notification_with_id)
         db_record
       else
         identifying_fields =
@@ -87,5 +90,26 @@ defmodule Notifications.Notification do
 
   defp log_creation(notification) do
     Logger.warn("Notification created new_notification=#{inspect(notification)}")
+  end
+
+  defp link_notification_to_users(notification) do
+    now = NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
+    user_ids = User.user_ids_for_route_ids(notification.route_ids)
+
+    notification_user_maps =
+      Enum.map(user_ids, fn user_id ->
+        %{
+          notification_id: notification.id,
+          user_id: user_id,
+          state: :unread,
+          inserted_at: now,
+          updated_at: now
+        }
+      end)
+
+    Skate.Repo.insert_all(
+      DbNotificationUser,
+      notification_user_maps
+    )
   end
 end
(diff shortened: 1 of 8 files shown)


phildarnowsky:
@lemald this still needs tests, but is feature-complete and seems to work when I exercise it manually--so I think it's complete enough to start reviewing.

lib/notifications/notification.ex
```diff
@@ -72,6 +74,7 @@ defmodule Notifications.Notification do
       if db_record.id do
         notification_with_id = %__MODULE__{notification_without_id | id: db_record.id}
         log_creation(notification_with_id)
+        link_notification_to_users(notification_with_id)
```
lemald, line 77: Might be good to put this inside of a transaction along with the `insert!` so that we don't get rows that are created but not linked to users.

lib/notifications/notification.ex
```diff
@@ -88,4 +91,26 @@ defmodule Notifications.Notification do
...
+    user_ids = User.user_ids_for_route_ids(notification.route_ids)
+
+    notification_user_maps =
+      Enum.map(user_ids, fn user_id ->
+        %{
+          notification_id: notification.id,
+          user_id: user_id,
+          state: :unread,
+          inserted_at: now,
+          updated_at: now
```
lemald, line 107: Very minor thing, but is there some way to use the current time from the database? If so that might save us a call to `utc_now` and so on.

lib/skate/settings/user.ex
```diff
@@ -14,13 +13,18 @@ defmodule Skate.Settings.User do
...
-        where: fragment("?::varchar[] && ?", ^route_ids, rs.selected_route_ids),
-        select: u.username
-      )
+    Skate.Repo.all(from(u in users_for_route_ids_query(route_ids), select: u.username))
+  end
+
+  defp users_for_route_ids_query(route_ids) do
+    from(u in DbUser,
+      join: rs in assoc(u, :route_settings),
+      where: fragment("?::varchar[] && ?", ^route_ids, rs.selected_route_ids)
```
lemald, line 27: Hmm, I'm having sort of a hard time parsing out exactly what this fragment does, could you explain it? I'm guessing it checks to see if the route IDs are contained in `selected_route_ids` somehow?

lib/notifications/notification.ex
phildarnowsky: There are current-time functions, but then we'd have to re-write this insert as a mess of SQL fragments and bindings. This way is simpler and easier to follow than that would be, IMHO.

lemald: Fair - for some reason I thought there was an easier way of accomplishing that besides fragments but if so I'm not able to dig it up currently.

lib/skate/settings/user.ex
phildarnowsky: The `&&` is a Boolean operator on arrays that returns true if the arrays have at least one element in common (https://www.postgresql.org/docs/12/functions-array.html). So this fragment gives us DB records that have at least one of the selected routes.

lemald: Ah, okay. I was just totally unfamiliar with the Postgres array syntax I guess. Thanks!

lib/notifications/notification.ex
```diff
@@ -66,26 +68,54 @@ defmodule Notifications.Notification do
...
+        }
+      end)
+
+    Skate.Repo.insert_all(
+      DbNotificationUser,
+      notification_user_maps
+    )
```
lemald, line 119: So, one other thing we could consider is actually using `users_for_route_ids_query` directly and building that into a big SQL query to insert the relevant rows, without having the relevant logic to make `notification_user_maps` on the Elixir side. That said, that's a lot of rewriting stuff by trial and error in weird Ecto query DSL form and / or fragments and we're not really trying to solve for a race condition or anything, so I think it's fine as-is.

lemald (approved):
👍

PR #951, opened 2020-12-07
Return unexpired notifications on channel join

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402591

When a user opens Skate, they get any notifications that are for routes they subscribe to, and that have an `end_time` no more than 8 hours in the past.

### DIFF
--- a/test/notifications/notification_server_test.exs
+++ b/test/notifications/notification_server_test.exs
@@ -369,7 +369,7 @@ defmodule Notifications.NotificationServerTest do
         end)
 
       assert_n_notifications_in_db(1)
-      assert String.contains?(log, "new_notification") == false
+      refute log =~ "new_notification"
     end
   end
 end
--- a/test/notifications/notification_server_test.exs
+++ b/test/notifications/notification_server_test.exs
@@ -369,7 +369,7 @@ defmodule Notifications.NotificationServerTest do
         end)
 
       assert_n_notifications_in_db(1)
-      assert log == ""
+      assert String.contains?(log, "new_notification") == false
     end
   end
 end
--- a/test/skate/application_test.exs
+++ b/test/skate/application_test.exs
@@ -2,6 +2,15 @@ defmodule Skate.ApplicationTest do
   use ExUnit.Case, async: true
   import Test.Support.Helpers
 
+  # TODO: some of these tests sometimes hit live AWS servers, causing
+  # errors such as the following:
+  #
+  #   1) test load_runtime_config replaces {:secret, var_name} with secret from AWS SecretsManager (Skate.ApplicationTest)
+  #    test/skate/application_test.exs:15
+  #    ** (ExAws.Error) ExAws Request Error!
+  #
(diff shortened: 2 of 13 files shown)


test/notifications/notification_server_test.exs
```diff
@@ -369,7 +369,7 @@ defmodule Notifications.NotificationServerTest do
         end)
 
       assert_n_notifications_in_db(1)
-      assert log == ""
+      assert String.contains?(log, "new_notification") == false
```
lemald, line 372: Is there a reason to use `String.contains?(...)` rather than `assert log =~ "thing we don't want"`?

lemald: Sorry, missing a `not` in there.

test/skate/application_test.exs
```diff
@@ -2,6 +2,15 @@ defmodule Skate.ApplicationTest do
...
 
+  # TODO: some of these tests sometimes hit live AWS servers, causing
+  # errors such as the following:
+  #
+  #   1) test load_runtime_config replaces {:secret, var_name} with secret from AWS SecretsManager (Skate.ApplicationTest)
+  #    test/skate/application_test.exs:15
+  #    ** (ExAws.Error) ExAws Request Error!
+  #
+  #    {:error, {:http_error, 400, %{body: "{\"__type\":\"AccessDeniedException\",\"Message\":\"User: arn:aws:iam::[REDACTED]:user/pdarnowsky is not authorized to perform: secretsmanager:GetSecretValue on resource: arn:aws:secretsmanager:us-east-1:[REDACTED]:secret:TEST-test-variable-REDACTED\"}", headers: [{"Date", "Mon, 07 Dec 2020 12:03:39 GMT"}, {"Content-Type", "application/x-amz-json-1.1"}, {"Content-Length", "246"}, {"Connection", "keep-alive"}, {"x-amzn-RequestId", "d2ed6c3d-c632-4eb8-9d88-439b30188920"}], status_code: 400}}}
+
```
lemald, line 13: Was this meant to be included in the PR? We might be more likely to actually address it if there's a ticket in Asana rather than a TODO in the code.

phildarnowsky: That's a better idea...at that point I just wanted to note it quickly and not break my flow, but yeah a ticket would make more sense.

lemald:
Also, general question - how picky are we usually about adding typespecs on the Elixir side in Skate? It looks like we have them in some places but not others.

phildarnowsky:
@lemald Re typespecs: Sky was better about adding them than I am, partly since I'm rather skeptical about their value, but we don't have a set policy on them. Do you have any thoughts on what the best practice might be?

lemald:
> @lemald Re typespecs: Sky was better about adding them than I am, partly since I'm rather skeptical about their value, but we don't have a set policy on them. Do you have any thoughts on what the best practice might be?

I have mixed feelings because Dialyzer is pretty "meh" as far as type systems goes and it rarely catches things proactively in my experience. That said, I do kind of like having typespecs for documentation if nothing else. I guess don't worry about it a ton for this PR but I think in the future it may be good to start requiring them at least for public functions.

phildarnowsky:
@lemald 

> I have mixed feelings because Dialyzer is pretty "meh" as far as type systems goes and it rarely catches things proactively in my experience. That said, I do kind of like having typespecs for documentation if nothing else. I guess don't worry about it a ton for this PR but I think in the future it may be good to start requiring them at least for public functions.

That's about how I feel about Dialyzer, but I think you have a good point there about typespecs' value as documentation, so let it be written, so let it be done. Wonder if we could find an automatic linter for that...

lemald (approved):
👍

PR #960, opened 2020-12-17
Hide non-revenue invalid buses

lemald:
Ticket: [Stop showing pulling back buses on route ladders using revenue logic](https://app.asana.com/0/1148853526253426/1198565919621605/f)


assets/src/components/routeLadder.tsx
```diff
@@ -130,7 +130,14 @@ const RouteLadder = ({
   }
 
   const byPosition: VehiclesByPosition = groupByPosition(
-    vehiclesAndGhosts,
+    vehiclesAndGhosts?.filter(
+      (vehicleOrGhost) =>
+        !(
```
phildarnowsky, line 135: @lemald the `!` gets visually lost here and makes the function hard to understand, introducing an intermediate variable and negating that instead would help readability.

phildarnowsky (approved):
@lemald good contingent on one minor change

assets/src/components/routeLadder.tsx
lemald: Updated.

PR #962, opened 2020-12-18
Update read/unread state of notifications

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402587

The user can now mark notifications as read or unread, and the state of them is persisted to the database.

### DIFF
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -44,11 +44,9 @@
 }
 
 .m-notification-drawer__card {
-  background-color: $color-secondary-light;
-  border: 1px solid $color-secondary-dark;
   border-radius: 4px;
-  box-shadow: $box-shadow-secondary;
   margin-bottom: 0.5rem;
+  position: relative;
   text-align: left;
   width: 100%;
 
@@ -56,4 +54,65 @@
   &:hover {
     background-color: $color-secondary-medium;
   }
+}
+
+.m-notification-drawer__card--read {
+  background-color: $white;
+  box-shadow: $subtle-box-shadow;
+}
+
+.m-notification-drawer__card--unread {
+  background-color: $color-secondary-light;
+  border: 1px solid $color-secondary-dark;
+  box-shadow: $box-shadow-secondary;
+}
+
+.m-notification-drawer__mark-all-read-link {
+  color: $color-secondary-dark;
+  position: absolute;
+  right: 0;
+}
+
+.m-notification-drawer__n-unread {
+  color: $color-component-medium;
+  text-transform: uppercase;
+}
+
+.m-notification-drawer__read-section-header {
+  margin-bottom: 0.5rem;
+  position: relative;
+}
+
+.m-notification-drawer__submenu {
+  background-color: $white;
+  border-radius: 4px;
+  bottom: 1.5rem;
+  box-shadow: 0 2px 4px 0 rgba(126, 126, 126, 0.5);
+  padding: 1rem 2rem 1rem 1rem;
+  position: absolute;
+  right: 1rem;
+  text-align: left;
+
+  a {
+    color: $color-secondary-dark;
+  }
+}
+
+.m-notification-drawer__submenu-icon {
+  bottom: 0.5rem;
+  position: absolute;
+  right: 1rem;
+
+  svg {
+    fill: $color-component-medium;
+    height: 1rem;
+    width: 1rem;
+  }
+}
+
+.m-notification-drawer__unread-section-header {
+  color: $color-component-medium;
+  margin-bottom: 0.5rem;
+  margin-top: 0.5rem;
+  text-transform: uppercase;
 }
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -88,7 +88,7 @@
   background-color: $white;
   border-radius: 4px;
   bottom: 1.5rem;
-  box-shadow: 0 2px 4px 0 rgba(126, 126, 126, 0.5);
+  box-shadow: $submenu-box-shadow;
   padding: 1rem 2rem 1rem 1rem;
   position: absolute;
   right: 1rem;
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -69,6 +69,7 @@
 
 .m-notification-drawer__mark-all-read-link {
   color: $color-secondary-dark;
+  cursor: pointer;
   position: absolute;
   right: 0;
 }
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -72,6 +72,10 @@
   cursor: pointer;
   position: absolute;
   right: 0;
+
+  &:hover {
+    font-weight: 700;
+  }
 }
 
 .m-notification-drawer__n-unread {
@@ -96,6 +100,10 @@
 
   a {
     color: $color-secondary-dark;
+
+    &:hover {
+      font-weight: 700;
+    }
   }
 }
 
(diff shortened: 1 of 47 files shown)


phildarnowsky:
@lemald clearly needs a lot more test coverage, but complete enough to start reviewing.

lib/notifications/db/notification.ex
```diff
@@ -21,6 +21,7 @@ defmodule Notifications.Db.Notification do
     field(:service_id, :string)
     field(:start_time, :integer)
     field(:end_time, :integer)
+    field(:state, :string, virtual: true)
```
lemald, line 24: Maybe I'm missing something, but what's `state` on `Notification` (as opposed to `NotificationUser`) for? The places I'm seeing so far reference the field on `NotificationUser`, which is what I'd expect.

assets/tests/testHelpers/mockHelpers.ts
```diff
@@ -21,3 +21,23 @@ export const mockUseStateOnce = <T>(mockInitialState: T): void => {
...
   spyUseState.mockImplementationOnce(() => actualUseState(mockInitialState))
 }
+
+/*
+ * Like mockUseStateOnce, but for useReducer
+/*
+export const mockUseReducerOnce = <StateT, ActionT>(
+  reducer: (state: StateT, action: ActionT) => StateT,
+  mockInitialState: StateT
+): void =>
```
lemald, line 31: Can get rid of this part of the comment.

ahakuta:
@phildarnowsky as we discussed on Slack, I'm seeing behavior where the most recent notification appears whenever a route ladder is added or removed from my route ladder view.

assets/tslint.json
```diff
@@ -10,7 +10,7 @@
   "rules": {
     "interface-name": [true, "never-prefix"],
     "jsx-no-lambda": false,
-    "object-literal-sort-keys": { "options": "match-declaration-order-only" },
+    "object-literal-sort-keys": false,
```
lemald, line 13: Are we sure we want to get rid of this? It's sort of annoying when making changes but IMO does help readability with objects that have a ton of keys.

ahakuta:
@phildarnowsky a few more observations and suggestions for improvement. I'm categorizing these as either IMPORTANT or NICE-TO-HAVE to reflect relative priority. If the NICE-TO-HAVE's are easy to implement, awesome. If they will take significant re-thinking or work, let's carve out that work as a ticket and plan on shipping without them.

IMPORTANT
- The tap/click target for the "..." menu feels pretty small. I miss it as often as I hit it on desktop. Can we make it a bit larger to make it easier for users to hit? I'd specifically want it to be a bit wide, and maybe a tiny bit taller.
Screen Shot 2020-12-18 at 1 53 38 PM

NICE-TO-HAVE
- Multiple three dot menus can be opened which feels like a somewhat weird interaction that isn't really helpful. Can we make it so that only one can be opened at a time? 
- Can we add a hover state to the three dot menu to indicate when the user is hovering over the "mark as unread" box? Maybe something as simple as bolding the text on hover, returning it to normal if it's not being hovered. We may want to do a better hover state in the future but we're designer-less right now and I think that should suffice. 
- Right now, if the three dot submenu is open, if I click anywhere else in Skate, it still remains open. Can we make it so that it closes if the user clicks anywhere else?

lib/notifications/db/notification.ex
phildarnowsky: @lemald see the SQL query in `Notifications.Notification.unexpired_notifications_for_user`. Basically, this is the only way I could find to get Ecto to generate the SQL I want, which is something like:

`SELECT notifications.*, notifications_users.state FROM notifications INNER JOIN notifications_users ON notifications.id = notifications_users.notification_id WHERE [some stuff]`

The alternative seemed to involve spelling out every column I wanted from `notifications` (i.e. all of them), which seemed both inelegant and brittle.

assets/tslint.json
phildarnowsky: So in `notificationsContext.tsx` we have:

```
interface State extends ReducerState {
  dispatch: (action: Action) => void
}
```

The `ReducerState` that we're extending has two other fields, so we end up with an interface with three fields. So far so good. But when you have a `State` in the code, there is no order of the three fields that will satisfy this linter rule. I assume that's due to a bug in the linter.

We could turn it off on a case by case basis but this `State` gets used a lot--the number of `disable` and `enable` comments I was adding was ridiculous. I figured switching off the rule entirely was a lesser evil.

lemald: Ahh, okay, now that you mention it I feel like I've run into either this or a similar linter issue before, though I forget what I did in that case. That makes sense (and I agree that switching it off case-by-case is probably worse).

lib/notifications/db/notification.ex
lemald: Ah, okay. Yeah, I don't have any good theories for other ways to accomplish that, and manually specifying ever column would be bad. As long as this field doesn't cause problems for other queries that might not join to `NotificationUser` (which I don't think it will?) I think this is fine.

assets/css/_notification_drawer.scss
```diff
@@ -57,3 +55,65 @@
...
+.m-notification-drawer__read-section-header {
+  margin-bottom: 0.5rem;
+  position: relative;
+}
+
+.m-notification-drawer__submenu {
+  background-color: $white;
+  border-radius: 4px;
+  bottom: 1.5rem;
+  box-shadow: 0 2px 4px 0 rgba(126, 126, 126, 0.5);
```
lemald, line 91: Should this color maybe get defined in `_ui_kit.scss`?

phildarnowsky:
@lemald I just pushed something that seems to fix the reappearing-notification bug.

lemald:
The fix regarding the notifications reappearing looks good to me.

At this point I think I just have the one remaining review comment, and then I'm waiting for the test coverage to improve in addition to seeing any other changes to fix the issues that Akira noted above.

ahakuta:
@phildarnowsky one thing I just noticed which you should be able to replicate on skate-dev: when a new notifications comes in, if I add a route ladder to my route ladder view, the notification card closes. 

I tested on prod and this doesn't happen (which is correct behavior). I'll try and see if I can grab a screen recording of this before Eddie takes dev over at 5pm.

phildarnowsky:
@lemald @ahakuta I just pushed something that should fix the notifications disappearing when you add a ladder. Made a small mistake fixing the last bug.

When Eddie's done with dev this morning I'll put that on there for review.

assets/css/_notification_drawer.scss
phildarnowsky: Naturally, the hardest part of doing this was thinking of an identifier.

phildarnowsky:
@ahakuta So as far as embiggening the ellipsis icon, here are three possibilities from smallest to largest, LMK what you think

Screen Shot 2020-12-22 at 7 38 06 AM
Screen Shot 2020-12-22 at 7 39 02 AM
Screen Shot 2020-12-22 at 7 39 24 AM

phildarnowsky:
@ahakuta I did the bold-on-hover thing, that was trivially easy. My instinct is to leave the other two nice-to-haves as polish tasks for post-launch. I'm going to concentrate now on raising test coverage since that's the last outstanding thing I know of here.

ahakuta:
@phildarnowsky signing off on the PR based on what I'm seeing on dev now, 3:20pm on Tuesday.

I'll file the other improvements that I noted above as new tickets- thanks for tackling the bolding on hover.

lemald (approved):
Looks good to me once you get codecov to pass!

Also don't forget to revert my reversion of the other notifications work that I did to get the Hackney upgrade out.

PR #966, opened 2020-12-23
Merge read and unread notifications into one list

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1199674046684106

### DIFF
--- a/assets/src/components/notificationDrawer.tsx
+++ b/assets/src/components/notificationDrawer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from "react"
+import React, { Dispatch, SetStateAction, useContext, useState } from "react"
 import {
   NotificationsContext,
   otherNotificationReadState,
@@ -48,6 +48,8 @@ const Content = () => {
 
   const [, stateDispatch] = useContext(StateDispatchContext)
 
+  const notificationsDispatch = useContext(NotificationsContext).dispatch
+
   const openVPPForCurrentVehicle = (notification: Notification) => {
     stateDispatch(setNotification(notification))
   }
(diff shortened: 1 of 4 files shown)


phildarnowsky:
@lemald note that I'm going to be overriding Codecov's disapproval here, if the patch coverage is 100% then how could project coverage possibly go down? I assume Codecov just had LSD for breakfast today.

assets/src/components/notificationDrawer.tsx
```diff
@@ -56,66 +58,21 @@ const Content = () => {
     return <EmptyMessage />
   }
 
-  const newNotifications = notifications.filter(
+  const nUnreadNotifications = notifications.filter(
```
lemald, line 61: Was this meant to be `newUnreadNotifications`?

lemald: Oh, I see, it's `n` as in number - maybe `unreadNotificationsCount` or something?

phildarnowsky: no, that's `n` as in "the number of". convention I picked up from somewhere a while back.

assets/src/components/notificationDrawer.tsx
```diff
@@ -130,37 +87,38 @@ const NewNotifications = ({
...
   const { dispatch } = useContext(NotificationsContext)
   const otherReadState = otherNotificationReadState(notification.state)
   return (
-    <div className="m-notification-drawer__submenu">
+    <div
+      className="m-notification-drawer__submenu"
+      onClick={
+        /* istanbul ignore next */
+        (event) => event.stopPropagation()
+      }
```
lemald, line 115: My understanding from reading this is that it's intended to prevent a click in the area outside of the `<a>` but inside of the `<div>` from propagating and firing the main `onClick` event for the card, correct? How hard would it be to add a test for this? You'd basically have to pick out the right `<div>`, fire the `onClick` event, and assert that it opens the submenu and doesn't do any of the other stuff normally associated with clicking on the card.

lemald:
> @lemald note that I'm going to be overriding Codecov's disapproval here, if the patch coverage is 100% then how could project coverage possibly go down? I assume Codecov just had LSD for breakfast today.

I think `codecov/project` on its own won't prevent you from merging if `codecov/patch` is okay? But yeah, I've frequently found the `project` numbers that Codecov reports to be confusing.

lemald:
Oh, actually, one thought on the Codecov thing - this PR does have a net reduction of lines, so even if it's 100% covered it's at least mathematically possible for it to reduce the overall coverage of the project.

phildarnowsky:
@lemald 

> I think codecov/project on its own won't prevent you from merging if codecov/patch is okay?

I think you can actually override both, but I don't remember if that's correct.

assets/src/components/notificationDrawer.tsx
phildarnowsky: probably not so hard, I suspect that at the moment I wrote that `istanbul ignore next` it was late and I was tired of Codecov's bullshit. I'll give a proper test a try.

phildarnowsky:
@lemald this should respond to all your feedback

lemald (approved):
Looks good!

PR #967, opened 2020-12-24
Mark notifications as read when opening

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1192202587402595

Note: this is based on https://github.com/mbta/skate/pull/966

### DIFF
--- a/assets/src/components/notificationDrawer.tsx
+++ b/assets/src/components/notificationDrawer.tsx
@@ -46,15 +46,15 @@ const TitleBar = () => {
 }
 
 const Content = () => {
-  const { dispatch, notifications } = useContext(NotificationsContext)
+  const { notifications } = useContext(NotificationsContext)
+  const notificationsDispatch = useContext(NotificationsContext).dispatch
+
   const currentTime = useCurrentTime()
 
   const [, stateDispatch] = useContext(StateDispatchContext)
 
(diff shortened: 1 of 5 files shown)


assets/src/components/notificationDrawer.tsx
```diff
@@ -43,79 +46,36 @@ const TitleBar = () => {
 }
 
 const Content = () => {
-  const { notifications } = useContext(NotificationsContext)
+  const { dispatch, notifications } = useContext(NotificationsContext)
   const currentTime = useCurrentTime()
 
   const [, stateDispatch] = useContext(StateDispatchContext)
 
+  const notificationsDispatch = useContext(NotificationsContext).dispatch
```
lemald, line 54: Am I missing something or is `notificationsDispatch` the same value as `dispatch` (but distinct from `stateDispatch`)?

phildarnowsky: whoops...so it is.

lemald (approved):
Looks good to me. There's the question of what to do about the inactive notification modal when opened from the drawer rather than from the ladder page, but I think that's orthogonal and can be addressed in a different PR.

phildarnowsky:
@lemald anything else on this one?

PR #977, opened 2021-01-05
fix: highlight ghost/vehicle on ladder if selected from notification

lemald:
Card: [When successfully linking to a ghost or VPP for a vehicle from a notification, highlight the applicable vehicle or ghost on the ladder](https://app.asana.com/0/1148853526253426/1199224182441183/f)

This refactors things a bit so that the `RightPanel` component no longer has a separate idea of "selected vehicle" versus "vehicle from notification," and the matter is instead handled one level up so that the vehicle selected from a notification can also be passed through as the vehicle to highlight.


phildarnowsky (approved):

ahakuta:
This lgtm, thanks @lemald!

PR #980, opened 2021-01-08
Dispatcher Note timestamps absolute if >60m

dks-mbta:
https://app.asana.com/0/812686650013698/1199671366117688/f

### DIFF
--- a/assets/src/components/notificationContent.tsx
+++ b/assets/src/components/notificationContent.tsx
@@ -17,7 +17,11 @@ export const NotificationContent = ({
           {title(notification.reason)}
         </div>
         <div className="m-notification-content__age">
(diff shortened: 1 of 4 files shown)


lemald (approved):
Looks good to me!

Since it's only `codecov/project` that's complaining but `codecov/patch` is fine, you'll still be able to merge.

If you drag the Asana ticket into the "in review" column it will tag product / design to take a look. Then, you can coordinate deploying it to dev so that one of them can take a look. Finally, once it gets their approval you can go ahead and merge. Skate-dev auto-deploys when new commits are merged to master.

molinash:
All set to go -- checked the same card from 40 to 59 minutes, then it switched to the time vs. min. Tested against iphone, tablet, desktop.

PR #981, opened 2021-01-11
Duplicate notifications to block waivers

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1199525699944668

This PR does two things:

* Notifications created going forward will have block-waiver-specific fields duplicated to the new `block_waivers` table
* Existing notifications will get gradually backfilled with the same

This is the first part of a gradual refactoring of our notifications DB schema to support polymorphic associations.

### DIFF
--- a/lib/notifications/block_waiver_backfiller.ex
+++ b/lib/notifications/block_waiver_backfiller.ex
@@ -1,0 +1,47 @@
+defmodule Notifications.BlockWaiverBackfiller do
+  use GenServer
+  import Ecto.Query
+  alias Notifications.Db.Notification, as: DbNotification
+  require Logger
+
+  @sleep_interval 5000
+
+  @batch_size 5
+
+  @spec default_name() :: GenServer.name()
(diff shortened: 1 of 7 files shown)


lib/notifications/block_waiver_backfiller.ex
```diff
@@ -0,0 +1,47 @@
...
+    notifications_requiring_backfill_count =
+      Skate.Repo.one(
+        from(n in DbNotification, select: count(n), where: is_nil(n.block_waiver_id))
+      )
+
+    Logger.warn(
+      "backfill_batch notifications_requiring_backfill_count=#{
+        notifications_requiring_backfill_count
+      }"
+    )
```
lemald, line 36: Could this maybe be `info` instead of `warn`? i worry about having too many `warn` messages to wade through when troubleshooting things.

lemald (approved):
One small comment about the logging but looks good otherwise! 🎉

PR #983, opened 2021-01-12
Don't start backfiller until after migrations

phildarnowsky:


phildarnowsky:
@lemald note that I'm planning to override codecov's complaints here, you'll see why if you look at the report, and this code is temporary anyway.

lib/skate/application.ex
```diff
@@ -34,6 +34,10 @@ defmodule Skate.Application do
 
     Migrate.up()
 
+    if Application.get_env(:skate, :start_data_processes) do
+      send(Notifications.BlockWaiverBackfiller.default_name(), :backfill_batch)
+    end
```
lemald, line 39: So, one alternative that comes to mind, which might be simpler and not require any overriding of codecov, would be to simply remove the code that starts the `Notifications.BlockWaiverBackfiller` process, deploy that, and then re-add it and deploy it once the previous version has gone out and the migration has passed. Do you have any thoughts on the relative merits of that?

phildarnowsky: @lemald that would also work, and we were indeed talking about the merits of deploying migrations on their own, but in this case I'd rather save the extra deploy, especially as this code is coming back out again very soon anyway, so I'm not as concerned about a little hackishness.

phildarnowsky: @lemald I definitely think that, going forward though, it would be a good practice to migrate separately.

lemald (approved):
👍

lib/skate/application.ex
phildarnowsky: @lemald scratch that--you can't override the codecov patch check after all. I'll go ahead and do a new PR with just the migrations.

PR #984, opened 2021-01-12
Don't start backfiller before migrations

phildarnowsky:


phildarnowsky:
@lemald this is the first PR, which should fix deploys and run migrations.

lemald (approved):
👍

PR #985, opened 2021-01-12
And now start backfiller

phildarnowsky:


phildarnowsky:
@lemald this PR is to restore the backfiller once the migrations are run

lemald:
@phildarnowsky I think this would probably be a good place to also include the fix to `default_name` from #983 since that one isn't going to get merged.

phildarnowsky:
@lemald fixed the `default_name`

lemald:
@phildarnowsky I think this also needs to be rebased to remove the first commit, which is actually the commit from the other PR?

phildarnowsky:
@lemald this should do it--just have to remember not to merge it until we actually deploy the other, which is now merged

lemald (approved):
Looks good to me with the caveat you gave about the deploy process. 👍

PR #1007, opened 2021-01-26
Write to block_waivers first and read from it instead of notifications

lemald:
Accomplishes two interrelated things:
[On backend, write to block_waivers and then copy to notifications](https://app.asana.com/0/0/1199663255721072/f)
[On backend, read data from block_waivers instead of notifications columns](https://app.asana.com/0/0/1199663255721073/f)

Note that there's some kind of ugly duplication of code between a couple of `select_merge` statements, but I don't know off-hand of a good way to factor that out with the way Ecto queries work, and in any event we're about to stop merging those two things together into one flat structure in the next subtask anyway.


lemald:
@phildarnowsky the build is currently failing on Semaphore, but that will be fixed once I rebase this on top of the migration after merging it.

phildarnowsky (approved):
looks good provided you remove the one commented-out line

lemald:
I'm going to hold off on merging this until we deploy the necessary migration (which makes the fields on `notifications` no longer required).

PR #1008, opened 2021-01-27
Dependency updates January 2021

phildarnowsky:


lemald (approved):
Looks good to me as long as we run it on dev before merging and nothing breaks.

PR #1020, opened 2021-01-29
Log warning if data status goes to :outage

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1199651591931861

One subtlety here is that, on a deploy or other server restart, it will be normal to possibly have a couple of `:outage` statuses as data is in the process of being loaded. To avoid unnecessary pages, we should check the logs after deploying to dev, to get an idea of how many messages will be generated normally, and then we'll set the threshold for alerting somewhere above that.

### DIFF
--- a/lib/realtime/data_status_alerter.ex
+++ b/lib/realtime/data_status_alerter.ex
@@ -1,0 +1,38 @@
+defmodule Realtime.DataStatusAlerter do
+  use GenServer
+  require Logger
+
+  def default_name, do: __MODULE__
+
+  @spec start_link() :: GenServer.on_start()
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    name = Keyword.get(opts, :name, default_name())
+
+    subscribe_fn =
+      Keyword.get(
+        opts,
+        :subscribe_fn,
+        &Realtime.DataStatusPubSub.subscribe/0
+      )
+
+    initial_state = %{subscribe_fn: subscribe_fn}
+
+    GenServer.start_link(__MODULE__, initial_state, name: name)
+  end
+
+  @impl GenServer
+  def init(%{subscribe_fn: subscribe_fn}) do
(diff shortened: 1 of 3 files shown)


lib/realtime/data_status_alerter.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Realtime.DataStatusAlerter do
+  use GenServer
+  require Logger
+
+  def default_name, do: __MODULE__
+
+  @spec start_link() :: GenServer.on_start()
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
```
lemald, line 8: Double `@spec`.

test/realtime/data_status_alerter_test.exs
```diff
@@ -0,0 +1,45 @@
+defmodule Realtime.DataStatusAlerterTest do
+  use ExUnit.Case
+  import(ExUnit.CaptureLog)
```
lemald, line 3: Did you mean to use the syntax with parens here? I didn't even realize that worked.

test/realtime/data_status_alerter_test.exs
```diff
@@ -0,0 +1,45 @@
...
+      assert String.contains?(log, "Data outage detected data_outage_detected")
+    end
+
+    test "logs nothing if data status is :good" do
+      log =
+        capture_log(fn ->
+          DataStatusAlerter.handle_info({:new_data_status, :good}, nil)
+        end)
+
+      assert log == ""
```
lemald, line 42: It may be more robust to `refute log =~ "data_outage_detected"` in case a warning gets logged from another process while this test is running.

test/realtime/data_status_alerter_test.exs
phildarnowsky: That's funny, I think my editor put those parens in on its own initiative.

phildarnowsky:
@lemald ready for another look

lemald (approved):
👍

PR #1035, opened 2021-02-18
Vehicle adherence setting

lemald:
Ticket: [Allow user to toggle late and early adherence colors](https://app.asana.com/0/1148853526253426/1199900940371686/f)

The ability to set the setting via the UI will come in a future ticket, but to test locally you can manually update the `user_settings` table.


lib/skate/settings/user_settings.ex
```diff
@@ -34,7 +38,8 @@ defmodule Skate.Settings.UserSettings do
           user_id: user.id,
           # defaults, which won't get written if it exists
           ladder_page_vehicle_label: :run_id,
-          shuttle_page_vehicle_label: :vehicle_id
+          shuttle_page_vehicle_label: :vehicle_id,
+          vehicle_adherence_colors: :early_red
```
phildarnowsky, line 42: A general concern I'd like to put in the back of your mind, rather than address now, is that we're developing a pattern of custom database types coupled to Elixir "types" with various boilerplate to translate. I think this is a good pattern and we should continue using it, but the boilerplate is starting to get on my nerves. Let's think sometime about how to refactor.

assets/src/models/vehicleStatus.ts
```diff
@@ -119,12 +120,18 @@ export const headwaySpacingToString = (spacing: HeadwaySpacing): string => {
...
-export const statusClass = (status: DrawnStatus): string => {
-  switch (status) {
-    case "plain":
-      return ""
-
-    default:
-      return status
+export const statusClasses = (
+  status: DrawnStatus,
+  settings: UserSettings
```
phildarnowsky, line 125: rather than passing in the full `userSettings` I'd rather we pass in just the vehicle adherence color setting here, it'll be self-documentinger that way.

phildarnowsky (changes requested):

lib/skate/settings/user_settings.ex
lemald: Hmm, yeah, good point, it is kind of a lot of boilerplate. I don't have an immediate solution either, but I'll keep it in mind.

assets/src/models/vehicleStatus.ts
lemald: FWIW I followed the same pattern used by the `vehicleLabel` helper here, but in general it's probably better to pass the minimum amount of data that the function should depend on.

phildarnowsky (approved):

phildarnowsky:
@lemald changes to add the temporary UI look OK to me

PR #1041, opened 2021-03-04
fix: display route name (rather than ID) for notifications

lemald:
Ticket: [Ensure that notifications use the same route labels as the route ladders in Skate](https://app.asana.com/0/1148853526253426/1199900940371663/f)

Screen Shot 2021-03-04 at 3 37 09 PM
fixed

### DIFF
--- a/assets/tests/components/notificationDrawer.test.tsx
+++ b/assets/tests/components/notificationDrawer.test.tsx
@@ -3,18 +3,59 @@ import React from "react"
 import renderer from "react-test-renderer"
 import NotificationDrawer from "../../src/components/notificationDrawer"
 import { NotificationsContext } from "../../src/contexts/notificationsContext"
+import { RoutesProvider } from "../../src/contexts/routesContext"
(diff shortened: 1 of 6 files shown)


assets/tests/components/notificationDrawer.test.tsx
```diff
@@ -433,3 +463,22 @@ const readNotification: Notification = {
   id: "1",
   state: "read" as NotificationState,
 }
+
+const routes: Route[] = [
```
phildarnowsky, line 467: It seems weird to me to have this definition at the end of the file rather than near the beginning.

phildarnowsky (approved):
LGTM apart from one tiny thing

molinash:
All good to go!! Approved

assets/tests/components/notificationDrawer.test.tsx
lemald: Yeah, I thought this was weird as well but did it just because some of the other test data (`notification` and `readNotification`) was already defined down here. I can just move all of it to the top of the file, though.

PR #1042, opened 2021-03-05
Chelsea bridge notifications

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1199624386814205

### DIFF
--- a/lib/notifications/bridge.ex
+++ b/lib/notifications/bridge.ex
@@ -106,7 +106,7 @@ defmodule Notifications.Bridge do
     #   or
     # {"bridge":{"id":1,"name":"Chelsea St Bridge","bridgeStatusId":{"id":2,"status":"Raised","lift_estimate":{"estimate_time":"2019-09-26 06:47:05.0"}}}}
 
-    status = get_in(response, ["bridge", "bridgeStatusId", "status"])
+    raw_status = get_in(response, ["bridge", "bridgeStatusId", "status"])
 
     lowering_time =
       response
@@ -115,10 +115,16 @@ defmodule Notifications.Bridge do
       |> do_get_datetime()
 
     Logger.info(
-      "bridge_response status=#{inspect(status)}, lowering_time=#{inspect(lowering_time)}"
+      "bridge_response status=#{inspect(raw_status)}, lowering_time=#{inspect(lowering_time)}"
     )
 
-    {status, lowering_time}
+    status =
+      case raw_status do
+        "Lowered" -> :lowered
+        "Raised" -> :raised
+      end
+
+    {status, lowering_time && DateTime.to_unix(lowering_time)}
   end
 
   defp do_get_datetime({:ok, estimate_time}) do
--- a/lib/notifications/bridge.ex
+++ b/lib/notifications/bridge.ex
@@ -3,6 +3,8 @@ defmodule Notifications.Bridge do
   Queries and maintains the status of a bridge
   """
 
+  @type bridge_movement :: {:raised, integer()} | {:lowered, nil}
+
   use GenServer
   require Logger
 
@@ -71,6 +73,7 @@ defmodule Notifications.Bridge do
     Process.send_after(pid, :update, @fetch_ms)
   end
 
+  @spec fetch_status() :: bridge_movement() | nil
   defp fetch_status() do
     headers = [{"Authorization", get_auth_header()}]
 
@@ -79,6 +82,7 @@ defmodule Notifications.Bridge do
     |> parse_response()
   end
 
+  @spec parse_response({:ok | :error, HTTPoison.Response.t()}) :: bridge_movement() | nil
   def parse_response({:ok, %HTTPoison.Response{status_code: status, body: body}})
       when status >= 200 and status < 300 do
     case Jason.decode(body) do
--- a/lib/notifications/bridge.ex
+++ b/lib/notifications/bridge.ex
@@ -6,14 +6,6 @@ defmodule Notifications.Bridge do
   use GenServer
   require Logger
 
-  defstruct bridge1_status: nil,
-            bridge1_lowering_time: []
-
-  @type t :: %__MODULE__{
-          bridge1_status: String.t() | nil,
-          bridge1_lowering_time: DateTime.t() | nil
-        }
-
   @fetch_ms 60 * 1_000
 
   @spec default_name() :: GenServer.name()
@@ -44,15 +36,30 @@ defmodule Notifications.Bridge do
 
       _ ->
         schedule_update(self())
-        {:ok, %__MODULE__{}}
+        {:ok, nil}
     end
   end
 
-  def handle_info(:update, _state) do
+  def handle_info(:update, state) do
     schedule_update(self())
     new_state = fetch_status()
 
-    {:noreply, new_state}
+    if new_state do
+      if state && state != new_state do
+        bridge_movement_fn =
+          Application.get_env(
+            :notifications,
+            :notifications_server_bridge_movement_fn,
+            &Notifications.NotificationServer.bridge_movement/1
+          )
+
+        bridge_movement_fn.(new_state)
+      end
+
+      {:noreply, new_state}
+    else
+      {:noreply, state}
+    end
   end
 
   def handle_info(msg, state) do
@@ -64,7 +71,6 @@ defmodule Notifications.Bridge do
     Process.send_after(pid, :update, @fetch_ms)
   end
 
-  @spec fetch_status() :: __MODULE__ | nil
   defp fetch_status() do
     headers = [{"Authorization", get_auth_header()}]
 
@@ -73,8 +79,6 @@ defmodule Notifications.Bridge do
     |> parse_response()
   end
 
-  @spec parse_response({:ok | :error, HTTPoison.Response.t()}) ::
-          __MODULE__ | nil
   def parse_response({:ok, %HTTPoison.Response{status_code: status, body: body}})
       when status >= 200 and status < 300 do
     case Jason.decode(body) do
--- a/lib/notifications/db/notification.ex
+++ b/lib/notifications/db/notification.ex
@@ -31,7 +31,7 @@ defmodule Notifications.Db.Notification do
     belongs_to(:block_waiver, DbBlockWaiver)
   end
 
-  def changeset(notification, attrs \\ %{}) do
+  def block_waiver_changeset(notification, attrs \\ %{}) do
     block_waiver =
       notification
       |> Map.from_struct()
--- a/lib/notifications/db/notification.ex
+++ b/lib/notifications/db/notification.ex
@@ -2,8 +2,8 @@ defmodule Notifications.Db.Notification do
   use Ecto.Schema
   import Ecto.Changeset
 
-  alias Notifications.NotificationReason
   alias Notifications.Db.BlockWaiver, as: DbBlockWaiver
+  alias Notifications.Db.BridgeMovement, as: DbBridgeMovement
   alias Notifications.Db.NotificationUser, as: DbNotificationUser
   alias Skate.Settings.Db.User, as: DbUser
 
@@ -11,17 +11,6 @@ defmodule Notifications.Db.Notification do
 
   schema "notifications" do
     field(:created_at, :integer)
-    field(:reason, NotificationReason)
-    field(:route_ids, {:array, :string})
-    field(:run_ids, {:array, :string})
-    field(:trip_ids, {:array, :string})
-    field(:operator_id, :string)
-    field(:operator_name, :string)
-    field(:route_id_at_creation, :string)
-    field(:block_id, :string)
-    field(:service_id, :string)
-    field(:start_time, :integer)
-    field(:end_time, :integer)
     field(:state, :string, virtual: true)
     timestamps()
 
@@ -29,10 +18,11 @@ defmodule Notifications.Db.Notification do
     many_to_many(:users, DbUser, join_through: DbNotificationUser)
 
     belongs_to(:block_waiver, DbBlockWaiver)
+    belongs_to(:bridge_movement, DbBridgeMovement)
   end
 
   def block_waiver_changeset(notification, attrs \\ %{}) do
-    block_waiver =
+    block_waiver_values =
       notification
       |> Map.from_struct()
       |> Map.merge(attrs)
@@ -59,10 +49,20 @@ defmodule Notifications.Db.Notification do
     ])
     |> put_assoc(
       :block_waiver,
-      DbBlockWaiver.changeset(%DbBlockWaiver{}, block_waiver)
+      DbBlockWaiver.changeset(%DbBlockWaiver{}, block_waiver_values)
     )
     |> validate_required([
       :created_at
     ])
+  end
+
+  def bridge_movement_changeset(notification, attrs \\ %{}) do
+    notification
+    |> cast(attrs, [:id, :created_at, :bridge_movement_id])
+    |> put_assoc(
+      :bridge_movement,
+      DbBridgeMovement.changeset(%DbBridgeMovement{}, attrs)
+    )
+    |> validate_required(:created_at)
   end
 end
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -153,7 +153,7 @@ defmodule Notifications.Notification do
       if bridge_movement_values.status == :raised do
         {:chelsea_st_bridge_raised, bridge_movement_values.lowering_time}
       else
-        {:chelsea_st_bridge_lowered, db_record.created_at + @notification_expiration_threshold}
+        {:chelsea_st_bridge_lowered, nil}
       end
 
     notification = %__MODULE__{
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -3,8 +3,6 @@ defmodule Notifications.Notification do
 
   import Ecto.Query
 
-  alias Schedule.Block
-  alias Schedule.Gtfs.Service
   alias Schedule.Route
   alias Schedule.Trip
   alias Schedule.Hastus.Run
@@ -18,13 +16,14 @@ defmodule Notifications.Notification do
 
   @notification_expiration_threshold 8 * 60 * 60
 
+  @bridge_lowering_blackout 120
+  @chelsea_st_bridge_route_ids ~w[112 743]
+
   @type id :: integer()
 
   @type t() :: %__MODULE__{
           id: id() | nil,
           created_at: Util.Time.timestamp(),
-          block_id: Block.id(),
-          service_id: Service.id(),
           reason: NotificationReason.t(),
           route_ids: [Route.id()],
           run_ids: [Run.id()],
@@ -40,8 +39,6 @@ defmodule Notifications.Notification do
   @derive Jason.Encoder
 
   @enforce_keys [
-    :block_id,
-    :service_id,
     :created_at,
     :reason,
     :route_ids,
@@ -53,8 +50,6 @@ defmodule Notifications.Notification do
 
   defstruct [
     :id,
-    :block_id,
-    :service_id,
     :created_at,
     :reason,
     :route_ids,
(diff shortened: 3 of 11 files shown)


assets/src/components/notificationDrawer.tsx
```diff
@@ -201,7 +201,12 @@ const NotificationCard = ({
   return (
     <button
       className={`m-notification-drawer__card m-notification-drawer__card--${notification.state}`}
-      onClick={() => openVPPForCurrentVehicle(notification)}
+      onClick={() => {
+        if (window && window.FS) {
+          window.FS.event("Chelsea bridge notification clicked")
```
lemald, line 206: Won't this trigger when any notification is clicked, not just Chelsea St Bridge notifications?

assets/src/components/notifications.tsx
```diff
@@ -64,6 +79,9 @@ export const NotificationCard = ({
       <button
         className="m-notifications__card-info"
         onClick={() => {
+          if (window && window.FS) {
+            window.FS.event("Chelsea bridge notification clicked")
```
lemald, line 83: See previous comment on `notificationDrawer.tsx`.

assets/tests/components/notifications.test.tsx
```diff
@@ -26,17 +25,20 @@ jest.mock("../../src/laboratoryFeatures", () => ({
   default: () => true,
 }))
 
+const now = new Date(123_456_789)
```
lemald, line 28: Hmm, maybe call this something different just to avoid someone looking at the file from confusing it with the `now()` helper function?

lib/notifications/bridge.ex
```diff
@@ -64,7 +71,6 @@ defmodule Notifications.Bridge do
     Process.send_after(pid, :update, @fetch_ms)
   end
 
-  @spec fetch_status() :: __MODULE__ | nil
```
lemald, line 67: Even if we can no longer use `__MODULE__`, we should still have a typespec here.

lib/notifications/bridge.ex
```diff
@@ -73,8 +79,6 @@ defmodule Notifications.Bridge do
     |> parse_response()
   end
 
-  @spec parse_response({:ok | :error, HTTPoison.Response.t()}) ::
-          __MODULE__ | nil
```
lemald, line 77: See previous comment about typespec.

lib/notifications/db/notification.ex
```diff
@@ -2,37 +2,27 @@ defmodule Notifications.Db.Notification do
...
-    field(:route_ids, {:array, :string})
-    field(:run_ids, {:array, :string})
-    field(:trip_ids, {:array, :string})
-    field(:operator_id, :string)
-    field(:operator_name, :string)
-    field(:route_id_at_creation, :string)
-    field(:block_id, :string)
-    field(:service_id, :string)
-    field(:start_time, :integer)
-    field(:end_time, :integer)
```
lemald, line 24: Feels good to see this chunk of code finally go away. 🙂

lib/notifications/notification.ex
```diff
@@ -68,18 +63,29 @@ defmodule Notifications.Notification do
...
+
+    notification_values = block_waiver_values |> Map.delete(:block_id) |> Map.delete(:service_id)
 
     db_record =
       case Skate.Repo.transaction(fn ->
              with {:ok, db_record} <- insert(changeset, on_conflict: :nothing),
                   false <- is_nil(db_record.id),
-                  notification_with_id <- %__MODULE__{notification_without_id | id: db_record.id} do
+                  notification_with_id <-
+                    struct!(__MODULE__, Map.merge(notification_values, %{id: db_record.id})) do
```
lemald, line 81: Any particular reason to use `struct!` rather than the `%__MODULE__{...}` syntax? From my quick reading of the `struct!` documentation it looks like it's equivalent.

lib/notifications/notification.ex
```diff
@@ -92,61 +98,101 @@ defmodule Notifications.Notification do
...
+                bw.start_time == ^block_waiver_values.start_time and
+                  bw.end_time == ^block_waiver_values.end_time and
+                  bw.block_id == ^block_waiver_values.block_id and
+                  bw.service_id == ^block_waiver_values.service_id and
+                  bw.reason == ^block_waiver_values.reason
+            )
+          )
+      end
+
+    struct!(__MODULE__, Map.merge(notification_values, %{id: db_record.id}))
```
lemald, line 111: See above question re: `struct!`.

lib/notifications/notification.ex
```diff
@@ -92,61 +98,101 @@ defmodule Notifications.Notification do
...
+        if existing_record do
+          {:existing_record, existing_record}
+        else
+          changeset =
+            DbNotification.bridge_movement_changeset(
+              %DbNotification{created_at: created_at},
+              bridge_movement_values
+            )
+
+          {:ok, new_record} = Skate.Repo.insert(changeset)
```
lemald, line 144: Isn't there still a race condition here? The result of the query done above in `Skate.Repo.one` could have changed, and if I'm reading correctly there's no unique index on the `bridge_movements` table.

lib/notifications/notification.ex
```diff
@@ -92,61 +98,101 @@ defmodule Notifications.Notification do
...
+          link_notification_to_users(new_record.id, @chelsea_st_bridge_route_ids)
+          {:new_record, new_record}
+        end
+      end)
+
+    {reason, end_time} =
+      if bridge_movement_values.status == :raised do
+        {:chelsea_st_bridge_raised, bridge_movement_values.lowering_time}
+      else
+        {:chelsea_st_bridge_lowered, db_record.created_at + @notification_expiration_threshold}
```
lemald, line 154: This is probably the best approach is `end_time` is a required field, though it does make me think that perhaps it ought to be optional, since this is a clear example where it doesn't make a ton of sense and we're basically just replicating the expiration logic.

lib/notifications/notification.ex
```diff
@@ -186,4 +232,41 @@ defmodule Notifications.Notification do
       notification_user_maps
     )
   end
+
+  defp convert_from_db_notification(db_notification) do
```
lemald, line 236: Add typespec

lib/notifications/notification.ex
```diff
@@ -186,4 +232,41 @@ defmodule Notifications.Notification do
...
+        block_waiver ->
+          Map.from_struct(block_waiver) |> Map.delete(:id)
+
+        bridge_movement ->
+          {reason, end_time} =
+            if bridge_movement.status == :raised do
+              {:chelsea_st_bridge_raised, bridge_movement.lowering_time}
+            else
+              {:chelsea_st_bridge_lowered,
+               db_notification.created_at + @notification_expiration_threshold}
```
lemald, line 257: Shouldn't `db_notification.end_time` already be set at this point?

lib/notifications/notification_server.ex
```diff
@@ -23,6 +23,10 @@ defmodule Notifications.NotificationServer do
     GenServer.cast(server, {:new_block_waivers, new_waivers_by_block_key})
   end
 
+  def bridge_movement(bridge_movement, server \\ default_name()) do
```
lemald, line 26: Typespec

lib/notifications/notification_server.ex
```diff
@@ -108,7 +119,7 @@ defmodule Notifications.NotificationServer do
           nil -> {nil, nil, nil}
         end
 
-      %Notification{
+      %{
```
lemald, line 122: Any particular reason for turning this from a struct to a plain map?

priv/repo/migrations/20210216142615_add_vehicle_adherence_color_to_user_setting.exs
```diff
@@ -11,6 +11,6 @@ defmodule Skate.Repo.Migrations.AddVehicleAdherenceColorToUserSetting do
       add(:vehicle_adherence_colors, :vehicle_adherence_colors, default: "early_red")
     end
 
-    execute("UPDATE user_settings SET vehicle_adherence_colors = 'early_red'", fn -> end)
+    execute("UPDATE user_settings SET vehicle_adherence_colors = 'early_red'", fn -> nil end)
```
lemald, line 14: Extraneous change to a migration that already ran (did `mix format` do this maybe?)

lib/notifications/notification_server.ex
```diff
@@ -143,6 +154,16 @@ defmodule Notifications.NotificationServer do
     end
   end
 
+  defp convert_bridge_movement_to_notification(bridge_movement) do
```
lemald, line 157: Typespec

lib/notifications/notification_server.ex
```diff
@@ -143,6 +154,16 @@ defmodule Notifications.NotificationServer do
...
+    bridge_movement
+    |> get_db_values_from_bridge_movement
+    |> Notification.get_or_create_from_bridge_movement()
+  end
+
+  defp get_db_values_from_bridge_movement({bridge_status, lowering_time}) do
```
lemald, line 163: Typespec

lib/notifications/bridge.ex
phildarnowsky: I removed these two existing typespecs because of a problem I ran into with Dialyzer. I forget the exact location, but somewhere there's code like:

```
# foo is either a Foo struct or nil
if(foo), do: process_thing(foo)

@spec process_thing(Foo.t()) :: Bar.t()
def process_thing(foo) do
  # ...do some stuff
end 
```

The problem is that `process_thing` is not speced to take a `nil` value, and unlike Typescript, Dialyzer doesn't detect that it's impossible for `process_thing` to get called with a `nil` value due to the `if` check--so you get a spurious error.

In general, I am very dubious about the value of typespecs and Dialyzer. I see it as a worst-of-both-worlds situation: you have all the extra formalities and boilerplate of type annotations, without the assurances an actual static type system gives you.

lemald:
Also, can you split the migrations off into a separate PR?

assets/src/components/notifications.tsx
phildarnowsky: Guess which part of the code I wrote directly before leaving for the day on Friday.

lib/notifications/notification.ex
phildarnowsky: Hmmm, not sure if I had anything in particular in mind or if `MODULE` would work just as well...

lib/notifications/notification.ex
phildarnowsky: There's no unique index on `bridge_movements` because we don't have anything that we can actually index on, at least for the `Lowered` status. You might think that the bridge API would say when the bridge was last lowered: it does not. You might think that it would include a `Last-Modified` header: it does not. Hence the hack referred to in the comments.

I should be able to avoid the race condition by increasing the isolation level of the transaction, though that feels kind of like using a sledgehammer to swat a fly. But it should work.

lib/notifications/notification.ex
phildarnowsky: Good point, and I notice that we only actually do use it for bridge raises...will make it optional.

lib/notifications/notification.ex
phildarnowsky: There is no `db_notification.end_time` field, the `end_time` is on `Notification` structs rather than `DbNotification`s. We do have some duplicate code here but given how small and specialized it is I don't quite think it's worth extracting.

priv/repo/migrations/20210216142615_add_vehicle_adherence_color_to_user_setting.exs
phildarnowsky: I did this on purpose to stop a warning I was getting every time I ran. I think my linter is maybe set to stricter standards than yours, we should coordinate on that.

test/notifications/notification_test.exs
```diff
@@ -191,23 +195,59 @@ defmodule Notifications.NotificationTest do
...
+        })
 
-      assert user2_notification_ids == [
-               route_1_unexpired.id,
-               route_3_unexpired.id,
-               multiroute_unexpired.id
-             ]
+      Skate.Repo.all(from(n in DbNotification, select: n.id, order_by: n.id))
+
+      Skate.Repo.all(from(bw in DbBlockWaiver, select: bw.id, order_by: bw.id))
```
lemald, line 224: Am I missing something, or are these queries basically no-ops?

lib/notifications/notification_server.ex
phildarnowsky: Functionally it would work fine as a struct, but semantically it's not a notification itself--it's a collection of values that we are proposing to construct a notification from.

test/notifications/notification_test.exs
phildarnowsky: I bet I was `IO.inspect`ing these at some point, will remove.

lib/notifications/notification.ex
lemald: Ahh, okay.

priv/repo/migrations/20210216142615_add_vehicle_adherence_color_to_user_setting.exs
lemald: Oh, interesting. Yeah, we should figure out what's up with that at some point.

lib/notifications/bridge.ex
lemald: Hmm, that's weird, I was just playing around with some examples and something similar to this passed dialyzer in my case:

```
  @spec foo(integer() | nil) :: integer()
  defp foo(i) do
    if i do
      bar(i)
    else
      0
    end
  end

  @spec bar(integer()) :: integer()
  defp bar(i) do
    i + 1
  end
```

FWIW I see the point of dialyzer as much more documentation and much less the sort of assurances you would associate with a static type system normally.

lib/notifications/notification.ex
lemald: Right, sorry, I didn't mean that there ought to be a unique index on `bridge_movements`, just pointing out that there wasn't one like with `block_waivers`.

I think it _might_ be possible to avoid increasing the isolation by doing some complicated stuff where you have an `insert` that pulls its data from a `select [data you want to add] where not exists [statement selecting any potential duplicate that you want to avoid]`. I'm not sure if one of those is actually any more performant than the other. Changing the isolation level is at least easier.

phildarnowsky: Oh, yeah, I'm just blowing off some of my frustration with the limitations of the bridge API. Upping the serialization level shouldn't have any performance impact to speak of at the scale we're doing things at.

phildarnowsky:
> Also, can you split the migrations off into a separate PR?

I'll also include the frontend code with the new Chelsea content, to help mitigate the hybrid code problem.

lib/notifications/notification.ex
phildarnowsky: I remember now, there is a reason. You can't use `__MODULE__` here because it only works with literals, else you get this kind of thing:

```
** (CompileError) lib/notifications/notification.ex:81: expected key-value pairs in a map, got: Map.merge(notification_values, %{id: db_record.id})

```

lemald: Ahh, right. And you have to use `Map.merge` rather than just `%{notification_values | id: db_record.id}` because the original map might not have an `id` key, correct?

phildarnowsky:
@lemald ready for another look

lib/notifications/notification.ex
```diff
@@ -116,6 +116,8 @@ defmodule Notifications.Notification do
 
     {:ok, {source, db_record}} =
       Skate.Repo.transaction(fn ->
+        Skate.Repo.query!("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE")
```
lemald, line 119: So if I understand [the Postgres docs](https://www.postgresql.org/docs/9.5/transaction-iso.html), the serializable isolation level will cause the transaction attempting to insert a duplicate notification to fail, which, if I'm not mistaken, will cause `Repo.transaction` to return `{:error, stuff}` rather than `{:ok, stuff}`. Currently that will cause a crash, but I think we should basically treat it the same as if the query below returns an `existing_record`. Does that make sense, or am I missing something?

phildarnowsky: I learned a number of interesting things about Postgres isolation levels this morning. Two of my takeaways are this:

1. It's probably sufficient to just use the REPEATABLE READ isolation level here, since it seems like what we want to avoid is the "phantom read" phenomenon.
2. Postgres' SERIALIZABLE is not necessarily reliable due to a known bug, fixed in Postgres' master but not yet released. See https://jepsen.io/analyses/postgresql-12.3. 

I'm not clear from that paper whether the buggy behavior is a practical problem in our case, but it seems plausible, and if we can sidestep it by using a less rigorous serialization level that seems like a win-win.

lemald: Interesting! Yeah, a lot of this stuff is new to me so I'm learning as I read, but I agree that repeatable read sounds like essentially what we want.

The Postgres docs do still say:
> Applications using this level must be prepared to retry transactions due to serialization failures.

Does that have any implications for how we want to structure the Elixir code?

phildarnowsky: Now I'm reading through these docs again and perhaps REPEATABLE READ isn't strong enough? Will do more research.

lemald: It might be worth revisiting the idea of using a nested `SELECT` inside of an `INSERT` and using a `WHERE NOT EXISTS` clause under that. I _think_ it would achieve the same thing, and it may be a pain to implement via Ecto, but there's a possibility it would be easier than this? Just a thought, I'm fine with either approach if we can make it work.

assets/src/components/notifications.tsx
```diff
@@ -30,7 +30,10 @@ export const Notifications = () => {
 
   useEffect(() => {
     if (notificationToShow && window.FS) {
-      if (notificationToShow.reason in chelseaReasons) {
+      if (
+        notificationToShow.reason === "chelsea_st_bridge_lowered" ||
+        notificationToShow.reason === "chelsea_st_bridge_raised"
+      ) {
```
lemald, line 36: What's the reason for avoiding the use of `chelseaReasons`?

test/notifications/notification_test.exs
```diff
@@ -194,30 +195,72 @@ defmodule Notifications.NotificationTest do
...
-          created_at: baseline_time - eight_hours + 1
+          created_at: baseline_time - eight_hours + 1234
         })
 
-      filled_in_bridge_unexpired = %Notification{
-        bridge_unexpired
-        | start_time: baseline_time - eight_hours + 1,
-          end_time: baseline_time + 1,
+      Skate.Repo.query!(
+        "UPDATE bridge_movements SET inserted_at = inserted_at - interval '1 hour'"
```
lemald, line 210: Oh huh I didn't realize Postgres let you do stuff like `interval '1 hour'`, that looks handy!

assets/src/components/notifications.tsx
phildarnowsky: As I've tripped over before, `in` uses `==`, which means that it is basically useless with strings--you need `===` to compare equality with strings in the sane way.

test/notifications/notification_test.exs
phildarnowsky: Postgres can do _tons_ of stuff, it's always worth a look at the function reference if you want to do something fancy. The odds of finding something useful are good.

assets/src/components/notifications.tsx
lemald: Even if you avoid `in`, can't you do something like `chelseaReasons.some((reason) => reason === notificationToShow.reason)` rather then re-writing the string values?

lib/notifications/notification.ex
phildarnowsky: One thing we could do is write a DB trigger on insert into bridge_movements that raises an error if the blackout period is violated, but if bumping the isolation level is like using a sledgehammer to swat a fly, writing a DB trigger feels like using a hydraulic press.

lemald: Haha yeah that seems like way too much.

phildarnowsky: Yeah I did an experiment with REPEATABLE READ and a couple of Postgres consoles and it's not strong enough. It's going to have to be SERIALIZABLE and we'll have to figure out the serialization anomaly question.

lemald: Huh, interesting, well good to know. For what it's worth I think the serialization anomaly shouldn't be _too_ hard to handle, but it might involve some slightly annoying rearranging of the Elixir code.

phildarnowsky:
@lemald that should address all your feedback, and coverage is green, I'll work on the rebase

assets/src/components/notifications.tsx
phildarnowsky: hmmm yes that would work

phildarnowsky:
@lemald that should _really_ respond to all your feedback, it's all rebased and ready to go

lemald (approved):
🌉 👍

phildarnowsky:
Thanks @lemald , will wait till Monday to merge and deploy this.

PR #1046, opened 2021-03-10
change tooltips to tippy.js

dks-mbta:


dks-mbta:
I still need to add the FullStory triggers. Feel free to look over these changes so far.

assets/tests/setup.tsx
```diff
@@ -11,11 +11,13 @@ jest.mock("react-leaflet-fullscreen", () => ({
...
 }))
 
-jest.mock("react-tooltip", () => ({
+jest.mock("@tippyjs/react", () => ({
   __esModule: true,
   default: jest
     .fn()
-    .mockImplementation(() => <div className="mock-react-tooltip" />),
+    .mockImplementation((props) => (
+      <div className="mock-tippy">{props.children}</div>
```
lemald, line 19: One thing that I didn't consider when suggesting this approach is that, since react-tooltip used the various `data-*` attributes on other elements, the contents of the tooltips themselves made it into the test snapshots that way. To make sure that they still make it into the snapshots with Tippy, I think you can change this to something like:
```
jest.mock("@tippyjs/react", () => ({
  __esModule: true,
  default: jest.fn().mockImplementation((props) => (
    <div className="mock-tippy">
      <div className="mock-tippy-content">{props.content}</div>
      {props.children}
    </div>
  )),
}))
```
and then rebuild the snapshots.

lemald:
One small comment about the tests, but everything else looks good! I'll just wait for the FullStory triggers to make it in before making a final approval.

lemald:
Changes look good to me. I'm basically ready to approve once the build passes on Semaphore.

dks-mbta:
@lemald Semaphore passed, though Codecov didn't. Do you have suggestions for things to test, or should we approve it?

lemald:
> @lemald Semaphore passed, though Codecov didn't. Do you have suggestions for things to test, or should we approve it?

Ah, it looks like it's just the FullStory events triggered via the `onShow` property of `Tippy` that aren't tested. Since actually triggering those would involve testing with the real Tippy tooltip, which we've decided we aren't doing, I'm fine with excepting those particular parts from the coverage. You can just add `/* istanbul ignore next */` in, I believe before the two `if (window.FS) {` lines, and it won't count those against the coverage.

molinash:
This looks good to go. Approved.

lemald:
@dks-mbta hmm, perhaps try putting the `/* istanbul ignore next */` line before the `onShow=...` lines instead / in addition?

lemald (approved):
Looks good to me!

I don't think it's worth spending too much more time on the codecov issue. Maybe try putting the `istanbul ignore next` line before the start of the `<Tippy ... />` element, but if that doesn't work I say we go ahead and override and merge.

PR #1048, opened 2021-03-12
Frontend and migrations for bridge notifications

phildarnowsky:


lemald (approved):
Compared against the other branch with `git diff` and this looks good! 👍

PR #1049, opened 2021-03-12
chore: remove unused columns on notifications table

lemald:
Ticket: [Remove block waiver attributes from notifications table](https://app.asana.com/0/1148853526253426/1199546435658878/f)

I have had this running locally and notifications are getting created without any errors. This is just the schema change, the backfiller was already removed in a #1007.


phildarnowsky (approved):
Why is deleting stuff so satisfying? Anyway, LGTM.

PR #1056, opened 2021-03-22
Block overload notifications

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1191786286510557

### DIFF
--- a/assets/src/components/notificationContent.tsx
+++ b/assets/src/components/notificationContent.tsx
@@ -67,6 +67,8 @@ export const title = (reason: NotificationReason): string => {
       return "CHELSEA ST BRIDGE RAISED"
     case "chelsea_st_bridge_lowered":
       return "CHELSEA ST BRIDGE LOWERED"
+    case "block_overload":
+      return "ADDED SERVICE"
     default:
       return reason.toUpperCase().replace("_", " ")
   }
@@ -110,6 +112,10 @@ const description = (
       )}.`
     case "chelsea_st_bridge_lowered":
       return "OCC reported that the Chelsea St bridge has been lowered."
+    case "block_overload":
+      return `An extra bus has been added to the ${
+        routeNameAtCreation || routeNames
+      }.`
     case "other":
     default:
       return `OCC created a dispatcher note for the ${routeNames}.`
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -32,7 +32,7 @@ defmodule Notifications.Notification do
           operator_name: String.t() | nil,
           route_id_at_creation: Route.id() | nil,
           start_time: Util.Time.timestamp(),
-          end_time: Util.Time.timestamp(),
+          end_time: Util.Time.timestamp() | nil,
           state: NotificationState.t() | nil
         }
 
@@ -45,7 +45,6 @@ defmodule Notifications.Notification do
     :run_ids,
     :trip_ids,
     :start_time,
-    :end_time
   ]
 
   defstruct [
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -44,7 +44,7 @@ defmodule Notifications.Notification do
     :route_ids,
     :run_ids,
     :trip_ids,
-    :start_time,
+    :start_time
   ]
 
   defstruct [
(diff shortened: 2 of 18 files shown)


phildarnowsky:
@lemald I am still waiting to hear from Akira and/or Ashli about what should cause the Fullstory event to fire, but otherwise this is ready to go.

assets/src/components/notificationContent.tsx
```diff
@@ -67,6 +67,8 @@ export const title = (reason: NotificationReason): string => {
       return "CHELSEA ST BRIDGE RAISED"
     case "chelsea_st_bridge_lowered":
       return "CHELSEA ST BRIDGE LOWERED"
+    case "block_overload":
+      return "ADDED SERVICE"
```
lemald, line 71: Random thought, but maybe we could sentence-case these and then use `text-transform: uppercase;` on the CSS end go achieve the effect we want? Not necessarily important for this ticket, especially given that we've already uppercased several other notification types here.

phildarnowsky: We could, that would work, but I don't see a strong advantage to changing it so I'd as soon leave it as it is.

lemald: Yeah, it's fine to leave as-is. That was just my over-active brain coming up with cool ideas that were largely orthogonal to the code review at hand. 🙂

test/notifications/notification_server_test.exs
```diff
@@ -538,4 +539,188 @@ defmodule Notifications.NotificationServerTest do
...
+        Skate.Repo.all(DbNotificationUser) |> Enum.map(& &1.user_id) |> Enum.sort()
+
+      expected_user_ids = Enum.sort([user1.id, user2.id])
+
+      assert actual_user_ids == expected_user_ids
+
+      assert String.contains?(log, "reason: :block_overload")
+      assert String.contains?(log, "route_ids: [\"1\", \"2\"]")
+      assert String.contains?(log, "run_ids: [\"run1\"]")
+      assert String.contains?(log, "trip_ids: [\"trip2\"]")
```
lemald, line 630: Any reason to use `String.contains?` as opposed to the `=~` operator? I don't mind, just curious if there's some relevant difference.

lemald:
@phildarnowsky the one major thing I'm seeing is that we should split the migrations out into a separate PR as usual.

Other than that I think this looks basically good, I just want to wait until it's had a chance to run on dev for a while and generate some real-world examples.

test/notifications/notification_server_test.exs
phildarnowsky: Not really no, I was copypasting existing tests to jump-start these and they use `String.contains?`, and I don't think there originally was any deep reason for using it. Probably just the first thing that would work that I thought of.

phildarnowsky:
@lemald 

> the one major thing I'm seeing is that we should split the migrations out into a separate PR as usual.

Yeah I think we should consider this standard practice henceforth, and the frontend on a case-by-case basis--which, in this case, I do think we pre-deploy it. Maybe someday we'll have a robust solution to the hybrid code problem and can stop including frontend code in the pre-deploy.

lemald:
> @lemald
> 
> > the one major thing I'm seeing is that we should split the migrations out into a separate PR as usual.
> 
> Yeah I think we should consider this standard practice henceforth, and the frontend on a case-by-case basis--which, in this case, I do think we pre-deploy it. Maybe someday we'll have a robust solution to the hybrid code problem and can stop including frontend code in the pre-deploy.

Right, the front-end concerns me less in this particular instance - worst-case scenario (as I understand it) is that one of the notifications will come through and just be formatted slightly weirdly, right?

phildarnowsky:
@lemald 

> Right, the front-end concerns me less in this particular instance - worst-case scenario (as I understand it) is that one of the notifications will come through and just be formatted slightly weirdly, right?

Pretty much yes. There are default cases for notification titles, content, etc. that should be able to accommodate unrecognized notification types without throwing an error at least.

lib/notifications/notification.ex
```diff
@@ -172,6 +171,83 @@ defmodule Notifications.Notification do
...
+               db_record
+             else
+               _ -> Skate.Repo.rollback(nil)
+             end
+           end) do
+        {:ok, db_record} when not is_nil(db_record) ->
+          db_record
+
+        {:error, _} ->
+          created_on = Map.fetch!(changeset_values, :created_on)
```
phildarnowsky, line 233: note to self: can just use the existing `created_on`, etc. variables

PR #1057, opened 2021-03-24
feat: restyle minischedules and always show origin

lemald:
Ticket: [Implement mini schedules formatting changes](https://app.asana.com/0/1148853526253426/1200074946607957/f)

Should match the designs in Figma except for the fact that (per conversation with Tyler) we're actually keeping the "Run 123-4567" header at the very top.

This also removes all of the frontend references to the minischedule trip label setting. I've created a [tech debt ticket](https://app.asana.com/0/1152340551558956/1200090125693229/f) to also clean it up on the backend, but I don't want to do that as part of this same PR due to the hybrid code issue.


phildarnowsky (approved):
LGTM assuming it passes QA

PR #1062, opened 2021-03-30
Proof-of-concept code for swings data

phildarnowsky:
@lemald I don't want to open a proper PR since we don't want to merge this code as-is into master, but I did want to get your feedback on it before I call this card done.


lemald:
@phildarnowsky this looks basically good to me! The thing (not introduced here) where the contents of the `trips` of a piece can be either trips themselves or IDs of trips isn't great, but we can dig into why that is separately.

phildarnowsky:
@lemald the thing about the IDs I think is there for a reason, Sky told me once but I forget.

PR #1066, opened 2021-04-01
Allow admin users to run reports on DB contents

lemald:
Ticket: [Allow whitelisted users to run reports against Skate DB](https://app.asana.com/0/1148853526253426/1200065191483852/f)

There are a lot of little changes, so it may help to look at this one commit-by-commit. I tried to keep them reasonably separated out.


lib/report.ex
```diff
@@ -0,0 +1,21 @@
...
+  Exectutes a report against a given repo, returning the results or an error
+  """
+  @type t :: module()
+
+  @callback run() :: {:ok, [map()]} | :error
+  @callback short_name() :: String.t()
+  @callback description() :: String.t()
+
+  @spec report_to_csv(t()) :: {:ok, String.t()} | :error
+  def report_to_csv(report) do
```
phildarnowsky, line 12: small point here, I would suggest the function just be called `to_csv`. The `report` in `report_to_csv` is redundant with the module name.

test/support/data_case.ex
```diff
@@ -32,6 +32,10 @@ defmodule Skate.DataCase do
       Ecto.Adapters.SQL.Sandbox.mode(Skate.Repo, {:shared, self()})
     end
 
+    on_exit(fn ->
+      Ecto.Adapters.SQL.Sandbox.checkin(Skate.Repo)
```
phildarnowsky, line 36: what's this about now?

lib/skate_web/controllers/report_controller.ex
```diff
@@ -0,0 +1,32 @@
...
+      |> Enum.map(fn report -> {report.description(), report.short_name()} end)
+
+    conn
+    |> assign(:reports, reports)
+    |> render("index.html", layout: {SkateWeb.LayoutView, "reports.html"})
+  end
+
+  @spec run(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def run(conn, params) do
+    report = Enum.find(Report.all_reports(), &(&1.short_name() == params["short_name"]))
```
phildarnowsky, line 22: the way we're using `find` here makes me think `Report.all_reports` should return a map rather than a list of tuples

lib/skate_web/controllers/report_controller.ex
```diff
@@ -0,0 +1,32 @@
...
+
+    if is_nil(report) do
+      send_resp(conn, 404, "no report found")
+    else
+      {:ok, results} = Report.report_to_csv(report)
+
+      send_download(conn, {:binary, results}, filename: report.short_name() <> ".csv")
```
phildarnowsky, line 29: should we consider adding a timestamp to the filename to help differentiate between files?

test/skate_web/router_test.exs
```diff
@@ -2,42 +2,28 @@ defmodule SkateWeb.RouterTest do
...
-        |> put_req_header("authorization", "bearer: " <> token)
-        |> get("/")
+      conn = get(conn, "/")
 
       assert html_response(conn, 200) =~ "div id=\"app\""
     end
   end
 
   describe "GET /search (client-side route)" do
+    @tag :authenticated
```
phildarnowsky, line 15: we should have made this a tag a long time ago

phildarnowsky (changes requested):

test/support/data_case.ex
lemald: Each time you check out a sandbox it runs everything within a transaction. Previously, we weren't checking them in in between tests, so changes from one test would accumulate and potentially interfere with another test. The options were to have each test I was writing manually delete the contents of tables that it cared about before doing anything, or to just make it so that each test gets a copy of the DB unaffected by the changes from other tests.

lib/report.ex
lemald: Good point, I'll make the change.

lib/skate_web/controllers/report_controller.ex
lemald: Yes, that's a good idea, I'll add something.

test/skate_web/router_test.exs
lemald: It's one of the things I copied over from Signs UI and it made our tests there soooo much smoother than they would have been otherwise.

lib/skate_web/controllers/report_controller.ex
lemald: Hmm, yeah, that would probably be better.

test/skate_web/router_test.exs
arkadyan: Oh wow, this is so great! I'm totally stealing it.

phildarnowsky (approved):

PR #1070, opened 2021-04-06
Downgrade stale-data errors to warning

phildarnowsky:
Card: https://app.asana.com/0/1148853526253426/1200131012280204


lemald (approved):
👍 

Could probably even be an `info` but `warn` is certainly better than `error`.

PR #1074, opened 2021-04-08
Disable vehicle hover tooltip on touch devices

dks-mbta:


lemald (approved):
Looks good to me! It's probably a good idea to put this up on dev briefly and ask Ashli to test it out on the inspector tablet (if you haven't done that already).

PR #1080, opened 2021-04-14
Backend changes to support operator first names

phildarnowsky:


phildarnowsky:
@lemald verified that with these backend changes only, the frontend is displaying last names like you would expect.

lemald (approved):
Did some `git diff` comparison and this looks good! 👍

PR #1081, opened 2021-04-15
Memoize several frontend components

phildarnowsky:
Card: https://app.asana.com/0/1200180014510248/1200166345135753

### DIFF
--- a/assets/src/components/vehicleIcon.tsx
+++ b/assets/src/components/vehicleIcon.tsx
@@ -7,7 +7,8 @@ import { DrawnStatus, statusClasses } from "../models/vehicleStatus"
 import { AlertIconStyle, IconAlertCircleSvgNode } from "./iconAlertCircle"
 import { runIdToLabel } from "../helpers/vehicleLabel"
 import { isGhost } from "../models/vehicle"
-import { VehicleOrGhost } from "../realtime.d"
+import { RunId, VehicleOrGhost } from "../realtime.d"
+import { BlockId, ViaVariant } from "../schedule.d"
 import { scheduleAdherenceLabelString } from "./propertiesPanel/header"
 
 export enum Orientation {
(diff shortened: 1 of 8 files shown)


phildarnowsky:
@lemald I can generate a proper performance profile for before and after if you want (didn't have time before standup today) but you can also compare what it feels like to use, e.g., skate-prod with 15 routes open vs. skate-dev with 45 open. I'm noticing that in such a case prod feels very sluggish compared to dev.

assets/src/components/vehicleIcon.tsx
```diff
@@ -46,35 +47,48 @@ const Y_CENTER_TO_POINT = 20
...
+        <VehicleIconSvgNode {...props} />
+      </svg>
+    )
+  }
+)
 
 export const VehicleTooltip = ({
   vehicleOrGhost,
   children,
 }: TooltipProps): ReactElement<HTMLElement> => {
```
lemald, line 80: Just curious, what's the reason for not memoizing this one directly? Does it have something to do with `children` being hard to compare for memoization purposes?

assets/src/components/vehicleIcon.tsx
```diff
@@ -173,251 +206,255 @@ export const viewBox = ({
...
+export const VehicleIconSvgNode = React.memo(
+  ({
+    size,
+    orientation,
+    label,
+    variant,
+    status,
+    alertIconStyle,
+  }: Props): ReactElement<SVGElement> => {
+    const [{ userSettings }] = useContext(StateDispatchContext)
```
lemald, line 218: Did we determine exactly how this will play along with the memoization? I guess `userSettings` won't change very often, though the rest of `StateDispatchContext` might.

lemald:
> @lemald I can generate a proper performance profile for before and after if you want (didn't have time before standup today) but you can also compare what it feels like to use, e.g., skate-prod with 15 routes open vs. skate-dev with 45 open. I'm noticing that in such a case prod feels very sluggish compared to dev.

Yeah, in terms of being convinced that this PR helps I think the subjective comparison of prod vs. dev is enough, though having some hard numbers to look at would also be cool!

assets/src/components/vehicleIcon.tsx
phildarnowsky: The logic below that calculates `label`, etc. has to live somewhere. If not here, then we'd have to repeat it in the multiple places that we call `VehicleTooltip`--not very DRY. So we're kind of stuck with `vehicleOrGhost` here, unless we want to do a much bigger refactoring, and as a rule of thumb I've assumed that `Vehicle`s and `Ghost`s will change too often to be worth memoizing.

assets/src/components/vehicleIcon.tsx
phildarnowsky: Per the `React.memo` docs:

> If your function component wrapped in React.memo has a useState, useReducer or useContext Hook in its implementation, it will still rerender when state or context change.

phildarnowsky:
@lemald it looks like deploys to dev right now are broken due to an invalidated S3 access key, which I'm guessing is fallout from the Codecov situation. I have a profile saved for dev with the optimizations, but can't deploy master to dev at the moment to get the "before" picture.

I'm about to leave for the day anyway so let's pick up the profiling question again tomorrow.

assets/src/components/vehicleIcon.tsx
lemald: Will that happen whenever anything from `StateDispatchContext` changes, or just `userSetings`? I wonder if we can get even more of an improvement by moving the `useContext` higher up, extracting `userSettings` from the state, and passing it down as a prop.

phildarnowsky: > Will that happen whenever anything from StateDispatchContext changes, or just userSetings?

That's an interesting question...I wonder if Stackoverflow would know, not sure how to phrase the search though. In any case we can probably lift the `useContext`, will give it a try.

phildarnowsky:
@lemald hoisted up the `useContext`. Deploying to skate-dev is still borked and I haven't been following that discussion closely so I don't know what has to be done to fix it, so no profiling for the moment. If you're good with the code and your subjective impression of the speedup, I'd like to get this merged in before I go on vacation.

lemald (approved):
Looks good to me! Let's see if we can resolve the dev issues, but if not I think this looks good to merge.

phildarnowsky:
@lemald /cc @molinash Dev deploys are fixed and I got the profile. My main takeaway is that, when new vehicle data for a route comes in to the frontend, and thus the frontend needs to update what it shows, that update process is now roughly 8X faster. Since the frontend has lots of these updates coming in constantly, that saves a considerable amount of CPU time that can be used to respond to user input.

PR #1091, opened 2021-04-29
Swings view beta

lemald:
Asana ticket: [⚙️ Swings view beta](https://app.asana.com/0/1200109164313812/1200108636020588/f)

This is mostly feature-complete but has some outstanding issues:
1. Some styling needs to be improved.
2. ~Fullstory events need to be added.~
3. A couple of minor bugs that Ashli found need addressing.
4. ~Front-end test coverage needs to be brought up.~

All of that said, I think the code so far is in a pretty reviewable state, just watch for subsequent pushes.


lib/schedule/swing.ex
```diff
@@ -0,0 +1,88 @@
...
+    :to_run_id,
+    :to_trip_id,
+    :time
+  ]
+
+  @type by_schedule_id_and_route_id :: %{{Service.id(), Route.id()} => [t()]}
+
+  @spec from_minischedule_blocks(Minischedule.Block.by_id(), Trip.by_id()) ::
+          by_schedule_id_and_route_id()
+  def from_minischedule_blocks(minischedule_blocks, trips_by_id) do
```
phildarnowsky, line 34: this function is a bit difficult to follow due to its bulk, I suggest extracting some helper functions here

lib/skate_web/controllers/swings_controller.ex
```diff
@@ -0,0 +1,25 @@
...
+    swings_fn = Application.get_env(:skate_web, :swings_fn, &Schedule.swings_for_route/3)
+    now_fn = Application.get_env(:skate_web, :now_fn, &Util.Time.now/0)
+
+    now = now_fn.()
+
+    swings =
+      route_ids
+      |> Enum.flat_map(fn route_id -> swings_fn.(route_id, now, now) end)
+      |> Enum.sort_by(fn swing -> {swing.from_run_id, swing.to_run_id} end)
+      |> Enum.dedup_by(fn swing -> {swing.from_run_id, swing.to_run_id} end)
```
phildarnowsky, line 21: If we're only sorting so that we can `dedup_by`, we could use `Enum.uniq` instead

test/schedule/swing_test.exs
```diff
@@ -0,0 +1,167 @@
...
+  alias Schedule.Trip
+
+  describe "from_minischedule_blocks/2" do
+    test "translates block information to swings" do
+      minischedule_blocks = %{
+        "A12-34" => %Minischedule.Block{
+          schedule_id: "a",
+          id: "A12-34",
+          pieces: [
+            %Minischedule.Piece{
```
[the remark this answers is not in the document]
phildarnowsky: Maybe even one or two more pieces.

config/config.exs
```diff
@@ -33,6 +33,8 @@ config :skate,
   redirect_http?: false,
   static_href: {SkateWeb.Router.Helpers, :static_path},
   timezone: "America/New_York",
+  swings_beta_usernames: {:system, "SWINGS_BETA_USERNAMES"},
+  beta_username_prefix: {:system, "BETA_USERNAME_PREFIX"},
```
phildarnowsky, line 37: what's with this prefix? also, you refer to it as a "suffix" in the commits message

assets/tests/hooks/useVehiclesForRunIds.test.ts
```diff
@@ -0,0 +1,144 @@
+import { renderHook } from "@testing-library/react-hooks"
+import useVehicleForRunIds from "../../src/hooks/useVehiclesForRunIds"
+import { VehicleData } from "../../src/models/vehicleData"
+import { dateFromEpochSeconds } from "../../src/util/dateTime"
+import { makeMockChannel, makeMockSocket } from "../testHelpers/socketHelpers"
+
+const vehicleData: VehicleData = {
```
phildarnowsky, line 7: we've really got to start moving towards factories, there's no reason either of us should spend time writing the literally 60 fields this object needs by hand.

phildarnowsky: not to mention the noise and bulk it adds to PRs.

assets/src/hooks/useSwings.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect, useState } from "react"
+import { fetchSwings } from "../api"
+import { Swing } from "../schedule.d"
+
+const useSwings = (): Swing[] | null => {
```
phildarnowsky, line 5: this hook is small and could be folded into `SwingsView`, removing the need to unit-test it separately from `SwingsView`

assets/src/components/swingsView.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { ByRouteId, Route, Swing } from "../schedule"
+import { selectVehicle, toggleSwingsView } from "../state"
+import {
+  formattedScheduledTime,
+  now,
+  serviceDaySeconds,
+} from "../util/dateTime"
+
+const SwingsView = (): ReactElement<HTMLElement> => {
+  const [, dispatch] = useContext(StateDispatchContext)
```
phildarnowsky, line 23: one of these days we should rename this `StateDispatchContext` and friends, and/or break it into smaller pieces, we have so many states and so many dispatch functions now

assets/src/components/swingsView.tsx
```diff
@@ -0,0 +1,177 @@
...
+      })
+    : []
+
+  const swingRunIds = uniq(
+    flatten(activeSwings.map((swing) => [swing.fromRunId, swing.toRunId]))
+  )
+
+  const { socket } = useContext(SocketContext)
+  const swingVehicles = useVehiclesForRunIds(socket, swingRunIds)
+  let swingVehiclesByRunId: ByRunId<VehicleOrGhost> = {}
```
phildarnowsky, line 39: we can eliminate this `let` here (which is almost always a good idea) by replacing the `if` with a ternary operator, turning it into an expression

assets/src/components/swingsView.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <th className="m-swings-view__table-header-cell swing-off">
+              Swing Off<div className="run-subheader">Run</div>
+              <div className="route-subheader">Route</div>
+            </th>
+            <th className="m-swings-view__table-header-cell">Vehicle No.</th>
+          </tr>
+        </thead>
+        <tbody>
+          {swings
+            .sort((swing1, swing2) => {
```
phildarnowsky, line 105: let's move this `sort` out of the template and introduce a `sortedSwings` const, to avoid having logic in the JSX as far as possible

assets/src/components/tabBar.tsx
```diff
@@ -40,6 +44,12 @@ const TabBar = ({
         break
     }
   }
+  let swingsBetaEnabled = false
```
phildarnowsky, line 47: we can get rid of this `let` too

assets/src/hooks/useVehiclesForRunIds.ts
```diff
@@ -0,0 +1,30 @@
...
+  vehicleOrGhostFromData,
+} from "../models/vehicleData"
+import { VehicleOrGhost } from "../realtime"
+import { RunId } from "../realtime"
+import { useChannel } from "./useChannel"
+
+const parser = (data: VehicleOrGhostData[]): VehicleOrGhost[] =>
+  data.map(vehicleOrGhostFromData)
+
+const useVehiclesForRunIds = (
```
phildarnowsky, line 14: this too could be folded into `SwingsView`

phildarnowsky: or rather, into `swings_view.tsx`, in a helper function

assets/src/realtime.d.ts
```diff
@@ -84,6 +84,10 @@ export type RouteStatus = "on_route" | "laying_over" | "pulling_out"
 
 export type RunId = string
 
+export interface ByRunId<T> {
+  [runId: string]: T
```
phildarnowsky, line 88: would `[runId: RunId]: T` work?

assets/tests/components/tabBar.test.tsx
```diff
@@ -31,7 +31,7 @@ describe("tabBar", () => {
     const tree = renderer
       .create(
         <BrowserRouter>
-          <TabBar pickerContainerIsVisible={true} />
+          <TabBar pickerContainerIsVisible={true} swingsViewIsVisible={false} />
```
phildarnowsky, line 46: could we make `swingsViewIsVisible` an optional prop, so that it defaults to `undefined` which is falsy, rather than add `swingsViewIsVisible={false}` everywhere?

config/config.exs
lemald: So for some reason, on the dev user pool the usernames are of the form `ActiveDirectory_jsmith` whereas on prod they're of the form `mbta-active-directory_jsmith`. And yes, if I used "suffix" in the commit message that was in error.

assets/src/hooks/useVehiclesForRunIds.ts
lemald: So the thing with `useVehiclesForRunIds` is that it's also used by the `useVehicleForNotification` hook, so even if it's small it's common to both places (I actually created this module by factoring out some of the code from `useVehicleForNotification`).

assets/src/realtime.d.ts
lemald: One would thing so, but sadly it would not:
```
src/realtime.d.ts:88:4 - error TS1336: An index signature parameter type cannot be a type alias. Consider writing '[runId: string]: T' instead.
```

assets/src/hooks/useSwings.ts
lemald: Hmm, I'm kind of of two minds on this one. On the one hand we have other hooks that are in the same modules as components that use them, and this hook is unlikely to be used by other components anytime soon. On the other hand, it's part of the data plumbing rather than helping to implement some piece of behavior on the component, and I actually personally like having that unit tested separately. I'll think about it.

assets/src/hooks/useVehiclesForRunIds.ts
phildarnowsky: ah OK I missed the other usage

assets/src/realtime.d.ts
phildarnowsky: yeah, figured you probably would have tried that...

assets/tests/components/swingsView.test.tsx
```diff
@@ -221,7 +221,16 @@ describe("SwingsView", () => {
     expect(tree).toMatchSnapshot()
   })
 
-  test("opens VPP when clicking an active run", () => {
+  test("opens VPP when clicking an active run and sends Fullstory event", () => {
+    const originalFS = window.FS
+    const originalUsername = window.username
+    window.FS = { event: jest.fn(), identify: jest.fn() }
```
phildarnowsky, line 227: this may not be the time to do it, but we have enough Fullstory custom events in the frontend that we would benefit by writing helpers for them

phildarnowsky (approved):

PR #1095, opened 2021-05-05
feat: mobile styling of swings view

lemald:
Asana ticket: [⚙️ Implement mobile design for swings beta](https://app.asana.com/0/1200109164313812/1200255331565835/f)

### DIFF
--- a/assets/css/_swings_view.scss
+++ b/assets/css/_swings_view.scss
@@ -57,7 +57,7 @@
 
   .subheader {
     position: absolute;
-    top: calc(50% - 7px / 2 + 8px);
+    top: calc(50% + 0.25rem);
     left: 1.5rem;
     font-size: 0.5rem;
     font-weight: 400;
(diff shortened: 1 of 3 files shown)


assets/css/_swings_view.scss
```diff
@@ -135,3 +135,53 @@
...
+      }
+    }
+
+    .swing-on {
+      display: none;
+    }
+
+    .swing-on-time {
+      position: absolute;
+      top: calc(50% - 7px / 2 + 8px);
```
phildarnowsky, line 163: not wild about this use of `px` in here, can we re-express in terms of `rem`?

phildarnowsky (changes requested):

assets/css/_swings_view.scss
lemald: Good call. This and some other values were copied from Figma. I re-did them in `rem` (or `rem` values that are almost exactly equivalent).

lemald: There was actually a bit of a visual difference so I decided to use the exact, if somewhat clunky, `rem` values.

phildarnowsky (approved):

PR #1106, opened 2021-05-10
Use test factory libraries

lemald:
Asana ticket: [⚙️ Decide on test factory libraries for back-end and front-end](https://app.asana.com/0/1148853526253426/1177725653917921/f)

I've gone ahead with Fishery on the front-end and ExMachina on the back-end and used each to replace some existing repetitive test data.


phildarnowsky (approved):
SO NICE

PR #1113, opened 2021-05-18
fix: run data status check every 30s, even if no updates come in

lemald:
Asana Ticket: [Make data outage logic independent of new data coming in](https://app.asana.com/0/1148853526253426/1200062807015666/f)

I considered making this a completely independent GenServer that references the ETS table, but because the actual name of the ETS table is only accessible by calling the `Realtime.Server` process to begin with, I decided to just keep it in there.

### DIFF
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -12,6 +12,8 @@ defmodule Realtime.Server do
   alias Schedule.Route
 
   alias Realtime.{
+    DataStatus,
+    DataStatusPubSub,
     Ghost,
     Vehicle,
     VehicleOrGhost
@@ -171,6 +173,8 @@ defmodule Realtime.Server do
   def init(_opts) do
     ets = :ets.new(__MODULE__, [:set, :protected, {:read_concurrency, true}])
 
+    :timer.send_interval(30_000, self(), :check_data_status)
+
     {:ok, %__MODULE__{ets: ets}}
   end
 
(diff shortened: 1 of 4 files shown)


lib/realtime/server.ex
```diff
@@ -179,6 +183,19 @@ defmodule Realtime.Server do
   def handle_info({reference, _}, %__MODULE__{} = state) when is_reference(reference),
     do: {:noreply, state}
 
+  def handle_info(:check_data_status, %__MODULE{ets: ets} = state) do
+    all_vehicles =
+      {ets, :all_vehicles}
+      |> lookup()
+      |> Enum.filter(fn vehicle_or_ghost -> match?(%Vehicle{}, vehicle_or_ghost) end)
+
+    data_status_fn = Application.get_env(:skate, :data_status_fn) || (&DataStatus.data_status/1)
```
phildarnowsky, line 192: I believe that `get_env` takes a default argument third, so `data_status_fn = Application.get_env(:skate, :data_status_fn, &DataStatus.data_status/1)`

test/realtime/server_test.exs
```diff
@@ -1,80 +1,38 @@
...
+             id: "v1",
+             label: "v1-label",
+             run_id: "123-9048",
+             operator_id: "71041",
+             operator_first_name: "FRANK",
+             operator_last_name: "FRANCIS",
+             operator_name: "FRANCIS"
+           )
+
+  @inactive_block build(:vehicle,
```
phildarnowsky, line 19: let's call this `@vehicle_on_inactive_block` to be explicit

phildarnowsky (changes requested):
Looks good apart from some small things

phildarnowsky (approved):

PR #1115, opened 2021-05-20
Allow subscribing to vehicles channel by block IDs

lemald:
Asana ticket: [Back-end changes to allow subscribing to vehicles by block ID](https://app.asana.com/0/0/1200347448522880/f)

These are just the back-end changes so that we can merge and deploy those first. I'll open another PR with the front-end changes shortly.

PR #1116, opened 2021-05-20
Block-based swings view vehicle numbers

lemald:
Asana ticket: [Make Swings View more block-oriented for determining vehicle number ](https://app.asana.com/0/1200109164313812/1200341164386661/f)

Once you review #1115 I can merge it and then TypeScript changes will be all that's left on this PR.

### DIFF
--- a/assets/tests/hooks/useVehiclesForBlockIds.test.ts
+++ b/assets/tests/hooks/useVehiclesForBlockIds.test.ts
@@ -1,73 +1,12 @@
 import { renderHook } from "@testing-library/react-hooks"
 import useVehicleForBlockIds from "../../src/hooks/useVehiclesForBlockIds"
-import { VehicleData } from "../../src/models/vehicleData"
-import { dateFromEpochSeconds } from "../../src/util/dateTime"
+import { vehicleFromData } from "../../src/models/vehicleData"
 import { makeMockChannel, makeMockSocket } from "../testHelpers/socketHelpers"
-
-const vehicleData: VehicleData = {
-  id: "y1234",
-  label: "1234",
-  run_id: "123-4567",
-  timestamp: 1600946090,
-  latitude: 42.32172773,
-  longitude: -71.11216123,
-  direction_id: 0,
-  route_id: "39",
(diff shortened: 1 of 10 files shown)


phildarnowsky:
@lemald I'll wait on reviewing this one until #1115 is merged, to help keep things straight.

assets/tests/hooks/useVehiclesForBlockIds.test.ts
```diff
@@ -0,0 +1,144 @@
+import { renderHook } from "@testing-library/react-hooks"
+import useVehicleForBlockIds from "../../src/hooks/useVehiclesForBlockIds"
+import { VehicleData } from "../../src/models/vehicleData"
+import { dateFromEpochSeconds } from "../../src/util/dateTime"
+import { makeMockChannel, makeMockSocket } from "../testHelpers/socketHelpers"
+
+const vehicleData: VehicleData = {
+  id: "y1234",
```
phildarnowsky, line 8: I see that we build `VehicleData`s at several points in the test suite as a whole, so let's make a factory for them too. Conceivably we could extract a lot of stuff out of the `Vehicle` factory and use it in both.

phildarnowsky (changes requested):

assets/tests/hooks/useVehiclesForBlockIds.test.ts
lemald: Ah, yes, I had the inkling feeling I might have forgotten an opportunity to factory-ize things.

lemald: Just pushed up some changes for this.

> Conceivably we could extract a lot of stuff out of the Vehicle factory and use it in both.
So I really _wanted_ to do this, but I couldn't really figure out how the `sequence` values would play along and how we would make sure that we could get a `Vehicle` and `VehicleData` object that we could compare like-for-like. My ultimate solution still leaves some duplication and is perhaps unsatisfying, but it's better than the way it was before.

phildarnowsky (approved):

PR #1118, opened 2021-05-27
Refactor VPP open

phildarnowsky:
This changes the code to open the VPP to take a `Vehicle` object rather than a vehicle ID. Doing so allows us to both simplify the code, and make it possible to e.g. open a vehicle when we have a `Vehicle` object in the swings view, but that vehicle is not yet in the map of active vehicles that the ladder page uses.


phildarnowsky:
@lemald note that there are still two `todo`'ed tests in there, which correspond to a feature that I'm going to make a case for removing. I'll bring that up at standup and see how that goes.

assets/src/components/appStateWrapper.tsx
```diff
@@ -3,33 +3,22 @@ import { NotificationsProvider } from "../contexts/notificationsContext"
...
-  )
-
   return (
     <StateDispatchProvider state={state} dispatch={dispatch}>
       <SocketProvider socketStatus={socketStatus}>
         <RoutesProvider routes={routes}>
           <NotificationsProvider>
-            <VehicleForNotificationProvider
-              vehicleForNotification={vehicleForNotification}
-            >
```
lemald, line 30: Getting rid of nested context providers? I'm liking this PR already.

assets/src/components/propertiesPanel.tsx
```diff
@@ -21,16 +23,26 @@ export const hideMeIfNoCrowdingTooltip = (hideMe: () => void) => {
 
 const PropertiesPanel = ({ selectedVehicleOrGhost }: Props) => {
   const [, dispatch] = useContext(StateDispatchContext)
+  const { socket } = useSocket()
+  const liveVehicle = useVehicleForId(socket, selectedVehicleOrGhost.id)
```
lemald, line 27: If we're pulling the `liveVehicle` here in the VPP itself, what's the rationale behind having the full vehicle in the state, rather than just the ID? Is it just a performance thing, so there isn't an initial lag while it fetches the `liveVehicle` for the first time?

(Maybe something further down the PR will illuminate this decision, but I wanted to flag it now while it's fresh in my mind).

assets/src/components/propertiesPanel/headwayDiagram.tsx
```diff
@@ -35,12 +35,12 @@ const minutes = (seconds: number): string => {
 
 const OtherVehicle = ({ vehicle }: { vehicle: Vehicle }) => {
   const [{ userSettings }, dispatch] = useContext(StateDispatchContext)
-  const { id, viaVariant } = vehicle
+  const { viaVariant } = vehicle
```
lemald, line 38: We should really just do the tech debt task to clean up the headway stuff.

assets/src/components/shuttleMapPage.tsx
```diff
@@ -30,18 +30,12 @@ export const allTrainVehicles = (
   trainVehiclesByRouteId: ByRouteId<TrainVehicle[]>
 ): TrainVehicle[] => flatten(Object.values(trainVehiclesByRouteId))
 
-const findSelectedVehicle = (
-  vehicles: Vehicle[],
-  selectedVehicleId: VehicleId | undefined
-): Vehicle | undefined =>
-  vehicles.find((vehicle) => vehicle.id === selectedVehicleId)
```
lemald, line 37: So good to get rid of this repetitive and confusing code.

assets/tests/hooks/usePersistedStateReducer.test.ts
```diff
@@ -89,25 +88,6 @@ describe("usePersistedStateReducer", () => {
     expect(state).toEqual(expectedState)
   })
 
-  test("stores persisted keys in localstorage when they change", () => {
```
lemald, line 92: Is localstorage still used for anything else besides the selected vehicle?

lib/realtime/server.ex
```diff
@@ -110,11 +110,16 @@ defmodule Realtime.Server do
     subscribe(server, {:block_ids, block_ids})
   end
 
-  def peek_at_vehicles(run_ids, server \\ default_name()) do
+  def peek_at_vehicles_by_run_ids(run_ids, server \\ default_name()) do
```
lemald, line 113: Hooray for better names! 🎉

assets/src/components/propertiesPanel.tsx
phildarnowsky: > Is it just a performance thing, so there isn't an initial lag

That's the basic idea. You have a perfectly good vehicle here in memory, so you can use that as your initial state before the whole subscribe-to-vehicle/respond-with-vehicle-data cycle completes.

assets/src/components/propertiesPanel/headwayDiagram.tsx
phildarnowsky: I don't really know much about the headway stuff, and I've never heard of anyone using it. We can probably make the case to finally cut it.

assets/tests/hooks/usePersistedStateReducer.test.ts
phildarnowsky: Oh...yes, I think so. I'll restore the balance of the test.

lemald:
@phildarnowsky can you separate out the back-end stuff into its own PR to do the whole song and dance of deploying that first, then the front-end stuff?

phildarnowsky:
@lemald restored that one test

lemald (approved):
Okay, I think this looks good! As we discussed we should probably hold off on deploying until Monday.

PR #1119, opened 2021-05-27
Swings type size changes

lemald:
Asana ticket: [⚙️ Implement Swings View type size refactor](https://app.asana.com/0/1200109164313812/1200341164386746/f)

### DIFF
--- a/assets/src/components/swingsView.tsx
+++ b/assets/src/components/swingsView.tsx
@@ -115,9 +115,7 @@ const SwingsTable = ({
               </div>
             </th>
             <th className="m-swings-view__table-header-cell m-swings-view__table-header-cell-swing-on">
-              <span className="m-swings-view__table-header-cell-swing-on-text">
-                Swing On
-              </span>
+              Swing On
               <div className="m-swings-view__table-header-cell-subheaders">
                 <div className="m-swings-view__table-header-cell-subheader">
                   Run
--- a/assets/src/components/swingsView.tsx
+++ b/assets/src/components/swingsView.tsx
@@ -105,19 +105,27 @@ const SwingsTable = ({
         <thead className="m-swings-view__table-header">
           <tr>
             <th className="m-swings-view__table-header-cell">
-              Swing On <span className="swing-on-time">Time</span>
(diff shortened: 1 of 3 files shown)


assets/src/components/swingsView.tsx
```diff
@@ -105,19 +105,27 @@ const SwingsTable = ({
         <thead className="m-swings-view__table-header">
           <tr>
             <th className="m-swings-view__table-header-cell">
-              Swing On <span className="swing-on-time">Time</span>
+              <div>
+                Swing On
+                <div className="subheaders">
```
phildarnowsky, line 110: @lemald let's stick to BEM conventions for class names as we mostly have throughout Skate: https://css-tricks.com/bem-101/. it's wordier but is a big help in keeping CSS organized.

phildarnowsky (changes requested):
@lemald the CSS class names aside, looks fine to me

assets/src/components/swingsView.tsx
lemald: We do actually break this convention a few places in Skate (for instance the classes `early` and `late` on vehicle icons)... that said, just because we do it one place doesn't necessarily mean it's best practice, so I'll go ahead and change this tomorrow. 🙂

phildarnowsky: It's a helpful practice, a workaround for the fact that all class names in CSS go into a global namespace effectively, so it's very easy to accidentally apply CSS rules to elements due to a collision in class names, leading to all kinds of layout messes big and small.

lemald: Hmm, that page you linked to discusses an interesting pattern for using SASS with BEM that we don't seem to take advantage of. I wonder if that's worth exploring in the future.

lemald: I also realized that I hadn't adequately cleaned up some of the old mobile-specific CSS that's now broken, and in fact it looks like I can remove most of it.

phildarnowsky (approved):
Much BEMmier. Approved.

PR #1122, opened 2021-05-28
Make swings view column headers sticky

phildarnowsky:


lemald (approved):
Looks good to me! We should just make sure to QA on tablet and mobile if you haven't done so already.

phildarnowsky:
> We should just make sure to QA on tablet and mobile if you haven't done so already.

It's the swings view--desktop only, right?

lemald:
> > We should just make sure to QA on tablet and mobile if you haven't done so already.
> 
> It's the swings view--desktop only, right?

Nah - in fact, if I had to guess, most users are on tablet, and we put together specific mobile designs and CSS for it.

phildarnowsky:
@lemald looks good on the inspector tablet, iPhone, and Android phone.

PR #1139, opened 2021-06-03
Add workflow to deploy to dev-blue

phildarnowsky:
Largely copypasta from the API dev-blue config.

I did remove one thing that the API config has, which is a call to a Slack webhook that I assume notifies the pertinent channel when a deploy is done. I actually would be in favor of adding such a hook for Skate but that seems to be out of scope of this ticket.


lemald:
Agreed on the Slack notification, that seems like a nice-to-have but we don't necessarily need to do it now. There's also the question of whether we'll want Slack notifications for deploys to every dev environment or just for prod.

lemald (approved):
I'm not (yet) an expert on GitHub Actions or ECS, but this looks good to me!

paulswartz:
I don't know why you'd remove the Slack notification, when it's part of the [workflow template](https://github.com/mbta/.github/blob/main/workflow-templates/deploy-ecs.yml) for deploying to ECS?

phildarnowsky:
@paulswartz re the notification: "I did remove one thing that the API config has, which is a call to a Slack webhook that I assume notifies the pertinent channel when a deploy is done. I actually would be in favor of adding such a hook for Skate but that seems to be out of scope of this ticket."

paulswartz:
> @paulswartz re the notification: "I did remove one thing that the API config has, which is a call to a Slack webhook that I assume notifies the pertinent channel when a deploy is done. I actually would be in favor of adding such a hook for Skate but that seems to be out of scope of this ticket."

Yep, I can see that :) I see it differently though: you removed part of the default configuration for deploying to ECS. If you removed the Credo or Formatting steps from the CI because it "wasn't part of the ticket" I'd also ask why you made that decision.

phildarnowsky:
@paulswartz There's a logic to it but never mind. I'll figure out how to set up the webhook.

phildarnowsky:
@ianwestcott @paulswartz actually, how _do_ you set up the webhook in question? That should probably go in the ECS deployment guide as well.

ianwestcott:
`SLACK_WEBHOOK` is already defined as an organizaztion secret, so you shouldn't have to configure anything. You can see that here, depending on your repo permissions: https://github.com/mbta/skate/settings/secrets/actions

phildarnowsky:
@paulswartz @ianwestcott aha...I didn't know the webhook already was defined, I thought that was a separate thing that had to be done on a per-repo basis.

phildarnowsky:
@lemald could you please set the `DOCKER_REPO` secret to the value for `skate` from the ECR dashboard at https://console.aws.amazon.com/ecr/repositories?region=us-east-1?

phildarnowsky:
@ianwestcott @paulswartz do you have anything further on this PR or are we good to merge?

paulswartz (approved):
lgtm, pending adding the `DOCKER_REPO` secret.

lemald:
While working on the deploy to prod and dev via Beanstalk, I noticed that the `Dockerfile` expects the Sentry secrets to be provided via environment variable. We'll want to make sure that those somehow get set in / before the `build-push-ecr` action.

lemald:
> While working on the deploy to prod and dev via Beanstalk, I noticed that the `Dockerfile` expects the Sentry secrets to be provided via environment variable. We'll want to make sure that those somehow get set in / before the `build-push-ecr` action.

Update: was chatting with @ianwestcott earlier and it looks like the Sentry secrets that Docker adds at build time aren't actually used anymore. Will do some testing to confirm.

phildarnowsky:
@lemald @ianwestcott I figured out where those Sentry secrets get used. If we look at this: https://github.com/mbta/skate/blob/573673307ee2d78875e707df8e8bdbdceddc14c7/assets/webpack.config.js#L16-L22 we see that Webpack loads a Sentry plugin in production (i.e. when the flag is true). That plugin uploads sourcemaps of our JS to Sentry, and by default gets its credentials through the env vars we're setting. So we need to get those variables set somehow if we still want sourcemaps to work.

phildarnowsky:
@ianwestcott @lemald I think I see how to set the env vars appropriately, will push shortly

phildarnowsky:
@lemald @ianwestcott OK, so if I understand correctly, the latest revision I just pushed should get the Sentry credentials from Secrets Manager, and we'd just have to make sure they're actually there.

lemald:
> @lemald @ianwestcott OK, so if I understand correctly, the latest revision I just pushed should get the Sentry credentials from Secrets Manager, and we'd just have to make sure they're actually there.

So, there is one remaining issue, which is that `build-push-ecr` would need to be updated to pass those in via the `--build-arg` argument to `docker build`. We still also need to solve the problem of running `upload_assets.sh`. @ianwestcott's (IMO correct) point is that running the Sentry stuff as part of the Docker build is in and of itself not super ideal.

phildarnowsky:
@lemald @ianwestcott It may not be ideal to run the Sentry stuff as part of the Docker build, but the Sentry plugin we're using is tightly coupled to Webpack. There's probably an alternative; it's not clear to me right now what that would be.

lemald:
> @lemald @ianwestcott It may not be ideal to run the Sentry stuff as part of the Docker build, but the Sentry plugin we're using is tightly coupled to Webpack. There's probably an alternative; it's not clear to me right now what that would be.

For what it's worth there is a standalone Sentry CLI thingamy (that the webpack plugin depends on) that we could invoke directly. Of course then that's more work to refactor more parts of our deploy process.

phildarnowsky:
@lemald I thought about the `sentry-cli` route but I _think_ we run into basically the same problems. Unless...what if we wrote an action to push sourcemaps to Sentry? That action could maybe run right after `build_ecs`, get the proper environment variables out of Secrets Manager, and not have to worry about passing the env vars through Docker as build args. That might work.

phildarnowsky:
@lemald @ianwestcott OK so I know that last night we ended up saying that we're going to go the modified `build-push-ecr` action route. But I did have one more idea I want to try first.

What if we add an optional `docker-additional-args` input to the global definition of `build-push-ecr`, with a blank default, like this:

```
inputs:
  docker-additional-args:
    description: Additional arguments to pass to call to docker
    required: false
    default: ''
```

And then we modify the appropriate `run` line to use those args:

`run: docker build ${{ inputs.docker_additional_args }} --pull -t ${{ steps.docker.outputs.tag }} ${{ inputs.dockerfile-path }}`

And tweak this PR correspondingly. Do you see any reason that wouldn't work?

paulswartz:
What are the additional arguments you're looking to pass? You can always run `docker build` outside of the `build-push-ecr`  task if, for example, you're trying to run a command inside the container.

phildarnowsky:
@paulswartz a number of `build_arg` options, to make the Sentry credentials available at build time as env vars, which then are used by a Webpack plugin to send sourcemaps off to Sentry. I'll look into if what you suggest would work.

phildarnowsky:
@paulswartz I think I see how to do what you suggest, but wouldn't that incur the overhead of doing an entire extra build for the sake of running one command? Am I missing something?

paulswartz:
Gotcha, that seems reasonable. I thought you were trying to run a separate command to get the sourcemaps out of the Docker container.

lemald:
@phildarnowsky I'd say that if Paul and Ian are okay with this approach, I'm okay with it too!

ianwestcott:
@phildarnowsky I like it! Let's give it a shot!

phildarnowsky:
@lemald @ianwestcott @paulswartz What I've just pushed should take care of both the outstanding issues: getting the `build_arg`s into our call to `docker build`, and having `upload_assets.sh` run and upload our static assets to S3. At least, I think so...not sure how to test it apart from trying it and seeing if it works.

Dockerfile
```diff
@@ -30,3 +30,4 @@ RUN npm ci && npm run deploy
 
 WORKDIR /root
 RUN mix do phx.digest, release
+RUN upload_assets.sh
```
lemald, line 33: Does the Docker image have the right credentials to upload to S3, or do we need to add those as well?

lemald:
> @lemald @ianwestcott @paulswartz What I've just pushed should take care of both the outstanding issues: getting the `build_arg`s into our call to `docker build`, and having `upload_assets.sh` run and upload our static assets to S3. At least, I think so...not sure how to test it apart from trying it and seeing if it works.

One question / comment, but once we resolve that I vote in favor of trying it out!

lemald:
@paulswartz looks like Ian is out today and tomorrow, any chance you could look into what AWS access key we'd need the Docker build to have in order to upload to the static content bucket? If you'd rather wait until Ian is back that's fine, it just means we won't get the deploys onto GitHub Actions until next week.

paulswartz:
It looks like Skate has access keys configured as secrets: if you want to pass them to the Docker container, can you use the new arguments you added to the `build-push-ecr` script? Alternatively, you can pull the files out of the container like I did in this Dotcom PR: https://github.com/mbta/dotcom/pull/975

PR #1147, opened 2021-06-09
Use updated Webpack option

phildarnowsky:
The recent update of Webpack caused it to break, but only in dev, due to one option being renamed.


phildarnowsky:
NVM thought this fixed it but I was wrong

PR #1165, opened 2021-06-28
build: use npm rather than nodejs-npm from APK

lemald:
We're now on Alpine 3.14 and it looks like nodejs-npm no longer
exists: https://superuser.com/a/1424979


lemald:
I don't love the use of APK for this in general so I created a [followup ticket](https://app.asana.com/0/1148853526253420/1200529988666924/f).

paulswartz:
Is there a ticket for switching Skate to a multi-stage build?

phildarnowsky (approved):

PR #1181, opened 2021-07-08
build(deps): upgrade @tippyjs/react

lemald:
Asana ticket: [⚙️ Consider updating major versions of front-end libraries](https://app.asana.com/0/1152340551558956/1200423135553550/f)

Actually a minor version but Dependabot doesn't seem to have opened a PR for it for whatever reason.

Tested on Chrome and Firefox on desktop as well as on a tablet.

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -1193,9 +1193,9 @@
       }
     },
     "@popperjs/core": {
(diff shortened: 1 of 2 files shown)


lemald:
Never mind, Dependabot actually did get to it in #1184

PR #1196, opened 2021-07-14
Add late ghosts to late view

phildarnowsky:
https://app.asana.com/0/1200180014510248/1200565967622361

### DIFF
--- a/assets/src/components/lateView.tsx
+++ b/assets/src/components/lateView.tsx
@@ -13,6 +13,20 @@ import {
 } from "../util/dateTime"
 import { runIdToLabel } from "../helpers/vehicleLabel"
 
+const compareGhosts = (a: Ghost, b: Ghost): number => {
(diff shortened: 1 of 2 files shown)


assets/src/components/lateView.tsx
```diff
@@ -24,20 +38,28 @@ const LateView = (): ReactElement<HTMLElement> => {
   const lateBusThreshold = 60 * 15
   const missingLogonThreshold = 60 * 45
 
+  const withinMissingLogonThreshold = (ghost: Ghost) =>
+    currentTime - (ghost.scheduledLogonTime as number) <= missingLogonThreshold
```
lemald, line 42: How sure are you about doing `as number` here? Should we add a case specifically for when the value is null?

lemald: Never mind, just scrolled down a couple of lines. Carry on. 🙂

lemald (approved):

PR #1197, opened 2021-07-19
Run-based ghosts

lemald:
Asana ticket: [⚙️ Switch to run-based ghosts](https://app.asana.com/0/1187728323414541/1200568590873231/f)

### DIFF
--- a/lib/realtime/ghost.ex
+++ b/lib/realtime/ghost.ex
@@ -1,7 +1,7 @@
 defmodule Realtime.Ghost do
   alias Schedule.{Block, Route, Trip}
   alias Schedule.Gtfs.{Direction, RoutePattern, StopTime}
-  alias Schedule.Hastus.Run
+  alias Schedule.Minischedule.Run
   alias Realtime.{BlockWaiver, BlockWaiverStore, RouteStatus, TimepointStatus, Vehicle}
 
   @type t :: %__MODULE__{
@@ -11,7 +11,7 @@ defmodule Realtime.Ghost do
           trip_id: Trip.id(),
           headsign: String.t(),
           block_id: Block.id(),
-          run_id: Run.id() | nil,
+          run_id: Schedule.Hastus.Run.id() | nil,
           via_variant: RoutePattern.via_variant() | nil,
           layover_departure_time: Util.Time.timestamp() | nil,
           scheduled_timepoint_status: TimepointStatus.timepoint_status(),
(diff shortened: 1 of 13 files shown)


lib/realtime/ghost.ex
```diff
@@ -53,31 +53,53 @@ defmodule Realtime.Ghost do
...
+          [trip | trips]
+
+        %Schedule.Minischedule.Piece{trips: trips} ->
+          trips
+
+        _ ->
+          []
+      end
+      # exclude AsDirected
+      |> Enum.filter(&is_binary(&1))
```
phildarnowsky, line 97: I'm not clear on the relationship between `is_binary` and as-directed trips.

lib/schedule/minischedule/load.ex
```diff
@@ -37,13 +37,37 @@ defmodule Schedule.Minischedule.Load do
...
-  def run(run_key, activities, trips, all_trips_by_block) do
+  def run(run_key, activities, trips, all_trips_by_block, trips_by_id) do
     {schedule_id, run_id} = run_key
     activities = activities || []
     trips = trips || []
 
+    service_ids =
+      trips
+      |> Enum.map(& &1.trip_id)
+      |> Enum.map(&Map.get(trips_by_id, &1))
```
phildarnowsky, line 56: `Map.take` here?

lib/schedule/minischedule/load.ex
```diff
@@ -37,13 +37,37 @@ defmodule Schedule.Minischedule.Load do
...
+      |> Enum.filter(&(!is_nil(&1)))
```
phildarnowsky, line 57: At least IMHO it's desirable to avoid `!` for negation when there's a reasonable alternative: I find it tends to get swallowed up visually, leading to confusion. Here and just below we could `Enum.reject(& is_nil(&1))`

phildarnowsky (changes requested):

lib/schedule/minischedule/load.ex
lemald: Will also need a `Map.values` but the end result will probably be more readable.

lib/realtime/ghost.ex
lemald: The type of the `trips` value on `Piece` is `[Trip.id() | Trip.t() | AsDirected.t()]`. These are non-hydrated runs so none of them will be `Trip.t()`, and the trip IDs will be binaries as opposed to structs. I guess I could do an `Enum.filter` with a `match?` which would be more verbose but hopefully more readable. Let me know what you think.

phildarnowsky: yeah I think let's do the more verbose version, it's not _terribly_ more verbose and it will be a good deal clearer

lemald:
@phildarnowsky okay, all comments should be addressed.

phildarnowsky (approved):

PR #1198, opened 2021-07-20
fix: make swings view and late view mutually exclusive

lemald:
Asana ticket: [⚙️ Make Swings and Late View mutually exclusive](https://app.asana.com/0/1148853526253420/1200610597179066/f)


phildarnowsky (approved):
nice refactoring

PR #1199, opened 2021-07-21
Replace Minischedule.Trip with Schedule.Trip

phildarnowsky:
In the process of doing so, we also make a better distinction between human-readable and non-human-readable place names.


lib/schedule/minischedule/piece.ex
```diff
@@ -89,8 +88,7 @@ defmodule Schedule.Minischedule.Piece do
         ) :: mid_route_swing()
   defp hydrate_mid_route_swing(mid_route_swing, trips_by_id, timepoint_names_by_id) do
     trip_id = mid_route_swing.trip
-    full_trip = trips_by_id[trip_id]
-    minischedule_trip = Trip.from_full_trip(full_trip, timepoint_names_by_id)
-    %{mid_route_swing | trip: minischedule_trip}
+    trip = trips_by_id[trip_id] |> Trip.set_pretty_names(timepoint_names_by_id)
```
lemald, line 91: IIRC we try to avoid single-pipe pipelines and just write out the function call conventionally instead.

lemald: _Although_, it may be preferable to use `Map.get` anyway.

phildarnowsky:
@lemald corrected the one issue, got anything else on this?

lemald (approved):
Okay, I think this looks good! Very satisfying to have part of this taken care of, and I appreciate also making the distinction between timepoint ID and human-readable timepoint name more explicit.

My one comment is to just let it run for a while (maybe overnight?) on one of the dev environments before merging, just in case something goes wrong.

phildarnowsky:
@lemald 

> My one comment is to just let it run for a while (maybe overnight?) on one of the dev environments before merging, just in case something goes wrong.

Probably a smart idea

lemald:
> @lemald
> 
> > My one comment is to just let it run for a while (maybe overnight?) on one of the dev environments before merging, just in case something goes wrong.
> 
> Probably a smart idea

You may also want to rebase now that I've merged run-based ghosts; I think there will be one or two small merge conflicts to resolve.

PR #1207, opened 2021-07-27
Dispatcher user group for Late View

lemald:
Asana ticket: [⚙️ Set up dispatcher user group ](https://app.asana.com/0/1187728323414541/1200381936995618/f)


assets/src/components/tabBar.tsx
```diff
@@ -214,4 +215,13 @@ const showAppcue = (appcueId: string): void => {
   }
 }
 
+const readDispatcherFlag = (): boolean => {
+  const data = appData()
+  if (!data || data === undefined || data.dispatcherFlag === undefined) {
```
phildarnowsky, line 220: the `data === undefined` is redundant because `!undefined` is `true`

phildarnowsky (changes requested):
good one thing aside

assets/src/components/tabBar.tsx
lemald: Now that you point it out, the `data.dispatcherFlag === undefined` is also redundant. I'll remove both.

phildarnowsky (approved):

PR #1240, opened 2021-08-18
Move Schedule.Minischedule.Run to Schedule.Run

phildarnowsky:
This eliminates the distinction, in the main backend state, between `runs` and `minischedule_runs`. With that no longer a factor, we were able to make some significant simplifications to the code.


phildarnowsky:
@lemald new stuff is from d757fd8 on

lib/schedule.ex
```diff
@@ -270,6 +269,15 @@ defmodule Schedule do
...
 
+  def handle_call(:peek, _from, {:loaded, gtfs_data} = state) do
+    {:reply, gtfs_data, state}
+  end
+
+  # You probably shouldn't use this in app code, but it is handy for debugging in the REPL
+  def peek(server \\ __MODULE__) do
+    call_catch_timeout(server, :peek, :peek, [])
+  end
+
```
lemald, line 280: Should we remove this before merging? I'd sort of prefer we not leave it if it's only useful in the REPL. Also, you can always do `:sys.get_state([process])` to accomplish basically the same thing.

lib/schedule/block.ex
```diff
@@ -1,101 +1,95 @@
...
-  In order to uniquely identify a block, you need a Service.id() in addition to a Block.id()
+  In order to uniquely identify a block, you need a Schedule.id() in addition to a Block.id()
   """
-  @type key :: {id(), Service.id()}
+  @type key :: {Hastus.Schedule.id(), id()}
   @type by_id :: %{key() => t()}
 
   @type t :: %__MODULE__{
           id: id(),
           service_id: Service.id(),
```
lemald, line 20: Is `service_id` still guaranteed to be non-nil? In practice I'm guessing it always is non-nil, but you do theoretically have the case of a block that's all nonrevenue trips that don't exist in GTFS.

lib/schedule/block.ex
```diff
@@ -1,101 +1,95 @@
...
+  @spec revenue_trips(t()) :: [Trip.t()]
+  def revenue_trips(%__MODULE__{pieces: pieces}) do
+    pieces
+    |> Enum.flat_map(& &1.trips)
+    |> Enum.reject(&match?(%AsDirected{}, &1))
+    |> Enum.filter(&Trip.is_revenue_trip?/1)
+  end
+
+  @spec get(by_id(), id(), Hastus.Schedule.id()) :: t() | nil
+  def get(by_id, schedule_id, block_idd) do
```
lemald, line 91: Is `block_idd` intentional or a typo? Also the arguments and on this line and in the typespec on the previous line seem to be in the opposite order.

lib/schedule/block.ex
```diff
@@ -133,17 +127,17 @@ defmodule Schedule.Block do
...
         # Block is scheduled to have finished
-        List.last(block.trips)
+        block |> revenue_trips |> List.last()
 
       true ->
         # Either the current trip or the trip that is about to start
         # If it's between the end of the last trip and the end of the block, use the last trip
-        Enum.find(block.trips, fn trip -> trip.end_time > now end) ||
-          List.last(block.trips)
+        Enum.find(revenue_trips(block), fn trip -> trip.end_time > now end) ||
```
lemald, line 139: Can pipeline this (`block |> revenue_trips |> Enum.find(...)`).

lib/schedule/hastus/activity.ex
```diff
@@ -75,4 +82,352 @@ defmodule Schedule.Hastus.Activity do
...
+  @spec activities_to_breaks([__MODULE__.t() | Piece.t()]) :: [Break.t() | Piece.t()]
+  defp activities_to_breaks(activities_and_pieces) do
+    Enum.map(activities_and_pieces, fn
+      %Piece{} = piece -> piece
+      %__MODULE__{} = activity -> Break.from_activity(activity)
+    end)
+  end
+
+  @spec trip_in_operator(__MODULE__.t(), Hastus.Trip.t()) :: boolean()
+  defp trip_in_operator(%__MODULE__{activity_type: "Operator"} = activity, trip) do
```
lemald, line 324: Should this actually be `trip_in_operator?`?

lib/schedule/hastus/activity.ex
```diff
@@ -75,4 +82,352 @@ defmodule Schedule.Hastus.Activity do
...
+    String.contains?(trip.block_id, activity.partial_block_id) and
+      trip.start_time >= activity.start_time and
+      trip.start_time <= activity.end_time
+  end
+
+  @spec operator_is_as_directed?(__MODULE__.t()) :: boolean()
+  defp operator_is_as_directed?(%__MODULE__{activity_type: "Operator"} = activity) do
+    String.contains?(activity.partial_block_id, "ad")
```
lemald, line 332: Would it make sense to just explicitly test for `rad` or `wad` rather than doing this "match anything that contains `ad`" logic?

lemald:
> @lemald new stuff is from [d757fd8](https://github.com/mbta/skate/commit/d757fd89e0a0f63e39ec8d3cb7fff2e9b337cc4f) on

...I may have missed this comment and just looked through the whole thing. 🤦‍♂️ 

In any event, I left a bunch of small comments, but overall this is looking good!

lib/schedule.ex
phildarnowsky: I was thinking of leaving it in, but then, I didn't know about `get_state`, so will remove.

lib/schedule/block.ex
phildarnowsky: So I can't 100% say "guaranteed", but I can say I did look through the data, and saw no such blocks. If you think about it, that would correspond to pulling the bus out, driving around for a while without picking up a single passenger, then pulling back to the garage, and I don't see why we would ever do that. Maybe @dks-mbta could say definitively if there's a valid reason to have a block like that.

Generally, apart from non-revenue trips, there is one other important case to be aware of where a trip might have a nil `service_id` (maybe you're aware already). That's when the trip is in the data but is for a rating other than the current one, and we do filter that out via `Piece.from_non_current_rating?/1`, which also has a big comment by it explaining why those trips are there.

lib/schedule/block.ex
phildarnowsky: Typo. And this is why I can't wait for Gleam to be production-ready, will fix the spec.

lib/schedule/hastus/activity.ex
phildarnowsky: makes sense

lib/schedule/block.ex
lemald: Yeah, I guess this gets at a question of whether our types should reflect the desired shape of the data or, like, all possible values of the data. Perhaps in a more strongly-typed language we would solve this by having a function from a "all possible states of the input" type to a "desired internal data representation OR error" type that does the relevant parsing, but I suppose we're not writing in such a language and, as you point out, Gleam is not yet production ready. I guess this is fine as-is.

lib/schedule/piece.ex
```diff
@@ -53,42 +51,27 @@ defmodule Schedule.Piece do
...
+  #
+  # These trips will have a nil service_id. So do all non-revenue trips, but
+  # since a piece should always have at least one revenue trip, if all trips
+  # in a piece have a nil service_id, we can assume that all those trips are
+  # from a non-current rating.
+  @spec from_non_current_rating?(t()) :: boolean()
+  def from_non_current_rating?(piece) do
+    piece.trips
+    |> Enum.reject(&match?(%AsDirected{}, &1))
+    |> Enum.all?(&is_nil(&1.service_id))
```
lemald, line 75: Question: I know trips from a past rating won't have a `service_id` (at least once the old trips are gone from GTFS), but what about trips from a future rating (once it's in GTFS at least?). There are two possibilities here:
1. These trips will have a `service_id`, in which case `from_non_current_rating?` is maybe a _slightly_ weird name... `from_non_past_rating?` perhaps? Though as I pointed out trips from a future rating might not have a `service_id` _yet_ if the rating hasn't hit GTFS...
2. These trips won't have a `service_id`, in which case we could have a problem with how the application handles rolling over on the first day of a new rating, and we'd somehow have to trigger a re-load of the schedule data.

phildarnowsky:
@lemald this should address all your feedback except for the good point about _future_ ratings, which I'll have to think about a minute

lib/schedule/piece.ex
phildarnowsky: excellent question, @dks-mbta would you know this one too?

lemald: Well, I think this is more of a question about how Skate processes the data than it is about the HASTUS export or GTFS itself, right? We know that HASTUS will include the trips for the present and future rating, and we know that the trips for the future rating will (eventually) be in GTFS, so it's a question of whether or not they get paired up with a service ID like trips in the current rating do.

phildarnowsky: OK I think I misunderstood your question at first. But suppose that today is during rating N and the data for rating N+1 just got added to GTFS, Skate rebooted and ingested the new data, and those trips don't have `service_id`s. Those trips will get filtered out for now: fine. Unless I'm missing something, though, at some point before rating N+1 actually starts, someone is going to have to update the GTFS so that these trips _do_ have `service_id`s, because all revenue trips for a current rating are supposed to have one.

So we really only have to consider case #1 there, and a name change makes sense then. Does that all add up, or do you see anything wrong with my reasoning?

lemald: > Unless I'm missing something, though, at some point before rating N+1 actually starts, someone is going to have to update the GTFS so that these trips do have service_ids, because all revenue trips for a current rating are supposed to have one.

I actually think the trips will have a service ID in the GTFS from the start (it's a required field per [the spec](http://gtfs.org/reference/static#tripstxt)). In which case... I think we're fine? Again, the main way it could go wrong is on the Skate end if we somehow filter some of that information out, but I don't believe we currently do.

phildarnowsky: I'll change the name anyhow, that makes sense to me.

phildarnowsky:
@lemald @dks-mbta per our conversation about as-directed trips, I do see one place that might not be handled correctly, will investigate.

phildarnowsky:
@lemald added a few more commits for your review

PR #1264, opened 2021-09-14
Stacked late view, and related CSS changes

phildarnowsky:
Card: https://app.asana.com/0/1200180014510248/1200941478186880


assets/src/components/lateView.tsx
```diff
@@ -5,11 +5,9 @@ import { VehiclesByRouteIdContext } from "../contexts/vehiclesByRouteIdContext"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import {
   bangIcon,
-  busBangIcon,
   lateViewGhostIcon,
   lateViewGhostWithWaiverIcon,
   upRightIcon,
-  userXIcon,
```
lemald, line 12: *question:* were these the only uses of these SVGs? If so, can we remove them from the codebase?

assets/src/components/lateView.tsx
```diff
@@ -86,10 +84,9 @@ const LateView = (): ReactElement<HTMLElement> => {
       <div className="m-late-view__title">Late View</div>
       <div className="m-late-view__panels">
         <div className="m-late-view__panel m-late-view__missing-logons">
-          <h2 className="m-late-view__panel-header m-late-view__missing-logons-panel-header">
-            {userXIcon("m-late-view__panel-header-icon")}
+          <div className="m-late-view__panel-header m-late-view__missing-logons-panel-header">
```
lemald, line 87: **question:** is the move from using an `h2` to a `div` just to avoid some styling we apply to `h2` elements elsewhere? I'd kind of rather keep the semantically-correct element here if possible.

assets/src/components/lateView.tsx
phildarnowsky: I believe so, and agreed, if we need them later we can always pull them out of the git history.

assets/src/components/lateView.tsx
phildarnowsky: IIRC yes but you're right, it would be better to keep the semantics clean.

phildarnowsky:
@lemald brought flush with your feedback

lemald (approved):
👍

lemald:
Latest changes still look good to me. 👍

PR #1276, opened 2021-09-23
hide late view rows

phildarnowsky:


assets/src/components/lateView.tsx
```diff
@@ -261,4 +444,72 @@ const MissingLogonRow = ({
...
+  unhideRecentlyHidden,
+  clearRecentlyHidden,
+}: {
+  nRecentlyHidden: number
+  unhideRecentlyHidden: () => void
+  clearRecentlyHidden: () => void
+}): ReactElement<HTMLElement> => {
+  const ref = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
```
phildarnowsky, line 488: @lemald this is the start of the code I've been referring to that's giving me trouble.

lemald: Hmm, no immediate theories come to mind but I will definitely keep thinking about it.

phildarnowsky: @lemald cool, thanks for looking, this is puzzling because I directly stole this code from elsewhere in the project (`notificationDrawer.tsx` IIRC). Anyway, I proposed an alternative interaction to Tyler that should be more straightforward, and he said it would work, so change to come here.

phildarnowsky:
@lemald the styling here is still ugly, but the logic is mostly ready for review in case you want to get a head start on that. There is one bug that I'm aware of and am looking for, where hidden runs are getting promoted to the permanently-hidden list shortly after getting hidden, rather than after 45 minutes.

phildarnowsky:
@lemald as discussed, here is the feature-complete but ugly version, styling to follow

assets/src/components/lateView.tsx
```diff
@@ -36,8 +58,187 @@ const compareGhosts = (a: Ghost, b: Ghost): number => {
   return runIdToLabel(a.runId!).localeCompare(runIdToLabel(b.runId!))
 }
 
+const idIn = (runId: RunId, runIds: RunId[]): boolean =>
+  runIds.some((idInList) => idInList === runId)
```
lemald, line 62: *suggestion (non-blocking):* `runIdIn` may be a better name if it's specific to that type.

assets/src/components/lateView.tsx
```diff
@@ -261,4 +528,112 @@ const MissingLogonRow = ({
...
+  attachedIds,
+}: {
+  attachedIds: RunId[]
+}): ReactElement<HTMLElement> => {
+  const { selectedIds, setSelectedIds } = useContext(LateViewContext)
+
+  const selectedAttachedIds = selectedIds.filter((id) =>
+    attachedIds.includes(id)
+  )
+  const selectionState =
```
lemald, line 605: **suggestion (non-blocking):** this is probably typed as `String` and, if we wanted to be super pedantic, could be explicitly typed as `"none" | "all" | "some"`.

assets/src/components/lateView.tsx
```diff
@@ -261,4 +526,104 @@ const MissingLogonRow = ({
   )
 }
 
+const HideCheckbox = ({
+  vehicleOrGhost,
+}: {
+  vehicleOrGhost: VehicleOrGhost
+}): ReactElement<HTMLElement> | null => {
+  const { selectedIds, toggleCheckedState } = useContext(LateViewContext)
```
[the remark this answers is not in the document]
phildarnowsky: I think I overestimated how much prop drilling the context would save, will pass arguments the regular way.

assets/tests/components/lateView.test.tsx
```diff
@@ -175,4 +182,602 @@ describe("LateView", () => {
...
+      wrapper
+        .find('.m-late-view__data-row--selected input[type="checkbox"]')
+        .first()
+        .simulate("click")
+    })
+    expect(
+      wrapper.find(
+        '.m-late-view__late-buses .m-late-view__data-row--selected input[type="checkbox"]'
+      )
+    ).toHaveLength(1)
```
lemald, line 419: **suggestion:** I think it would also make sense to make an assertion about the state of the master checkbox after unchecking the individual late row.

assets/tests/components/lateView.test.tsx
```diff
@@ -175,4 +182,602 @@ describe("LateView", () => {
...
+    expect(wrapper.find(".m-late-view__data-row--unselected")).toHaveLength(2)
+
+    act(() => {
+      wrapper
+        .find(".m-late-view__unhide-popup button")
+        .first()
+        .simulate("click")
+    })
+    expect(wrapper.find(".m-late-view__unhide-popup button")).toHaveLength(0)
+    expect(wrapper.find(".m-late-view__data-row--selected")).toHaveLength(0)
```
lemald, line 611: **suggestion:** Also add some sort of assertion that the rows were in fact unhidden (which, if I'm not mistaken, is basically that there should be 2 unselected rows?).

lemald:
Okay, looking pretty good! There are a couple of comments I left in previous reviews that I don't _think_ are resolved but feel free to correct me if I'm wrong about that. I'll sleep on the props / context thing as mentioned and take another look tomorrow.

assets/src/components/lateView.tsx
```diff
@@ -294,148 +281,163 @@ const LateView = (): ReactElement<HTMLElement> => {
...
-                    <MissingLogonRow
-                      ghost={missingLogon}
-                      key={missingLogon.id}
+    <div className="m-late-view">
+      <div className="m-late-view__content-wrapper">
+        <div className="m-late-view__title">Late View</div>
+        <div className="m-late-view__panels">
+          <div className="m-late-view__panel m-late-view__missing-logons">
+            <h2 className="m-late-view__panel-header m-late-view__missing-logons-panel-header">
+              Missing logons{" "}
```
lemald, line 290: Extraneous `{" "}` (probably introduced by prettier).

lemald, line 290: Extraneous `{" "}` (probably inserted by prettier).

lemald: Also appears in at least one other spot lower down.

lemald (approved):
🎉

lemald:
@phildarnowsky latest commit is looking good!

PR #1281, opened 2021-09-27
Poll schedule data periodically, not just at application startup

lemald:
Asana ticket: [⚙️  Have Skate proactively poll for and load new GTFS](https://app.asana.com/0/1152340551558956/1200443690692995/f)

### DIFF
--- a/lib/schedule/gtfs/fetcher.ex
+++ b/lib/schedule/gtfs/fetcher.ex
@@ -9,18 +9,21 @@ defmodule Schedule.Gtfs.Fetcher do
   }
 
   @default_opts [
-    poll_interval_ms: 60 * 60 * 1_000,
+    poll_interval_ms: 5 * 60 * 1_000,
     health_server: Health.Server.default_server(),
     updater_function: &Schedule.update_state/1,
     files_source: :remote
   ]
+
+  @type files_source :: :remote | {:mocked_files, Schedule.mocked_files()}
 
   @type state :: %{
           poll_interval_ms: integer(),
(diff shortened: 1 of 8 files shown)


lemald:
@phildarnowsky I've still got to do some work on test coverage at least, and maybe refactor things a bit more (the new module should really be `Schedule.Fetcher` rather than `Schedule.GTFS.Fetcher`), but I figured I'd let you start taking a look.

lib/schedule/gtfs/fetcher.ex
```diff
@@ -164,28 +196,72 @@ defmodule Schedule.Gtfs.Fetcher do
...
+    case HTTPoison.get(
+           url,
+           if !is_nil(latest_timestamp) do
+             [{"if-modified-since", latest_timestamp}]
+           else
+             []
+           end
+         ) do
+      {:ok, %HTTPoison.Response{status_code: 200, body: zip_binary, headers: headers}} ->
+        last_modified =
```
phildarnowsky: **suggestion**: clarify using `find_value`

A more straightforward way to write this would be:

```
last_modified = Enum.find_value(headers, fn {header, value} -> if String.downcase(header) == "last-modified", do: value end)
```
```

phildarnowsky (approved):
**praise:** overall very good, just some minor stylistic beefs.

I'm approving this one before changes are made on the theory that the changes I suggest are pretty straightforward, meaning that if you don't implement them before I go on vacation, it's probably fine to merge them without me re-reviewing.

PR #1315, opened 2021-11-02
Foundational work for presets and tabs

lemald:
Asana ticket: [⚙️ Live prototype: Implement logic + interactions for Presets tab ](https://app.asana.com/0/1200180014510248/1201087271304728/f)

This PR contains a few somewhat-disparate pieces of groundwork for presets and tabs, and the commits should probably be reviewed separately. ~~One of them is a database migration, but since the code that references these database fields isn't written yet, I feel okay not separating it out into its own PR.~~ Never mind, there are uses of the new tables. I have reopened #1311 as a separate PR. Probably the meatiest commit is the last one, which adds the basics of tab interactions to the front-end. Future PRs will sync this up with the tab data on the back-end, allow saving as presets, etc.

### DIFF
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -58,13 +58,13 @@ const LadderTab = ({
   )
 }
 
-const AddLadderButton = ({
-  addLadder,
+const AddTabButton = ({
+  addTab,
 }: {
-  addLadder: () => void
+  addTab: () => void
 }): ReactElement<HTMLDivElement> => {
   return (
-    <div className="m-ladder-page__add-tab-button" onClick={addLadder}>
+    <div className="m-ladder-page__add-tab-button" onClick={addTab}>
       +
     </div>
   )
@@ -153,7 +153,7 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
           />
         ))}
 
-        <AddLadderButton addLadder={() => dispatch(createRouteTab())} />
+        <AddTabButton addTab={() => dispatch(createRouteTab())} />
       </div>
 
       <RouteLadders
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -2,6 +2,7 @@ import React, { ReactElement, useContext } from "react"
 import RoutesContext from "../contexts/routesContext"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import useTimepoints from "../hooks/useTimepoints"
+import { RouteTab, currentRouteTab } from "../models/routeTab"
 import { allVehiclesAndGhosts } from "../models/vehiclesByRouteId"
 import { VehicleId, VehicleOrGhost } from "../realtime.d"
 import { ByRouteId, Route, RouteId, TimepointsByRouteId } from "../schedule.d"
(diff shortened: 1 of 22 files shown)


assets/src/components/ladderPage.tsx
```diff
@@ -24,9 +38,53 @@ export const findSelectedVehicleOrGhost = (
...
+        tab.isCurrentTab ? "m-ladder-page__tab-current" : "m-ladder-page__tab"
+      }
+      onClick={() => selectTab()}
+    >
+      {title}
+    </div>
+  )
+}
+
+const AddLadderButton = ({
```
phildarnowsky, line 61: **nitpick**: from context it looks like this was supposed to be called `AddTabButton`

assets/src/components/ladderPage.tsx
```diff
@@ -24,9 +38,53 @@ export const findSelectedVehicleOrGhost = (
...
+
+const LadderPage = (): ReactElement<HTMLDivElement> =>
+  featureIsEnabled("presets_workspaces") ? (
+    <LadderPageWithTabs />
+  ) : (
+    <LadderPageWithoutTabs />
+  )
+
+const LadderPageWithoutTabs = (): ReactElement<HTMLDivElement> => {
+  const [state, dispatch] = useContext(StateDispatchContext)
```
phildarnowsky, line 81: **thought**: another thing about reducers, we have some in the jumbo reducer associated with `StateDispatchContext`, and then some more special-purpose reducers here and there. I wonder if there would be any benefit to consolidating those loose reducers into the big one rather than having them scattered.

assets/tests/components/routePicker.test.tsx
```diff
@@ -24,7 +22,13 @@ describe("RoutePicker", () => {
     const tree = renderer
       .create(
         <RoutesProvider routes={routes}>
-          <RoutePicker selectedRouteIds={selectedRouteIds} />
+          <RoutePicker
+            selectedRouteIds={selectedRouteIds}
+            // tslint:disable-next-line: no-empty
+            selectRoute={() => {}}
```
phildarnowsky, line 28: **suggestion**: instead of writing the trivial function `() => {}`, how about let's call `jest.fn()`, and then we don't have to keep doing the tslint disable.

phildarnowsky (changes requested):
this is good stuff overall, I have just a few minor corrections to suggest

assets/src/models/routeTab.ts
```diff
@@ -3,7 +3,27 @@ import { LadderDirections } from "./ladderDirection"
...
+
+export const newRouteTab = (): RouteTab => ({
+  isCurrentTab: true,
+  selectedRouteIds: [],
+  ladderDirections: {},
+  ladderCrowdingToggles: {},
+})
+
+export const currentRouteTab = (routeTabs: RouteTab[]): RouteTab => {
+  const currentTab = routeTabs.find((routeTab) => routeTab.isCurrentTab)
```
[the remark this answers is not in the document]
lemald: Good catch. FWIW I'm probably going to have to rework this logic at some point anyway, because simply calling `newRouteTab()` doesn't actually add it to the state, and I need to figure out a graceful spot to put that.

assets/src/state.ts
```diff
@@ -111,6 +111,80 @@ export const toggleLadderCrowding = (
   payload: { routeId },
 })
 
+interface CreateRouteTabAction {
```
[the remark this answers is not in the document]
lemald: Yeah this is definitely one of the more boilerplate-y parts of our codebase... not sure exactly what to do about it but I'm open to suggestions.

assets/src/components/ladderPage.tsx
lemald: Ah, yup, you're absolutely right.

assets/src/components/ladderPage.tsx
lemald: Right, we've got a few. I kind of like the way that the search reducer works, where it's part of the main state but has its own reducer function that's defined in a separate module etc.

phildarnowsky (approved):

PR #1323, opened 2021-11-15
Persist tabs

lemald:
Asana ticket: [⚙️ Live prototype: Implement logic + interactions for Presets tab ](https://app.asana.com/0/1200180014510248/1201087271304728/f)

I probably need to add a couple more tests to get coverage up to 100% but figured I would open a PR for review.

### DIFF
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -1,9 +1,15 @@
 import { useEffect, useReducer, useState } from "react"
-import { putRouteSettings } from "../api"
+import { putRouteSettings, putRouteTabs } from "../api"
 import appData from "../appData"
 import { loadState, saveState } from "../localStorage"
 import { defaultRouteSettings, RouteSettings } from "../routeSettings"
-import { Dispatch, initialState, reducer, State } from "../state"
+import {
+  Dispatch,
+  initialState,
+  reducer,
+  State,
+  updateRouteTabs,
+} from "../state"
 import {
   defaultUserSettings,
   putLadderVehicleLabel,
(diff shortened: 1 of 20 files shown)


assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -48,6 +59,16 @@ const usePersistedStateReducer = (): [State, Dispatch] => {
     }
   }, [selectedRouteIds, ladderDirections, ladderCrowdingToggles])
 
+  useEffect(() => {
+    if (firstLoadDone && pendingRouteTabs !== undefined) {
+      putRouteTabs(pendingRouteTabs).then((routeTabs) => {
```
phildarnowsky, line 64: **thought**: this might take design assets we don't have yet, but pursuant to our recent talk about error-handling, should we handle network errors here more gracefully? and/or do we need a spinner?

phildarnowsky (changes requested):

assets/src/hooks/usePersistedStateReducer.ts
lemald: Indeed, this is probably one of the first cases where we would definitely want to handle it. I'm not really sure about the design resources and exactly what should happen when the network operation fails, though - do we still make the update in the front-end but just warn the user of the failure? Don't make the update at all and just revert the state of the application to what it was before the user initiated the change? I'd like to have at least a high-level answer to these questions before proceeding.

test/skate/settings/route_tab_test.exs
```diff
@@ -1,53 +1,66 @@
...
 
-  describe "set/2" do
-    test "updates existing route_tab" do
-      route_tab = RouteTab.create("charlie")
+  describe "update_all_for_user!/2" do
+    test "adds a new tab entry" do
+      route_tab =
+        build(:route_tab, %{preset_name: "some routes", selected_route_ids: ["1", "28"]})
+
+      [_new_route_tab] = RouteTab.update_all_for_user!("charlie", [route_tab])
```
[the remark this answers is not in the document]
lemald: Good idea. The one test where I still think it makes sense to omit such an assertion is for deleting, since the first half of that is just doing an insertion of a new record, a thing that's handled in the first test.

assets/src/hooks/usePersistedStateReducer.ts
phildarnowsky: yeah I think the answer here is, that's something we should have built in from the design stage, so let's leave it for now until we approach the question of errors more systematically

phildarnowsky: Agreed, we should bring it up but it's at the moment still basically a product question.

phildarnowsky (approved):
good once you're satisfied with the coverage

PR #1327, opened 2021-11-22
Style route picker

lemald:
Asana ticket: [⚙️ Implement styling to route picker](https://app.asana.com/0/1200180014510248/1201148067873468/f)

This should be largely feature-complete but note that I still haven't quite figured out how to test the focus-dependent clear route filter button, so there's still a failing test in there. Everything else should be ready for review though and I'll revisit that test tomorrow.


lemald:
@phildarnowsky update: I got the tests working, so this should be good to go for review.

phildarnowsky (approved):

PR #1329, opened 2021-11-30
Save and list presets

phildarnowsky:


phildarnowsky:
@lemald still needs tests updated, but feature-complete

assets/src/models/routeTab.ts
```diff
@@ -38,11 +56,18 @@ export const parseRouteTabData = (
...
     ladderCrowdingToggles: routeTabData.ladder_crowding_toggles,
   }))
 }
+
+export const isPreset = (routeTab: RouteTab): boolean =>
+  routeTab.ordering === undefined
+export const isNotPreset = (routeTab: RouteTab): boolean => !isPreset(routeTab)
+
+const nullToUndefined = <T>(data: T | null): T | undefined =>
+  data === null ? undefined : data
```
lemald, line 73: **suggestion (non-blocking):** it might be nice to put this in a module under the `helpers` directory, though I don't think it really fits into any of the existing modules.

lib/skate_web/controllers/route_tabs_controller.ex
```diff
@@ -4,6 +4,12 @@ defmodule SkateWeb.RouteTabsController do
   alias SkateWeb.AuthManager
   alias Skate.Settings.RouteTab
 
+  def create(conn, %{"route_tab" => route_tab} = _params) do
+    username = AuthManager.Plug.current_resource(conn)
+    RouteTab.create_for_user!(username, format_tab(route_tab))
+    json(conn, %{})
+  end
```
lemald, line 11: **question:** how is this actually invoked from the front end? It isn't mapped in `router.ex` and I also don't see any changes to `api.ts`.

assets/src/state.ts
```diff
@@ -532,9 +561,21 @@ const routeTabsReducer = (
...
+    case "CREATE_PRESET":
+      return [state, [...state, newPreset(action.payload.sourceTab)]]
+    case "INSTANTIATE_PRESET":
+      return [
+        state,
+        [
+          ...state.map((existingRouteTab) => {
+            return { ...existingRouteTab, isCurrentTab: false }
+          }),
+          tabFromPreset(action.payload.preset, state.length),
```
lemald, line 573: **suggestion:** if you rewrite `isPreset` to instead reference `presetName`, we don't need to duplicate the entry whenever we create or instantiate a preset - you'd only need to create a second entry when the user begins to make unsaved changes to the contents of the preset, thereby necessitating storing different information.

lemald: I guess semantically to me it doesn't make sense to have the `routeTabs` be one array if saved presets and open tabs are going to be two totally separate subsets of it, rather than overlapping.

assets/src/state.ts
```diff
@@ -532,9 +561,21 @@ const routeTabsReducer = (
           newRouteTab(state.length),
         ],
       ]
+    case "CREATE_PRESET":
+      return [state, [...state, newPreset(action.payload.sourceTab)]]
```
lemald, line 565: **question:** See below comment, but one thing to consider about this interaction - this means that your existing tab will stay as "untitled" rather than becoming an instantiation of the preset you just created. Are we sure that's what we want?

lemald, line 565: **issue:** as it stands, this logic means the tab you have open will stay "untitled" rather than becoming an instantiation of the new preset. This seems to contradict the clickable Figma prototype, which shows the tab changing from "Untitled" to "Cabot-1" once the preset is saved.

lemald: Note that I think this is easy to fix - one option is to go with the suggestion in my comment below, but it's probably also possible to just carry the `presetName` from the new preset back over to the current tab.

lib/skate_web/controllers/route_tabs_controller.ex
phildarnowsky: It's not...that can come out.

assets/src/state.ts
phildarnowsky: OK, so, this comment and your other ones below puzzled me for a while--but now I think I have it. You're thinking of a `RouteTab` as something that can serve as an open tab and/or preset at the same time, whereas as you say, in my implementation they're mutually exclusive--a preset is a prototype for a route tab, not a route tab itself.

There are pros and cons to both of these approaches. For instance, the approach I've taken here lets you instantiate a single preset several times easily, and edit each of those independently. Your approach makes it easier to update the tab name on preset save as you mention.

However, it seems like conflating route tabs and presets is already creating confusion for us. I think a better idea would be to explicitly treat presets and tabs as different things, and introduce a third concept called maybe "tab settings". Tab settings would consist of the open routes, ladder toggles, and crowding toggles, but not e.g. the ordering or preset name. Presets have tab settings, and route tabs have tab settings, and when you instantiate a preset you copy the tab settings into the newly created tab, setting the rest of the fields as appropriate.

It would involve reworking our schema, but there'll never be an easier time to do that than now, and I feel like if we continue down this present path we're going to regret it.

lemald: Okay, having thought this over I like your idea. My hope was that the combined `routeTabs` would be concise while still easy to understand, but apparently that's not so much the case. I think we can create an additional table / thing (I like `routeTabPreset` or something as a name - seems clearer to me than "tab settings" and also mirrors what it's called i nthe UI). There's some risk of duplicated code because a lot of the schema, utility functions, APIs, etc. will be a quite similar, but hopefully it won't feel too boilerplate-y.

lemald: One thing to keep in mind is that however we implement this, we'll want to precisely nail down the interactions between saved presets, open tabs, and workspaces - for instance, if we're separating the two concepts, should a workspace be a collection of saved presets or open tabs? I'm kind of leaning towards the former but we'd have to think it through.

PR #1332, opened 2021-12-08
feat: style route tabs

lemald:
Asana ticket: [⚙️ Implement styling to tabs bar](https://app.asana.com/0/1200180014510248/1201448223729748/f)

Most of this should be pretty straightforward. The one exception is that I had to use a very specific application of `box-shadow` and `z-index` to cover up the border dividing adjacent tabs on one side of the active tab. As far as I can tell the best alternative would be to add a special class to the inactive tab immediately to the right of the active tab - this seems like it would actually be a worse hack, but let me know if the CSS I wrote is _so_ confusing that the special class thing seems better. Basically tree-position-based pseudo classes are great but can't quite let you peek into the details of sibling elements.


lemald:
@phildarnowsky for the purposes of review, you can ignore the code in my "temp fix, do not merge" commit aside from not giving final approval until I remove it. 😅

phildarnowsky (approved):

PR #1354, opened 2021-12-17
fix: change fake Ueberauth strategy

lemald:
Ueberauth was recently updated and the new version requires CSRF
protection, or explicitly setting the ignore_csrf_attack flag. Doing
the latter since this isn't a real strategy.


lemald:
See https://github.com/ueberauth/ueberauth/pull/136

dks-mbta (approved):

PR #1355, opened 2021-12-17
remove headway feature

dks-mbta:


lemald:
@dks-mbta this looks mostly good in terms of the actual diffs, but I'm seeing a ton of unaffected files producing diffs due to line ending changes - `package.json`, several SVGs, and a bunch of the Ecto migrations. Could you make sure these get reverted? Thanks.

lemald:
@dks-mbta one thing we'll want to confirm is that if an instance of Skate with the old front-end code that still references the headway fields on `vehicle` hits an instance of the Skate back-end with those fields removed doesn't hit any errors. Looking at the code I think it will be okay but we should test it to be sure.

lemald:
@dks-mbta hmm, I'm still seeing the newline changes show up in the diff here on GitHub. Also, the last couple of commits on `master` see to have been added as commits to this branch, rather than being added as a merge commit or rebased.

lemald (approved):
👍

PR #1359, opened 2021-12-20
Update Node and Dockerfile

lemald:
Asana ticket: [⚙️ Update node.js](https://app.asana.com/0/1152340551558956/1200415301013517/f)

I ended up also largely redoing the Dockerfile as part of this ticket. This makes the build process much nicer and hopefully reduces the size of the resulting container as well.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 elixir v1.10.2-otp-22
 erlang 22.3.2
-nodejs 11.10.0
+nodejs 14.18.2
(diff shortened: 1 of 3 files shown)


paulswartz (approved):
🍰 
LGTM assuming it deployed safely to dev-green?

lemald:
> 🍰 LGTM assuming it deployed safely to dev-green?

It deployed successfully to dev, though actually I'll try dev-blue (which is in ECS) as well real quick just to be on the safe side.

PR #1363, opened 2021-12-29
Correct race condition and possible duplicate orderings in route tabs

phildarnowsky:
This reworks how we sync the route tabs open in a user's browser to the backend DB.

One key aspect is that, after the route tabs have been initialized on page load from the `data-route-tabs` JSON in the HTML, updates only flow from the browser to the backend. When we PUT an update of the route tabs state to the backend, we disregard the server's response, apart from checking its status code. The database IDs of route tabs are as a consequence no longer of interest to the frontend. To put it another way, the definitive route-tabs state is that in the client's browser; we are only interested in saving that state to the backend so that we can use it to initialize future sessions.

That still leaves one race condition possible. Imagine that the route-tabs state in the browser starts out in state S0. The user then adds two tabs, and consequently, the in-browser state goes from S0 to S1 and then to S2. The obvious implementation is to have each change of state fire off a PUT, so that we PUT first S1 and then S2 to the backend in separate requests R1 and R2. But just because we start R1 before R2 doesn't mean that R1 will necessarily complete before R2. If R2 completes first, then R1 overwrites the backend state with S1, a stale state. We add a simple lock-and-queue system to guard against this, ensuring that only one such request can be issued at a time.

There was also a separate bug where tabs might have ended up with a duplicate ordering. I fixed that while I was in here.

Tests to follow, this one is going to take some testing.

### DIFF
--- a/assets/tests/hooks/usePersistedStateReducer.test.ts
+++ b/assets/tests/hooks/usePersistedStateReducer.test.ts
@@ -152,7 +152,6 @@ describe("usePersistedStateReducer", () => {
     expect(state.ladderCrowdingToggles).toEqual({ "83": true })
     expect(state.routeTabs).toEqual([
       routeTabFactory.build({
-        id: "1",
         ordering: 0,
         presetName: "some name",
         isCurrentTab: true,
@@ -237,7 +236,6 @@ describe("usePersistedStateReducer", () => {
 
   test("sends updated route tabs to backend on changes", () => {
     const routeTab = routeTabFactory.build({
-      id: "1",
       isCurrentTab: true,
       selectedRouteIds: [],
       ladderDirections: {},
--- a/assets/tests/hooks/usePersistedStateReducer.test.ts
+++ b/assets/tests/hooks/usePersistedStateReducer.test.ts
@@ -15,6 +15,7 @@ import {
   State,
   toggleLadderCrowding,
   createRouteTab,
+  startingRouteTabsPush,
 } from "../../src/state"
 import {
   VehicleLabelSetting,
@@ -242,7 +243,7 @@ describe("usePersistedStateReducer", () => {
       ordering: 0,
     })
     ;(putRouteTabs as jest.Mock).mockImplementationOnce(
-      () => new Promise(() => [routeTab])
+      () => new Promise(jest.fn())
     )
     const { result } = renderHook(() => usePersistedStateReducer())
     const [, dispatch] = result.current
@@ -261,6 +262,87 @@ describe("usePersistedStateReducer", () => {
     ])
     const [state] = result.current
     expect(state.routeTabs).toEqual([routeTab])
+  })
+
+  test("saves updated route tabs to push later if a push is currently in progress", () => {
+    const { result } = renderHook(() => usePersistedStateReducer())
+    const [, dispatch] = result.current
+
+    act(() => {
+      dispatch(startingRouteTabsPush())
+      dispatch(createRouteTab())
+    })
+
+    const [state] = result.current
+
+    expect(state.routeTabs).toEqual([
+      routeTabFactory.build({
+        ordering: 0,
+        isCurrentTab: true,
+      }),
+    ])
+    expect(state.routeTabsToPush).toEqual(state.routeTabs)
+    expect(state.routeTabsPushInProgress).toEqual(true)
+  })
+
+  test("retries on HTTP error if not outdated", () => {
+    const badResponse = { ok: false }
+    const fakePromise = {
+      then: (callback: (data: any) => void) => {
+        callback(badResponse)
+        return { catch: jest.fn() }
+      },
+    }
+
+    const { result } = renderHook(() => usePersistedStateReducer())
+    const [, dispatch] = result.current
+
+    ;(putRouteTabs as jest.Mock).mockImplementationOnce(() => fakePromise)
+
+    act(() => {
+      dispatch(createRouteTab())
+    })
+
+    const [state] = result.current
+
+    expect(state.routeTabs).toEqual([
+      routeTabFactory.build({
+        ordering: 0,
+        isCurrentTab: true,
+      }),
+    ])
+    expect(state.routeTabsToPush).toEqual(state.routeTabs)
+    expect(state.routeTabsPushInProgress).toEqual(false)
+  })
+
+  test("retries on client error if not outdated", () => {
+    const fakePromise = {
+      then: () => ({
+        catch: (callback: () => void) => {
+          callback()
+        },
+      }),
+    }
+
+    const { result } = renderHook(() => usePersistedStateReducer())
+    const [, dispatch] = result.current
+
+    ;(putRouteTabs as jest.Mock).mockImplementationOnce(() => fakePromise)
+
+    act(() => {
+      dispatch(createRouteTab())
+    })
+
+    const [state] = result.current
+
+    expect(state.routeTabs).toEqual([
+      routeTabFactory.build({
+        ordering: 0,
+        isCurrentTab: true,
+      }),
+    ])
+    expect(state.routeTabsToPush).toEqual(state.routeTabs)
+    expect(state.routeTabsPushInProgress).toEqual(false)
   })
 })
 
--- a/assets/tests/hooks/usePersistedStateReducer.test.ts
+++ b/assets/tests/hooks/usePersistedStateReducer.test.ts
@@ -20,7 +20,6 @@ import {
   VehicleLabelSetting,
   VehicleAdherenceColorsSetting,
 } from "../../src/userSettings"
-import { instantPromise } from "../testHelpers/mockHelpers"
 import routeTabFactory from "../factories/routeTab"
 
 // tslint:disable: react-hooks-nesting
@@ -242,8 +241,8 @@ describe("usePersistedStateReducer", () => {
       ladderCrowdingToggles: {},
       ordering: 0,
     })
-    ;(putRouteTabs as jest.Mock).mockImplementationOnce(() =>
-      instantPromise([routeTab])
+    ;(putRouteTabs as jest.Mock).mockImplementationOnce(
+      () => new Promise(() => [routeTab])
     )
     const { result } = renderHook(() => usePersistedStateReducer())
     const [, dispatch] = result.current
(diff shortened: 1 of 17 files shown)


lemald:
**qutesion:** as I understand it, having the client-side ignore IDs means that we'll simply blow away all of the individual route tabs and recreate them in the database every time there's an update, correct? Currently we use Ecto's association handling via the `User` struct's `route_tabs` property to avoid this. The result should be correct either way in terms of the contents of database reflecting the user's tabs, but have we determined if there's any performance impact? I'm thinking of both the time spent updating the database and any increased storage needs in Postgres (though for the latter the `route_tabs` property is set to `on_replace: :delete` which I _assume_ means it won't be a problem).

An additional related concern: will this be any harder in terms of observability and debugging? I find it nice to be able to look at the database directly via Postgres queries in my local dev environment and I could see this making that a bit more complicated.

lemald:
**suggestion:** One possible approach that could preserve the IDs and also get rid of race conditions would be to set a flag that just disallows the user from changing their tabs while waiting for a response from the server, probably with a timeout. This might have its own complications, though (in particular we'd want to do some design to figure out how to visually indicate that the change is pending and they can't make changes (probably by greying out some stuff on the page).

phildarnowsky:
@lemald per our conversation, here are some other concerns that I left out of the original PR description because it was already very long. Some of them are related to points you raised, some not.

As you say [here](https://github.com/mbta/skate/pull/1363#issuecomment-1002734440), this approach blows away all existing `route_tabs` for the user and recreates them. And as you say, it's a correct but unsophisticated strategy and not so performant. But as always we have the luxury of working at internal-tool scale as opposed to public-Web scale, and I think for the size of our userbase it'll be performant enough.

One alternative that should work in theory is to recognize that, although the default in Ecto, inherited from Rails, is for the primary key of each DB table to be an autoincrementing serial number, there's no rule that says that ~~a dog can't pla~~ we can't use a different kind of primary key. In particular, a route tab can be uniquely identified by the pair `(user_id, ordering)`, giving us a natural composite key. (Note to self: make sure there's a unique index on that pair.) I didn't go too far down this road, but I did verify that Ecto has at least some basic support for composite keys. If we decided to make this switch, we could presumably get Ecto to update all of a user's route tabs in one go with `INSERT...ON CONFLICT UPDATE`. But then, I'm also a little dubious about how much actual performance bump we'd get.

Another option that I briefly considered was ditching the entire DB infrastructure for route tabs, and turning `route_tabs` into a single `text` field on (or associated one-to-one with) `users`. It's an approach I've used in the past, it has its pros and cons that should be fairly obvious, and I ruled out that idea after not much consideration.

I would suggest we don't lock the frontend until the sync finishes. I've observed a pattern in many apps recently where syncing is retried on failure indefinitely, and if syncing fails for N seconds running, a warning to the effect of "There's a problem, your data may not be saved, last save X seconds ago" appears.

Probably also we want to throw in some sort of throttle/debounce (always get them confused) on the sync retry rather than let it proceed willy-nilly.

lemald:
@phildarnowsky I guess I'd be curious about hard performance numbers - I agree it's not likely to be a problem for the stability of the application as a whole, but I would want to make sure that the updates can process roughly as fast as a user does them. Like, if I user is going through and doing a whole bunch of tab manipulations, would we end up with just one or two things in queue, or an ever-growing backlog of stuff that takes a couple of seconds to clear out? It doesn't actually block the UI obviously but keeping the queue from getting too big should reduce the chances of something too weird happening.

The primary key idea is an interesting one. How much effort do you think it would be? My gut instinct is that it's not really worth it and might create its own issues.

Agreed that we shouldn't just stuff this into a text field on the `users` table.

Also agreed now that you mention it re: how to handle this in the UI, that would be a lot more elegant.

phildarnowsky:
@lemald 

> Like, if I user is going through and doing a whole bunch of tab manipulations, would we end up with just one or two things in queue, or an ever-growing backlog of stuff that takes a couple of seconds to clear out?

Ah, so, it was sort of misleading of me to call what we have here a "queue" of changes. `routeTabsToPush` is a `RouteTab[] | null`, not a `RouteTab[][]`. At any given moment, it either contains either a `RouteTab[]` representing the route tabs the user had open at the moment they made a change that we want to sync, or it contains `null`, an absence of such a set. When a change is made while `routeTabsPushInProgress` is `true`, we update `routeTabsToPush` with the current tab state, but any state there already is overwritten.

Suppose we call the starting tabs state S0 again. You add a tab, changing the front-end tabs state to S1, and trigger a request R0 to push S1 to the backend. But R0 takes a very long time due to computers being awful, and in the meantime you make lots more changes to your tab state, taking your frontend in turn through states S2, S3, S4, S5, and all the way up to S100 by the time R0 finally completes. At this point, `routeTabsToPush` contains S100: each of S2, S3, etc. spent a moment in `routeTabsToPush` before the next state change overwrote it. With R0 complete, we can now make request R1, pushing S100 to the backend, and thus making 2 requests instead of 100.

Given that the pushing is not as aggressive as you perhaps thought, does that change your level of concern about performance?

> The primary key idea is an interesting one. How much effort do you think it would be? My gut instinct is that it's not really worth it and might create its own issues.

[This makes it sound fairly straightforward](https://hexdocs.pm/phoenix/1.3.0-rc.3/custom_primary_key.html#content). I'm kind of thinking now that it would be desirable all else being equal, not for performance reasons as much as it being a sounder design. It doesn't seem _that_ important though, so I think what I'll do is try that out Monday morning but back off if I encounter nontrivial resistance.

In any case, for future DB schema design, we should keep in the back of our minds the fact that, while a synthetic serial key is generally a sensible default, we have other options available when those make sense.

phildarnowsky:
@lemald I experimented with a composite key on `route_tabs`. There is in fact a blocker: `has_one` (and other associations in Ecto models) don't support composite foreign keys, and I don't think the benefit of having a tidier schema outweighs the benefit of being able to use associations. There is a PR open for Ecto for associations to support composite foreign keys (https://github.com/elixir-ecto/ecto/pull/3638) so that's something we may be able to do in the future, but for now that's as far as I want to take that.

Not sure if you saw the question buried in my long comment above--given that the tempo of pushes to the backend is less aggressive than you might think, do you still have reservations about the performance implications of dropping and recreating the whole state at once?

lemald:
@phildarnowsky yes, sorry for not responding sooner, your explanation of how the not-quite-a-queue actually works makes a lot of sense, I should have paid more attention to the code.

Unfortunate about the Ecto schema limitations (and agreed that that means it's probably not worth it to make the change).

lemald:
@phildarnowsky if I'm not mistaken this means I'm all caught up on the back-and-forth around my feedback from last week and I'm just waiting for final tests and such to come through before re-reviewing and approving, correct?

phildarnowsky:
@lemald that's correct

phildarnowsky:
@lemald there is also the question of what to do if syncing fails for a significant time. I think though that will take some discussion and grooming (especially because adding one failure case like this might make us think of others we want to add), probably a conversation between you me Ashli and Tyler. I'll make a ticket.

lemald:
> @lemald there is also the question of what to do if syncing fails for a significant time. I think though that will take some discussion and grooming (especially because adding one failure case like this might make us think of others we want to add), probably a conversation between you me Ashli and Tyler. I'll make a ticket.

Yeah that's not addressed in the current design / specs so I think fair to say not something you need to worry about at the moment.

phildarnowsky:
@lemald actually, I'm thinking about it, perhaps it would be overkill to have a separate warning just for this one sync. Consider the most likely scenarios for why a sync would fail:

1. Bug in the backend and/or frontend code. Sentry notifies us, we fix it, no more sync failing.
2. Entire server down with no endpoints working. We have a modal for this already (see below).
3. Network trouble somewhere between the end user's device and the server. Same modal appears.

The modal in question:

Screen Shot 2022-01-03 at 2 56 09 PM

lemald:
> @lemald actually, I'm thinking about it, perhaps it would be overkill to have a separate warning just for this one sync. Consider the most likely scenarios for why a sync would fail:
> 
>     1. Bug in the backend and/or frontend code. Sentry notifies us, we fix it, no more sync failing.
> 
>     2. Entire server down with no endpoints working. We have a modal for this already (see below).
> 
>     3. Network trouble somewhere between the end user's device and the server. Same modal appears.
> 
> 
> The modal in question:
> 
> Screen Shot 2022-01-03 at 2 56 09 PM

Makes sense to me!

phildarnowsky:
@lemald tests are complete. I did end up omitting tests for two paths (when a push fails because of an HTTP error, and there is another set of tabs waiting to push; and the same, but where the error is client-side instead). I can walk you through the details of why if you're interested, but in brief, it turned out to be surprisingly hard to simulate the particular circumstances that would send the code down those paths.

lemald:
> @lemald tests are complete. I did end up omitting tests for two paths (when a push fails because of an HTTP error, and there is another set of tabs waiting to push; and the same, but where the error is client-side instead). I can walk you through the details of why if you're interested, but in brief, it turned out to be surprisingly hard to simulate the particular circumstances that would send the code down those paths.

Hmm okay looks pretty good overall. I'd like to at least give some thought to those two test cases to see if I can think of any ideas, but it's the end of the day and I'm pretty tired, so I'll come back to that in the morning.

assets/tests/hooks/usePersistedStateReducer.test.ts
```diff
@@ -237,15 +236,14 @@ describe("usePersistedStateReducer", () => {
...
       isCurrentTab: true,
       selectedRouteIds: [],
       ladderDirections: {},
       ladderCrowdingToggles: {},
       ordering: 0,
     })
-    ;(putRouteTabs as jest.Mock).mockImplementationOnce(() =>
-      instantPromise([routeTab])
+    ;(putRouteTabs as jest.Mock).mockImplementationOnce(
+      () => new Promise(jest.fn())
```
lemald, line 246: **question:** should this still use `instantPromise` but instead return a response object with `ok: true` so that we hit line 72 of `usePersistedStateReducer.ts`?

phildarnowsky: actually `instantPromise` won't work because of the `catch`, so a fake promise it is

phildarnowsky:
@lemald adjusted that one test as you suggested, ready for another look

lemald (approved):
LGTM!

PR #1364, opened 2021-12-30
feat: log time spent sending data message from Schedule.Fetcher

lemald:
Asana ticker: [⚙️ Log time spent in GenServer.call from Schedule.Fetcher to Schedule](https://app.asana.com/0/1200180014510248/1201523184137409/f)

I also did other experimenting but none of it was really fruitful in terms of improving performance - see my comment on the Asana ticket.

Note that we're still only doing the schedule fetch on app startup, so we'll only see this logging when a new instance starts up.

### DIFF
--- a/test/schedule/fetcher_test.exs
+++ b/test/schedule/fetcher_test.exs
@@ -57,6 +57,7 @@ defmodule Schedule.FetcherTest do
           end
         )
 
+      assert log =~ ~r/Sent updated schedule data to receiving process, time_in_ms=\d+/
       assert log =~ "Successfully loaded schedule data"
     end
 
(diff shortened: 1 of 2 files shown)


test/schedule/fetcher_test.exs
```diff
@@ -57,6 +57,7 @@ defmodule Schedule.FetcherTest do
           end
         )
 
+      assert log =~ "Sent updated schedule data to receiving process"
```
phildarnowsky, line 60: **suggestion**: this is fine as far as it goes, but we could make a stronger assertion: `assert log =~ /Sent updated schedule data to receiving process, time_in_ms=\d+/`

phildarnowsky (approved):
Looking good contingent on a minor change I flagged

PR #1369, opened 2022-01-12
Don't create virtual ghost if trip should have started more than an hour ago

dks-mbta:


assets/tests/models/vehiclesByPosition.test.ts
```diff
@@ -197,7 +197,7 @@ describe("groupByPosition", () => {
     })
   })
 
-  test("does not generate a virtual ghost for incoming buses that are late on their previous trip", () => {
+  test("does not generate a virtual ghost for incoming buses whose scheduled location isn't on the ladder route", () => {
```
lemald, line 200: **praise:** thanks for catching and correcting this.

lemald (approved):
Looks good to me!

PR #1370, opened 2022-01-14
feat: use Logster for better request logging

lemald:
To aid in [⚙️Benchmark tabs / presets API performance](https://app.asana.com/0/1200180014510248/1201651613270218/f)

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -53,6 +53,9 @@ config :logger, :console,
   format: "$time $metadata[$level] node=$node $message\n",
   metadata: [:request_id]
 
+# "code" is the secret value returned by AWS to /auth/cognito/callback
+config :logster, :filter_parameters, ~w(password code)
+
 # Configure Ueberauth to use Cognito
 config :ueberauth, Ueberauth,
   providers: [
--- a/mix.exs
+++ b/mix.exs
@@ -76,7 +76,8 @@ defmodule Skate.MixProject do
       {:postgrex, "~> 0.15"},
       {:ex_aws_rds, "~> 2.0.2"},
       {:fast_local_datetime, "~> 1.0"},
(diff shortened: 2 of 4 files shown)


mix.exs
```diff
@@ -76,7 +76,8 @@ defmodule Skate.MixProject do
       {:postgrex, "~> 0.15"},
       {:ex_aws_rds, "~> 2.0.2"},
       {:fast_local_datetime, "~> 1.0"},
-      {:ex_machina, "~> 2.7.0", only: :test}
+      {:ex_machina, "~> 2.7.0", only: :test},
+      {:logster, "~> 1.0"}
```
arkadyan, line 80: Not related to this line specifically, but…

`logger` isn't defined here in the `deps` list, but it is still above in the list of `extra_applications`. Is it still required to be there for other libraries or Phoenix or something?

config/prod.exs
```diff
@@ -53,6 +53,9 @@ config :logger, :console,
   format: "$time $metadata[$level] node=$node $message\n",
   metadata: [:request_id]
 
+# "code" is the secret value returned by AWS to /auth/cognito/callback
```
arkadyan, line 56: 👍

arkadyan (approved):
Small question just to check whether Logger still needs to be running, but otherwise looks good. 👍

mix.exs
lemald: My understanding is that Logster is just a plug that calls Logger with a nicer format than the default, so we still want that running.

lemald: It looks like the API [also leaves the `:logger` process running](https://github.com/mbta/api/blob/master/apps/api_web/mix.exs#L28), for a point of comparison, so I feel pretty good about merging this as-is.

PR #1373, opened 2022-01-18
Temporarily disable tabs backend

lemald:
Asana ticker: [⚙️ Get existing tabs and presets work into a good state ](https://app.asana.com/0/1200180014510248/1201646081199337/f)

Tabs and presets is a feature that is not yet live to our actual users, being developed behind a feature flag. I need to make some slight changes to the data model, mainly using client-generated UUIDs as the primary key. Since nobody is actually using this feature, the easiest way to handle the transitional state during the deploy is to simply make the application no longer use those tables. Once the changes are made and the relevant utility functions are updated I'll undo the changes in this PR.


arkadyan (approved):
👌

PR #1374, opened 2022-01-18
Redo route tabs schema

lemald:
Asana ticket: More of [⚙️ Get existing tabs and presets work into a good state ](https://app.asana.com/0/1200180014510248/1201646081199337/f)

Following #1373, I'm redoing the `route_tabs` table by merging in the columns from `tab_settings` and, more importantly, replacing the Postgres auto-incrementing integer ID with a UUID. This will allow the client to generate a unique ID for new tabs which will make some concurrency and data consistency issues much easier to deal with.


arkadyan (approved):
👍 Still getting my environment set back up so I can play with this locally, but looks good.

lemald:
> 👍 Still getting my environment set back up so I can play with this locally, but looks good.

Cool, I don't think there have been any major changes recently in terms of what you need for a local dev environment setup but let me know if you need a hand with anything.

PR #1376, opened 2022-01-19
Front-end support for route tabs UUID

lemald:
Asana ticket: [⚙️ Get existing tabs and presets work into a good state](https://app.asana.com/0/1200180014510248/1201646081199337/f)

This should be the last major piece of the schema refactor and move to UUIDs.

### DIFF
--- a/assets/tests/hooks/usePersistedStateReducer.test.ts
+++ b/assets/tests/hooks/usePersistedStateReducer.test.ts
@@ -126,7 +126,7 @@ describe("usePersistedStateReducer", () => {
       }),
       routeTabs: JSON.stringify([
         {
-          id: "1",
+          uuid: "1",
           ordering: 0,
           preset_name: "some name",
           is_current_tab: true,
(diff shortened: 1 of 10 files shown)


assets/tests/hooks/usePersistedStateReducer.test.ts
```diff
@@ -281,11 +269,11 @@ describe("usePersistedStateReducer", () => {
 
     const [state] = result.current
 
-    expect(state.routeTabs).toEqual([
-      routeTabFactory.build({
+    expect(state.routeTabs).toMatchObject([
```
skyqrose, line 272: **praise:** TIL about `toMatchObject`

assets/tests/state.test.ts
```diff
@@ -447,10 +447,16 @@ describe("reducer", () => {
       State.createRouteTab()
     )
 
+    const expectedNewTab = routeTabFactory.build({
+      isCurrentTab: true,
+      ordering: 5,
+    })
+    delete expectedNewTab.uuid
```
skyqrose, line 454: **suggestion (non-blocking):** could this be `expectedNewTabFields = {isCurrentTab:true, ordering:5}`, so you can skip building the rest of the object?

skyqrose (approved):

assets/tests/state.test.ts
lemald: So I considered that, but I want to make sure that we do match on any other default fields that a new route tab might have, and those will be added to the `routeTab` factory if we ever add more.

PR #1377, opened 2022-01-20
Save and list presets, close tabs

lemald:
Asana tickets: [⚙️Presets panel - list the existing presets + save / open them ](https://app.asana.com/0/1200180014510248/1201367915520953/f) and [⚙️  Implement close tab interaction ](https://app.asana.com/0/1200180014510248/1201448223729774/f)

To the reviewer: there's more documentation in Notion about this feature, just let me know if you'd like links for more background. This functionality is still behind a feature flag so not yet going out to users. Styling and some other parts of the tabs / presets interactions will be in forthcoming PRs.

### DIFF
--- a/assets/src/components/presets.tsx
+++ b/assets/src/components/presets.tsx
@@ -8,7 +8,7 @@ const Presets = () => {
   const presets = routeTabs.filter(isPreset)
 
   return (
-    <div>
+    <div className="m-presets-panel">
       <button
         onClick={() =>
           dispatch(
--- a/assets/src/components/presets.tsx
+++ b/assets/src/components/presets.tsx
@@ -10,7 +10,14 @@ const Presets = () => {
   return (
     <div>
       <button
-        onClick={() => dispatch(createPreset(currentRouteTab(routeTabs)))}
+        onClick={() =>
+          dispatch(
+            createPreset(
+              currentRouteTab(routeTabs).uuid,
+              `Preset ${Math.floor(Math.random() * 10000)}`
+            )
+          )
+        }
       >
         Save as preset
       </button>
@@ -20,7 +27,7 @@ const Presets = () => {
           {presets.map((preset) => (
             <button
               key={preset.uuid}
-              onClick={() => dispatch(instantiatePreset(preset))}
+              onClick={() => dispatch(instantiatePreset(preset.uuid))}
             >
               {preset.presetName}
             </button>
--- a/assets/src/components/presets.tsx
+++ b/assets/src/components/presets.tsx
@@ -1,0 +1,9 @@
+import React from "react"
+
+const Presets = () => (
+  <div>
+    <button>Save as preset</button>
+  </div>
+)
+
+export default Presets
--- a/assets/src/components/presets.tsx
+++ b/assets/src/components/presets.tsx
@@ -1,9 +1,34 @@
-import React from "react"
+import React, { useContext } from "react"
+import { createPreset, instantiatePreset } from "../state"
+import { StateDispatchContext } from "../contexts/stateDispatchContext"
+import { currentRouteTab, isPreset } from "../models/routeTab"
 
-const Presets = () => (
-  <div>
-    <button>Save as preset</button>
-  </div>
-)
+const Presets = () => {
+  const [{ routeTabs }, dispatch] = useContext(StateDispatchContext)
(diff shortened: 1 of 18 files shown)


assets/src/models/routeTab.ts
```diff
@@ -40,11 +46,108 @@ export const parseRouteTabData = (
...
+
+  if (tabToClose.isCurrentTab) {
+    const nextTabToRight = newRouteTabs
+      .filter(
+        (routeTab) =>
+          routeTab.ordering && routeTab.ordering > (tabToClose.ordering || 0)
+      )
+      .sort((a, b) => a.ordering || 0 - (b.ordering || 0))[0]
+
+    if (nextTabToRight) {
```
arkadyan, line 133: **thought:** I wonder if you could use `highestExistingOrdering` combined with the order of the `tabToClose` to check whether there is a tab to the right, without having to do the filtering and sorting first.

assets/tests/models/routeTab.test.ts
```diff
@@ -16,3 +43,195 @@ describe("currentRouteTab", () => {
...
+    const routeTab = routeTabFactory.build({
+      presetName: undefined,
+      ordering: 1,
+    })
+
+    expect(isOpenTab(routeTab)).toBeTruthy()
+  })
+})
+
+describe("instantiatePresetByUUID", () => {
```
arkadyan, line 87: **praise:** These tests are helpful for figuring out this function.

assets/src/components/presets.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const presets = routeTabs.filter(isPreset)
+
+  return (
+    <div className="m-presets-panel">
+      <button
+        onClick={() =>
+          dispatch(
+            createPreset(
+              currentRouteTab(routeTabs).uuid,
+              `Preset ${Math.floor(Math.random() * 10000)}`
```
arkadyan, line 17: **suggestion (non-blocking):** I could see the state reducer (or maybe better yet a function on the `RouteTab` model…) generating the default name if one isn't provided. (But I can also imagine that this might be changing drastically soon, so I'm not really worried about it.)

arkadyan (approved):
Nice! I added a few random suggestions, but all of the "take 'em or leave 'em" variety.

assets/src/models/routeTab.ts
lemald: Hmm, I'll have to consider that. Overall this bit of logic here is definitely the hairiest part of the PR so I'm definitely open to ideas for how to make it less confusing.

assets/tests/models/routeTab.test.ts
lemald: Glad they helped! This and `closeTabByUUID` presented themselves as great opportunities for some TDD, so they actually helped me a lot in writing this function as well.

assets/src/components/presets.tsx
lemald: Yeah, that would arguably be the more correct way of doing it. This is very much temporary code, though, because there's going to be a modal that prompts the user for a name, which will be a required part of the preset creation process - I'll probably end up implementing it with some sort of callback that gets sent to the modal component to invoke with the specified name.

assets/src/models/routeTab.ts
arkadyan: I pre-prefix this by saying that this doesn't feel important for this PR.

Then I'll prefix it by saying that the following is an initial thought that I haven't investigated at all. It's quite likely it won't make sense in practice.

That said, part of what that initial thought came from was the notion that maybe you could bury all this logic in a helper function, called `nextTab` or `fallbackTab` or something that made more sense. Sort of a refactor to both hide but also "describe with a function name" kind of approach. Then maybe `highestExistingOrdering` could be checked and if there was a higher order number than the current `tabToClose`, that means there's a tab to the right (I think?) and you could just call this existing "next tab to right" logic, otherwise, you could call the "next tab to left" logic. It doesn't change a lot, but maybe it's a hair clearer? And with a function, you could test the logic in isolation.

Or just decide none of this is worth the effort and just ignore it.

lemald: Over all I like that idea (and you have a good point about function names as documentation), but I'm not sure it's really worth the effort for this PR. Thanks for the idea, though!

PR #1379, opened 2022-01-21
feat: use `persistent_term` to store the schedule data

paulswartz:
https://erlang.org/doc/man/persistent_term.html

This stores the data in a way that it's available to all processes
without copying. Giving this a shot to see if it works


lemald:
Interesting, I hadn't heard of `persistent_term`! Optimized for reading versus writing / updating sounds like exactly what we're looking for, though.

paulswartz:
@lemald it's looking good on dev so far: it uses 30% of the memory and 50% the CPU.

paulswartz:
I do want to check how the updating works: I'll do that next week.

lemald:
> I do want to check how the updating works: I'll do that next week.

Right, though the good news with this solution is that, even if updating in place causes too much blocking, it should be fairly simple to swap between two different terms (relative to juggling many different Mnesia tables).

paulswartz:
@lemald this is ready for review. 
- Updating the data takes ~3 seconds in the case where the result has changed (simulated here by using the current system time). It takes about ~2 seconds in the case where the data ends up being the same.
- Total memory use is about 40% of the current main branch. This isn't specifically reflected in the ehmon report, as the memory `:persistent_term` uses isn't included. We log that memory separately when we update the state.

lemald:
@paulswartz this is looking really good, and I'm pleasantly surprised at how fast the updates are!

While you're here, could you re-enable the periodic fetching in the `Schedule.Fetch` GenServer? It should basically involve reverting commit `ede2372c`.

I still want to do some testing with an in-place update happening while someone is using the app to see how it performs overall (i.e. make sure we don't see a bunch of various GenServers timing out and crashing). I probably won't get to that by the end of the day tomorrow but I can do it tomorrow.

lemald:
One thing I'm looking at is the time it takes the miniature copy of Concentrate to process new data, since the pipelines for both TripUpdates and VehiclePositions do some referencing of the schedule data. For the last 15 minutes, the longest this process took on dev was 1312 ms, whereas on prod it was 796 ms. The average is also slightly higher on dev, but that might be explained in part by doing much more polling / loading than we would normally. Overall I think these numbers should be acceptable but I want to record them here.

lemald:
@paulswartz okay, looked over the changes as well as did some more looking at Splunk, I think we're good to pull out the temporary stuff and look at steady-state performance. Once we do that I'll approve the PR itself assuming no issues suddenly crop up.

paulswartz:
@lemald the version which doesn't fetch constantly is up on dev. I'll give you some relevant Splunk queries to compare the memory and scheduler usage.

lemald:
@paulswartz just for my own edification, can you give me a brief rundown on how you arrived at the combination of parameters for each GenServer?

As an additional note, I assume this means we'll have to be careful when introducing any other GenServers that use schedule data, though I'm not sure where to best document that (just thinking of how to avoid getting tripped up by this in six months when I've forgotten about this change).

paulswartz:
@lemald would you prefer expanding on that in the commit message, in the code comments, or something else?

lemald:
> @lemald would you prefer expanding on that in the commit message, in the code comments, or something else?

Hmm given that we usually squash and merge let's go with comments.

lib/realtime/block_waiver_store.ex
```diff
@@ -27,7 +27,8 @@ defmodule Realtime.BlockWaiverStore do
   def start_link(opts \\ []) do
     name = Keyword.get(opts, :name, default_name())
 
-    GenServer.start_link(__MODULE__, %__MODULE__{}, name: name)
+    # always fullsweep: this keeps us from having the old Schedule data stick around
+    GenServer.start_link(__MODULE__, %__MODULE__{}, name: name, spawn_opt: [fullsweep_after: 0])
```
paulswartz, line 31: @lemald it turns out Elixir 1.11 has the wrong type for `spawn_opt` :( 

I can either keep us on 1.11 and ignore the Dialyzer error, or I can bump us up to 1.12 or 1.13 (which have the right type): up to you.

lemald: I was actually just thinking the other day that we should bump the Elixir version in Skate sometime, so if it ends up not causing any other difficulties I'd say go ahead and upgrade.

lemald (approved):
Whew, thanks for sticking with this throughout all of the various challenges that came up. Looks good to me!

PR #1387, opened 2022-01-25
feat: style toggle between routes and presets in picker

lemald:
Asana ticket: [⚙️ Style route picker toggle between routes and presets view](https://app.asana.com/0/1200180014510248/1201403136252736/f)

### DIFF
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -166,18 +166,30 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
 
       <PickerContainer>
         <>
-          <button
-            id="m-ladder-page__routes_picker_button"
-            onClick={() => setCurrentDrawerContent("route_picker")}
-          >
-            Routes
-          </button>
-          <button
-            id="m-ladder-page__presets_picker_button"
-            onClick={() => setCurrentDrawerContent("presets")}
-          >
-            Presets
-          </button>
+          <div className="m-ladder-page__routes-presets-toggle">
+            <button
+              id="m-ladder-page__routes_picker_button"
+              className={
+                currentDrawerContent === "route_picker"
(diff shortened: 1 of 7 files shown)


assets/src/components/ladderPage.tsx
```diff
@@ -166,18 +166,30 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
...
-          </button>
-          <button
-            id="m-ladder-page__presets_picker_button"
-            onClick={() => setCurrentDrawerContent("presets")}
-          >
-            Presets
-          </button>
+          <div className="m-ladder-page__routes-presets-toggle">
+            <button
+              id={"m-ladder-page__routes_picker_button"}
```
skyqrose, line 171: **nitpick (non-blocking):** curly braces not necessary here.

assets/src/components/routePicker.tsx
```diff
@@ -37,19 +37,38 @@ const RoutePicker = ({
     garageFilterData
   )
 
-  return (
+  return featureIsEnabled("presets_workspaces") ? (
```
skyqrose, line 40: **question (non-blocking):** How long are you planning to leave this flag up? There's some duplication between the two branches that could drift apart over time, but if this is planned to be released for everybody soon, then it's not really an issue. If it's gonna be up for a while, then at least the `RoutesList` could be pulled out of the flag check.

skyqrose (approved):
A couple comments but neither of them are blocking.

When reviewing UI PRs like this, it's nice to be able to see before/after screenshots of the feature. It was fine not having them on this one, but it could make future PRs easier, especially when I don't have as much context on the feature.

assets/src/components/routePicker.tsx
lemald: The idea is that this will be released to all Skate users in the coming weeks, so it won't hang around for a long time.

PR #1389, opened 2022-01-26
use ADRs, document switch to :persistent_term

paulswartz:
Offering this as a proposal to document internal engineering choices, for our use as well as to help onboard future engineers/PMs.

### DIFF
--- a/documentation/adr/0002-use-persistent-term-to-store-schedule-state.md
+++ b/documentation/adr/0002-use-persistent-term-to-store-schedule-state.md
@@ -1,0 +1,37 @@
+# 2. Use :persistent_term to store schedule state
+
+Date: 2022-01-26
+
+## Status
+
+Accepted
+
+## Context
+
+Asana: https://app.asana.com/0/116562505129567/1201514090100824
+
+The `Schedule.Fetcher` GenServer makes a call to the `Schedule GenServer` (which a
+lot of Skate calls) when it has new state to provide. Usually this works
+smoothly, but sometimes it causes us to log `GenServer Schedule.Fetcher
+terminating` followed by a longer message `** (stop) exited in:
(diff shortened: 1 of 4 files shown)


documentation/adr/0002-use-persistent-term-to-store-schedule-state.md
```diff
@@ -0,0 +1,37 @@
...
+
+The `Schedule.Fetcher` GenServer makes a call to the `Schedule GenServer` (which a
+lot of Skate calls) when it has new state to provide. Usually this works
+smoothly, but sometimes it causes us to log `GenServer Schedule.Fetcher
+terminating` followed by a longer message `** (stop) exited in:
+GenServer.call(Schedule, [...state])`. This is then followed by some logging on
+the `Schedule` side, with a bunch of warnings like `module=Elixir.Schedule
+function=trip error=timeout` in the couple of minutes after the other error.
+
+It is not purely a deadlock, causes a degradation of performance.
```
lemald, line 21: **issue:** I think this is supposed to read "It is not purely a deadlock, **but it** causes..."

lemald:
Over all, I really like this idea - do you have any thoughts about how it would fit into other places where we document engineering choices? I know we're also talking about making tech specs for product features, for instance.

paulswartz:
> Over all, I really like this idea - do you have any thoughts about how it would fit into other places where we document engineering choices? I know we're also talking about making tech specs for product features, for instance.

This is something that came out of the discussions I've had with Ryan and Ruth about technical specs related to products. The goal is to have documentation about some of the technical choices that are made to meet product requirements, but still be lightweight to avoid generating too much extra work for the tech lead or other engineers.

lemald (approved):

arkadyan:
@paulswartz Thanks for introducing this—ADRs seem like a useful approach I haven't come across before.

PR #1391, opened 2022-01-28
fix: don't crash if we can't access instance id

paulswartz:
If `curl` fails, ignore it with `true` which will give an empty instance ID.


lemald (approved):

paulswartz:
I'll wait for the deploy to finish and then merge.

PR #1395, opened 2022-01-31
Edit presets

lemald:
Asana ticket: [⚙️ Ability to update an existing preset](https://app.asana.com/0/1200180014510248/1201367915520955/f)

### DIFF
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -64,7 +64,14 @@ const LadderTab = ({
       onClick={() => selectTab()}
     >
       <div className="m-ladder-page__tab-contents">
-        <div className="m-ladder-page__tab-title">{title}</div>
+        <div
+          className={
+            "m-ladder-page__tab-title" +
+            (tab.saveChangesToTabUuid ? " m-ladder-page__tab-title-edited" : "")
+          }
+        >
+          {title}
+        </div>
         {tab.isCurrentTab ? saveIcon("m-ladder-page__tab-save-icon") : null}
         <CloseButton onClick={() => closeTab()} />
       </div>
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -79,7 +79,9 @@ const LadderTab = ({
         <div
           className={
             "m-ladder-page__tab-title" +
-            (tab.saveChangesToTabUuid ? " m-ladder-page__tab-title-edited" : "")
+            (tab.saveChangesToTabUuid
+              ? " m-ladder-page__tab-title--edited"
+              : "")
           }
         >
           {title}
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -147,7 +147,11 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
   }, [JSON.stringify(routeTabs)])
 
   const { selectedRouteIds, ladderDirections, ladderCrowdingToggles } =
-    currentRouteTab(routeTabs)
+    currentRouteTab(routeTabs) || {
+      selectedRouteIds: [] as string[],
+      ladderDirections: {},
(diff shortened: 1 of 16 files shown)


assets/src/components/ladderPage.tsx
```diff
@@ -64,8 +76,24 @@ const LadderTab = ({
       onClick={() => selectTab()}
     >
       <div className="m-ladder-page__tab-contents">
-        <div className="m-ladder-page__tab-title">{title}</div>
-        {tab.isCurrentTab ? saveIcon("m-ladder-page__tab-save-icon") : null}
+        <div
+          className={
+            "m-ladder-page__tab-title" +
+            (tab.saveChangesToTabUuid ? " m-ladder-page__tab-title-edited" : "")
```
arkadyan, line 82:
```suggestion
            (tab.saveChangesToTabUuid ? " m-ladder-page__tab-title--edited" : "")
```

**nitpick:** should this be a double-dash before the modifier? (I'm rusty on my BEM and it never stopped being at least a little confusing for me, so I might well be wrong. )

arkadyan:
Side observation while I have the app open locally: I'm seeing the "incoming box" at the bottom of the route ladders being pushed partially off the bottom of my screen. I assume that's not related to this PR, but is that a known issue (maybe due to the tabs)? Just wanted to mention it while it was in front of me.

assets/src/components/ladderPage.tsx
```diff
@@ -210,6 +242,14 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
               tab={routeTab}
               selectTab={() => dispatch(selectRouteTab(routeTab.uuid))}
               closeTab={() => dispatch(closeRouteTab(routeTab.uuid))}
+              showSaveIcon={!isPreset(routeTab) || isEditedPreset(routeTab)}
+              saveTab={() => {
+                if (isEditedPreset(routeTab)) {
```
arkadyan, line 247: **thought:** If you passed the entire `routeTab`, could you remove the condition here, call e.g. `savePreset` all of the time, and let either the action or maybe better yet the reducer figure out the "create new vs update existing" question here?

assets/src/components/ladderPage.tsx
```diff
@@ -210,6 +242,14 @@ const LadderPageWithTabs = (): ReactElement<HTMLDivElement> => {
...
+                  dispatch(savePreset(routeTab.uuid))
+                } else if (!isPreset(routeTab)) {
+                  dispatch(createPreset(routeTab.uuid, "Preset name"))
```
arkadyan, line 250: **question:** It seems like all presets are currently named, "Preset name". Is that a known issue that will be addressed later?

arkadyan (approved):
👍

Included a few suggestions and a question, but nothing blocking.

assets/src/components/ladderPage.tsx
lemald: Yes - [this ticket](https://app.asana.com/0/1200180014510248/1201286872028647/f) will implement modals to confirm deletion of a preset as well as allowing the user to input the name to save a preset under. This also means these callbacks are going to have to change a bit - we're going to have to dispatch a change to the state that triggers the modal to open that itself takes a callback to invoke once the user has supplied a name and clicked "save."

assets/src/components/ladderPage.tsx
lemald: Good question, I was just poking through the BEM documentation and I believe you're right. I'll admit I haven't been the most consistent about making sure we're doing BEM naming by the book, so this gives me another bit of motivation to go through and clean up our class naming conventions at some point.

assets/src/components/ladderPage.tsx
lemald: This is a really good point - since I'm going to be moving around these callbacks and code associated with them as part of implementing the modals I mentioned in another comment, I'll leave a note on that ticket to put this logic in the reducer while I'm at it.

PR #1396, opened 2022-01-31
build: Upgrade patch versions of languages

arkadyan:
No ticket, had trouble building the older version of Erlang on a new machine.

Use the latest 1.10.x version of Elixir, 22.3.x version of Erlang, and
14.x version of Node.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-elixir v1.10.2-otp-22
-erlang 22.3.2
-nodejs 14.18.2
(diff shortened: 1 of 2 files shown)


lemald:
@arkadyan are you sure you have the latest version of Skate? Paul just updated the Erlang version while doing some other work and my branch should be rebased onto the current master.

arkadyan:
@lemald Yep! Sorry, you are both correct and quick.

PR #1397, opened 2022-02-01
Style presets panel

lemald:
Asana ticket: [⚙️ Implement styling to presets panel ](https://app.asana.com/0/1200180014510248/1201262012057883/f)

Here's the resulting styling (once I do the next ticket to actually provide a modal where the user can enter a real name when saving presets I'll be able to get some even better screenshots):
Screen Shot 2022-02-01 at 4 15 46 PM

### DIFF
--- a/assets/css/_presets.scss
+++ b/assets/css/_presets.scss
@@ -1,0 +1,50 @@
+.m-presets-panel {
+  ul {
+    margin-top: 0rem;
+    margin-bottom: 0.5rem;
+
+    li {
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      padding-bottom: 1rem;
+      line-height: 0.75rem;
+
+      .m-presets-panel__preset-button-container {
+        button {
+          @include font-small;
+          color: #0041ea;
+          text-decoration-line: underline;
+        }
+      }
+
+      .m-close-button {
+        @include button-icon(0.625rem);
+        font-size: unset;
+      }
+    }
+  }
+
+  .m-presets-panel__save-as-preset-button {
+    @include font-small;
+    font-weight: 700;
+    color: #fafafb;
+    background: $color-primary;
+    border-radius: 4px;
+    padding: 0.25rem 0.5rem;
+    display: flex;
+    align-items: center;
+
+    .m-presets-panel__save-as-preset-button-icon {
+      padding-right: 0.25rem;
+      width: 0.75rem;
+      height: 0.75rem;
+
+      svg {
+        path {
+          stroke: #fafafb;
+        }
+      }
+    }
+  }
+}
--- a/assets/css/_presets.scss
+++ b/assets/css/_presets.scss
@@ -6,7 +6,7 @@
     li {
       display: flex;
       justify-content: space-between;
-      align-items: center;
+      align-items: flex-start;
       padding-bottom: 1rem;
       line-height: 0.75rem;
 
@@ -15,6 +15,7 @@
           @include font-small;
           color: #0041ea;
           text-decoration-line: underline;
+          text-align: start;
         }
       }
 
--- a/assets/css/_presets.scss
+++ b/assets/css/_presets.scss
@@ -18,6 +18,7 @@
           text-align: start;
           hyphens: auto;
           overflow-wrap: anywhere;
+          word-break: break-all;
         }
       }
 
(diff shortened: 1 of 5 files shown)


assets/src/components/presets.tsx
```diff
@@ -1,15 +1,34 @@
...
 const Presets = () => {
   const [{ routeTabs }, dispatch] = useContext(StateDispatchContext)
   const presets = routeTabs.filter(isPreset)
 
   return (
     <div className="m-presets-panel">
+      <ul>
+        {presets.map((preset) => (
+          <li key={preset.uuid}>
+            <div className="m-presets-panel__preset-button-container">
```
skyqrose, line 17: **question:** could you assign this class to the button instead, or is there a reason you need a wrapping div?

skyqrose (commented):
A few questions.

assets/src/components/presets.tsx
```diff
@@ -1,15 +1,34 @@
...
+              <button onClick={() => dispatch(instantiatePreset(preset.uuid))}>
+                {preset.presetName}
+              </button>
+            </div>
+            <CloseButton
+              onClick={() => {
+                return
```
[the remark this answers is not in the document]
lemald: Yes, I'm actually going to do that as part of the ticket I'm working on now.

assets/src/components/presets.tsx
lemald: I wanted `flex-grow: 1` to apply to the container but not to the button, because if it applies to the button than the entire empty horizontal space between the preset title and the "x" button is a clickable part of the preset name.

assets/css/_presets.scss
```diff
@@ -0,0 +1,51 @@
+.m-presets-panel {
+  ul {
+    margin-top: 0rem;
+    margin-bottom: 0.5rem;
+
+    li {
+      display: flex;
+      align-items: center;
+      padding-bottom: 1rem;
+      line-height: 0.75rem;
```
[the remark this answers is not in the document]
lemald: `line-height` just determines the minimum overall height of the line, and the default was making each entry to tall despite the fact that none of the sub-elements were forcing a larger height. Padding or margin would just add space on top of this minimum height (or I guess I could use negative padding, but using negative padding to offset a value that appears to be a browser default could introduce weird client-specific layout bugs). And even with the line set to `0.75rem`, the icons are (slightly) smaller than that so I still want them centered vertically.

assets/src/components/presets.tsx
skyqrose: Wouldn't it be good to have as big a click target as possible? I'm remembering this discussion about a similar thing from last month: https://mbta.slack.com/archives/C10EPAU6Q/p1642090431014400?thread_ts=1642088728.010000&cid=C10EPAU6Q

If not, `justify-content: space-between` might do the spacing without requiring `flex-grow` or a wrapper.

skyqrose (approved):
one more unresolved conversation, but if you decide that it's fine as it is then 👍

skyqrose:
Oh, another thought! How does this behave for long preset names? Will it wrap gracefully?

assets/src/components/presets.tsx
lemald: Ah, good point about `space-between`, that probably would work.

Regarding the click area itself, Tyler is going to take a look at this and I'll see what he has to say.

lemald:
> Oh, another thought! How does this behave for long preset names? Will it wrap gracefully?

Probably not, as there's nothing specifically in the designs about that so I didn't implement it. Will check in with product / design before merging.

assets/src/components/presets.tsx
lemald: Okay well I tried this and discovered the hard way that there is another reason this container was helping me out: the close button uses `currentColor` to match its surrounding text, and having a sibling with the blue color we're using was causing the close button to also be blue. Of course I can override that, but then because of selector specificity I have to either copy-paste the `:focus` and `:hover` states into this CSS or resort to hacks like `!important`. I'm open to other ideas but unless I think of something else I think I'm going to keep the container.

lemald: I can at least use `justify-content: space-between` on the `li` instead of `flex-grow: 1` on the container though, so there's that. 😅

skyqrose (approved):

PR #1400, opened 2022-02-03
Remove user settings trip label from database

dks-mbta:


priv/repo/migrations/20220203210855_remove_user_settings_trip_label.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Skate.Repo.Migrations.RemoveUserSettingsTripLabel do
+  use Ecto.Migration
+
+  def change do
+    alter table(:user_settings) do
+      remove :minischedules_trip_label
```
lemald, line 6: **suggestion:** you can use `remove/3` to do the removal in a reversible way (i.e. so that rollbacks work automatically): https://hexdocs.pm/ecto_sql/Ecto.Migration.html#remove/3

lemald (approved):
👍

PR #1410, opened 2022-02-11
Notifications from tabs rather than route settings

dks-mbta:


lemald (approved):
Ready to merge for real now!

PR #1413, opened 2022-02-14
feat: process to migrate route_settings to route_tabs

lemald:
Asana ticket: [⚙️ Migrate user settings for tabs + presets](https://app.asana.com/0/1200180014510248/1201481603842987/f)

Note that this doesn't actually enable the process in the supervision tree, that will be in the PR to roll tabs and presets out.

The logic ended up being complicated enough that I opted for this GenServer rather than a real Ecto migration, not to mention the fact that it's not _truly_ reversible so putting it in a migration with an `up` and `down` function would have been slightly disingenuous. This means we have to take care to look at `inserted_at` and `updated_at` values to make sure we're not blowing away subsequent changes by users, given that we don't have the tracking in the `schema_migrations` table to check whether or not the migration has run yet.

### DIFF
--- a/lib/skate/settings/db/user.ex
+++ b/lib/skate/settings/db/user.ex
@@ -28,6 +28,7 @@ defmodule Skate.Settings.Db.User do
       :username
     ])
     |> cast_assoc(:route_tabs, with: &DbRouteTab.changeset/2)
+    |> cast_assoc(:route_settings, with: &DbRouteSettings.changeset/2)
     |> validate_required([
       :username
     ])
--- a/lib/skate/settings/route_settings.ex
+++ b/lib/skate/settings/route_settings.ex
@@ -1,6 +1,5 @@
 defmodule Skate.Settings.RouteSettings do
-  import Ecto.Query
-  import Skate.Repo
+  alias Skate.Repo
 
   alias Schedule.Route
   alias Skate.Settings.Db.RouteSettings, as: DbRouteSettings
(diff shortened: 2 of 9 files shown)


arkadyan (approved):
This is great, but you missed a great opportunity to use some postgres triggers to let folks bounce back and forth! 😉 

Is there a ticket written up yet as a reminder to rip this out once it's no longer needed?

lemald:
> This is great, but you missed a great opportunity to use some postgres triggers to let folks bounce back and forth! 😉
> 
> Is there a ticket written up yet as a reminder to rip this out once it's no longer needed?

Hah, I was initially thinking of leaning more heavily on Postgres features and putting the code in an actual migration, but then I remembered that performance doesn't matter a ton here (it has to happen just once and we don't have too many users) and I had already written a lot of utility helper for working with this data.

Re: cleanup, that's a good point - we already had a [general-purpose ticket](https://app.asana.com/0/1152340551558956/1201442401995240/f) for cleaning up old / unused code after going live with this feature, but I added removal of this module as a specific punchlist item.

lemald:
@arkadyan just worked some things out from debugging, requesting a quick re-review for the recent changes.

lemald:
@arkadyan okay, I have pushed yet another commit to address an issue discovered while testing (surrounding the `updated_at` column on `route_settings` and the fact that `update_all` doesn't actually take that column into account). Note that while this won't fix the column for users until they actually log in and make new changes, everyone who isn't on the Skate team will fall into the case for users with no route tabs yet anyway. This is just making absolutely sure that if we deploy, roll back, and _then_ roll forward again all settings get carried over correctly.

arkadyan (approved):
@lemald Makes sense, looks good. I appreciate all testing, especially with as tricky as it seems.

PR #1424, opened 2022-02-23
Enable tabs and presets

lemald:
Asana ticket: [⚙️ Enable tabs and presets](https://app.asana.com/0/1200180014510248/1201796368174139/f)

Note that I'm initially just forcing the `featureIsEnabled` function to return `true` for this feature. This way I don't have to touch much of the existing code or remove the dead code associated with the old pre-tabs functionality. This means a rollback would be small in terms of the amount of code involved. However, cleaning up that dead code once we're sure we're not rolling back is of course important, so I have [a ticket](https://app.asana.com/0/1152340551558956/1201442401995240/f) ready for that as well.


arkadyan (approved):
👍

PR #1425, opened 2022-02-24
feat: leave garage filter collapsed by default on page load

lemald:
Asana ticket: [⚙️Make garage filter default to collapsed](https://app.asana.com/0/1200180014510248/1201862941782408/f)

The (now-default) collapsed state is shown here:
Screen Shot 2022-02-24 at 11 19 00 AM
The Asana ticket includes a recording of a weird user interaction that results from the garage filter taking up a lot of vertical space - leaving it collapsed by default makes this less likely, while still hopefully being prominent enough that users see that it's there to expand if they need it.


arkadyan (approved):
👍

PR #1426, opened 2022-02-24
chore: disable route settings migration

lemald:
Cleanup from #1424. The migration is written such that running multiple times on application startup shouldn't be an issue, but it has now served its purpose so I'm going to take a "better safe than sorry" approach. The module itself remains for now in case we need to roll back quickly (looking highly unlikely!), but that will be removed as part of the general post-launch cleanup ticket.


arkadyan (approved):
👍

PR #1428, opened 2022-02-25
fix: base routes in swings view off of route_tabs table

lemald:
Asana ticket: [⚙️ Swings view - base swings to show on open tabs](https://app.asana.com/0/1200180014510248/1201880022483800/f)

### DIFF
--- a/lib/skate/settings/route_tab.ex
+++ b/lib/skate/settings/route_tab.ex
@@ -61,6 +61,9 @@ defmodule Skate.Settings.RouteTab do
     |> Enum.map(&db_route_tab_to_route_tab(&1))
   end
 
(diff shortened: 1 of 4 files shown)


skyqrose (approved):
Will the swings view still work for users without the tabs feature turned on?

lemald:
> Will the swings view still work for users without the tabs feature turned on?

The tabs feature is live for all users, so unless we roll the whole thing back (in which case I'll also roll back this change), we're good.

lemald:
@skyqrose thanks for making me think about that, though - added reverting this PR to my written checklist of things to do if we do have to roll the feature back for some reason. 🙏

PR #1430, opened 2022-02-25
fix: include vehicles from all tabs in VehiclesByRouteIdProvider

lemald:
Asana ticket: [⚙️ Late view - base routes on tabs](https://app.asana.com/0/1200180014510248/1201880022483803/f)

The Late View uses the contents of the `VehiclesByRouteIdContext`, and Late View users want to be able to monitor routes across all of their open tabs simultaneously. This will somewhat increase the amount of network traffic coming across the WebSocket relative to only subscribing to the current tab's worth of routes, but since the data on each individual vehicle is relatively small and only updates every few seconds, I don't think this is a huge problem. Also, having this data already loaded makes switching back and forth between tabs slightly more instantaneous.

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -23,7 +23,7 @@ import LateView from "./lateView"
 import { OpenView } from "../state"
 import featureIsEnabled from "../laboratoryFeatures"
 import { isOpenTab } from "../models/routeTab"
-import { flatten } from "../helpers/array"
+import { flatten, uniq } from "../helpers/array"
 
 const AppRoutes = () => {
   useAppcues()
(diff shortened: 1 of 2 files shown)


assets/src/components/app.tsx
```diff
@@ -38,7 +39,9 @@ const AppRoutes = () => {
     openView === OpenView.Late || location.pathname === "/"
 
   const routesForVehicles = featureIsEnabled("presets_workspaces")
-    ? currentRouteTab(routeTabs)?.selectedRouteIds || []
+    ? flatten(
```
arkadyan, line 42: **suggestion (non-blocking):** Would you want to apply `uniq` to this list in addition to `flatten` in case the user has a route open in multiple tabs? I don't know whether that would actually matter or not in practice.

arkadyan (approved):
One small question, but otherwise looks good. I agree that the extra data seems well worth it for the usability gains.

assets/src/components/app.tsx
lemald: Oh, that's actually a really good point - each route gets turned into a subscription to the relevant channel, and weird things can happen when you double-subscribe to a single channel - see "Duplicate Join Subscriptions" on [this page](https://hexdocs.pm/phoenix/js/). Will update.

PR #1431, opened 2022-02-28
Refresh swings on routes change

lemald:
Asana ticket: [⚙️ Bug: Skate Swings View should update when you add/remove route ladders](https://app.asana.com/0/1148853526253437/1201547695983756/f)

Note that in addition to fixing up the hook logic on the front end, I ran into a race condition because of the fact that the swings API endpoint was basing its selection of routes off of the user-selected routes in the database. Since we're making the API query right as a user changes which routes they have open, this means that the swings view would end up one update behind the open routes in a user's tabs. Now, the swings endpoint supports explicitly specifying route IDs. Once this is deployed and has been running for about a day I'll open another PR to remove the old version of the endpoint that doesn't accept arguments, I just don't want to do that immediately because we need to wait for users to get the new front end code.


assets/src/api.ts
```diff
@@ -154,9 +154,9 @@ export const fetchNearestIntersection = (
     defaultResult: null,
   })
 
-export const fetchSwings = (): Promise<Swing[] | null> =>
+export const fetchSwings = (routeIds: RouteId[]): Promise<Swing[] | null> =>
   apiCall({
-    url: "/api/swings",
+    url: `/api/swings?route_ids=${routeIds.join(",")}`,
```
arkadyan, line 159: **thought:** Given that our route IDs are pretty short, this probably isn't a concern in the real world, but theoretically could this be a problem if the number of routes was long enough that we exceeded the URL length "limit"?

arkadyan (approved):
Looks good. I think my one thought there can be ignored, I don't think it's truly a realistic concern.

assets/src/api.ts
lemald: Hmm, I don't think that's much of a concern immediately but I could see that being an issue for our endpoints in general. I'll definitely keep it in mind.

PR #1433, opened 2022-03-02
chore: remove old swings API endpoint that doesn't take argument

lemald:
Asana ticket: [⚙️Remove old version of swings API](https://app.asana.com/0/1200180014510248/1201890023661626/f)

Quick followup to #1431 now that that's been released.

### DIFF
--- a/lib/skate_web/controllers/swings_controller.ex
+++ b/lib/skate_web/controllers/swings_controller.ex
@@ -1,9 +1,6 @@
 defmodule SkateWeb.SwingsController do
   use SkateWeb, :controller
 
-  alias SkateWeb.AuthManager
-  alias Skate.Settings.RouteTab
-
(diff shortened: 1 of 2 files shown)


lib/skate_web/controllers/swings_controller.ex
```diff
@@ -1,9 +1,6 @@
 defmodule SkateWeb.SwingsController do
   use SkateWeb, :controller
 
-  alias SkateWeb.AuthManager
-  alias Skate.Settings.RouteTab
-
   def index(conn, %{"route_ids" => route_ids}) do
```
arkadyan, line 4: **question:** It doesn't really matter since it's an internal API, but I'm presuming if someone requests this route without passing `route_ids` they'll just get a 404 or some sort of error back?

arkadyan (approved):
👍

PR #1434, opened 2022-03-03
feat: go to tab with route when opening VPP from notification

lemald:
Asana ticket: [⚙️ Switch to appropriate tab when opening vehicle from notification](https://app.asana.com/0/1200180014510248/1201862941782388/f)

### DIFF
--- a/assets/src/models/routeTab.ts
+++ b/assets/src/models/routeTab.ts
@@ -69,6 +69,18 @@ export const isEditedPreset = (routeTab: RouteTab): boolean =>
 export const isOpenTab = (routeTab: RouteTab): boolean =>
   routeTab.ordering !== undefined
 
+export const selectTabByUUID = (
+  routeTabs: RouteTab[],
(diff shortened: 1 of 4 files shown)


arkadyan (approved):
Nice, this feels very readable.

One non-blocking suggestion: I believe that if the route is selected in the currently open tab, but also in a tab that's more to the left than the current one we want to leave the current tab open. Assuming I'm getting that right, it might be useful to have an explicit test as a way of documenting that behavior.

lemald:
> Nice, this feels very readable.
> 
> One non-blocking suggestion: I believe that if the route is selected in the currently open tab, but also in a tab that's more to the left than the current one we want to leave the current tab open. Assuming I'm getting that right, it might be useful to have an explicit test as a way of documenting that behavior.

Ah, thanks for catching this - I believe you're correct about the desired behavior, but I think as implemented it will do the wrong thing. I'll fix that and add a test.

PR #1436, opened 2022-03-07
Keyboard shortcuts for presets modals

lemald:
Asana ticket: [⚙️ Key shortcuts: User can "save" by clicking the "return key"](https://app.asana.com/0/1200180014510248/1201862253651608/f)

In terms of being able to press the "return" key, I'm mainly relying on autofocusing the corresponding button to just rely as much as possible on default browser behavior instead of ad-hoc custom event handlers. Similarly, I made the create preset modal into a form so that the return key would trigger a submit, just moving the already-existing logic into a custom `onSubmit` handler. Since all of the modals share the same basic functionality with the "escape" key I went ahead and made a shared top-level modal component and defined the handler there.


skyqrose (approved):
👍 clever

PR #1439, opened 2022-03-08
chore: clean up pre-route tabs frontend code

lemald:
Asana ticket: [⚙️ Remove old non-tabs ladder page code](https://app.asana.com/0/1152340551558956/1201442401995240/f)

This is the first step of cleaning up the old code. Up next, I'll be removing the server-side code and then running a DB migration to drop the old table.

This also fixes a couple of small bits of functionality (basically just directionality of arrows in the VPP and minischedule) that hadn't yet been updated to work with route tabs.


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -605,7 +607,12 @@ const RevenueTrip = ({
     " ",
     trip.startPlace || "",
   ].join("")
-  const [{ ladderDirections }] = useContext(StateDispatchContext)
+  const [{ routeTabs }] = useContext(StateDispatchContext)
+
+  const currentTab = currentRouteTab(routeTabs)
+  const ladderDirections = currentTab
```
skyqrose, line 613: **suggestion (non-blocking):** There's a bunch of variables getting set here, and this long function might be easier to follow if there wasn't quite so much logic spread over 3 variables.

I noticed some similar login in `header.tsx`. Could you extract this into a `import { ladderDirectionsFromRouteTabs } from "routeTab.ts"` function?

assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -23,11 +23,6 @@ jest.mock("../../../src/hooks/useMinischedule", () => ({
   useMinischeduleBlock: jest.fn(),
 }))
 
-jest.mock("../../../src/laboratoryFeatures", () => ({
```
skyqrose, line 26: **thought:** There were other changes in `minischedules.tsx`, but they didn't require test changes. I guess there's no test that icon directions match the ladder direction. Should there be?

assets/tests/components/__snapshots__/ladderPage.test.tsx.snap
```diff
@@ -27,12 +27,27 @@ exports[`LadderPage renders the empty state 1`] = `
         />
       </button>
     </div>
+    <div
```
skyqrose, line 30: **thought:** Ideally, when a snapshot test changes, you can read the snapshot, understand the changes, and verify that they're what you want. This file has 1400 lines of changes, so I, uh, have not reviewed it.

Perhaps there are other ways to rewrite the tests so they explicitly make a check, instead of depending on manually reading the snapshot. For example, "renders with selectedRoutes in different order than routes data" could query for route headers, grab the route names from them, and assert that the route names show up in the right order.

That's shouldn't be part of this PR though.

skyqrose (approved):
👍 minor comments

assets/src/components/propertiesPanel/minischedule.tsx
lemald: The main pieces of this logic that aren't already part of `currentRouteTab` are a) pulling `routeTabs` from the context and b) handling the case where `routeTabs` is null. Regarding a, calling `useContext` from a utility function in the `routeTabs` module would violate the [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions). As for b, the types don't actually guarantee that there will be a unique current tab, hence why null is a possible return value, but based on the application logic that should never happen, so I suppose I could simply raise an exception if that happens and remove null as an option from the return type, simplifying things a little bit. I'm not sure that I want to do that as part of this PR, though.

assets/tests/components/propertiesPanel/minischedule.test.tsx
lemald: Okay, turns out it's actually kind of difficult - we don't have good ways of testing things related to the actual contents of the SVGs, and we don't have much in the way of accessibility metadata to make it easy to select said elements in React Testing Library. The alternative is a snapshot test but those don't actually include the rendered contents of the SVGs and there isn't other information in existing classnames or anything to differentiate the icon types in a snapshot.

PR #1440, opened 2022-03-08
chore: clean up pre-route tabs backend code

lemald:
Asana ticket: [⚙️ Remove old non-tabs ladder page code](https://app.asana.com/0/1152340551558956/1201442401995240/f)

This will only be merged and deployed after #1439.


test/skate_web/controllers/swing_controller_test.exs
```diff
@@ -28,9 +27,7 @@ defmodule SkateWeb.SwingsControllerTest do
   ]
 
   describe "GET /api/swings" do
-    setup %{user: username} do
-      if username, do: RouteSettings.get_or_create(username)
-
+    setup do
```
skyqrose, line 30: Why did you need route_settings here, and do you need to make route_tabs here now? Is there a reference to route settings in the `/api/swings` code that should have been removed in this PR? (I don't need an answer, I just want to make sure you understand it.)

skyqrose (approved):
Wow it's really easy to review a PR that has literally no new code.

test/skate_web/controllers/swing_controller_test.exs
lemald: Originally, the way that this controller worked is that it took no arguments, looked up the routes that a user had selected, and then returned swings for those routes. As part of the change over to route tabs, we made the API endpoint actually take an argument with a list of route IDs so that the backend isn't also doing the work of figuring out which routes to return. The references to route settings in that file were already removed as part of the switch over, and since the endpoint just gets its route information from the argument provided by the client we don't need a reference to route tabs here.

PR #1441, opened 2022-03-08
feat: disabled state for save as preset button

lemald:
Asana ticket: [⚙️ Grey out "Save as Preset" button when inactive](https://app.asana.com/0/1200180014510248/1201862941782394/f)

Active state (already implemented previously):
Screen Shot 2022-03-08 at 4 24 01 PM

New disabled state:
Screen Shot 2022-03-08 at 4 24 05 PM

### DIFF
--- a/assets/css/_presets.scss
+++ b/assets/css/_presets.scss
@@ -50,5 +50,18 @@
         }
       }
     }
+
+    &:disabled {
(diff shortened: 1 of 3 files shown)


boringcactus (approved):
👍

PR #1443, opened 2022-03-09
Upgrade to Webpack 5

lemald:
Asana ticket: [⚙️ Upgrade Webpack from 4 to 5 in Skate](https://app.asana.com/0/1152340551558956/1200553848190003/f)

I mainly just followed the [official guide](https://app.asana.com/0/1152340551558956/1200553848190003/f) for moving from 4 to 5, then went in and updated all of the plugins and loaders that had previously been stuck on older versions due to Webpack 5 being required to upgrade. We do now get one `DEP_WEBPACK_COMPILATION_OPTIMIZE_CHUNK_ASSETS` warning, but that's from a recursive dependency so I can't directly fix it. In addition to testing locally, I have deployed to dev and looked at the resulting output from GitHub Actions to make sure that all of the static content / source maps / etc. that we expected are still being generated and uploaded to the appropriate places.


boringcactus (approved):
👍

PR #1452, opened 2022-03-11
Upgrade TypeScript and other minor JS dependencies

lemald:
Asana ticket: [⚙️ Upgrade TypeScript to 4.x](https://app.asana.com/0/1152340551558956/1200561402021164/f)

Most of the TypeScript upgrade itself was pretty straightforward. A couple of minor notes:
- We had been using an argument to `reload`. This is apparently [only supported on Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload#location.reload_has_no_parameter) and even then doesn't really do anything for our purposes. The types were causing issues so I went ahead and just removed the parameter.
- The `resize-observer-polyfill` package also had some issues, see [here](https://github.com/que-etc/resize-observer-polyfill/issues/80). However, from discussion with my PM we agreed to drop support for older browsers that don't have native ResizeObserver support (which would be pretty out-of-date at this point anyway). If this becomes an issue we can also use [`@juggle/resize-observer`](https://github.com/juggle/resize-observer), which seems better supported.
- I also upgraded a handful of other JS dependencies that looked like low-hanging fruit.

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -464,6 +464,12 @@
       "integrity": "sha512-ws57AidsDvREKrZKYffXddNkyaF14iHNHm8VQnZH6t99E8gczjNN0GpvcGny0imC80yQ0tHz1xVUKk/KFQSUyA==",
       "dev": true
     },
+    "@gar/promisify": {
+      "version": "1.1.3",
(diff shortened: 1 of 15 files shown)


boringcactus (approved):
👍

PR #1457, opened 2022-03-14
fix: JSON.stringify useEffect hook dependency

lemald:
Asana ticket: [⚙️ [extra] Fix infinite useEffect loop](https://app.asana.com/0/1200180014510248/1201957254872274/f)

I also tested this out locally with some `console.log` statements and made sure that I was able to reproduce the original issue and that the fix then actually resolved it.

### DIFF
--- a/assets/tests/hooks/useNotifications.test.tsx
+++ b/assets/tests/hooks/useNotifications.test.tsx
@@ -1,6 +1,6 @@
 import { renderHook } from "@testing-library/react-hooks"
 import { Socket } from "phoenix"
-import React, { ReactElement } from "react"
+import React, { ReactNode } from "react"
 import { SocketProvider } from "../../src/contexts/socketContext"
 import { StateDispatchProvider } from "../../src/contexts/stateDispatchContext"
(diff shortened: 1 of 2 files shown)


assets/tests/hooks/useNotifications.test.tsx
```diff
@@ -131,29 +137,80 @@ describe("useNotifications", () => {
...
+
+    expect(mockChannel.join).toHaveBeenCalledTimes(1)
+  })
+
+  test("rejoins channel if selected routes change", () => {
+    const mockAddNotification = jest.fn()
+    const mockSetNotifications = jest.fn()
+    const mockSocket = makeMockSocket()
+    const mockChannel = makeMockChannel("ok", {
+      initial_notifications: ["notification1", "notification2"],
```
skyqrose, line 175: **nitpick:** Is there any reason to include the notifications here? Could you just repsond `[]`? (Same for above.)

lemald: I guess not strictly necessary, I can remove them.

PR #1459, opened 2022-03-16
ERL_FLAGS in example env, minor README improvements

lemald:
With the changes in #1379, we had to bump the memory available to the literal allocator in the release. This PR just does the same via the `ERL_FLAGS` environment variable for local dev work via the sample direnv configuration, to avoid having to pass it in separately.

Additionally, made some minor updates to the README to reflect currently reality.


boringcactus (approved):
👍

PR #1001, opened 2021-01-23
fetch bridge status

dks-mbta:

PR #1468, opened 2022-03-25
Upgrade Phoenix and Elixir

lemald:
Asana ticket: [⚙️ upgrade to Erlang 24 / Elixir 1.13](https://app.asana.com/0/1152340551558956/1201728814381286/f)

This was mostly pretty straightforward. On the Phoenix upgrade to 1.6, I followed [the guide](https://gist.github.com/chrismccord/2ab350f154235ad4a4d0f4de6decba7b), but ran into some issues moving to heex templates and decided not to add that to the scope of this PR. Since the LiveView dependency seems to be otherwise unneeded I didn't include it, but correct me if I'm wrong.

Also had to make a small tweak to the CI workflow because it turns out that the cache action can [hit an error with uploading the cache](https://github.com/mbta/skate/runs/5697292481?check_suite_focus=true#step:13:3) and leave it in a bad state.


test/support/conn_case.ex
```diff
@@ -20,7 +20,8 @@ defmodule SkateWeb.ConnCase do
   using do
     quote do
       # Import conveniences for testing with connections
-      use Phoenix.ConnTest
+      import Plug.Conn
+      import Phoenix.ConnTest
```
paulswartz, line 24: suggestion: you could also inport `Plug.Test` here if you didn't want to make the `Plug.Test.init_session` change.

test/concentrate/producer/http/state_machine_test.exs
```diff
@@ -176,7 +176,7 @@ defmodule Concentrate.Producer.HTTP.StateMachineTest do
         end)
 
       # only one message (some content before, some content after)
-      assert [_, _] = String.split(log, "[warn]")
+      assert [_, _] = String.split(log, "[warning]")
```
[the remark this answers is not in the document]
lemald: Good question, I just looked through all of the Splunk alerts set up for Skate and confirmed that none of them need updating.

test/support/conn_case.ex
lemald: Even better - it looks like `Phoenix.ConnTest` now has its own `init_test_session/2` that's just [delegated to `Plug.Test`](https://github.com/phoenixframework/phoenix/blob/v1.6.6/lib/phoenix/test/conn_test.ex#L253), so I can avoid the direct use of `Plug.Test` entirely.

paulswartz (approved):
🍰

I'm not too worried about it, but probably a good idea to put this on Dev for a bit to make sure nothing unexpected happens.

lemald:
> 🍰
> 
> I'm not too worried about it, but probably a good idea to put this on Dev for a bit to make sure nothing unexpected happens.

Yeah, I had this running on dev over the weekend and I just deployed the updated version to run on dev for a bit more, I'll give it another day or two just to be on the safe side. The logs from over the weekend look good, I didn't find anything concerning in them.

lemald:
@paulswartz adding one more fix, based on [this discussion of the same issue](https://elixirforum.com/t/warning-description-authenticity-is-not-established-by-certificate-path-validation/41588). I think we explicitly set the SSL version due to some problem with the old version of Elixir / Erlang being unable to talk to the bridge API server (see #1001 for the initial implementation). My understanding is that older versions of TLS are no longer supported by default so there shouldn't be any security concerns, but let me know if that's not the case.

lemald:
Update: still seeing some `Authenticity is not established by certificate path validation` errors on dev-blue, and I notice we also do some SSL configuration in `config.exs`. Will continue digging further.

Dockerfile
```diff
@@ -57,6 +60,8 @@ COPY --from=app-builder --chown=skate:skate /root/priv/static ./priv/static
 # Add application artifact compiled in app-builder container
 COPY --from=app-builder --chown=skate:skate /root/_build/prod/rel/skate .
 
+ADD --chown=skate:skate /root/aws-cert-bundle.pem ./priv/aws-cert-bundle.pem
```
paulswartz, line 63: suggestion: you want to `COPY` this from the builder, instead of `ADD`ing it from the local filesystem.

lemald:
@paulswartz okay, finally got a working deploy with the updated SSL verification function and accompanying cert bundle.

Re: removing the SSL protocol configuration in `config.exs`, it looks like it was added as part of #843, but the errors in question still happen. Based on running `:ssl.versions` it looks like the current version of the SSL library only has TLS 1.2 and 1.3 enabled out of the box, not 1.1, so I think from a security perspective we should be good to remove that line if it's not helping solve other problems.

Dockerfile
```diff
@@ -31,22 +30,28 @@ FROM elixir-builder as app-builder
...
 COPY --from=assets-builder /root/priv/static ./priv/static
 
 RUN mix do compile --force, phx.digest, release
 
-FROM alpine:3.13.6
+FROM alpine:3.15.0
 
-RUN apk add --update libssl1.1 ncurses-libs bash curl dumb-init \
-  && rm -rf /var/cache/apk
+RUN apk add --update libssl1.1 libstdc++ libgcc \
```
paulswartz, line 47: Suggestion (non-blocking): you can use --no-cache instead of update with a cache delete

Dockerfile
```diff
@@ -31,22 +30,28 @@ FROM elixir-builder as app-builder
 
 ENV LANG="C.UTF-8" MIX_ENV=prod
 
+RUN apk add --no-cache --update curl
+
 WORKDIR /root
 
+RUN curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
+    -o aws-cert-bundle.pem
```
paulswartz, line 38: question: I don't suppose there's a SHA256 sum or something we can use to avoid a supply chain attack like we had with Codecov?

lemald: Good question... I'm not seeing anything on AWS's site. Of course I could just download it on my own machine, look at the checksum, and hardcode it into the Dockerfile, but that's still dependent on `truststore.pki.rds.amazonaws.com` having a legitimate certificate when I do the download, so I'm not sure how much extra security that buys us.

paulswartz (approved):

Dockerfile
lemald: True, though we do do deploys (and therefore rebuild the image) fairly often. Thinking about it from that angle I guess it would be a good thing to have such a check.

lemald: Added a check, and verified if I replaced the correct SHA256 sum with some gibberish and tried to build locally it caused the build to fail, rather than proceeding along silently.

PR #1471, opened 2022-03-28
Allow scrolling route ladders horizontally with scroll wheel

lemald:
Asana ticket: [⚙️ Make default route ladders scroll direction horizontal](https://app.asana.com/0/1148853526253445/1201914420173870/f)

There are some [pure HTML/CSS ways of doing this](https://stackoverflow.com/a/39573512) but they tend to be pretty hacky, and I'm not sure if they'd play along nicely with touch-based scrolling. Given that there was a fairly straightforward JS-based solution I decided to go with that.

Note that this specifically fires on a [wheel event](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) rather than a [scroll event](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event) so as to not trigger if a mobile user tries to scroll vertically with touch gestures.

I have opted to call [scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) rather than directly setting [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft). I considered setting `behavior` to `"smooth"`, doing so somewhat paradoxically results in rather jerky movement with multiple wheel events firing in quick succession. The resulting default behavior isn't _quite_ as smooth as the browser default scrolling but it should be perfectly serviceable.


skyqrose (approved):
How does this interact with real mouse-based scrolling horizontally? E.g. with a fancy mouse or if you two finger scroll horizontally on a macbook touchpad.

1. Will it prevent that scrolling, because real horizontal scrolling won't have a `deltaY`? Do you need to only trigger the handler on vertical scroll?
2. Might scrolling be smoother if you reissued the wheel event, but modified to be a horizontal scroll event with a `deltaX`, rather than setting the scroll with JS?

lemald:
> How does this interact with real mouse-based scrolling horizontally? E.g. with a fancy mouse or if you two finger scroll horizontally on a macbook touchpad.
> 
>     1. Will it prevent that scrolling, because real horizontal scrolling won't have a `deltaY`? Do you need to only trigger the handler on vertical scroll?
> 
>     2. Might scrolling be smoother if you reissued the wheel event, but modified to be a horizontal scroll event with a `deltaX`, rather than setting the scroll with JS?

Normal horizontal scrolling with my MacBook touchpad seems to work as expected (it's just that you can _also_ use the vertical scrolling because that apparently sends wheel events). I guess wheel events do support a `deltaY` but I don't have a fancy mouse that would produce such an event for me to test with. Maybe I should remove the `e.preventDefault()` to allow default scrolling behavior to work in that case.

Re: 2, that's a really cool idea and I went to try it, but it turns out that sadly scroll events just represent a generic indication that the element has been scrolled, conforming to the basic `Event` interface but without `deltaX` and `deltaY` attributes.

skyqrose:
1. Shift+wheel causes a horizontal scroll. Idk what kind of event it emits under the hood. Maybe that'd be a way to test it?

2. Sorry, I mixed up scroll and wheel. I meant reissue a wheel event with `deltaX`, not a scroll event.

Of course, trying either of those might be more time to test than they're worth for a PR that's already approved and works, so don't bother with them if you don't want to.

lemald:
> 1. Shift+wheel causes a horizontal scroll. Idk what kind of event it emits under the hood. Maybe that'd be a way to test it?
> 
>     2. Sorry, I mixed up scroll and wheel. I meant reissue a wheel event with `deltaX`, not a scroll event.
> 
> 
> Of course, trying either of those might be more time to test than they're worth for a PR that's already approved and works, so don't bother with them if you don't want to.

Ah, right, forgot about shift + wheel - tried that out and it still seems to work fine. In any event I think the alternate idea is cool but like you said the current approach works and sending a modified wheel event could be tricky, so I'm going to leave this be.

PR #1478, opened 2022-04-04
fix: use flex display for button-icon mixin

lemald:
Asana ticket: [⚙️ Fix up button-icon CSS mixin](https://app.asana.com/0/1152340551558956/1201979636832835/f)

Previously, the `button-icon` mixin used `inline-block` display, which is fine but didn't really line up well with most of the places in the application that use the mixin, resulting in a lot of weird hacks to override the styles it set, as well as some places with minor styling bugs. In particular, the bounds of the containing `<span>` element often ended up not lining up well with the `<svg>` element inside. This uses flex display for the mixin, cleans up the existing hacks / overrides, and fixes a couple of minor pre-existing layout bugs. The comments of the tickets include screenshots of all of the locations in the application where this mixin is used (albeit mainly to highlight them for QA, not necessarily in their final form after the changes in this PR).

### DIFF
--- a/assets/css/_drawer_tab.scss
+++ b/assets/css/_drawer_tab.scss
@@ -17,15 +17,4 @@
   line-height: 1.5rem;
   margin: 0;
   padding: 2rem 0;
-
-  span {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-
-    svg {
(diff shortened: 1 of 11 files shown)


arkadyan (approved):
👍

PR #1480, opened 2022-04-05
Replace hardcoded colors with variables

lemald:
Asana ticket: [⚙️ CSS: replace hardcoded color hex strings with variables.](https://app.asana.com/0/1152340551558956/1201994849078244/f)

This should have no impact on the actual appearance of the app, except for a couple of places where we had #fafafb instead of #fafafa (I went and confirmed in the original designs that the latter is actually correct - one more reason to use variables for these things...). I have confirmed with `git grep` that all existing references to the new palette of colors have been replaced with the new variables.

Also to be clear this doesn't replace _all_ hardcoded colors, just those that appear in our new color palette. I'll be working with design to go through all of the other remaining instances of one-off hex values to determine what to do with them.


lemald:
In coordination with design I added an additional commit that standardizes a handful of colors from throughout the application that were _very close_ to matching the new color palette. The commit messages gives the details, but none of the differences are all that visually perceptible.

skyqrose (approved):
👍

PR #1494, opened 2022-04-11
Move from TSLint to ESLint

lemald:
Asana ticket: [⚙️ Convert tslint to eslint](https://app.asana.com/0/1152340551558956/1158352058280617/f)

The configuration I got out of [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config) didn't actually very closely match the old TSLint behavior in practice, so I figured I may as well try to move towards a more standard linter configuration if I'm going to have to fix a bunch of existing errors / warnings anyway. I have tried to strike a balance between getting rid of antipatterns where ESLint / TypeScript ESLint identifies them while also not going too far down the rabbit hole of refactoring existing code.

One important ESLint plugin that I want to enable, but did not as part of this PR, is [the React hooks one](https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks). That one flags a bunch of warnings for existing issues, and given that it's easy to accidentally break things with hook dependencies I'm going to hold off on making those changes. There is, however, a [preexisting ticket](https://app.asana.com/0/1152340551558956/1201957254872278/f) to fix some of those problems, and I added a note to enable the ESLint plugin as part of it.


arkadyan (approved):
👍 That was more involved than I would have suspected it would be.

PR #1508, opened 2022-04-19
Initial nav code rework

lemald:
Asana ticket: [⚙️ Refactor existing TabBar nav in prep for further Skate nav work](https://app.asana.com/0/1200180014510248/1202131476892151/f)


lemald:
@skyqrose I can give more context, but this is prep work for a big redo of the Skate navigation. There is going to be a beta group for a gradual rollout, and the purpose of the `<Nav />` component is to contain the logic about which version of the navigation components to render.

assets/src/components/app.tsx
```diff
@@ -48,23 +48,24 @@ const AppRoutes = () => {
...
-          />
-          <BrowserRoute
-            exact={true}
-            path="/settings"
-            component={SettingsPage}
-          />
-          <BrowserRoute exact={true} path="/search" component={SearchPage} />
-          {openView === OpenView.Late ? <LateView /> : null}
+          >
+            <BrowserRoute exact={true} path="/" component={LadderPage} />
```
skyqrose, line 55: React Router v6 [(docs)](https://reactrouter.com/docs/en/v6/upgrading/v5) has a whole bunch of significant changes to the way routing is done. Have you considered whether/when you want to upgrade, and how those changes might interact with this PR, or how they might affect your plans for how the code will be refactored?

This approach of putting nav around the routes works for v5, but it might be incompatible with v6.

As an example, here's Glides' upgrade, and it turned out pretty messy and significant: https://github.com/mbta/glides/pull/529

This might be easier to talk about synchronously.

assets/tests/components/nav.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const tree = renderer
+      .create(
+        <BrowserRouter>
+          <Nav pickerContainerIsVisible={true} openView={OpenView.None}>
+            Hello, world!
+          </Nav>
+        </BrowserRouter>
+      )
+      .toJSON()
+    expect(tree).toMatchSnapshot()
```
skyqrose, line 30: This is a very big snapshot. It's too big to read by hand and spot changes or errors in, and it'll be a chore to keep updated. Instead, you could make some assertions about what you expect to be in there. e.g. "expect the tree to have the content "Hello, world".", or "expect this to contain an element with the class m-tab-bar".

I know a big testing rewrite to address all of these big snapshots in the codebase is probably too big to be worth it, but it'd be nice to use this approach in new code.

I'm less concerned about getting the testing framework right or using best practices as I am about writing down what's significant that this test is supposed to be checking for.

(Same for the test below. That should specifically assert that the late view icon exists.)

skyqrose (commented):
I know there's a lot of complexity around all the side bars and top bars (disconnection notice, tab bar expanded/collapsed, route filter, etc). I can't review that those are all positioned correctly in all the possible states, so I'm assuming you've done that or will do that before this is shipped. (Less of a comment for this specific PR and more a comment for the whole epic of work.)

I also haven't seen the designs that you're moving towards, so I can't really evaluate how well it meets those goals.

assets/src/components/app.tsx
lemald: Hmm, I hadn't been keeping up with React Router so I was unaware that there was a new version with potentially-breaking changes. Let me dig into the docs and PR you linked to a bit and see if it's obvious whether or not this approach will work with 6.

assets/tests/components/nav.test.tsx
lemald: Okay, so I did some cleanup here to use React Testing Library and assert the important things rather than using snapshots. Note that RTL is rather opinionated about how to select things in order to encourage you to write good HTML, but the tab bar is old and doesn't really adhere to best practices, so the best way to select the late view icon is to use the `data-testid` escape hatch. This means in turn that it changes some _other_ snapshots, but I think on net this is the least bad thing. Also the existing `TabBar` component is going away once we release the new navigation so we'll be able to get rid of those old snapshots anyway.

lemald:
Also @skyqrose it's probably more than you want to digest as part of this review, but if you're curious we do have a [one-stop shop](https://www.notion.so/mbta-downtown-crossing/Navigation-7262cd8ba8744102a9621e480a9919a2) of all of the documentation on the new navigation, including a link to the designs.

assets/src/components/app.tsx
lemald: Okay, so after reviewing the [upgrade guide](https://reactrouter.com/docs/en/v6/upgrading/v5) I think that upgrading to version 6 will require some minor changes in a few places, but won't require any major refactors. It looks like Glides' use of React Router is somewhat more complicated than Skate's already, so there are really only a couple of basic components and features that we're using.

lemald:
> I can't review that those are all positioned correctly in all the possible states, so I'm assuming you've done that or will do that before this is shipped. (Less of a comment for this specific PR and more a comment for the whole epic of work.)

For what it's worth, the changes in this PR change the structure of the React components but result in the same HTML, so there won't be a user-facing change. All work on the future navigation setup will be gated behind a beta user group and won't go out to anybody until we've done a bunch of QA on the dev, product, and engineering sides.

assets/src/components/app.tsx
skyqrose: Yeah, Glides parses path params, and I guess Skate doesn't anywhere.

I'm particularly worried about `Route` elements having to be direct children of `Routes` elements, while this PR takes a step away from that by nesting `Route` inside `Nav`, but that's a problem that will be totally solvable after this navigation redesign.

skyqrose (approved):
👍

skyqrose:
Access denied to the notion page. Can you share it with me? I'm curious to see it, but won't really need to unless you send me more nav PRs to review.

assets/src/components/app.tsx
lemald: Ideally I'd like to continue having a separate `<Nav>` component even once it's no longer doing the work of gating the beta group. Do you see anything specific in the release notes about `<Route>` elements needing to be direct children? It works in 5 so I would assume they would call it out if it doesn't work in 6 anymore. In any event I could always go in the other direction (move all of the router / routes stuff to the inside of the `<Nav>`), so I think that one way or another I have a path forward.

skyqrose: [Here's the details.](https://gist.github.com/mjackson/d54b40a094277b7afdd6b81f51a0393f)

Oh, it only applies with `<Switch>` (v5) or `<Routes>` (v6) which you aren't using. I didn't realize you could have a `<Route>` without a `<Switch>`, but even if you had to add `<Routes>` during your upgrade to v6, you could add it inside the `<Nav>`, so this won't end up being an issue.

I believe the `<BrowserRouter>` can be far up the tree, so that's okay, too.

False alarm.

lemald: Okay, whew!

In any event thanks for the heads up, this got me looking at the upgrade guide and thinking about it.

PR #1509, opened 2022-04-21
Nav beta group

lemald:
Asana ticket: [⚙️ Add support for beta group in Nav component](https://app.asana.com/0/1200180014510248/1202131476892155/f)

In addition to running this locally I've put it on dev-blue and tested it there by adding myself to the group.

This will allow us to develop the new navigation incrementally behind the beta group flag, but we won't actually be adding any users to it in prod until it's actually up to the level of quality that we want to be providing.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -34,7 +34,7 @@ config :skate, SkateWeb.AuthManager, secret_key: "dev key"
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Skate.Ueberauth.Strategy.Fake, []}
+    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispathcer"]]}
   ]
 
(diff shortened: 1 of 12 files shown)


config/dev.exs
```diff
@@ -34,7 +34,7 @@ config :skate, SkateWeb.AuthManager, secret_key: "dev key"
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Skate.Ueberauth.Strategy.Fake, []}
+    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispathcer"]]}
```
lxhughes, line 37: typo in dispatcher

lxhughes (changes requested):
just the one typo - line noted

arkadyan (approved):
👍

lxhughes (approved):

PR #1521, opened 2022-04-27
chore: logging when schedule health checks fail

lxhughes:
Asana: [Better Logging of Schedule Health Check Failures](https://app.asana.com/0/1152340551558956/1201994849078241/f)

Adds warning-level logging when schedule health checks fail: routes checker, timepoint checker, and trip stop times checker. Adds CaptureLog testing to associated tests.

### DIFF
--- a/lib/schedule/health/checkers/routes_checker.ex
+++ b/lib/schedule/health/checkers/routes_checker.ex
@@ -13,12 +13,12 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
     routes_fn = Application.get_env(:skate_web, :routes_fn, &Schedule.all_routes/0)
 
     length = routes_fn.() |> length()
-    passfail = length >= min_length
+    pass? = length >= min_length
 
-    if(!passfail) do
-      Logger.warning("Routes Checker failed. min_legth=#{min_length} length=#{length}")
+    if !pass? do
+      Logger.warning("Routes Checker failed. min_length=#{min_length} length=#{length}")
     end
 
-    passfail
+    pass?
   end
 end
--- a/lib/schedule/health/checkers/routes_checker.ex
+++ b/lib/schedule/health/checkers/routes_checker.ex
@@ -2,6 +2,7 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
   @moduledoc """
   Check that Schedules returns at least a minimum number of routes.
   """
+  require Logger
 
   @type config :: %{
           min_length: non_neg_integer()
@@ -11,6 +12,13 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
   def healthy?(%{min_length: min_length}) do
     routes_fn = Application.get_env(:skate_web, :routes_fn, &Schedule.all_routes/0)
 
-    routes_fn.() |> length() >= min_length
+    length = routes_fn.() |> length()
+    passfail = length >= min_length
+
+    if(!passfail) do
+      Logger.warning("Routes Checker failed. min_legth=#{min_length} length=#{length}")
+    end
+
+    passfail
   end
 end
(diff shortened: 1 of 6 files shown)


lib/schedule/health/checkers/routes_checker.ex
```diff
@@ -11,6 +12,13 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
   def healthy?(%{min_length: min_length}) do
     routes_fn = Application.get_env(:skate_web, :routes_fn, &Schedule.all_routes/0)
 
-    routes_fn.() |> length() >= min_length
+    length = routes_fn.() |> length()
+    passfail = length >= min_length
```
lemald, line 16: One common idiom in Elixir is to name booleans / functions that return booleans with a `?` at the end, so in this case you might choose a name like `pass?`.

lib/schedule/health/checkers/routes_checker.ex
```diff
@@ -11,6 +12,13 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
...
+
+    if(!passfail) do
```
lemald, line 18: Note that the parentheses are optional here and usually not included in Elixir unless required (for the form of if that fits on a single line mainly).

lib/schedule/health/checkers/trip_stop_times_checker.ex
```diff
@@ -19,7 +20,16 @@ defmodule Schedule.Health.Checkers.TripStopTimesChecker do
   @spec healthy_route?(timepoint_config()) :: boolean
   defp healthy_route?(%{route_id: route_id, min_length: min_length}) do
     with {:trip, %Trip{stop_times: stop_times}} <- trip(route_id) do
-      length(stop_times) >= min_length
+      length = length(stop_times)
+      passfail = length >= min_length
+
+      if(!passfail) do
+        Logger.warning(
+          "Trip Stop Times Checker failed on route #{route_id}. min_legth=#{min_length} length=#{length}"
```
lemald, line 28: One thought - since it's checking stop times for a particular trip, it might be good to log the trip ID as well (you can just additionally pull that out of the `%Trip{...}` struct on line 22).

lib/schedule/health/checkers/timepoints_checker.ex
```diff
@@ -18,6 +19,15 @@ defmodule Schedule.Health.Checkers.TimepointsChecker do
     timepoints_on_route_fn =
       Application.get_env(:skate_web, :timepoints_on_route_fn, &Schedule.timepoints_on_route/1)
 
-    timepoints_on_route_fn.(route_id) |> length() >= min_length
+    length = timepoints_on_route_fn.(route_id) |> length()
+    passfail = length >= min_length
+
+    if(!passfail) do
+      Logger.warning(
+        "Timepoints Checker failed on route #{route_id}. min_legth=#{min_length} length=#{length}"
```
lemald, line 27: Here and a couple of other places: prefer `route_id=#{route_id}` instead of sentence-style formatting to make Splunk queries easier.

lib/schedule/health/checkers/routes_checker.ex
```diff
@@ -11,6 +12,13 @@ defmodule Schedule.Health.Checkers.RoutesChecker do
...
+      Logger.warning("Routes Checker failed. min_legth=#{min_length} length=#{length}")
```
lemald, line 19: Optional suggestion: rather than saying "Routes Checker failed", you can use the special `__MODULE__` variable, as in `"#{__MODULE__} checker failed" which will automatically include the full name of the module to make it easy to find the source when someone sees this message.

lemald (changes requested):
Left a few stylistic comments and suggestions, but overall this looks great!

lemald (approved):
👍

PR #1529, opened 2022-04-29
Bump Alpine version

lemald:
Additionally adding a run of `apk upgrade` to catch any upgrades to the packages in the base image.


arkadyan (approved):
👍

PR #1533, opened 2022-05-05
Sentry: Correctly set environment name

lxhughes:
[Asana](https://app.asana.com/0/1152340551558956/1201890023661630/f)


assets/tests/helpers/sentryInit.test.ts
```diff
@@ -0,0 +1,9 @@
+import sentryInit from "../../src/helpers/sentryInit"
+
+const opts = {
+  environment: "test_env",
+}
+
+test("Sentry Init runs", () => {
+  expect(sentryInit(opts)).toHaveProperty("environment", "test_env")
```
lemald, line 8: So this is testing that the `sentryInit` function returns the options passed in, but what we'd really like to test is that `Sentry.init` gets called with the options. We can do this with a combination of imports and use of `jest.mock` like so (it's sort of complicated so I'll just paste the whole result rather than trying to walk through the changes individually):

```
import sentryInit from "../../src/helpers/sentryInit"
import * as Sentry from "@sentry/react"

jest.mock("@sentry/react", () => ({
  __esModule: true,
  init: jest.fn(),
}))

const opts = {
  environment: "test_env",
}

test("Sentry Init runs", () => {
  sentryInit(opts)
  expect(Sentry.init).toHaveBeenCalledWith(opts)
})
```

With that change made we can also remove the `return` and make `sentryInit` return `void` since the return value isn't actually used anywhere.

Also, if you're here anyway, it might be nice to write a test for `Sentry.setUser` part.

lemald, line 8: So this is testing that the sentryInit function returns the options passed in, but what we'd really like to test is that Sentry.init gets called with the options. We can do this with a combination of imports and use of jest.mock like so (it's sort of complicated so I'll just paste the whole result rather than trying to walk through the changes individually):

```
import sentryInit from "../../src/helpers/sentryInit"
import * as Sentry from "@sentry/react"

jest.mock("@sentry/react", () => ({
  __esModule: true,
  init: jest.fn(),
}))

const opts = {
  environment: "test_env",
}

test("Sentry Init runs", () => {
  sentryInit(opts)
  expect(Sentry.init).toHaveBeenCalledWith(opts)
})
```

With that change made we can also remove the return and make sentryInit return void since the return value isn't actually used anywhere.

Also, if you're here anyway, it might be nice to write a test for Sentry.setUser part.

test/skate_web/controllers/page_controller_test.exs
```diff
@@ -79,5 +80,17 @@ defmodule SkateWeb.PageControllerTest do
 
       assert html_response(conn, 200) =~ "div id=\"app\""
     end
+
+    @tag :authenticated
+    setup do
+      reassign_env(:skate, :record_sentry, true)
+      reassign_env(:skate, :sentry_environment, "test_env")
+    end
```
lemald, line 88: Note that the contents of the `setup` block are run before every `test` inside of a given `describe` - see the docs [here](https://hexdocs.pm/ex_unit/1.13/ExUnit.Callbacks.html#setup/1). I'm not actually sure exactly how that behaves when the `setup` block is placed partway through the test cases instead of at the top, but in any event since this setup is only relevant for the one Sentry-related test you should probably move those two calls to `reassign_env` into the test case itself and get rid of the `setup` block.

lemald (changes requested):
Okay, I have a couple of test-related comments but overall the basic stuff looks good! Before merging we should also deploy to dev-blue and make sure that it works as intended there.

assets/tests/helpers/sentryInit.test.ts
lxhughes: updated thanks! also added a test for Sentry.setUser and moved the window.username into an input to the function for easier testing

test/skate_web/controllers/page_controller_test.exs
lxhughes: done!

lemald (approved):
Looks good!

PR #1544, opened 2022-05-16
fix: specify size of logo icon SVG so it doesn't get too large

lemald:
Asana ticket: related to [⚙️ Implement top nav (desktop / tablet version)](https://app.asana.com/0/1200180014510248/1201682620805412/f)

Here is an example of what the logo SVG looked like before:
Screen Shot 2022-05-16 at 3 25 11 PM
This covered up other elements of the page making them un-clickable.

### DIFF
--- a/assets/css/nav/_top_nav.scss
+++ b/assets/css/nav/_top_nav.scss
@@ -15,6 +15,11 @@
 .m-top-nav__logo {
   width: 2.5rem;
   height: 1.5rem;
+
+  svg {
+    width: 100%;
+    height: 100%;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


lxhughes (approved):
I just fixed this in the mobile nav today as well! I also found that adding width:100% to the svg fixed it as well (and does not require syncing with the container class should it change)

lemald:
> I just fixed this in the mobile nav today as well! I also found that adding width:100% to the svg fixed it as well (and does not require syncing with the container class should it change)

Ah, yes, that would be better, thanks for the tip!

PR #1549, opened 2022-05-18
adjust minischedules for overload offset

dks-mbta:


assets/src/realtime.d.ts
```diff
@@ -114,6 +114,7 @@ export interface Vehicle {
   operatorFirstName: string
   operatorLastName: string
   operatorLogonTime: Date | null
+  overloadOffset?: number | undefined
```
lemald, line 117: If it's already called `overloadOffset?` I believe the question mark makes the `| undefined` redundant.

lemald (changes requested):
I flagged one minor thing where a typespec is somewhat redundant.

Additionally, I think it would be good to add a few test cases in `assets/tests/components/propertiesPanel/minischedule.test.ts`. The existing tests should provide a good template, you can just take the `vehicle` defined in the test file and modify it to have an overload offset.

It would also be nice to be able to manually test this and see if it works but I know that overloads aren't very common right now.

lemald:
> Additionally, I think it would be good to add a few test cases in assets/tests/components/propertiesPanel/minischedule.test.ts. The existing tests should provide a good template, you can just take the vehicle defined in the test file and modify it to have an overload offset.

Note though that we don't need to test all of the edge cases there, since you already have extensive unit tests for the `formattedScheduledTime` that contains most of the logic.

dks-mbta:
Added one test, working on some more.

assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -214,6 +214,57 @@ const vehicle: Vehicle = {
   crowding: null,
 }
 
+const vehicleWithOffset: Vehicle = {
```
lemald, line 217: Note that if this is mostly the same as `vehicle` you can use the spread operator to write this more concisely, something like `const vehicleWithOffset = { ...vehicle, overloadOffset: 8 }`.

dks-mbta: Yeah, I made this change in the next commit.

dks-mbta:
@lemald The additional tests are ready for review.

lemald (approved):
Looks good to me!

PR #1562, opened 2022-05-25
Lh top nav mobile

lxhughes:
https://app.asana.com/0/1200180014510248/1202131476892171/f

Mobile only top bar with collapsing menu. 

MobileMenuIsOpen global added; other landing pages added a blur feature when true.

### DIFF
--- a/assets/css/_nav.scss
+++ b/assets/css/_nav.scss
@@ -1,5 +1,4 @@
-.m-nav--wide,
-.m-nav--narrow {
+.m-nav--wide {
   display: grid;
   gap: 0;
   grid-template-columns: max-content minmax(0, 1fr);
@@ -12,6 +11,13 @@
 }
 
 .m-nav--narrow {
+  display: grid;
+  gap: 0;
+  grid-template-columns: max-content minmax(0, 1fr);
+  grid-template-rows: max-content minmax(0, 1fr);
+  grid-template-areas:
+    "top top"
+    "left content";
   width: 100%;
   height: 100%;
 }
--- a/assets/css/_nav.scss
+++ b/assets/css/_nav.scss
@@ -1,4 +1,4 @@
-.m-nav--wide {
+.m-nav--wide, .m-nav--narrow {
   display: grid;
   gap: 0;
   grid-template-columns: max-content minmax(0, 1fr);
@@ -6,6 +6,11 @@
   grid-template-areas:
     "top top"
     "left content";
+  width: 100%;
+  height: 100%;
+}
+
+.m-nav--narrow {
   width: 100%;
   height: 100%;
 }
--- a/assets/css/_nav.scss
+++ b/assets/css/_nav.scss
@@ -1,4 +1,5 @@
-.m-nav--wide, .m-nav--narrow {
+.m-nav--wide,
+.m-nav--narrow {
   display: grid;
   gap: 0;
   grid-template-columns: max-content minmax(0, 1fr);
--- a/assets/css/nav/_top_nav_mobile.scss
+++ b/assets/css/nav/_top_nav_mobile.scss
@@ -9,8 +9,15 @@
   border-bottom: 1px solid $color-gray-300;  
 }
 
-.m-top-nav__right-items, .m-top-nav__left-items {
-  max-width: 20px;
+.m-top-nav-mobile__icon svg, .m-top-nav-mobile__menu-icon svg {
+    fill: $color-gray-600;
+  &:hover, &:active, &:focus {
+    fill: $color-gray-500;
+  }
+}
+
+.m-top-nav-mobile__notifications-icon svg {
+  fill: $color-eggplant-500;
 }
 
 .m-top-nav-mobile__close {
@@ -25,8 +32,7 @@
 }
 
 .m-top-nav-mobile__menu {
-  width: 80%;
-  min-width: 292px;
+  width: 18.125rem;
   max-width: 100%;
   height: 100%;
   position: absolute;
@@ -51,6 +57,10 @@
   background-color: $color-gray-100;
 }
 
+.m-top-nav-mobile__header-text {
+  color: $color-gray-900;
+}
+
 .m-top-nav-mobile__menu-link {
   text-decoration: none;
   width: 100%;
@@ -58,6 +68,9 @@
 
 .m-top-nav-mobile__menu-button, .m-top-nav-mobile__menu-link {
   color: $color-gray-600;
+  &:hover, &:active, &:focus {
+    color: $color-gray-500;
+  }  
   width: 100%;
   height: 3rem;
   line-height: 3rem;
@@ -72,7 +85,6 @@
   padding-left: 1.5rem;
   padding-right: 1.5rem;
   svg {
-    fill: $color-gray-600;
     width: 1.5rem;
     height: 1.5rem;
   }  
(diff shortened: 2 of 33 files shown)


assets/src/components/nav/topNavMobile.tsx
```diff
@@ -0,0 +1,153 @@
...
+              title="About Skate"
+            >
+              {questionMarkIcon("m-top-nav-mobile__menu-icon")}
+              About Skate
+            </button>
+          </li>
+
+          <li>
+            <NavLink
+              className="m-top-nav-mobile__menu-link"
```
lemald, line 100: Perhaps more of a design question, but should this also have a separate active state and corresponding `activeClassName`?

assets/src/components/nav/topNavMobile.tsx
```diff
@@ -0,0 +1,153 @@
...
+      </div>
+
+      <div
+        data-testid="mobile-overlay" 
+        className={
+          "m-top-nav-mobile-overlay" +
+          (mobileMenuIsOpen ? " m-top-nav-mobile-overlay__open" : "")
+        }
+        onClick={() => dispatch(toggleMobileMenu())}
+      ></div>
```
lemald, line 120: If there are no children I think you could condense this ever so slightly to a self-closing tag, i.e. `<div [...attributes...] />`.

assets/tests/components/nav.test.tsx
```diff
@@ -71,7 +71,6 @@ describe("Nav", () => {
       </BrowserRouter>
     )
 
-    expect(result.queryByText("Mobile nav placeholder.")).not.toBeNull()
```
lemald, line 74: Should probably also update the test description here and add an assertion that the mobile nav is rendered.

assets/css/nav/_top_nav_mobile.scss
```diff
@@ -0,0 +1,108 @@
...
+  background-color: $color-gray-50;
+  border-bottom: 1px solid $color-gray-300;
+  z-index: 108;
+}
+
+.m-top-nav-mobile-overlay {
+  position: fixed;
+  padding: 0px;
+  margin: 0px;
+  display: none;
```
lemald, line 15: For what it's worth, historically for rendering things that should only display in some situations we have used conditionals in the JSX rather than the method of setting / unsetting `display: none` via CSS. I think this approach is fine too, and I'd actually be curious if there's a strong reason to prefer one over the other going forward.

assets/src/components/nav/topNavMobile.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const { routeTabs, notificationDrawerIsOpen, mobileMenuIsOpen } = state
+
+  const bellIconClasses = notificationDrawerIsOpen
+    ? ["m-top-nav__notifications-icon", "m-top-nav__notifications-icon--active"]
+    : ["m-top-nav__notifications-icon"]
+
+  let tabName = "Untitled"
+  const showTabName = location.pathname === "/"
+  const currentTab = currentRouteTab(routeTabs)
+  if (showTabName && currentTab) tabName = currentTab.presetName || "Untitled"
```
lemald, line 34: Given that we now have logic both here and in `LadderPage` for defaulting the display name of a tab to "Untitled" if it isn't saved as a preset, I wonder if we could split some of this out into a small helper function in `models/routeTab.ts`.

lemald (changes requested):
Okay, I have a bunch of comments on individual things but overall I think this looks really good.

I really like the comprehensive tests of the top nav component itself. If possible it might be nice to also test the components where you added the conditional blur class logic. Also, it's a minor detail, but separate tests for the new toggle to the mobile menu state in `state.ts` would be good.

assets/css/_nav.scss
```diff
@@ -10,6 +11,11 @@
   height: 100%;
 }
 
+.m-nav--narrow {
+  width: 100%;
+  height: 100%;
+}
```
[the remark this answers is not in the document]
lxhughes: resolved in a817525; separated narrow and wide entirely as noted for future building

assets/css/nav/_top_nav_mobile.scss
```diff
@@ -0,0 +1,108 @@
...
+  &:focus {
+    fill: $color-gray-500;
+  }
+}
+
+.m-top-nav-mobile__notifications-icon svg {
+  fill: $color-eggplant-500;
+}
+
+.m-top-nav-mobile__menu-open {
```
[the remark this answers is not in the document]
lxhughes: resolved in 1e2ca8b

assets/css/nav/_top_nav_mobile.scss
```diff
@@ -0,0 +1,108 @@
...
+  max-width: 100%;
+  height: 100%;
+  position: absolute;
+  left: 0;
+  z-index: 110;
+  padding: 0px;
+  background-color: $color-gray-50;
+}
+
+.m-top-nav-mobile__menu-closed {
```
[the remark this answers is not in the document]
lxhughes: resolved in 1e2ca8b

assets/src/components/nav/topNavMobile.tsx
lxhughes: I added a matching activeClassName to Left Nav in 29da05c

assets/src/components/nav/topNavMobile.tsx
lxhughes: fixed in 710a1cf

assets/src/components/nav/topNavMobile.tsx
lxhughes: added this in 5760972

assets/src/models/routeTab.ts
```diff
@@ -75,6 +75,31 @@ export const isEditedPreset = (routeTab: RouteTab): boolean =>
...
+  )
+}
+
+export const pageOrTabName = (routeTabs: RouteTab[]): string => {
+  let tabName = "Skate"
+  if (location.pathname === "/") tabName = currentTabName(routeTabs)
+  else
+    tabName = toTitleCase(location.pathname.replace("/", "").replace("-", " "))
+  return tabName
+}
```
lemald, line 101: A couple of things on these functions:
1. It's probably better to use the `useLocation()` hook from `react-router-dom` instead of referencing so that the component rerenders if the location changes, which might be relevant with how React router does things internally. Note that since you can't call hooks from functions that aren't components this means you'd need to take the location as an argument to this function and put the `useLocation` call in the relevant component(s).
2. I'm not sure that the logic about the page title really belongs in `routeTab.ts` - it might make sense to just put these in the same file as the mobile top nav component as helper functions if that's the only place in the app that needs this logic.

lxhughes: fixed in 3bac2c8

assets/src/components/nav/topNavMobile.tsx
```diff
@@ -13,10 +13,28 @@ import {
...
 
+export const toTitleCase = (str: string): string => {
+  return str.replace(
+    /\w\S*/g,
+    (txt: string): string =>
+      txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()
+  )
+}
+
+export const pageOrTabName = (location: any, routeTabs: RouteTab[]): string => {
```
lemald, line 28: I think it's possible to give `location` a more specific type than `any` here - see `src/helpers/appCue.ts` for an example.

lxhughes: Resolved in eb41669 - also in the process I moved a function to create a test Location out of the appCue test and into a test helper.

lemald (approved):
Looks good to me, pending product and design review!

lemald:
@lxhughes okay, I've had a chance to look over your latest changes and they all look good to me. 👍

PR #1585, opened 2022-06-13
Google Analytics

lemald:
Asana ticket: [⚙️ Set up Google Analytics for Skate and port FullStory custom events](https://app.asana.com/0/1148853526253445/1202175790100734/f)

This PR takes care of:
1. Passing the Google Tag Manager ID to the front-end.
2. Including the GTM scripts on the page and initializing everything.
3. Firing events for all of the FullStory events we decided we wanted to port over.
4. Generating a UUID for each user and passing it to the front-end.
5. Using this UUID to identify the user to Google and Clarity.
6. Implementing a report that internal users can use to determine which UUID corresponds to which username.

Note that right now it also includes the schema change for the UUID field so that builds pass and I can deploy to dev-blue. Once #1584 is merged I'll rebase this PR.


assets/src/hooks/useGarageFilter.tsx
```diff
@@ -82,6 +83,7 @@ export const GarageFilter = ({
                 onClick={() => {
                   if (!filteredGarages.includes(garage) && window.FS) {
                     window.FS.event("User filtered routes by garage")
+                    tagManagerEvent("filtered_routes_by_garage")
```
lxhughes, line 86: Can tagManagerEvent() be called even if window.FS is undefined?

lxhughes (approved):
just one question, not a showstopper

PR #1589, opened 2022-06-15
fix: Shape on map in VPP not updating when trip changes

lxhughes:
https://app.asana.com/0/1148853526253437/1202207261666275/f

Simple fix - adding tripID to dependency array in useMapShape hook

### DIFF
--- a/assets/src/hooks/useShapes.ts
+++ b/assets/src/hooks/useShapes.ts
@@ -56,7 +56,7 @@ export const useTripShape = (tripId: TripId | null): Shape[] => {
         setShape(shapeResult)
       )
     }
-  }, [])
+  }, [tripId])
 
   return shape === null ? [] : [shape]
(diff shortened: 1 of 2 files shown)


assets/src/hooks/useShapes.ts
```diff
@@ -56,7 +56,7 @@ export const useTripShape = (tripId: TripId | null): Shape[] => {
         setShape(shapeResult)
       )
     }
-  }, [])
+  }, [tripId])
```
lemald, line 59: With React Hooks Testing Library, it's actually possible to test that this behaves in the way that you want it to. See the `useSwings` hook tests for examples that show how to test that the hook doesn't re-fire when the requested routes don't change, but that it does re-fire when they do change.

lemald (approved):
👍

PR #1592, opened 2022-06-17
feat: persistent state on reload for views and filter visibility

lxhughes:
Created a `showGaragesFilter` global variable.

Added the following to local storage:

1. `notificationDrawerIsOpen`
2. `openView`
3. `selectedVehicleOrGhost`
4. `pickerContainerIsVisible`
5. `showGaragesFilter`

### DIFF
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -31,7 +31,6 @@ const LOCALLY_PERSISTED_KEYS: Key[] = [
   ["pickerContainerIsVisible"],
   ["selectedShuttleRouteIds"],
   ["selectedShuttleRunIds"],
-  ["selectedVehicleOrGhost"],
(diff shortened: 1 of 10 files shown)


assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -26,9 +26,14 @@ const APP_STATE_KEY = "mbta-skate-state"
 type Key = string[]
 
 const LOCALLY_PERSISTED_KEYS: Key[] = [
+  ["notificationDrawerIsOpen"],
+  ["openView"],
+  ["pickerContainerIsVisible"],
   ["selectedShuttleRouteIds"],
   ["selectedShuttleRunIds"],
+  ["selectedVehicleOrGhost"],
```
lemald, line 34: Just curious, how does this play along with cases where the vehicle is no longer active when the user opens the application again?

lemald (approved):

PR #1599, opened 2022-06-24
upgrade: React Router 6

lxhughes:
[[Asana](https://app.asana.com/0/1152340551558956/1202313309645325/f)]

Upgrades `react-router-dom` to version 6.3.0 from v5, following [instructions from reactrouter.com](https://reactrouter.com/docs/en/v6/upgrading/v5#upgrade-to-react-router-v6). 

* Surrounds `<Route>` elements with `<Routes>`
* Replaces `<Route component>` with `<Route element>`
* Removes `useHistory`
* Removes `<NavLink exact>` attribute
* Removes `<NavLink activeClassName>` attribute
* Rewrites tests to always include `<BrowserRouter>` in renders
* Removes `optimize-css-assets-webpack-plugin` as it is not compatible
* Recreates `package.json` and `package-lock.json`


assets/package.json
```diff
@@ -67,7 +66,6 @@
     "jest": "^26.6.3",
     "mini-css-extract-plugin": "^2.6.0",
     "node-sass": "^7.0.1",
-    "optimize-css-assets-webpack-plugin": "^6.0.1",
```
lemald, line 70: There's still a reference to this in the Webpack configuration, causing building the static content to fail.

lemald: Also according to the `optimize-css-assets-webpack-plugin`, there's another plugin you can use with newer versions of Webpack called [css-minimizer-webpack-plugin](https://github.com/webpack-contrib/css-minimizer-webpack-plugin), so it might be worth switching to that rather than moving the plugin completely.

Followup question: was there some additional incompatibility with React Router 6, or was this because of Webpack 5? I'm surprised this didn't become an issue earlier (I upgraded from Webpack 4 to 5 just a few months ago).

lemald (changes requested):
Looks good overall, but there are a couple of specific parts that need minor changes.

assets/src/components/rightPanel.tsx
```diff
@@ -16,10 +16,12 @@ const RightPanel = ({
...
   // TODO delete when notifications are viewable from anywhere
-  const history = useHistory()
-  if (history) {
+  const location = useLocation()
+  if (location) {
     /* istanbul ignore next */
-    history.listen(() => dispatch(setNotification(undefined)))
+    useEffect(() => {
+      dispatch(setNotification(undefined))
+    }, [location])
```
lemald, line 24: Additionally, if you do end up keeping this new `useEffect` hook, you should make sure to `git rebase master` so that you pull in the ESLint React hooks rule.

lemald: Since there are conflicts in `package-lock.json`, you'll probably want to accept the changes from `master` and then re-run the `npm install` command(s) you used.

assets/src/components/rightPanel.tsx
```diff
@@ -16,10 +16,12 @@ const RightPanel = ({
 
   // close notification if you move away from ladder page
   // TODO delete when notifications are viewable from anywhere
-  const history = useHistory()
-  if (history) {
+  const location = useLocation()
```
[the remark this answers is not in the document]
lxhughes: Removed this code in 32c4785d, didn't see any issues in cursory tests but lmk if there are specific tests you want me to make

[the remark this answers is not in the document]
lemald: That's probably fine. I'm thinking we can deploy this to dev-blue once you get the Webpack issues sorted out and then I can play around with it a bit there to see if I notice anything unusual.

assets/package.json
lxhughes: I installed css-minimizer-webpack-plugin and updated webpack.config.js in 07823ee8

lemald: For what it's worth, you can try `npm run deploy` to try out the Webpack build without doing a full build of the Docker image.

assets/src/components/rightPanel.tsx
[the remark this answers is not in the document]
lxhughes: The only warnings were about the size of app.js, presumably pre-existing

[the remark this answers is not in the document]
lemald: Yeah, those have been there for a while. I have a ticket in the tech debt backlog to experiment with things like code splitting to reduce the size of the main JS file.

lemald (approved):
Looks good to me!

PR #1645, opened 2022-07-15
chore: remove secretsmanager

lxhughes:
Asana: https://app.asana.com/0/1152340551558956/1201977461483227/f


lemald (approved):
Code looks good, but we should definitely test this one out on a dev environment, when one is available, prior to merging.

PR #1656, opened 2022-07-22
Fix: Chelsea Bridge Notification Popup

lxhughes:
Asana: https://app.asana.com/0/1148853526253437/1202535232184138/f

When the Chelsea St Bridge notification is clicked, this fix replaces the unending/uncloseable loading message with a simple pop-up, currently with the same text as the notification card. Pop-up can be closed. Read/unread capability works as normal.


assets/tests/components/notificationModals/chelseaRaisedNotificationModal.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react"
+import { mount } from "enzyme"
```
lemald, line 2: **issue:** when adding new tests we should use React Testing Library since Enzyme is no longer well-supported (and I'd like to gradually transition our old tests over).

assets/tests/components/notificationModals/chelseaRaisedNotificationModal.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import ChelseaRaisedNotificationModal from "../../../src/components/notificationModals/chelseaRaisedNotificationModal"
+import { Notification, NotificationState } from "../../../src/realtime.d"
+
+describe("ChelseaRaisedNotificationModal", () => {
```
lemald, line 6: **issue:** should also add a test for the close button interaction.

assets/src/components/notificationModals/chelseaLoweredNotificationModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { StateDispatchContext } from "../../contexts/stateDispatchContext"
+import { closeIcon } from "../../helpers/icon"
+import { setNotification } from "../../state"
+
+const ChelseaLoweredNotificationModal = () => {
+  const [, dispatch] = useContext(StateDispatchContext)
+
+  const closeModal = () => {
+    dispatch(setNotification(undefined))
+  }
```
lemald, line 11: **issue:** most of this component doesn't need tests but the close interaction should get one.

lemald (changes requested):
Overall looks good! Just some comments on testing.

assets/tests/components/notificationModals/chelseaRaisedNotificationModal.test.tsx
lxhughes: fixed in 15a8e5f2

assets/tests/components/notificationModals/chelseaRaisedNotificationModal.test.tsx
lxhughes: fixed in 1213e1a

assets/src/components/notificationModals/chelseaLoweredNotificationModal.tsx
lxhughes: fixed in 1213e1a

assets/src/components/modal.tsx
```diff
@@ -23,6 +25,22 @@ const Modal = (): ReactElement | null => {
...
+      <ChelseaRaisedNotificationModal notification={selectedNotification} />
+    )
+  }
+
+  if (
+    selectedNotification &&
+    selectedNotification.reason == "chelsea_st_bridge_lowered"
+  ) {
+    return <ChelseaLoweredNotificationModal />
+  }
```
lemald, line 42: **issue:** sorry, one more thing that I should have caught in my original review, but `modal.test.tsx` should also have tests to make sure it renders the correct modals depending on the selected notification.

lemald (approved):

PR #1661, opened 2022-07-26
Mobile portrait nav changes

lemald:
Asana ticket: [⚙️ Nav | Implement mobile portrait changes](https://app.asana.com/0/1200180014510248/1202630445285825/f)

As part of this I split the components up a bit to create a separate `MobilePortraitNav` component that takes care of the logic to show / hide the top + bottom nav depending on whether or not a view is open. The "is a view open" logic is actually kind of complicated. As far as I can tell, for now it will be specific to mobile nav but in the future it might be good to make it a utility function... somewhere, I'm not sure what the most logical spot would be.

### DIFF
--- a/assets/css/_nav.scss
+++ b/assets/css/_nav.scss
@@ -23,6 +23,11 @@
   height: 100%;
 }
 
+.m-nav--narrow.m-nav--covered {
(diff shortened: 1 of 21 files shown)


lxhughes (approved):
This is a really clever way of handling this!

PR #1663, opened 2022-07-28
chore: Visually differentiate navlinks and views

lxhughes:
Asana: https://app.asana.com/0/1200180014510248/1202630445285838/f

Desktop only. VIsual changes to left nav, as well as moving the Notifications from top nav to left nav.


assets/css/nav/_left_nav.scss
```diff
@@ -57,9 +61,45 @@
...
+  padding: 0.25rem 0.5rem;
+  box-sizing: border-box;
+  width: 100%;
+  height: 2.25rem;
+  border-radius: 0.375rem;
+  font-weight: 400;
+  @include font-body;
+  color: $color-gray-600;
+  text-decoration: none;
+  background: #ffffff;
```
lemald, line 88: **nitpick:** it doesn't matter as much as with some of the other hex values, but you can use `$white` here.

lemald (changes requested):
A couple of minor comments and one heads up about a merge conflict. Also, one thing I noticed in the Figma but don't see here is the change to the Eggplant-200 hex value.

assets/css/nav/_left_nav.scss
lxhughes: fixed in 995b9260

assets/css/nav/_left_nav.scss
```diff
@@ -57,9 +61,45 @@
...
+  border: 1px solid #ecedee;
```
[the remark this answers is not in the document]
lxhughes: fixed in 995b9260

lemald:
@lxhughes looking good to me! I'll hold off on actually approving until we can get #1661 merged into master and then make sure that the diff for this PR still looks good.

PR #1676, opened 2022-08-04
fix: don't make late view take up full screen on old nav

lemald:
Asana ticket: [🐞 [extra] Late view taking up entire screen](https://app.asana.com/0/1200180014510248/1202707315274268/f)

With the old nav combined with the changes to make late view open on the right, we were setting both `left` and `right` values for the late view that caused it to take up basically the whole screen.


lxhughes (approved):
works in my tests!

PR #1688, opened 2022-08-15
chore: display new nav to all

lemald:
Asana ticket: [⚙️ Remove beta group and ship to GA](https://app.asana.com/0/1200180014510248/1202140732775596/f)

This does a fairly minimal set of changes to display the new navigation to everyone, without going through and cleaning up all of the code related to the old navigation. This way it will be relatively quick and easy to switch back to the old navigation in case of issues once we release it. Once it has run for a few days, we have [a ticket to go in and thoroughly clean up the old navigation code](https://app.asana.com/0/1200180014510248/1202131476892177/f).

### DIFF
--- a/assets/src/components/nav.tsx
+++ b/assets/src/components/nav.tsx
@@ -1,9 +1,7 @@
 import React, { useContext } from "react"
 import { OpenView, toggleMobileMenu } from "../state"
-import TabBar from "./tabBar"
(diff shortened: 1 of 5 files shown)


KaylaBrady (approved):
👍 Haven't seen feature flags in elixir yet - laboratory seems cool!

PR #1689, opened 2022-08-15
Expand tap targets and other nav styling changes

lemald:
Asana ticket: [⚙️ Implement color and tap target size changes](https://app.asana.com/0/1200180014510248/1202801429560543/f)


KaylaBrady (approved):
👍 Will this create conflicts with the snapshot test changes in https://github.com/mbta/skate/pull/1688?

lemald:
> 👍 Will this create conflicts with the snapshot test changes in #1688?

I don't think so - the snapshot tests just look at the rendered HTML and not the CSS. If we had needed to change the classes on some elements it might have caused issues.

PR #1691, opened 2022-08-16
chore(deps): remove pigeon

lemald:
Asana ticket: [⚙️[extra] remove Pigeon code](https://app.asana.com/0/1200180014510248/1202801429560548/f)

We had experimented a bit with the pigeon-maps library a while back, which initially looked promising but ended up not being quite full-featured enough for our needs. This PR just removes some code from when we thought we were adopting Pigeon to clear things up before doing other map work.


KaylaBrady (approved):
👍

PR #1699, opened 2022-08-22
Remove old nav code

lemald:
Asana ticket: [⚙️  Remove code for old version of navigation](https://app.asana.com/0/1200180014510248/1202131476892177/f)

This is mostly just removing existing code. The one place where it looks like there are substantial additions `_ladder_page.scss` is actually a side-effect if taking a bunch of Sass code out of block and thereby changing its indentation.

The changes are broken down topically commit-by-commit if that's an easier way to do review.


assets/src/components/viewHeader.tsx
```diff
@@ -36,8 +35,7 @@ const ViewHeader: ViewHeaderType = ({
     <div className="m-view-header">
       {backlinkToView &&
       backlinkToView !== OpenView.None &&
-      deviceType === "mobile" &&
-      featureIsEnabled("nav_beta") ? (
+      deviceType === "mobile" ? (
```
KaylaBrady, line 38: question (non-blocking): Does this mean there is a behavior change on the front-end for mobile devices as part of this PR? Or is `featureIsEnabled("nav_beta")` always returning true at this point?

KaylaBrady (approved):
👍 appreciate the commit-by-commit breakdown!

assets/src/components/viewHeader.tsx
lemald: (I totally didn't put two and two together on this when going through and mechanistically deleting the code related to the feature flag, so thanks for pointing it out.)

PR #1702, opened 2022-08-24
fix(formattedScheduledTime): Interpret given offset in seconds

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1202737931015251/f

Interprets the offset given to `formattedScheduledTime` in seconds to match the unit as defined in the TM XML API.

### DIFF
--- a/assets/src/util/dateTime.ts
+++ b/assets/src/util/dateTime.ts
@@ -55,8 +55,8 @@ export const formattedScheduledTime = (
   time: number,
   offset?: number | undefined
 ): string => {
-  const offsetInMinutes = offset ? Math.floor(offset / 60) : 0
-  let minutes = Math.floor(time / 60) + offsetInMinutes
+  const offsetSeconds = offset || 0
+  let minutes = Math.floor((time + offsetSeconds) / 60)
   if (minutes < 0) {
     /* if the offset shifts the time before midnight */
     minutes += 1440
--- a/assets/src/util/dateTime.ts
+++ b/assets/src/util/dateTime.ts
@@ -49,13 +49,14 @@ export const formattedTimeDiffUnderThreshold = (
     : formattedTime(b)
 }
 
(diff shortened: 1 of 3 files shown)


assets/src/util/dateTime.ts
```diff
@@ -49,13 +49,14 @@ export const formattedTimeDiffUnderThreshold = (
...
-/** Takes a time of day in seconds since midnight, offset in minutes (reasonably -720 to 720)
+/** Takes a time of day in seconds since midnight, offset in seconds (reasonably -8640 to 8640 seconds, or -12 to 12 hours)
  */
 export const formattedScheduledTime = (
   time: number,
   offset?: number | undefined
 ): string => {
-  let minutes = Math.floor(time / 60) + (offset ? offset : 0)
+  const offsetInMinutes = offset ? Math.floor(offset / 60) : 0
+  let minutes = Math.floor(time / 60) + offsetInMinutes
```
KaylaBrady, line 59: question: would the given offset ever not be a multiple of 60? Thinking there could be an edge case if we are doing `Math.floor` separately on the offset and the time. Maybe this should be `let minutes = Math.floor((time + offset) / 60)`?

lemald: Good catch! I'm not actually sure if it's possible to do non-whole increments of minutes in the UI, though if it's easy enough to combine into one `Math.floor` call that would probably be best to safeguard us from future bugs.

assets/tests/util/dateTime.test.ts
```diff
@@ -124,12 +124,12 @@ describe("formattedScheduledTime", () => {
     expect(formattedScheduledTime(90900)).toEqual("1:15 AM")
   })
   test("applies offset minutes", () => {
```
lemald, line 126: **nitpick:** the test should probably now be called "applies offset seconds."

lemald (approved):

PR #1705, opened 2022-08-26
New close button designs

lemald:
Asana ticket: [⚙️  Implement new close button](https://app.asana.com/0/1200180014510248/1202728285417108/f)

This is split up into more digestible commits for easier review. With respect to the enumeration of styles available in the `CloseButton` component, I opted to separate out size and color into different CSS classes under the hood for reusability, but I'm only exposing the permutations of those that are actually in the designs via the `CloseButtonType` type, which I think makes sense in terms of what options to expose to other parts of the application, but I'm open to pushback.


assets/css/_presets.scss
```diff
@@ -6,7 +6,7 @@
     li {
       display: flex;
       justify-content: space-between;
-      align-items: flex-end;
+      align-items: center;
```
KaylaBrady, line 9: question (non-blocking): Intentional change to align-items center? Seems like it could affect more than the close button

KaylaBrady (approved):

assets/css/_presets.scss
lemald: Correct - previously the preset name and the close button were about the same height and `flex-end` gave us the alignment we wanted, but with the larger button and slightly larger font it looked pretty weird.

assets/src/components/closeButton.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import { closeXIcon } from "../helpers/icon"
+
+export type CloseButtonType =
+  | "s_light"
+  | "l_light"
+  | "s_dark"
+  | "l_dark"
+  | "l_green"
+  | "xl_light"
```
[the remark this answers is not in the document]
lemald: That's a good catch, this code and the designs have kind of been co-evolving so I should do a final pass to make sure the list lines up with what's in Figma.

KaylaBrady (approved):

PR #1706, opened 2022-08-26
refactor(tests): convert a subset of enzyme tests to use react testing library

KaylaBrady:
A subset of files from https://app.asana.com/0/1152340551558956/1201708616021563/f.

### DIFF
--- a/assets/tests/components/app.test.tsx
+++ b/assets/tests/components/app.test.tsx
@@ -1,4 +1,3 @@
-import { mount } from "enzyme"
 import React from "react"
 import renderer from "react-test-renderer"
 import { render } from "@testing-library/react"
@@ -27,7 +26,7 @@ describe("App", () => {
   })
 
   test("shows disconnected modal if the socket is disconnected", () => {
-    const wrapper = mount(
+    const result = render(
       <SocketProvider
         socketStatus={{
           socket: undefined,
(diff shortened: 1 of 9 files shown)


assets/tests/components/app.test.tsx
```diff
@@ -37,7 +36,9 @@ describe("App", () => {
         <App />
       </SocketProvider>
     )
-    expect(wrapper.exists(".c-modal")).toBeTruthy()
+    expect(
+      result.queryByText(/Your connection to Skate has expired./)
+    ).not.toBeNull()
```
KaylaBrady, line 41: I used `.not.toBeNull` here for consistency with other tests. I've previously seen `toBeVisible()` used from [jest-dom](https://github.com/testing-library/jest-dom#tobevisible), which I found helpful for brevity when first learning react testing, but I don't feel strongly enough about it to break the existing testing conventions here!

lemald: I actually wasn't aware of `toBeVisible`, that actually looks better in at least some respects (its assertions cover more cases that might lead to the element not showing up on the page, and the naming is more clear about what we're looking to test). Let's use it going forward!

lemald:
@KaylaBrady I see this is still marked as a draft, so I'll hold off on approving it until you mark it as ready for review, but this is looking really good so far!

lemald (approved):
👍

PR #1707, opened 2022-08-29
fix: better handle partially- or fully-missing operator information

lemald:
Asana ticket: [⚙️ Update data null message](https://app.asana.com/0/1200180014510248/1202883819004130/f)

Previously, when we were missing operator information, the application would display "null null #null" in that field. This provides a more user-friendly "Not Available" message as well as handling intermediate cases where some pieces of the operator information are available but others aren't.


arkadyan (approved):
👍

PR #1710, opened 2022-08-30
feat: more responsive layout for Late View table

lemald:
Asana ticket: [🎨⚙️ Make tables responsive](https://app.asana.com/0/1200180014510248/1202729338901503/f)

This incorporates a variety of changes discussed between dev and design that add up to a more responsive table. Adjustments include:
1. Using percentage widths for `<th>` column header elements, making the table shrink / grow more gracefully.
2. Reducing the width of the "Scheduled Logon" column so that the title wraps by default.
3. The container has fixed-width on larger screens, but on smaller screens now has `width: 100%`.
4. The table container panels can scroll horizontally if your screen is _really_ small.

### DIFF
--- a/assets/css/_late_view.scss
+++ b/assets/css/_late_view.scss
@@ -4,10 +4,15 @@ $late-view-background-color: $color-bg-light;
   @include modal-content;
   background-color: $late-view-background-color;
   height: 100%;
+  width: 33.5rem;
   position: absolute;
   right: 0;
   top: 0;
   z-index: 1001;
+
+  @media screen and (max-width: 700px) {
+    width: 100%;
+  }
 }
 
 .m-late-view__content-wrapper {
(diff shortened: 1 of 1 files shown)


assets/css/_late_view.scss
```diff
@@ -4,10 +4,15 @@ $late-view-background-color: $color-bg-light;
   @include modal-content;
   background-color: $late-view-background-color;
   height: 100%;
+  width: 33.5rem;
   position: absolute;
   right: 0;
   top: 0;
   z-index: 1001;
+
+  @media screen and (max-width: 700px) {
```
arkadyan, line 13: question: This is close to, but not the same as `$mobile-max-width`. I assume it's intentional to use a different value in this one place?

arkadyan (approved):
👍 This all looks good on the code side.

assets/css/_late_view.scss
lemald: Correct, it's intentional based on how narrow the table can get and therefore when you start to run into issues with it being partially-hidden under the left nav bar, so specific to the content of this particular component.

PR #1719, opened 2022-09-08
fix(MobilePortraitNav): use single grid layout to prevent scroll position resetting when a view is opened

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1202946731207901/f

The issue seemed to be that when opening the VPP, the page was rendered with a new grid, which seemed to reset the scroll position of the ladder page. 

I'm not sure that the approach I've taken here is the cleanest one. My intention was to fix the issue without having to introduce storing the scroll position in state. By using one grid class and conditionally hiding the top / bottom navs, the scroll reset issue is no longer a problem. If there is a cleaner or more conventional approach here I am certainly open to changing!

I also wasn't sure of the best way to introduce a regression test for this - there is already test coverage to make sure the navs are conditionally shown / hidden, but maintaining scroll position seems like it might require using something like Cypress.

### DIFF
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -9,6 +9,7 @@
   top: 0;
   width: 23.5rem;
   z-index: 1001;
+
   @media screen and (max-width: $mobile-max-width) {
     width: 100%;
   }
@@ -38,6 +39,7 @@
   cursor: pointer;
   position: absolute;
   right: 0;
+
   &:hover {
     font-weight: 700;
   }
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -13,11 +13,10 @@
   @media screen and (max-width: $mobile-max-width) {
     width: 100%;
   }
-  @media screen and (max-width: $mobile-max-width) {
+  @media screen and (max-width: $mobile-portrait-max-width) {
     position: fixed;
     top: 0;
     bottom: 0;
-    width: 100%;
     overflow-y: scroll;
   }
 }
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -15,9 +15,7 @@
   }
   @media screen and (max-width: $mobile-portrait-max-width) {
     position: fixed;
-    top: 0;
     bottom: 0;
-    overflow-y: scroll;
   }
 }
 
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -9,9 +9,15 @@
   top: 0;
   width: 23.5rem;
   z-index: 1001;
-
   @media screen and (max-width: $mobile-max-width) {
     width: 100%;
+  }
+  @media screen and (max-width: $mobile-max-width) {
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    width: 100%;
+    overflow-y: scroll;
   }
 }
 
@@ -32,7 +38,6 @@
   cursor: pointer;
   position: absolute;
   right: 0;
-
   &:hover {
     font-weight: 700;
   }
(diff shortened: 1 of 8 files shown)


assets/src/components/nav/mobilePortraitNav.tsx
```diff
@@ -19,33 +19,37 @@ const MobilePortraitNav = ({
...
-        <div className="m-nav__app-content">{children}</div>
+  const isViewOpen = openView !== OpenView.None || selectedVehicleOrGhost
+
+  const navVisibilityStyle = isViewOpen ? "hidden" : "visible"
+
+  return (
+    <div className="m-nav--narrow">
+      <div
+        className="m-nav__nav-bar m-nav__nav-bar--top"
+        style={{ visibility: navVisibilityStyle }}
```
KaylaBrady, line 30: I opted to do inline styling here as a work around for this issue w/ [jest toBeVisible ](https://github.com/testing-library/jest-dom/issues/209) not working with separate stylesheets unless there is some css configuration work done. If there is a preferred alternative approach I'd be open to trying it!

lemald: My usual approach for these things has been to set a class based on the condition in question, and then just assert the presence / absence of the relevant classes on the element in my tests. This works, too, though, and `toBeVisible` actually tests the relevant property more directly.

assets/css/_notification_drawer.scss
```diff
@@ -13,6 +13,13 @@
   @media screen and (max-width: $mobile-max-width) {
     width: 100%;
   }
+  @media screen and (max-width: $mobile-max-width) {
```
lemald, line 16: Is `$mobile-max-width` actually the correct value? It's defined in `app.scss` as 768px and it was a breakpoint used in most old Skate designs, but it's not the same as the breakpoint for mobile portrait nav.

On the other hand, if it is supposed to be `$mobile-max-width`, this can be combined with the above selector / block.

assets/css/_notification_drawer.scss
```diff
@@ -13,6 +13,13 @@
...
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    width: 100%;
```
lemald, line 20: It should have the same outcome, but I wonder if `100vw` would be a bit clearer about the fact that the intent is to take up all of the horizontal space on the screen. I don't think we're actually consistent one way or another in the existing CSS about when we use `vw` / `vh`.

assets/css/_properties_panel.scss
```diff
@@ -18,6 +18,13 @@
       width: 100%;
     }
   }
+  @media screen and (max-width: $mobile-max-width) {
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    width: 100%;
```
lemald, line 25: Other comments from above apply here, but also note that the `width: 100%` here is redundant with `width: 100vw` in the block above.

assets/css/_swings_view.scss
```diff
@@ -27,6 +27,13 @@
       }
     }
   }
+  @media screen and (max-width: $mobile-max-width) {
```
lemald, line 30: See same question as above about `$mobile-max-width`.

assets/css/_notification_drawer.scss
KaylaBrady: Good call this is not the right value - as-is, shrinking to a window between 480-768px wide produces a broken mashup of navs. I'll introduce a new `$mobile-portrait-max-width` variable and reference that instead.

lemald: For what it's worth, the breakpoints for the new nav are defined in `src/hooks/useDeviceType.ts`. Since the nav logic happens largely in JavaScript, I didn't end up creating new SCSS constants for the breakpoints. I think eventually we want to retire the use of the existing `$mobile-max-width` and make all of the different breakpoints consistent.

lemald (approved):
👍

PR #1723, opened 2022-09-13
feat(views): sticky navigation headers on scroll

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1202832778108727/f

### DIFF
--- a/assets/css/_swings_view.scss
+++ b/assets/css/_swings_view.scss
@@ -56,7 +56,8 @@
   background-color: $color-bg-base;
   position: sticky;
   position: -webkit-sticky;
-  top: -1px;
+  // top offset to stick below the view-header
+  top: 2.5rem;
   z-index: 1002;
 
   .m-swings-view__table-header-cell-subheaders {
--- a/assets/css/_swings_view.scss
+++ b/assets/css/_swings_view.scss
@@ -55,9 +55,8 @@
   text-align: right;
   background-color: $color-bg-base;
   position: sticky;
-  position: -webkit-sticky;
   // top offset to stick below the view-header
-  top: 2.5rem;
+  top: $view-header-height;
   z-index: 1002;
 
   .m-swings-view__table-header-cell-subheaders {
--- a/assets/css/_view_header.scss
+++ b/assets/css/_view_header.scss
@@ -3,7 +3,7 @@
   top: 0;
   z-index: 1;
   width: 100%;
-  min-height: 2.5rem;
+  min-height: $view-header-height;
 
   display: flex;
   justify-content: center;
--- a/assets/css/_view_header.scss
+++ b/assets/css/_view_header.scss
@@ -1,5 +1,7 @@
 .m-view-header {
-  position: relative;
+  position: sticky;
+  top: 0;
+  z-index: 1;
   width: 100%;
   min-height: 2.5rem;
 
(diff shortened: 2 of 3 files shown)


assets/css/_view_header.scss
```diff
@@ -1,5 +1,7 @@
 .m-view-header {
-  position: relative;
+  position: sticky;
```
KaylaBrady, line 2: I saw that some other uses of `position: sticky` in this repo also specified `position: -webkit-sticky`. I don't see that prefix specified in the [mdn doc](https://developer.mozilla.org/en-US/docs/Web/CSS/position#browser_compatibility), so I'm thinking maybe it is an artifact of an old version and am not including. If it should be included though I'll add it!

assets/css/_view_header.scss
```diff
@@ -1,5 +1,7 @@
...
+  top: 0;
+  z-index: 1;
```
KaylaBrady, line 4: z-index: 1 here keeps the content from scrolling over the header. I'm sorting through why that works and if there could be a cleaner solution without specifying z-index

KaylaBrady: update: After a little more playing around, this seems to work because of the stack context. Looking only at the notifications drawer, I was able to get the sticky header without using z-index if alerting some of the `position: relative` and `position: absolute` configurations of elements within the sibling to this header. It seems like that would result in more significant changes & open up more surface area for possible regressions, so I'm opting to leave as-is for feedback.

lemald: If the only thing the z-index is doing here is triggering a new stacking context, could you explicitly set `isolation: isolate;`? I just learned about `isolation` from poking around MDN but it looks like it might do what we need: https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

assets/css/_view_header.scss
lemald: [Per Can I Use](https://caniuse.com/?search=sticky) it looks like `-webkit-sticky` was needed for certain versions of Safari but those are quite old at this point. We do actually have a [browser support policy](https://www.notion.so/mbta-downtown-crossing/Skate-Browser-Support-Policy-7cf57f38ea294afb889e63e95f73a300) that we put together a while back and we require a much newer version of Safari, so I'd say you're good to leave out `-webkit-sticky` or remove it if you're ever modifying code.

assets/css/_swings_view.scss
```diff
@@ -56,7 +56,8 @@
   background-color: $color-bg-base;
   position: sticky;
   position: -webkit-sticky;
-  top: -1px;
+  // top offset to stick below the view-header
+  top: 2.5rem;
```
lemald, line 60: One thing: for this, I'd prefer if we had a `$view-header-height` constant or something like that, which can then be referenced both here in `_view_header.scss` or anywhere else that needs to use that value for layout purposes.

assets/css/_view_header.scss
KaylaBrady: `isolation: isolate` doesn't seem to work in this case. I may be wrong here (stacking contexts is a new concept to me), but it  seems like the behavior of the current change in this PR is that setting z-index moves the view header up in the stacking context so that it overlays any element where `position` is set and `z-index` is not. 

That is why alternatively removing `position: relative` on other elements has the same effect: the current approach of the PR moves the header up a level in the stacking context, the alternative moves everything else down a level.

lemald: Yeah I'll confess I'm not great with stacking contexts.

In any event, I'd say it's preferable to go with the z-index solution here over the other one that involves changing the code in more places and introducing possible regressions.

lemald (approved):
Looks good to me on the code end, though it's probably worth a quick product / design review as well.

PR #1726, opened 2022-09-15
refactor(tests): convert another subset of enzyme tests to use react testing library

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1201708616021563/f

There are 10 more test files to go after this PR. This seemed like a meaningfully sized chunk for review, but I can keep adding to this PR if it would be preferable to review the larger set at once!

### DIFF
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -92,7 +92,7 @@ const LadderTab = ({
         {tab.isCurrentTab && showSaveIcon ? (
           <button
             className="m-ladder-page__tab-save-button"
-            title="save"
+            title="Save"
             onClick={(e) => {
               e.stopPropagation()
 
@@ -118,7 +118,7 @@ const AddTabButton = ({
   return (
     <button
       className="m-ladder-page__add-tab-button"
-      title="add tab"
+      title="Add Tab"
       onClick={() => {
         tagManagerEvent("new_tab_added")
         addTab()
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -92,7 +92,7 @@ const LadderTab = ({
         {tab.isCurrentTab && showSaveIcon ? (
           <button
             className="m-ladder-page__tab-save-button"
-            data-testid="save-button"
+            title="save"
             onClick={(e) => {
               e.stopPropagation()
 
@@ -118,7 +118,7 @@ const AddTabButton = ({
   return (
     <button
       className="m-ladder-page__add-tab-button"
-      data-testid="add-tab-button"
+      title="add tab"
       onClick={() => {
         tagManagerEvent("new_tab_added")
         addTab()
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -92,6 +92,7 @@ const LadderTab = ({
         {tab.isCurrentTab && showSaveIcon ? (
           <button
             className="m-ladder-page__tab-save-button"
+            data-testid="save-button"
             onClick={(e) => {
               e.stopPropagation()
 
@@ -117,6 +118,7 @@ const AddTabButton = ({
   return (
     <button
       className="m-ladder-page__add-tab-button"
+      data-testid="add-tab-button"
       onClick={() => {
         tagManagerEvent("new_tab_added")
         addTab()
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -53,7 +53,7 @@ const SearchForm = () => {
           />
           <button
             type="reset"
-            title="clear"
+            title="Clear"
             className="m-search-form__clear"
             onClick={clearTextInput}
           >
@@ -63,7 +63,7 @@ const SearchForm = () => {
 
         <button
           type="submit"
-          title="submit"
+          title="Submit"
           className="m-search-form__submit"
           onClick={subscribeToSearch}
           disabled={!isValidSearchQuery(query)}
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -17,8 +17,10 @@ const SearchForm = () => {
     },
     dispatch,
   ] = useContext(StateDispatchContext)
-  const handleTextInput = (event: React.FormEvent<HTMLInputElement>): void =>
-    dispatch(setSearchText(event.currentTarget.value))
+  const handleTextInput = (event: React.FormEvent<HTMLInputElement>): void => {
+    const value = event.currentTarget.value
+    dispatch(setSearchText(value))
+  }
 
   const clearTextInput = (event: React.FormEvent<EventTarget>): void => {
     event.preventDefault()
@@ -51,6 +53,7 @@ const SearchForm = () => {
           />
           <button
             type="reset"
+            title="clear"
             className="m-search-form__clear"
             onClick={clearTextInput}
           >
@@ -60,6 +63,7 @@ const SearchForm = () => {
 
         <button
           type="submit"
+          title="submit"
           className="m-search-form__submit"
           onClick={subscribeToSearch}
           disabled={!isValidSearchQuery(query)}
--- a/assets/src/hooks/useRouteFilter.tsx
+++ b/assets/src/hooks/useRouteFilter.tsx
@@ -61,7 +61,7 @@ export const RouteFilter = ({
           <button
             className="m-route-filter__clear"
             onClick={clearTextInput}
-            data-testid="clear-text"
+            title="Clear"
           >
             {circleXIcon()}
           </button>
--- a/assets/src/hooks/useRouteFilter.tsx
+++ b/assets/src/hooks/useRouteFilter.tsx
@@ -58,7 +58,11 @@ export const RouteFilter = ({
         />
 
         {filterText.length > 0 ? (
-          <button className="m-route-filter__clear" onClick={clearTextInput}>
+          <button
+            className="m-route-filter__clear"
+            onClick={clearTextInput}
+            data-testid="clear-text"
+          >
             {circleXIcon()}
           </button>
         ) : null}
--- a/assets/tests/components/ladderPage.test.tsx
+++ b/assets/tests/components/ladderPage.test.tsx
@@ -219,7 +219,7 @@ describe("LadderPage", () => {
       </StateDispatchProvider>
     )
 
-    await userEvent.click(result.getByTitle("save"))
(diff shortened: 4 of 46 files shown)


assets/src/components/ladderPage.tsx
```diff
@@ -117,6 +118,7 @@ const AddTabButton = ({
   return (
     <button
       className="m-ladder-page__add-tab-button"
+      data-testid="add-tab-button"
```
lemald, line 121: One thing we've done in the past is add a `title`, which React Testing Library can also query by. The `title` property has some issues but I think as long as you don't use it to encode information that isn't available via other means, and instead to just label some symbol / button, it's fine.

assets/src/components/ladderPage.tsx
```diff
@@ -92,6 +92,7 @@ const LadderTab = ({
         {tab.isCurrentTab && showSaveIcon ? (
           <button
             className="m-ladder-page__tab-save-button"
+            data-testid="save-button"
```
lemald, line 95: See the other note about `title`.

assets/tests/components/swingsView.test.tsx
```diff
@@ -1,7 +1,7 @@
-import { mount } from "enzyme"
```
lemald, line 1: Note that I've also made updates to `swingsView.test.tsx` in #1725 (and as part of that converted the old Enzyme tests to RTL). I had to do some substantial other changes because the other changes in that PR changed some stuff about hooks and how the page renders, so there probably isn't a ton of overlap between our changes (though I haven't compared them line-by-line yet).

lemald:
@KaylaBrady a couple of minor comments on what you've got so far, but otherwise looks good!

assets/tests/components/swingsView.test.tsx
KaylaBrady: Ah I forgot about that! I can remove it from this PR

assets/src/components/ladderPage.tsx
lemald: Yeah, I'd make it something that's formatted to be human-readable as it does show up as a tool tip and I believe in theory can also be referenced by things like screen readers. I usually don't involve design for things where the title seems fairly obvious but it wouldn't hurt to ask.

assets/src/components/ladderPage.tsx
```diff
@@ -92,7 +92,7 @@ const LadderTab = ({
         {tab.isCurrentTab && showSaveIcon ? (
           <button
             className="m-ladder-page__tab-save-button"
-            data-testid="save-button"
+            title="save"
```
lemald, line 95: We've usually used title case for the `title` attribute since that looks better in the tooltip. I'm not sure if that's a hard and fast rule, but it's at least the convention I've been following so far.

KaylaBrady:
@lemald I'm working resolving some console warning logs for one of the modified test files, but otherwise this encompasses the changes for all remaining non-swings test files. I pushed them all to the existing PR, but if you'd prefer I separate into a different PR I'll do so!

assets/src/components/searchForm.tsx
```diff
@@ -51,6 +53,7 @@ const SearchForm = () => {
           />
           <button
             type="reset"
+            title="clear"
```
lemald, line 56: Title should be capitalized

assets/src/components/searchForm.tsx
```diff
@@ -60,6 +63,7 @@ const SearchForm = () => {
 
         <button
           type="submit"
+          title="submit"
```
lemald, line 66: Capitalize title

assets/src/hooks/useRouteFilter.tsx
```diff
@@ -58,7 +58,11 @@ export const RouteFilter = ({
         />
 
         {filterText.length > 0 ? (
-          <button className="m-route-filter__clear" onClick={clearTextInput}>
+          <button
+            className="m-route-filter__clear"
+            onClick={clearTextInput}
+            data-testid="clear-text"
```
lemald, line 64: Could this be a title instead, or is there a reason that wouldn't be appropriate here?

assets/tests/components/routeVariantName.test.tsx
```diff
@@ -88,33 +91,35 @@ describe("RouteVariantName", () => {
       viaVariant: "_",
     }
 
-    const wrapper = shallow(<RouteVariantName vehicle={testVehicle} />)
+    const result = render(<RouteVariantName vehicle={testVehicle} />)
 
-    expect(wrapper.text()).toEqual("39_")
+    expect(result.getByTestId("variant-name")).toHaveTextContent("39_")
   })
+})
```
lemald, line 98: It looks like this closes the `describe(...)` call and the remaining tests are moved one level of indentation up, outside of a `describe`. Was this intentional?

lemald:
@KaylaBrady okay just a few more minor comments but overall this is looking great!

assets/tests/components/ladderPage.test.tsx
```diff
@@ -287,10 +287,10 @@ describe("LadderPage", () => {
       </StateDispatchProvider>
     )
 
-    expect(wrapper.find(".m-ladder-page__tab-save-button").length).toBe(0)
+    expect(result.queryByTestId("save-button")).toBeNull()
```
lemald, line 290: I think this needs to be updated to reference the title instead (the result should be null either way but right now I don't think it's actually an effective test if the test ID is no longer in place).

assets/tests/components/ladderPage.test.tsx
```diff
@@ -480,14 +480,14 @@ describe("LadderPage", () => {
...
-    const wrapper = mount(
+    const result = render(
       <StateDispatchProvider state={state} dispatch={jest.fn()}>
         <BrowserRouter>
           <LadderPage />
         </BrowserRouter>
       </StateDispatchProvider>
     )
-    expect(wrapper.find("#m-properties-panel").exists()).toBeFalsy()
+    expect(result.queryByText(/vehicles/)).toBeNull()
```
lemald, line 490: If I'm understanding this correctly (I assume it's looking for the text in the title bar?), I think the v should either be capitalized or the regex should be made case-insensitive.

assets/tests/components/ladderPage.test.tsx
```diff
@@ -496,14 +496,14 @@ describe("LadderPage", () => {
...
```
lemald, line 506: See above comment about `/vehicles/`.

assets/tests/components/lateView.test.tsx
```diff
@@ -135,7 +130,7 @@ describe("LateView", () => {
     expect(mockDispatch).toHaveBeenCalledWith(closeLateView())
   })
 
-  test("clicking ghost run number opens ghost and sends Fullstory event", () => {
+  test("clicking ghost run number opens ghost and sends Fullstory event", async () => {
```
lemald, line 133: Small nitpick, but if you're updating this line anyway it should probably refer to Google Tag Manager, not Fullstory.

assets/tests/components/lateView.test.tsx
```diff
@@ -282,66 +253,48 @@ describe("LateView", () => {
...
+    )
+    expectIndeterminateValue(result, missingMasterCheckboxId, true),
+      expectCheckedValue(result, missingMasterCheckboxId, false)
+
+    await userEvent.click(
+      result.getByTestId(`row-checkbox-${missingLogonGhost2.runId}`)
+    )
+
+    expectIndeterminateValue(result, missingMasterCheckboxId, false),
+      expectCheckedValue(result, missingMasterCheckboxId, true)
```
lemald, line 294: This was already a long block of code in the original test, but some comments highlighting the individual interactions being tested at different spots would probably make it a lot more readable if you're able to add them (totally optional!).

assets/tests/components/lateView.test.tsx
```diff
@@ -880,59 +774,35 @@ describe("LateView", () => {
...
       </StateDispatchProvider>
     )
-    expect(reloadWrapper.find(".m-late-view__data-row--selected")).toHaveLength(
-      0
-    )
-
-    expect(
-      reloadWrapper.find(".m-late-view__data-row--unselected")
-    ).toHaveLength(2)
+    expect(result.getAllByTestId(/row-checkbox/)).toHaveLength(2)
```
lemald, line 802: If I'm reading things correctly, the checkboxes get the `row-checkbox` test ID regardless of whether or not they're selected, so I think the assertion may need to be refined a bit to match the original test.

assets/tests/components/routeVariantName.test.tsx
```diff
@@ -99,9 +102,9 @@ describe("RouteVariantName", () => {
       isShuttle: true,
     }
 
-    const wrapper = shallow(<RouteVariantName vehicle={testVehicle} />)
+    const result = render(<RouteVariantName vehicle={testVehicle} />)
 
-    expect(wrapper.text()).toEqual("Shuttle")
+    expect(result.getByText("Shuttle")).toBeTruthy()
```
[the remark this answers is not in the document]
lemald: (To be clear, I mean this specifically in reference to places where there's a test that a certain element exists.)

assets/tests/components/lateView.test.tsx
KaylaBrady: Yeah good point! I debated between breaking these tests up more, but commenting seems like a helpful change w/ a lower surface area

assets/tests/components/lateView.test.tsx
KaylaBrady: Great catch, I'll update! I did think about including a notion of whether or not the row is selected in the testid, but opted not to in favor of using the `toHaveProperty` test. This is an example of where that choice seemed to get a bit clunkier

assets/tests/components/routeVariantName.test.tsx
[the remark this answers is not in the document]
KaylaBrady: I stumbled up this [helpful doc ](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) this AM when trying to debug the `act` error log which has a [section](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence) that suggests only using `query*` when testing the absence of an element. I do think the `toBeTruthy` check here isn't very descriptive here and will change to `toBeInTheDocument` if that seems like an acceptable alternative!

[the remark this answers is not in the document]
lemald: Ah, interesting, I hadn't seen that page before! In any event that makes sense.

lemald (approved):
👍

PR #1732, opened 2022-09-21
chore: Add dev-green deployment action

KaylaBrady:
Adds new dev-green deployment action, largely copied from the dev-blue action. 

As part of this new environment setup, a new dev-green environment will need to be configured for the repo and require reviewers set if deploy approvals are desired. I don't seem to have the permissions to enable that - @lemald can you?


lemald:
@KaylaBrady I've create a dev-green environment with the same deploy review settings as dev-blue.

lemald (approved):

PR #1739, opened 2022-09-27
chore: Upgrade React, Leaflet, React Leaflet

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1200578295576925/f

This PR build's off @lemald's branch `lem-leaflet-upgrade` to upgrade to React 17, Leaflet 1.9, and React Leaflet 3. 

Upgrading to React 18 + React Leaflet 4 will simplify aspects of the Map component, but is out of scope for this PR since React 18 is a more significant upgrade.
* Support for ref in `MapContainer` ([issue](https://github.com/PaulLeCam/react-leaflet/issues/841))
* Ability to use `react-custom-control` library for simpler `RecenterControl` implementation ([issue](https://github.com/chris-m92/react-leaflet-custom-control/issues/6))

Additional checks related to React 17 breaking changes from [release guide](https://github.com/chris-m92/react-leaflet-custom-control/issues/6)
* No manual calls to `document.addEventListener`, which would now not be called by components that use `e.stopPropagation`
* onScroll stopped bubbling - manually tested route ladder / VPP scroll behavior 
* Effect cleanup timing - would be caught by the `exhaustive-deps` eslint plugin. Confirmed cleanups don't use mutable values for hooks where this check is disabled


assets/tests/components/map.test.tsx
```diff
@@ -3,7 +3,7 @@ import "@testing-library/jest-dom"
 import { LatLng } from "leaflet"
 import React, { MutableRefObject } from "react"
 import { act } from "react-dom/test-utils"
```
lemald, line 5: Minor thing while you're already updating this file, but the React Testing Library docs recommend importing their `act` wrapper: https://testing-library.com/docs/react-testing-library/api/#act

assets/tests/components/map.test.tsx
```diff
@@ -14,13 +14,7 @@ import { Shape } from "../../src/schedule"
...
 import userEvent from "@testing-library/user-event"
 
-jest.unmock("leaflet")
-jest.unmock("react-leaflet-control")
-
-jest.mock("../../src/laboratoryFeatures", () => ({
-  __esModule: true,
-  default: jest.fn(() => false),
-}))
+window.scrollTo = jest.fn()
```
lemald, line 17: Is it possible to make this update temporarily and restore the old value once the tests finish running? I worry about weird interactions with other tests that might reference this same function.

assets/tsconfig.json
```diff
@@ -9,7 +9,7 @@
   ],
   "compilerOptions": {
     /* Basic Options */
-    "target": "es5",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "target": "es6",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
```
lemald, line 12: Question: what change makes this update necessary? It looks like es6 has [enough browser support](https://caniuse.com/?search=es6) that we can safely use it, but I'm curious about the exact reason.

assets/src/components/map.tsx
```diff
@@ -329,34 +381,38 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
...
         <FullscreenControl position="topright" />
-        <RecenterControl turnOnAutoCenter={() => setShouldAutoCenter(true)} />
+        <RecenterControlButton
+          position="topright"
+          recenter={() => setShouldAutoCenter(true)}
+        />
+
         <TileLayer
           url={`${tilesetUrl()}/{z}/{x}/{y}.png`}
           attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
```
lemald, line 418: It looks like at one point we explicitly set the attribution (probably just to make it shorter and take up less space?), but now it displays the default attribution with the Leaflet link (on both prod and dev-green). If it's possible, it would be nice to fix that as part of this upgrade.

assets/tsconfig.json
KaylaBrady: At one stage of this PR I was running into an error instantiating `RecenterControl`, which this update seemed to resolve. May have been a confounding change that actually did the trick though since resetting to "es5" did not resurface the old error.

assets/tests/components/map.test.tsx
KaylaBrady: Updated to only mock once in the 1 test where mocking is needed

assets/src/components/map.tsx
KaylaBrady: Setting `prefix={false}` in [8fc146a](https://github.com/mbta/skate/pull/1739/commits/8fc146a1c5fc42eaccc3cada0f59136eee32a31e) did the trick

PR #1740, opened 2022-09-27
chore: add email to users table

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1201189412881312/f

Subsequent PR will write to this new column

### DIFF
--- a/priv/repo/migrations/20220927170631_add_user_email.exs
+++ b/priv/repo/migrations/20220927170631_add_user_email.exs
@@ -5,5 +5,6 @@ defmodule Skate.Repo.Migrations.AddUserEmail do
     alter table(:users) do
       add(:email, :string)
     end
+    create unique_index(:users, :email)
   end
 end
--- a/priv/repo/migrations/20220927170631_add_user_email.exs
+++ b/priv/repo/migrations/20220927170631_add_user_email.exs
@@ -1,0 +1,9 @@
+defmodule Skate.Repo.Migrations.AddUserEmail do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add(:email, :string)
+    end
+  end
+end


priv/repo/migrations/20220927170631_add_user_email.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Skate.Repo.Migrations.AddUserEmail do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add(:email, :string)
```
KaylaBrady, line 6: I _think_ the email column should eventually have a unique constraint added, but it seems possible that during the migration if the format of the `username` field changes, there could be 2 user records with the same email

lemald: That's a good point, and now I realize that the step of actually de-duplicating users is actually missing from my original plan for how this would go. I believe my thinking was that, unless the username format changes on us mid-switchover, each email would correspond to exactly one existing user entry, and any old users that don't get an email associated with them over a ~1 month period could be dropped. That said, having to add the unique constraint later sounds like it's potentially fraught with challenges. Maybe start with the unique constraint, and make it so that the application gracefully handles the case where updating the field throws an error and just does nothing rather than crashing?

KaylaBrady: Makes sense that adding the unique constraint upfront could prevent headaches of trying to make data unique after the fact. One concern of doing nothing when an email address already exists in the DB is that users might unexpectedly their settings on migration. For example: 

1. First login: username: KBrady, email: kbrady@mbta.com. Save 1 preset 
2. Later login: username: newCognitoFormat\\KBrady. email is never set b/c it is already in use. I lose my preset
3. as ` newCognitoFormat\\KBrady`, I save 2 presets & change my colors
4. Migration to use email column is performed. I loose my latest settings again - reverted to only 1 preset & old colors

I realize I may be overly focused on an edge case here if it is unlikely that the username format will be changing again soon. If this is an edge case that is worth covering, perhaps if an existing email is found, we can overwrite the username to ensure there is still only 1 record for that user.

lemald: Yes, I think that's an accurate description of the edge case, and it's really important to map these sorts of cases out so that we know what we're getting ourselves into. On the whole, given that the changes in the username format don't happen _that_ often, and the amount of user configuration possible in Skate is currently low, my opinion is that it's an acceptable tradeoff to not handle this situation.

lemald (approved):

PR #1742, opened 2022-09-29
feat: Save user email on login

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1201189412881312/f

This moves the logic for creating/updating user records into the login flow and adds functionality for saving the user's email.

### DIFF
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -3,7 +3,11 @@ defmodule Skate.Settings.User do
 
   import Ecto.Query
 
-  def get_or_create(username) do
+  def get(username) do
+    Skate.Repo.get_by!(DbUser, username: username)
+  end
+
+  def upsert(username) do
     user =
       Skate.Repo.insert!(
         DbUser.changeset(%DbUser{}, %{username: username}),
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -2,20 +2,57 @@ defmodule Skate.Settings.User do
   alias Skate.Settings.Db.User, as: DbUser
 
   import Ecto.Query
+  require Logger
 
+  @spec get(String.t()) :: DbUser.t()
+  @doc """
+  Get a user by their username
+  """
   def get(username) do
     Skate.Repo.get_by!(DbUser, username: username)
   end
 
-  def upsert(username) do
+  @spec get_by_email(String.t()) :: DbUser.t() | nil
+  @doc """
+  Get a user with the matching email, if one exists
+  """
+  def get_by_email(email) do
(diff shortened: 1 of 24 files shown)


lib/skate/settings/user.ex
```diff
@@ -2,16 +2,61 @@ defmodule Skate.Settings.User do
...
+            conflict_target: [:username],
+            on_conflict: {:replace, [:email]}
+          )
+
+        # existing user has same username - no update needed
+        user_matching_email.username == username ->
+          user_matching_email
+
+        # username format has changed. Create a new record for this user without an associated email
+        # This record will be used only temporarily until user settings are universally fetched by email.
```
KaylaBrady, line 49: Handling of this case based on [this thread](https://github.com/mbta/skate/pull/1740#discussion_r981635623)

lib/skate/settings/user.ex
```diff
@@ -2,16 +2,61 @@ defmodule Skate.Settings.User do
...
+        user_matching_email.username != username ->
+          Skate.Repo.insert!(
+            DbUser.changeset(%DbUser{}, %{
+              username: username
+            }),
+            returning: true,
+            conflict_target: [:username],
+            on_conflict: {:replace, [:email]}
```
lemald, line 57: If we're intending to create a new entry, do we want something other than `:replace` for `on_conflict`? I suppose there's a bit of a race condition here if something in the `user` table changes between the query to get the matching user and this insert so we need to handle that somehow, but I'm not entirely sure of the best way to handle it.

test/skate/settings/user_test.exs
```diff
@@ -3,26 +3,91 @@ defmodule Skate.Settings.UserTest do
...
+      assert %DbUser{username: @username} = User.get(@username)
+    end
+
+    test "raises if user not found" do
+      assert_raise Ecto.NoResultsError, fn -> User.get("missinguser") end
+    end
+  end
 
-      Skate.Repo.insert!(DbUser.changeset(%DbUser{}, %{username: username}))
+  describe "get_by/1" do
```
lemald, line 22: Minor typo (should be `get_by_email/1`).

lib/skate/settings/user.ex
KaylaBrady: Yeah that is true - I'll change to do nothing instead. In the case where a conflict is found, there is really nothing to replace in this scenario, because there is no email to set.

KaylaBrady: Though I see that will not return the full user -- I `on_conflict: {:replace, [:username]}` is the convention I'll follow from `get_or_create` to effectively do a no-op and still return the full user

lemald (approved):
Looks good to me!

PR #1743, opened 2022-09-29
fix(Map): controls not blocked by iphone 'notch' in mobile landscape

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203057881715593/f

Note this adjusts the controls on mobile landscape whether the notch is on the left or righthand side. 

Tested using Xcode simulator on iphone 14 pro max

### DIFF
--- a/assets/css/_leaflet.scss
+++ b/assets/css/_leaflet.scss
@@ -3,6 +3,10 @@
   height: 100%;
 }
 
+.leaflet-right {
+  padding-right: env(safe-area-inset-right);
+}
+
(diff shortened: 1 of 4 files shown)


lemald (approved):
Looks good to me! I tried on my Android phone and it actually handles the notch / camera a bit differently - when in landscape mode the overall browser viewport is just truncated to no encroach on that part of the screen, and it much not do anything with `safe-area-inset-right` because the controls were in the same place. In any event, that's a totally fine outcome, so we should be good to go.

PR #1748, opened 2022-10-03
test: resolve interdependency between search page tests

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1203058157664665/f

This PR refactors `<RightPanel>` so that `Page` component tests are no longer also testing the `RightPanel` component. A separate ticket will further explore possible changes to `useSocket` based on the findings in the discussion of this PR.

<details>
  <summary>Previous WIP notes</summary>
  This PR currently includes some changes that resolved the console errors previously seen in the test file. I don't see this PR as it stands now as a comprehensive approach to resolving this issue, but want to document what I've found so far

Prior to these changes:
* running  "renders a selected vehicle" and "on mobile, allows you to toggle to the map view and back again" one at a time: both succeeded
* running all `SearchPage.test.tsx` at once with  "renders a selected vehicle" *before*  "on mobile, allows you to toggle to the map view and back again": both tests succeed, but console errors logged
* running all `SearchPage.test.tsx` at once with  "renders a selected vehicle" *after*  "on mobile, allows you to toggle to the map view and back again": both tests succeed, no console errors logged. 

Isolating the issue:
* removing the `selectedVehicleOrGhost` from state in "renders a selected vehicle": both tests succeed, no console errors logged
* removing `async/await` from  "on mobile, allows you to toggle to the map view and back again": both tests succeed, no console errors
* Mocking out `<RightPanel>`: both tests succeed, no console errors logged

After performing the above tests, I wondered if something about the call to `react-test-renderer.create()` in "renders a selected vehicle" wasn't being cleaned up properly before running the next test. 
* Switched to use `react-testing-library` renderer instead to avoid errors due to inconsistent renderers within the same file
* Mocked out additional API calls used in the VPP to resolve `Error: Error: connect ECONNREFUSED 127.0.0.1:80` errors
* **most potentially concerning change** : added cleanup function to `useSocket` to resolve `Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.`

While those fixes within the test file resolved the console errors, it raises a few more questions:
* Would moving `<RightPanel>` up a level so it is a sibling to these pages rather than a child help to keep these tests isolated?
* is the `useSocket` hook not having a cleanup function a real problem, or a symptom of something else missing in the test? I suspect it is a real problem, but have doubts since it is an older area of the code. Though this kind of issue might not have user-facing impact
* Is there an advantage I'm overlooking to using `react-test-renderer` for snapshots over `react-testing-library `render` + `asFragment`? 
</details>


assets/src/hooks/useSocket.ts
```diff
@@ -42,6 +42,10 @@ const useSocket = (): SocketStatus => {
       setConnectionStatus(ConnectionStatus.Disconnected)
     )
     setSocket(initialSocket)
+
+    return () => {
+      setConnectionStatus(ConnectionStatus.Disconnected)
+    }
```
lemald, line 48: I think you're correct that there should be some cleanup here, though I'm not entirely sure exactly what. This cleanup does the same thing as the socket's `onClose` callback, although that may still make a difference since there's nothing to call that `onClose` callback on the mock socket. One other possibility that comes to mind for the cleanup is invoking the [disconnect callback](https://hexdocs.pm/phoenix/js/#socketdisconnect). I also wonder about setting the socket state back to `undefined`, though presumably that doesn't matter much if the hook is being cleaned up due to being unmounted. Going further down the rabbit hole, I wonder if the socket object will simply drop the connection when it's cleaned up in memory management, meaning we just need to make sure it's no longer referenced.

lemald:
With regards to moving `<RightPanel>` up a level, at first I was undecided but now I'm in favor. My fear was that there could be pages where we don't want the right panel to appear, but then I remembered that currently you can access notifications from any page, and opening a notification entails being able to open the VPP. If that ever changes we'll need to make other significant changes to Skate, and independently rendering `<RightPanel>` in each separate existing page component is repeated code. So yeah, I'd say go for it, it will make the application code cleaner and it sounds like it will help with testing too.

KaylaBrady:
👍 Thanks @lemald, I'll rework this PR to move `<RightPanel>`. 

I *suspsect* that one possible user-facing added benefit of moving `<RightPanel>` up a level is that it the panel won't re-render when switching between modes. Right now:
1. open the VPP from search and open the block minischedule
2. navigate to the route ladder VPP reset to vehicle status

I expect with this change, the VPP would stay on the block minischedule when switching between modes. Before making this change I can confirm with product in the ticket whether the existing behavior is intended.

lemald:
> 👍 Thanks @lemald, I'll rework this PR to move `<RightPanel>`.
> 
> I _suspsect_ that one possible user-facing added benefit of moving `<RightPanel>` up a level is that it the panel won't re-render when switching between modes. Right now:
> 
>     1. open the VPP from search and open the block minischedule
> 
>     2. navigate to the route ladder. VPP reset to vehicle status
> 
> 
> I expect with this change, the VPP would stay on the block minischedule when switching between modes. Before making this change I can confirm with product in the ticket whether the existing behavior is intended.

Ah, yes, that's also a good point! I suspect you're correct about how it will work with the change, and it should be a nice (if small) UX improvement.

assets/src/hooks/useSocket.ts
KaylaBrady: I think this just unravelled another layer as to why the original test was producing `Warning: Can't perform a React state update on an unmounted component` - `PropertiesPanel` seems to be the only component that calls `useSocket` directly - all others use the socket from `SocketContext`. 

That indicates to me that in most component tests which don't use the `SocketProvider` (only `App.test.tsx` does), the socket would be undefined, so we'd never encounter this unmounting issue. I'll remove the changes to `useSocket` from this PR & add these details to the ticket -- maybe also worth considering if `PropertiesPanel` should use the socket from context as part of that.

KaylaBrady: Removed the socket changes in [cb2ca68](https://github.com/mbta/skate/pull/1748/commits/cb2ca68c8d7229ec74503cc0f292cbd85657b704) and created a [new ticket](https://app.asana.com/0/1148853526253420/1203118229899945/f) for investigating those considerations

lemald (approved):
Looks good to me!

assets/css/_properties_panel.scss
```diff
@@ -9,7 +9,7 @@
   right: 0;
   top: 0;
   width: 23.5rem;
-  z-index: 1002;
+  z-index: $right-panel-z-index;
```
KaylaBrady, line 12: I looked into adding a new `right-panel` class as a single location to set the z-index property, but this did not resolve the z-index issue for the properties panel. 

Opting to create a single z-index constant instead set by each of the possible right-panel classes to resolve the immediate issue with a marginal improvement in consistency.

lemald: Unfortunately, the `z-index` change does result in the header being visible again, but it changes the overall layout of the page versus before - now the right panel takes up 100% of the height of the browser and covers part of the top nav, as opposed to being positioned below the top nav. Interestingly, late view doesn't have this issue, and I'm not immediately sure why.

KaylaBrady: Ah, I missed that subtlety and will re-visit! LateView is a separate component from RightPanel -- I'll dig a bit more into how it is positioned & replicate for the RightPanel components

lemald: Bringing some consistency to our z-indices is still a really good addition, though!

KaylaBrady: Thanks! Okay, I think the latest on dev-green reflects what is expected here: right panel is below the top bar, but in the case of VPP, above the modal overlay shadow. 

Doing this change & running into these visual inconsistencies makes me think about adding some sort of visual regression tests, especially since there are multiple breakpoints to check! I suspect that would be out of scope for this particular PR but can add a backlog ticket if it seems worthwhile. Looks like there [is already one](https://app.asana.com/0/1152340551558956/1202357245709641/f) for adding cypress for a different use case (accessibility testing)

lemald: Yeah, agreed that that would be nice, especially given how different the various breakpoints are.

assets/css/_swings_view.scss
```diff
@@ -8,7 +8,6 @@
   position: absolute;
   right: 0;
   top: 0;
-  z-index: 1001;
```
lemald, line 11: Question: what's the rationale for completely removing the z-index here and in the notifications drawer? It looks like it still renders as I'd expect at least on desktop, but I've encountered a lot of weird bugs with seemingly-minor z-index changes in Skate so I just want to double-check.

KaylaBrady: I removed them while trying to isolate the differences between these two and the vpp (vpp has the modal overlay background, which seems to make the z-index of the vpp matter more), and also noticed it didn't seem to have an impact. I'll undo this change though to limit the change for regressions

lemald: Thanks!

lemald (approved):
Latest changes look good!

assets/css/_swings_view.scss
KaylaBrady: Updated $vpp-z-index to 1002 to match the original value as well.

PR #1752, opened 2022-10-05
Bus data map proof of concept

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203057881715584/f

For demonstration purposes, I added route, stop, and garage data onto the existing search page map. Before search criteria is entered, the map renders with all routes & stops. Once search criteria is entered, only the routes of the matching vehicles are shown.


lib/skate_web/controllers/shape_controller.ex
```diff
@@ -1,21 +1,59 @@
...
+    is_station = stop.is_station || !is_nil(stop.parent_station_id)
+
+    %{
+      id: stop.id,
+      lat: stop.latitude,
+      lon: stop.longitude,
+      name: stop.name,
+      isStation: is_station
+    }
+  end
```
lemald, line 58: If I understand correctly, does this mean that both the child stop and the parent stop will get points on the map of the same size, they will just happen to be at (approximately) the same location? If so that's fine for this proof of concept, I just want to make sure that I understand what sort of things we would need to work on to make it production-ready.

lemald: Also for what it's worth, sometimes child stops for bus bays at stations have the same coordinates as the station (I believe the Harvard Busway is an example), but in some cases (like Sullivan) there are more specific coordinates for the individual bus stops, so they won't be in quite the same location as the parent station.

lemald:
Overall this is looking great as a proof of concept! I'm playing around with it a bit (in Firefox on my laptop) and the performance is a little bit sluggish while showing all routes and all stops, but like you mentioned there's probably some deduplication we can do, as well as various other optimizations.

lib/skate_web/controllers/shape_controller.ex
KaylaBrady: Looking at only bus stops, it seems that no stop alone is considered a station (without the `!is_nil(stop.parent_station_id)` check, `is_station` is false for all records). If we did want to leave the `parent_station_id` check out here so that cases like Sullivan Station aren't all marked as station stops, we could alternatively add the `isStation` flag to the hardcoded subway stop files and do some work to ensure in deduplication that the `isStation` flag is preserved.

KaylaBrady:
Closing since there is no intention to merge this PR.

PR #1759, opened 2022-10-12
Upgrade to React 18

lemald:
Asana ticket: [⚙️ Upgrade to React 18](https://app.asana.com/0/1152340551558956/1203058157664654/f)

Unfortunately these changes are very heavily front-loaded into the first commit. However, that commit description does include a breakdown of what changes are included. The following commits are smaller but should be better broken down.

### DIFF
--- a/.github/workflows/report-coverage.yml
+++ b/.github/workflows/report-coverage.yml
@@ -45,7 +45,7 @@ jobs:
           unzip elixir-lcov.zip
           echo "PR_SHA=$(cat PR_SHA)" >> $GITHUB_ENV
           echo "PR_NUMBER=$(cat PR_NUMBER)" >> $GITHUB_ENV
-      - uses: actions/checkout@v2 # UNTRUSTED CODE - do not run scripts from it
+      - uses: actions/checkout@v3 # UNTRUSTED CODE - do not run scripts from it
         with:
           ref: ${{ env.PR_SHA }}
       - name: Upload coverage artifact and post comment
--- a/assets/package.json
+++ b/assets/package.json
@@ -28,32 +28,33 @@
     "@types/uuid": "^8.3.4",
     "core-js": "^3.21.1",
     "identity-obj-proxy": "^3.0.0",
-    "leaflet": "^1.9.0",
+    "leaflet": "^1.9.2",
     "leaflet-defaulticon-compatibility": "^0.1.1",
     "phoenix": "file:../deps/phoenix",
     "phoenix_html": "file:../deps/phoenix_html",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "react-leaflet": "^3.2.5",
-    "react-leaflet-fullscreen": "^2.0.2",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-leaflet": "^4.1.0",
     "react-router-dom": "^6.3.0",
     "uuid": "^8.3.2",
     "whatwg-fetch": "^3.6.2"
   },
   "devDependencies": {
+    "@babel/core": "^7.19.3",
+    "@babel/preset-env": "^7.19.3",
     "@testing-library/jest-dom": "^5.16.4",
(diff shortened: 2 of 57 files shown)


assets/tests/components/ladderPage.test.tsx
```diff
@@ -415,65 +407,16 @@ describe("LadderPage", () => {
...
-              <LadderPage />
-            </RoutesProvider>
-          </BrowserRouter>
-        </StateDispatchProvider>
-      )
-      .toJSON()
-    expect(tree).toMatchSnapshot()
-  })
-
-  test("renders with vehicles and selected vehicles", () => {
```
lemald, line 432: Note: I've removed this (and I think a couple of other places that test rendering selected vehicles) because #1748 moved the rendering of the VPP outside of the page component.

lemald:
Also note: I think the test "Shuttle Map Page renders with shapes selected" is actually broken (not just with the upgrade, the original snapshot doesn't look like it demonstrates what the name of the test says it's demonstrating). I'll see if I can fix that as part of this PR, or otherwise create a followup ticket.

lemald:
More generally, I'm going to give this another pass (probably tomorrow) and see if any of the snapshot tests can easily be reworked into more traditional React Testing Library tests so that we don't have to review those huge diffs anymore. I don't want to spend too much time trying to remove snapshot tests as part of this particular PR though just because it's already got a lot going on.

assets/src/components/map.tsx
```diff
@@ -251,49 +246,42 @@ export const autoCenter = (
...
-          role="button"
-          aria-label="Recenter Map"
+const RecenterControlButton = ({
+  position,
+  recenter,
+}: {
+  position: ControlPosition
+  recenter: () => void
+}): ReactElement => (
+  <Control position={position}>
```
KaylaBrady, line 256: When the VPP is open on the search page, the recenter button attaches to the search map instead of the VPP. I think the map ref needs to be passed in some way to the control with the `container` prop

assets/package.json
```diff
@@ -71,6 +73,7 @@
     "prettier": "^2.7.1",
     "react-test-renderer": "^18.2.0",
     "sass-loader": "^13.0.2",
+    "screenfull": "^6.0.2",
```
KaylaBrady, line 78: suggestion (non-blocking): Maybe worth adding an in-line comment as to why this is a dev dependency?

KaylaBrady (commented):
Looks good! Only blocking comment is about the recenter control

assets/src/components/map.tsx
lemald: After discussion, creating [⚙️ Improve map recenter control](https://app.asana.com/0/1152340551558956/1203142430154314/f) as a followup to improve this code and dropping the commit that did this particular refactor.

assets/package.json
lemald: I tried, but unfortunately it's a JSON file so it doesn't like comments in the syntax.

.github/workflows/report-coverage.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           unzip elixir-lcov.zip
           echo "PR_SHA=$(cat PR_SHA)" >> $GITHUB_ENV
           echo "PR_NUMBER=$(cat PR_NUMBER)" >> $GITHUB_ENV
-      - uses: actions/checkout@v2 # UNTRUSTED CODE - do not run scripts from it
+      - uses: actions/checkout@v3 # UNTRUSTED CODE - do not run scripts from it
```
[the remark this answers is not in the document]
lemald: This comes from the base `report-coverage.uml` in the mbta/.github repo: https://github.com/mbta/.github/blob/main/workflow-templates/report-coverage.yml#L48. I'm not sure exactly why that comment is there but not in other workflows, but I'm wary of getting rid of it without fully understanding its purpose.

KaylaBrady (approved):
Looks good!

PR #1763, opened 2022-10-18
feat: Bus stops added to the VPP

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203119002718952/f

This PR mostly pulls from https://github.com/mbta/skate/pull/1752 with additional testing.

I couldn't find a suitable way to test for the tooltip's appearance on stop hover in the map component. It seems that the tooltip is only added to the DOM on hover, and after trying a few ways to pass accessible selectors (`title`, `alt`), I didn't find a way to hover over the stop in testing to trigger the tooltip. This is one of the rare cases where I wish RTL had a workaround to find an element by class name! Open to suggestions for other ways to unit test this behavior.

### DIFF
--- a/test/schedule/data_test.exs
+++ b/test/schedule/data_test.exs
@@ -532,13 +532,13 @@ defmodule Schedule.DataTest do
     test "returns nil when no trip found", %{trip: trip, shape: shape} do
       data = %Data{trips: %{}, shapes: %{trip.route_id => [shape]}}
 
-      assert nil == Data.shape_with_stops_for_trip(data, trip.id)
+      assert is_nil(Data.shape_with_stops_for_trip(data, trip.id))
     end
 
     test "returns nil when no shape found", %{trip: trip} do
       data = %Data{trips: %{trip.id => trip}, shapes: %{}}
 
-      assert nil == Data.shape_with_stops_for_trip(data, trip.id)
+      assert is_nil(Data.shape_with_stops_for_trip(data, trip.id))
     end
 
     test "returns the shape without stops when no stops found", %{trip: trip, shape: shape} do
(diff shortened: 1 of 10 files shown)


test/schedule/data_test.exs
```diff
@@ -495,6 +496,89 @@ defmodule Schedule.DataTest do
...
+        ]
+      }
+
+      {:ok, %{trip: trip, stops: stops, shape: shape}}
+    end
+
+    test "returns nil when no trip found", %{trip: trip, shape: shape} do
+      data = %Data{trips: %{}, shapes: %{trip.route_id => [shape]}}
+
+      assert nil == Data.shape_with_stops_for_trip(data, trip.id)
```
lemald, line 535: Minor stylistic point, but you can also use `is_nil/1` here and other places to assert that the result is nil.

lemald (approved):
One non-blocking stylistic comment but overall looks good to me!

test/schedule/data_test.exs
```diff
@@ -495,6 +496,89 @@ defmodule Schedule.DataTest do
...
+    end
+
+    test "returns the shape without stops when no stops found", %{trip: trip, shape: shape} do
+      data = %Data{trips: %{trip.id => trip}, stops: %{}, shapes: %{trip.route_id => [shape]}}
+
+      assert %ShapeWithStops{id: shape.id, points: shape.points, stops: []} ==
+               Data.shape_with_stops_for_trip(data, trip.id)
+    end
+
+    test "returns  all stops when all stops are found", %{
```
firestack, line 551: nit: is there a reason there are two spaces here?

otherwise, 
```suggestion
    test "returns all stops when all stops are found", %{
```

firestack (approved):
I've got the _smallest_ of nitpicks but otherwise LGTM!

PR #1765, opened 2022-10-20
refactor: Notifications Hook Cleanup

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1203074341639621/f

This PR builds on @lemald's work to refactor the notifications hook & reducer to address the tight coupling & exhaustive-deps issues.


assets/src/hooks/useChannel.ts
```diff
@@ -16,13 +16,15 @@ export const useChannel = <T>({
   parser,
   loadingState,
   closeAfterFirstRead,
+  dependencies,
```
KaylaBrady, line 19: I think this may be questionable -- further comment about the possibility of removing this change below

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -183,9 +195,82 @@ const persistToggledNotificationReadState = (
...
+      ).initial_notifications.map(notificationFromData),
+    }
+  }
+  return {
+    type: "new",
+    payload: notificationFromData(notificationData as NotificationData),
+  }
+}
+
+const useNotifications = (socket: Socket | undefined, routeIds: RouteId[]) => {
```
KaylaBrady, line 222: I'm thinking about breaking this out into a separate file to simplify `useNotificationsReducer` & independently testing it

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -183,9 +195,82 @@ const persistToggledNotificationReadState = (
...
+  return useChannel<ReceivedNotifications>({
+    socket,
+    topic: "notifications",
+    event: "notification",
+    parser: parseNotifications,
+    dependencies: routeIds,
```
KaylaBrady, line 228: Okay this is the piece I'm not sure about. Originally `useNotifications` had a dependency on routeIds so that the channel is re-joined every time the open list of routeIds changes. I'm not sure if that is necessary though - my impression from[ this line](https://github.com/mbta/skate/blob/e17a6a19ad7ee2a4ce9294c3322d58411696fe04/lib/notifications/notification_server.ex#L184) in NotificationServer is that if they didn't re-join the channel, they would still get notifications for the new route_ids.

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -58,17 +72,17 @@ export const markAllAsRead = (): MarkAllAsReadAction => ({
 interface SetNotificationsAction {
   type: "SET_NOTIFICATIONS"
   payload: {
-    notificationsData: NotificationData[]
+    notifications: Notification[]
```
lemald, line 75: **praise:** thank you for making it consistent whether the actions for this reducer take `Notification`s or `NotificationData`.

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -183,9 +195,82 @@ const persistToggledNotificationReadState = (
...
+  const newRouteIds = allOpenRouteIds(routeTabs)
+
+  if (!equalByElements(routeIds, newRouteIds)) {
+    setRouteIds(newRouteIds)
+  }
+
+  const latestMessage: ReceivedNotifications = useNotifications(
+    socket,
+    routeIds
+  )
```
lemald, line 253: **praise:** I really like this approach for coupling the different reducers together, I was sort of grasping at something like this idea but couldn't quite put it together in a way that worked.

assets/src/hooks/useNotificationsReducer.ts
lemald: That would make sense (`useNotifications` lives to see another day! My original changes deleted it entirely).

assets/src/hooks/useNotificationsReducer.ts
lemald: Also if we don't need to rejoin we can undo the addition of the `dependencies` arg to `useChannel`.

KaylaBrady: > I was held up on the fact that you wouldn't get old notifications for a route when you add it, but we wouldn't have delivered those notifications to that user to begin with.

Interesting, I hadn't thought about rejoining to get the batch of all unexpired notifications for that new route as `initial_notifications`. To confirm, is the intended behavior that adding a new route means you'll only get new notifications for that route?

assets/tests/hooks/useNotificationsReducer.test.tsx
```diff
@@ -170,4 +192,228 @@ describe("useNotificationsReducer", () => {
...
+    )
+
+    expect(mockChannel.join).toHaveBeenCalled()
+
+    unmount()
+
+    expect(mockChannel.leave).toHaveBeenCalled()
+  })
+
+  test("console.error on join error", async () => {
```
lemald, line 278: **question:** this and a few other places seem to effectively duplicate tests from `useChannel`. Is there a reason to be testing these properties here as well?

assets/src/hooks/useNotificationsReducer.ts
lemald: > To confirm, is the intended behavior that adding a new route means you'll only get new notifications for that route?

Correct, if you didn't have the route open at the time that the notification was generated you shouldn't get it.

lemald: > To confirm, is the intended behavior that adding a new route means you'll only get new notifications for that route?

Correct, if you didn't have the route open at the time that the notification was generated you shouldn't get it.

lemald: > To confirm, is the intended behavior that adding a new route means you'll only get new notifications for that route?

Correct, if you didn't have the route open at the time that the notification was generated you shouldn't get it.

lib/skate_web/channels/notifications_channel.ex
```diff
@@ -36,6 +36,6 @@ defmodule SkateWeb.NotificationsChannel do
 
     initial_notifications = notification_fetch.(username)
 
-    {:ok, %{initial_notifications: initial_notifications}, socket}
+    {:ok, %{data: %{initial_notifications: initial_notifications}}, socket}
```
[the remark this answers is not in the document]
KaylaBrady: Great call. I'll add a temporary change to the `useChannel` hook to handle the old way of not getting `data:`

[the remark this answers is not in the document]
lemald: That should handle the case where a client with the new version of the front-end code is talking to a server with the old version of the back-end code. If a client with the old version of the front-end code is talking to a server with the new version of the back-end code I think that still might be a potential issue?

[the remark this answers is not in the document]
KaylaBrady: Absolutely true! New idea: 
1. first deploy a change to backend that has the payload `%{data: %{initial_notifications: initial_notifications}, initial_notifications: initial_notifications}`
2. deploy this PR & change the backend payload to `%{data: %{initial_notifications: initial_notifications}}`
Does that seem reasonable?

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -150,10 +158,14 @@ const showLatestNotificationReducer = (
...
-        .isInitialLoad
+      const { isInitialLoad, notifications } = (
+        action as SetNotificationsAction
+      ).payload
+
       if (isInitialLoad) {
-        return true
+        return (
+          true && notifications.length > 0 && notifications[0].state !== "read"
+        )
```
KaylaBrady, line 168: @lemald this is a change in logic, but I suspect may be needed since the initial load piece wasn't previously loading. In testing locally, I found it strange that a read notification was popping up on load. I was expecting that only a new notification would popup.

lib/skate_web/channels/notifications_channel.ex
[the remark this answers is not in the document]
lemald: Sounds good to me.

assets/src/hooks/useNotificationsReducer.ts
lemald: Makes sense, might be worth running by product / design real quick but I'm like 90% sure that's the right call.

assets/tests/hooks/useNotificationsReducer.test.tsx
KaylaBrady: Good point, removed in [c2bd7f2](https://github.com/mbta/skate/pull/1765/commits/c2bd7f2ebc9b4036021c2f05a14b1f2aa72ec8cf)

lib/skate_web/channels/notifications_channel.ex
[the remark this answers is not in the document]
lemald: Hmm actually regarding step 2, I think we only want to remove the old `initial_notifications` key _after_ the front-end changes in this PR go out to all users, so it would have to be another PR / release after this one. I'm not sure if that changes anything else about the process, but I'll think about it. It may be possible to do in two deploys rather than three.

lemald (approved):
Looks good to me, we'll just have to be careful about the order of merging and deploying this vs. the other PRs involved in the transition of the socket data format.

PR #1773, opened 2022-10-25
fix(Map): Fullscreen controls work on mobile safari

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1189935858798779/f

Note this PR focuses only on solving for the mobile safari case, not the desktop resizing case.


assets/css/_leaflet.scss
```diff
@@ -16,10 +16,17 @@
...
-  // You might think that the FullscreenControl library would apply
-  // different CSS classes to the icon depending on if it were in the
-  // expanded or collapsed state. But it doesn't, hence this hack.
-  &[title="Exit Full Screen"] {
+  &.leaflet-fullscreen-on {
     background-image: url("data:image/svg+xml,%3Csvg id='Layer_1' data-name='Layer 1' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 48 48'%3E%3Cpolygon points='13.07 13.07 0 13.07 0 16.8 13.07 16.8 16.8 16.8 16.8 13.07 16.8 0 13.07 0 13.07 13.07'/%3E%3Cpolygon points='0 31.2 0 34.93 13.07 34.93 13.07 48 16.8 48 16.8 34.93 16.8 31.2 13.07 31.2 0 31.2'/%3E%3Cpolygon points='34.93 13.07 34.93 0 31.2 0 31.2 13.07 31.2 16.8 34.93 16.8 48 16.8 48 13.07 34.93 13.07'/%3E%3Cpolygon points='31.2 31.2 31.2 34.93 31.2 48 34.93 48 34.93 34.93 48 34.93 48 31.2 34.93 31.2 31.2 31.2'/%3E%3C/svg%3E");
   }
 }
+// From https://github.com/brunob/leaflet.fullscreen/blob/master/Control.FullScreen.css - workaround for safari
+// mobile not natively supporting fullscreen API
```
KaylaBrady, line 24: I struggled with importing the styling directly in `app.scss`, `app.tsx`, and `map.tsx` (see failed deploy actions of this branch). I would prefer to import the style sheet directly and am open to alternative suggestions for how to do that! One silver lining of this approach is that only the mobile safari fullscreen button was broken, and this adds only the styling relevant to that case.

assets/css/properties_panel/_vehicle_properties_panel.scss
```diff
@@ -1,3 +1,4 @@
+$vpp-header-z-index: 10;
```
lemald, line 1: I'm slightly confused on the naming of this variable - it sounds like there VPP header should have this particular z-index, but as far as I can tell it doesn't, so I'm having a little bit of trouble wrapping my head around what it's supposed to mean. Apologies if I'm missing something obvious.

assets/css/_leaflet.scss
lemald: It's not perfect but I agree it's the best solution given all of the issues with the other approaches, plus the fact that it's actually not very much CSS.

assets/css/properties_panel/_vehicle_properties_panel.scss
KaylaBrady: That is absolutely fair! Playing around with some different options to make that more obvious.

KaylaBrady: @lemald in thinking about z-index variable organization, I tried a different approach for organizing z-index variables by context inspired by [this post ](https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/). It is a variation on some of the recommendations - instead of using naming conventions to indicate the context, I found it personally helpful to represent each context as a map so it is a bit easier to identify all the defined z-indexes within a single context. Since we are using pre-existing z-indexes it felt clearer to me to just use the existing values rather than try to make each value relative to eachother.

This change is a bit tangential to the PR - I could alternatively add a $properties-panel-header-z-index variable `app.scss` to follow the existing pattern.

assets/css/app.scss
```diff
@@ -8,10 +8,17 @@ $mobile-route-picker-width: 18.125rem;
...
+  "modal-overlay": 1000,
+  "properties-panel": 1002,
+  "navbar": 1100,
+);
 
+$z-properties-panel-context: (
+  "map": 9,
+  "header": 10,
+  "fullscreen-map": 11,
+);
```
firestack, line 21: This is really cool!

assets/css/_nav.scss
```diff
@@ -1,3 +1,4 @@
+@use "sass:map";
```
firestack, line 1: Just curious: Why is `sass:map` used here but not in any other document?

assets/css/properties_panel/_vehicle_properties_panel.scss
lemald: Okay, I think this looks good! I share @firestack's question though and I want to give this a second look tomorrow just to make sure that everything adds up with the z-indexes given how tricky this all is.

lemald (approved):
Okay yeah, looks good to me! Just curious about the `@use "sass:map";` thing.

assets/css/_nav.scss
KaylaBrady: Ah yes, good catch! It is an unnecessary leftover of trying to use the map incorrectly. Originally I was using `map.get` instead of `map-get` (frustratingly based on the [sass documentation](https://sass-lang.com/documentation/values/maps#look-up-a-value)) which didn't work, so this was an attempt at a fix that was ultimately unnecessary.

assets/css/properties_panel/_vehicle_properties_panel.scss
KaylaBrady: Thanks! Addressed that comment [here](https://github.com/mbta/skate/pull/1773#discussion_r1008055508) and removed that line

firestack (approved):
🎉

PR #1780, opened 2022-11-01
fix(components/map): ensure autocentering stays on when backgrounded

firestack:
When the `Autocenterer` calls `autoCenter` in rapid succession before the animation of the map pan/zoom is finished, the autocentering control gets disabled by out of order events recieved by `useMapEvents`.

This PR uses `dragstart` to detect _user_ input, which disables the autocentering mode. In addition, the movestart handler has been changed to `zoomstart` because there are no other _user specific_ events fired when the zoom level also changes, and `zoomstart` is fired by `[leaflet::]setView` when the zoom level needs to change.

---

Asana Ticket: https://app.asana.com/0/1148853526253437/1199678649889552/f

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -323,7 +323,7 @@ const EventAdder = ({
     dragstart() {
       setShouldAutoCenter(false)
     },
-    moveend: () => {
+    moveend() {
       // Wait until the auto centering is finished to start listening for manual moves again.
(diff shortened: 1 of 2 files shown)


lemald (approved):
Nice, I'm glad that you and @KaylaBrady managed to get to the bottom of this one, this had me stumped for a while.

PR #1781, opened 2022-11-01
Quick Superstruct fixes

lemald:
Asana ticket: followup on [⚙️ Adopt Superstruct](https://app.asana.com/0/1152340551558956/1201917242620476/f)

Fixes an issue discovered on dev after merging #1776, namely that routes on swings can actually be null in some cases. It looks like there's an underlying data integrity issue which I'll create a followup ticket to fix (and the fix should include removing the `nullable` on these fields as well), but in the meantime we'll be more forgiving with what we accept.

Also removes the default dev environment HASTUS URL since that was having issues and I needed to change it; going forward it should be stored in `.envrc` or elsewhere instead.


assets/src/api.ts
```diff
@@ -187,7 +187,6 @@ export const fetchSwings = (routeIds: RouteId[]): Promise<Swing[] | null> =>
     url: `/api/swings?route_ids=${routeIds.join(",")}`,
     dataStruct: array(SwingData),
     parser: nullableParser(swingsFromData),
-    defaultResult: null,
```
KaylaBrady, line 190: question (non-blocking): To make sure I understand the reasoning for this removal, is it because `defaultResult` affected how the parsing error presented?

KaylaBrady (approved):

assets/src/api.ts
lemald: Yeah, if `defaultResult` is provided and there's an error it just provides the default and doesn't throw an exception. Here the default result isn't actually useful so it doesn't improve things versus the case where an exception is thrown, and this way we'll see console output and / or a Sentry alert if there's an issue.

firestack (approved):

PR #1782, opened 2022-11-02
chore: migration to add test_groups and test_groups_users tables

lemald:
Asana ticket: [⚙️ Database work for user test groups](https://app.asana.com/0/1200180014510248/1202519045249746/f)

The one thing worth noting about this change is that normally for a join table like this I wouldn't include a primary key, but the way that Ecto implements `through` associations (see upcoming PR) seems to require a primary key on the intermediate table. In any event I don't think it hurts anything even if it later turns out we figure out a way to avoid using it.

### DIFF
--- a/priv/repo/migrations/20221031170029_add_test_group.exs
+++ b/priv/repo/migrations/20221031170029_add_test_group.exs
@@ -1,0 +1,38 @@
+defmodule Skate.Repo.Migrations.AddTestGroup do
+  use Ecto.Migration
+
+  def change do
+    create table(:test_groups) do
+      add(:name, :string, null: false)
+      timestamps()
+    end
+
+    create table(:test_groups_users) do
+      add(
+        :test_group_id,
+        references(:test_groups,
+          on_delete: :delete_all,
+          on_update: :update_all
+        ),
+        primary_key: true
+      )
+
+      add(
+        :user_id,
+        references(:users, on_delete: :delete_all, on_update: :update_all),
+        primary_key: true
+      )
+
+      timestamps()
+    end
+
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20221031170029_add_test_group.exs
```diff
@@ -0,0 +1,38 @@
...
+      add(:name, :string, null: false)
+      timestamps()
+    end
+
+    create table(:test_groups_users) do
+      add(
+        :test_group_id,
+        references(:test_groups,
+          on_delete: :delete_all,
+          on_update: :update_all
```
KaylaBrady, line 15: question (non-blocking): Is the `on_update` needed? What would be updated on this record if a test_group is updated?

KaylaBrady (approved):

priv/repo/migrations/20221031170029_add_test_group.exs
lemald: This made me curious so I dug up the [relevant section in the Postgres docs](https://www.postgresql.org/docs/13/ddl-constraints.html):

> Analogous to ON DELETE there is also ON UPDATE which is invoked when a referenced column is changed (updated). The possible actions are the same. In this case, CASCADE means that the updated values of the referenced column(s) should be copied into the referencing row(s).

If I'm reading this correctly, this basically means that this would trigger if `test_group.id` was updated and therefore any corresponding `test_groups_users.test_group_id` valyes needed updating to match the new value. I don't think we're going to go around changing the IDs of test groups but it seems like a harmless case to cover.

PR #1783, opened 2022-11-02
Test group DB utils

lemald:
Asana tickets: [⚙️ Database work for user test groups](https://app.asana.com/0/1200180014510248/1202519045249746/f)

This adds functions to allow for the easy creating and updating of test groups. I paid particular attention to make it possible to update the users associated with a test group in a way that doesn't require the caller to deal with the underlying schema much, so the controller modules for the admin pages will be able to implement functionality to add / remove people without much difficulty. That said, followup ticket may add functionality or slightly modify these utility functions as appropriate.

Currently based on top of the branch with the migration but I'll rebase on master once that's merged.

### DIFF
--- a/lib/skate/settings/db/test_group.ex
+++ b/lib/skate/settings/db/test_group.ex
@@ -1,5 +1,6 @@
 defmodule Skate.Settings.Db.TestGroup do
   use Ecto.Schema
+  import Ecto.Changeset
 
   alias Skate.Settings.Db.TestGroupUser, as: DbTestGroupUser
 
@@ -9,7 +10,14 @@ defmodule Skate.Settings.Db.TestGroup do
     field(:name, :string)
     timestamps()
 
-    has_many(:test_group_users, DbTestGroupUser)
+    has_many(:test_group_users, DbTestGroupUser, on_replace: :delete_if_exists)
     has_many(:users, through: [:test_group_users, :user])
+  end
+
+  def changeset(test_group, attrs \\ %{}) do
+    test_group
+    |> cast(attrs, [:name])
+    |> cast_assoc(:test_group_users)
+    |> validate_required([:name])
   end
 end
--- a/lib/skate/settings/db/test_group.ex
+++ b/lib/skate/settings/db/test_group.ex
@@ -1,0 +1,15 @@
+defmodule Skate.Settings.Db.TestGroup do
+  use Ecto.Schema
+
+  alias Skate.Settings.Db.TestGroupUser, as: DbTestGroupUser
+
+  @type t() :: %__MODULE__{}
+
+  schema "test_groups" do
+    field(:name, :string)
+    timestamps()
+
+    has_many(:test_group_users, DbTestGroupUser)
+    has_many(:users, through: [:test_group_users, :user])
+  end
+end
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -1,0 +1,39 @@
+defmodule Skate.Settings.TestGroup do
+  alias Skate.Settings.Db.TestGroup, as: DbTestGroup
+  alias Skate.Settings.Db.User, as: DbUser
+
+  @type t() :: %__MODULE__{
+          id: integer(),
+          name: String.t(),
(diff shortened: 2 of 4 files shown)


lib/skate/settings/test_group.ex
```diff
@@ -0,0 +1,72 @@
...
+    DbTestGroup
+    |> Skate.Repo.all()
+    |> Skate.Repo.preload(:users)
+    |> Enum.map(&convert_from_db_test_group(&1))
+  end
+
+  @spec update(t()) :: t()
+  def update(test_group) do
+    existing_test_group =
+      DbTestGroup |> Skate.Repo.get(test_group.id) |> Skate.Repo.preload(:test_group_users)
```
KaylaBrady, line 34: question: could `existing_test_group` be nil?

lib/skate/settings/db/test_group.ex
```diff
@@ -0,0 +1,23 @@
...
+  alias Skate.Settings.Db.TestGroupUser, as: DbTestGroupUser
+
+  @type t() :: %__MODULE__{}
+
+  schema "test_groups" do
+    field(:name, :string)
+    timestamps()
+
+    has_many(:test_group_users, DbTestGroupUser, on_replace: :delete_if_exists)
+    has_many(:users, through: [:test_group_users, :user])
```
KaylaBrady, line 14: Very cool - I've never seen the `through` association before!

lib/skate/settings/test_group.ex
```diff
@@ -0,0 +1,72 @@
...
+    existing_test_group
+    |> DbTestGroup.changeset(
+      test_group
+      |> Map.from_struct()
+      |> Map.put(
+        :test_group_users,
+        Enum.map(test_group.users, fn user ->
+          existing_test_group_user =
+            existing_test_group.test_group_users
+            |> Enum.find(&(&1.user_id == user.id))
```
KaylaBrady, line 45: suggestion (non-blocking): Maybe using `MapSet.new/2` to create a set of existing user ids could simplify determining if  a user is part of the existing group with a `member?` check

lib/skate/settings/test_group.ex
lemald: Theoretically it could be, but given that this is the `update` function I think it's safe to assume that we're being passed an existing test group that already exists in the database, and I'm generally okay with throwing an exception if that isn't true.

KaylaBrady: That's fair! A non-blocking suggestion - in the unlikely case that the group didn't exist, using  `Skate.Repo.get!` here might return a clearer exception message by detecting the unexpected `nil` earlier in the function. Likely not needed though based on your point that it is safe to assume the group will exist, so it is unlikely we'll see an exception from this anyway!

lemald: That's a good idea, actually - I'll make the switch.

lib/skate/settings/db/test_group.ex
lemald: I haven't used them much either. They have some limitations (Ecto is good at reading them but for writing them you have to implement much more of the logic yourself, as I did here), but definitely a good tool to have!

lib/skate/settings/test_group.ex
lemald: Hmm I'm not sure if there's a good way to use `MapSet` to make this better, since we don't just need the existing user IDs but also the ID of the `test_group_user` corresponding to them. This is important so that we don't just delete and recreate those entries every time we add or remove a different user. Let me know if I'm missing something about your idea though!

KaylaBrady: Ah yeah I see what you mean! Maybe a regular map then? Still a non-blocking suggestion though w/ some shorthand below!
```
# Above the changeset creation
existing_user_id_lookup = Map.new(existing_test_group, & %{&1.user_id, &1.id}

# as part of changeset creation
Enum.map(test_group_users, fn user -> 
    id = Map.get(existing_user_id_lookup, user.id, nil)
    %{id: id, user_id: user.id, test_group_id: existing_test_group.id}
end)
```

PR #1784, opened 2022-11-02
feat(Reports): New report for all user configuration

KaylaBrady:

A new report to understand how users will be affected by the above ticket. 

There is an existing `UserSettings` report that includes only settings with no username attached. I opted to create a new report for this specific use case, but can combine into the existing `UserSettings` report if that seems more appropriate.

### DIFF
--- a/lib/report.ex
+++ b/lib/report.ex
@@ -4,7 +4,7 @@ defmodule Report do
   """
   @type t :: module()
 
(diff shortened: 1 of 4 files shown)


lemald (approved):
👍

firestack (approved):
Too new to elixir for this review to feel meaningful but LGTM 👍🏻

PR #1786, opened 2022-11-03
fix(Report.UserConfigurations): routes per tab as comma separated list

KaylaBrady:
A fix for https://github.com/mbta/skate/pull/1784 - 
using `inspect` resulted in strings being turned for lists where the numbers in that list were all codepoints. For example, someone with a tab that has 7 routes and another with 8 would return '\a\b', instead of the expected "[7, 8]". 

Now, the number of routes is converted to a string & concatenated.


lemald (commented):
👍

PR #1788, opened 2022-11-04
feat(skate/test-groups): send test group information to frontend

firestack:
This implements access to a user's test groups via storing the information in the database schema, rendering it to the frontend, and providing a utility function to access the data.

---

Closes: https://app.asana.com/0/1200180014510248/1202519045249749/f

### DIFF
--- a/assets/src/userTestGroups.ts
+++ b/assets/src/userTestGroups.ts
@@ -1,8 +1,17 @@
+import { array, create, Infer, string } from "superstruct"
 import appData from "./appData"
 
+const TestGroups = array(string())
+type TestGroups = Infer<typeof TestGroups>
+
 const getTestGroups = (): string[] => {
-  const groups = appData()?.userTestGroups
-  const testGroups: string[] = groups === undefined ? [] : JSON.parse(groups)
+  const testGroupsJson = appData()?.userTestGroups
+
+  if (testGroupsJson === undefined) {
+    return []
+  }
+
+  const testGroups = create(JSON.parse(testGroupsJson) as unknown, TestGroups)
 
   return testGroups
 }
--- a/assets/src/userTestGroups.ts
+++ b/assets/src/userTestGroups.ts
@@ -16,7 +16,7 @@ const getTestGroups = (): string[] => {
   return testGroups
 }
 
-const inTestGroup = (key: string) => {
+const inTestGroup = (key: string): boolean => {
   return getTestGroups().includes(key)
 }
 
--- a/assets/src/userTestGroups.ts
+++ b/assets/src/userTestGroups.ts
@@ -1,12 +1,14 @@
 import appData from "./appData"
 
-function getTestGroups(): string[] {
+const getTestGroups = (): string[] => {
   const groups = appData()?.userTestGroups
   const testGroups: string[] = groups === undefined ? [] : JSON.parse(groups)
 
   return testGroups
 }
 
-export default function inTestGroup(key: string) {
+const inTestGroup = (key: string) => {
   return getTestGroups().includes(key)
 }
+
+export default inTestGroup
--- a/assets/src/userTestGroups.ts
+++ b/assets/src/userTestGroups.ts
@@ -1,0 +1,12 @@
+import appData from "./appData"
+
+function getTestGroups(): string[] {
+  const groups = appData()?.userTestGroups
+  const testGroups: string[] = groups === undefined ? [] : JSON.parse(groups)
+
+  return testGroups
+}
+
+export default function inTestGroup(key: string) {
+  return getTestGroups().includes(key)
+}
--- a/assets/tests/userTestGroups.test.ts
+++ b/assets/tests/userTestGroups.test.ts
@@ -1,24 +1,64 @@
+import { StructError } from "superstruct"
 import appData from "../src/appData"
 import inTestGroup from "../src/userTestGroups"
 
 jest.mock("appData")
 
-describe("User test groups helper", () => {
-  // test("Assert")
-  test("Assert detect user in group", () => {
-    const group_1 = "user-test-group-1"
-    const group_2 = "user-test-group-2"
-    const mockSettings = {
-      userTestGroups: JSON.stringify([group_1, group_2]),
-    }
+describe("User Test Groups API", () => {
+  test("raise error when given wrong json data type", () => {
     ;(appData as jest.Mock)
-      .mockImplementationOnce(() => {})
-      .mockImplementation(() => mockSettings)
+      .mockImplementationOnce(() => ({
+        // Wrong data type
+        userTestGroups: null,
+      }))
+      .mockImplementationOnce(() => ({
+        // Wrong data type
+        userTestGroups: "1.0",
+      }))
+      .mockImplementation(() => ({
+        // Wrong data type
+        userTestGroups: "{}",
+      }))
 
     // Start with no groups or keys
-    expect(inTestGroup(group_1)).toBe(false)
-    // Once groups are "loaded", check if in groups
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+  })
+
+  test("raise error when given invalid json", () => {
+    ;(appData as jest.Mock)
+      .mockImplementationOnce(() => ({
+        userTestGroups: "asdf1234!@#$",
+      }))
+      .mockImplementationOnce(() => ({
+        userTestGroups: [1, 2, 3],
+      }))
+      .mockImplementation(() => ({
+        userTestGroups: "",
+      }))
+
+    expect(() => inTestGroup("non-existent-group")).toThrow(SyntaxError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(SyntaxError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(SyntaxError)
+  })
+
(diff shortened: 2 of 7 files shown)


test/skate_web/controllers/page_controller_test.exs
```diff
@@ -52,6 +52,29 @@ defmodule SkateWeb.PageControllerTest do
...
+      user_struct = user |> Skate.Settings.User.get() |> Skate.Repo.preload(:test_groups)
+
+      conn = get(conn, "/")
+
+      html = html_response(conn, 200)
+
+      json =
+        Jason.encode!(user_struct.test_groups |> Enum.map(& &1.name))
+        |> Phoenix.HTML.html_escape()
+        |> Phoenix.HTML.safe_to_string()
```
firestack, line 73: Does this read okay? would it read more "idomatically" if instead it was like this?
```suggestion
      json =
      	user_struct.test_groups
      	|> Enum.map(& &1.name)
        |> Jason.encode!()
        |> Phoenix.HTML.html_escape()
        |> Phoenix.HTML.safe_to_string()
```

I like it being within the `encode` parentheses because it makes it clear what's happening to the `test_groups` first but maybe there's another way to format it and reduce cognitive overhead via indentation/shape?

test/skate_web/controllers/page_controller_test.exs
```diff
@@ -52,6 +52,29 @@ defmodule SkateWeb.PageControllerTest do
...
+      })
+
+      user_struct = user |> Skate.Settings.User.get() |> Skate.Repo.preload(:test_groups)
+
+      conn = get(conn, "/")
+
+      html = html_response(conn, 200)
+
+      json =
+        Jason.encode!(user_struct.test_groups |> Enum.map(& &1.name))
```
firestack, line 71: I'm also kinda concerned the test needs to handle the case of having multiple groups or groups returned in different orders?

If so maybe it needs to be a different data type?

test/skate_web/controllers/page_controller_test.exs
```diff
@@ -52,6 +52,29 @@ defmodule SkateWeb.PageControllerTest do
...
     end
 
+    @tag :authenticated
+    test "includes user test groups in HTML", %{conn: conn, user: user} do
+      user_struct = user |> Skate.Settings.User.get()
+
+      Skate.Settings.TestGroup.update(%{
+        Skate.Settings.TestGroup.create("html-test-group")
+        | users: [user_struct]
+      })
```
firestack, line 62: The test succeeds with multple test groups defined, but it's somewhat verbose. I'm wondering if anyone else thinks we should be testing with multiple groups in this test or create another one that does that?

test/skate_web/controllers/page_controller_test.exs
KaylaBrady: I personally tend to find it easier to read pipe chains when they start with a single value, like your suggested change does. The first thing that is happening to the test groups is that they are being mapped to only their name, so for me personally it is easier to see that as a distinct step. 

Much of my elixir preferences were informed by working on a project with "controversial" credo checks enabled, for example, the credo check for [pipe_chain_start](https://github.com/rrrene/credo/blob/master/lib/credo/check/refactor/pipe_chain_start.ex) which applies in this scenario. So I acknowledge that this may be a "controversial" preference, and that particular credo check is not enabled in this repo, so I wouldn't block a PR over it. Over all though, I found credo checks to be a helpful guide for picking up conventions as I was starting to learn elixir, and have adopted or rejected some of the "controversial" ones over time.

assets/src/userTestGroups.ts
```diff
@@ -0,0 +1,12 @@
+import appData from "./appData"
+
+function getTestGroups(): string[] {
```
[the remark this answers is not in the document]
lemald: For what it's worth we usually declare TypeScript functions with `const` rather than `function` in most of the codebase.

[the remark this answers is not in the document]
lemald: As for exporting I'd say it's fine to leave private for now and we can export it later if need be.

assets/src/userTestGroups.ts
```diff
@@ -0,0 +1,12 @@
...
+  const groups = appData()?.userTestGroups
+  const testGroups: string[] = groups === undefined ? [] : JSON.parse(groups)
```
lemald, line 5: Since `JSON.parse` returns an `any` type TypeScript doesn't have a problem with this, but it would still be nice to do some runtime checks to make sure the result of parsing is actually an array of strings like we'd expect.

test/skate_web/controllers/page_controller_test.exs
lemald: I also prefer starting the pipe with a single value.

assets/src/userTestGroups.ts
firestack: Typescript can't infer based on the `string[]` annotation?

Could/should this be done with superstruct in some form?

assets/src/userTestGroups.ts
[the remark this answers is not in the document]
firestack: > For what it's worth we usually declare TypeScript functions with `const` rather than `function` in most of the codebase.

I'm just curious, can you elaborate on why?

I understand that with `const` we get errors on reassignment, but we loose function hoisting, which doesn't matter in this case (and due to modules), but it seems strange to me that const functions would be preferred over the language keyword, but I feel like I'm missing something.

test/skate_web/controllers/page_controller_test.exs
lemald: I'm not overly concerned but testing with 0, 1, and N would certainly be nice. I don't think ExUnit really has built-in test parametrization tools, but your test can always wrap some assertions in a `for` block.

assets/src/userTestGroups.ts
firestack: I'm curious what you think of the revision!

lemald: I hadn't really be familiar with `create` as opposed to `assert` and `is` previously but I think that looks good!

lemald:
@firestack the one thing that sticks out to me that still needs doing is tests for `userTestGroups.ts`, but other than that I think this is close to ready for approval!

firestack:
Writing the typescript test and then I'll have y'all hopefuly able to final review 👍🏻

firestack:
I would still like to make [the elixir test more robust](https://github.com/mbta/skate/pull/1788#discussion_r1014168942) but this PR is ready for review and could be squash-merged as is.

assets/tests/userTestGroups.test.ts
```diff
@@ -0,0 +1,24 @@
+import appData from "../src/appData"
+import inTestGroup from "../src/userTestGroups"
+
+jest.mock("appData")
+
+describe("User test groups helper", () => {
+  // test("Assert")
```
lemald, line 7: **nitpick:** looks like this was left in unintentionally.

assets/tests/userTestGroups.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    ;(appData as jest.Mock)
+      .mockImplementationOnce(() => {})
+      .mockImplementation(() => mockSettings)
+
+    // Start with no groups or keys
+    expect(inTestGroup(group_1)).toBe(false)
+    // Once groups are "loaded", check if in groups
+    expect(inTestGroup(group_1)).toBe(true)
+    expect(inTestGroup(group_2)).toBe(true)
+  })
```
lemald, line 23: **suggestion:** it would also be nice to have a test for the case where the JSON data is malformed (as in it parses as valid JSON but isn't an array of strings).

lib/skate/settings/db/user.ex
```diff
@@ -19,6 +20,10 @@ defmodule Skate.Settings.Db.User do
     has_many(:notification_users, DbNotificationUser)
     many_to_many(:notifications, DbNotification, join_through: DbNotificationUser)
     has_many(:route_tabs, DbRouteTab, on_replace: :delete_if_exists)
+
+    # :test_groups goes through :test_group_users to find the :test_group's the user belongs to
```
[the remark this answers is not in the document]
lemald: I don't personally feel like this needs a comment, given that you can see it's used in the `has_many` declaration for `:test_groups`, though I could be convinced otherwise if it still feels highly unintuitive to others.

lemald:
@firestack okay left a couple of minor comments that I'd like to see addressed before merging but it's getting there. 👍

lib/skate/settings/db/user.ex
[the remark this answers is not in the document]
firestack: I found myself very confused when first reading over this, but this feels like a special case to elixir beginners.

It's personally confusing how something inside a function call can become a referencable member, but we don't document this in all the other similar places, so I don't believe it needs to stay.

assets/tests/userTestGroups.test.ts
```diff
@@ -1,24 +1,64 @@
...
+        userTestGroups: "{}",
+      }))
+
+    // Start with no groups or keys
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+    expect(() => inTestGroup("non-existent-group")).toThrow(StructError)
+  })
+
+  test("raise error when given invalid json", () => {
```
lemald, line 29: Sorry, what I meant to get at with my earlier comment wasn't only cases where the value is something that doesn't parse as JSON, but also when it's something like `"{"key": "value"}"` that is valid JSON, but not an array of strings when parsed.

KaylaBrady (approved):

lemald (approved):

PR #1789, opened 2022-11-04
feat: Reduce usage of username field for retrieving user data

KaylaBrady:

This ticket replaces places where user data is queried by username with the user's primary key: id. 

There are a few core places that username is still used which this PR does not address: 
* `notifications_channel`: This uses `AuthManager.username_for_socket!/1` to get the username, and which notifications to show the user are queried by username. It feels like there may be some consequences to changing this identifier used to subscribe to notifications that I don't fully understand yet. If there are no consequences, changing the notifications code still feels a bit riskier and can be addressed in a separate PR. 

* `AuthManager.reesource_for_claims` still returns `username`. The first commit in this PR modified that to return both username & user_id, but after reading the [Guardian docs](https://hexdocs.pm/guardian/introduction-implementation.html#basic-setup) I realized that would break all the existing user tokens and reverted from this PR and further developed in a separate draft PR [here](https://github.com/mbta/skate/pull/1790).
* FE user identification for 3rd party integrations

My inclination is to break those three out into separate tickets from this one since we know the `username` field will be sticking around a a while longer (at least until [this ticket](https://app.asana.com/0/1152340551558956/1203278600620523/f) is addressed)

In any case, I'm thinking there will be a few additional PRs needed to complete this current ticket (which could all be done before the above three bullets are addressed): 
1. `User.upsert` should take only email address, not email + username
2. Remove existing records from the DB where email is null
3. Add a not-null constraint to the email field

Please let me know if you have any concerns about that plan - any other PRs I'm missing?

### DIFF
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -7,6 +7,7 @@ defmodule Notifications.Notification do
   alias Schedule.Trip
   alias Schedule.Hastus.Run
   alias Skate.Settings.User
+  alias Skate.Settings.Db.User, as: DbUser
   alias Notifications.NotificationReason
   alias Notifications.NotificationState
   alias Notifications.Db.Notification, as: DbNotification
@@ -197,12 +198,14 @@ defmodule Notifications.Notification do
     |> Enum.map(&convert_from_db_notification/1)
   end
 
-  def update_read_states(username, notification_ids, read_state)
+  @spec update_read_states(DbUser.id(), [id()], NotificationState.t()) ::
+          {non_neg_integer(), nil | [t()]}
+  def update_read_states(user_id, notification_ids, read_state)
       when read_state in [:read, :unread, :deleted] do
     query =
       from(nu in DbNotificationUser,
         join: u in assoc(nu, :user),
-        where: u.username == ^username,
+        where: u.id == ^user_id,
         where: nu.notification_id in ^notification_ids
       )
 
--- a/lib/skate/settings/db/user.ex
+++ b/lib/skate/settings/db/user.ex
@@ -8,6 +8,7 @@ defmodule Skate.Settings.Db.User do
   alias Skate.Settings.Db.RouteTab, as: DbRouteTab
 
   @type t :: %__MODULE__{}
+  @type id :: integer()
 
   schema "users" do
     field(:username, :string)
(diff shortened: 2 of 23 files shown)


lemald:
Whew, there's a lot to digest here! I probably won't give it all justice on my first pass.

> `AuthManager.reesource_for_claims` still returns `username`. The first commit in this PR modified that to return both username & user_id, but after reading the [Guardian docs](https://hexdocs.pm/guardian/introduction-implementation.html#basic-setup) I realized that would break all the existing user tokens and reverted from this PR and further developed in a separate draft PR https://github.com/mbta/skate/pull/1790.

How badly does it break the existing tokens? If it just sends users back to the login page or something that should be fine. If it puts things into a state that's harder to recover from without doing something like clearing cookies though I agree it should happen in a separate PR.

KaylaBrady:
> How badly does it break the existing tokens? If it just sends users back to the login page or something that should be fine. If it puts things into a state that's harder to recover from without doing something like clearing cookies though I agree it should happen in a separate PR.

@lemald at least from my local testing it seemed to be the latter (clearing cookies needed), unless building in backwards compatibility for the old token format. Maintaining that backwards compatibility seemed sufficiently complicated to me where it feels worth keeping split in a separate PR for separate testing & deploying.

lemald:
> > How badly does it break the existing tokens? If it just sends users back to the login page or something that should be fine. If it puts things into a state that's harder to recover from without doing something like clearing cookies though I agree it should happen in a separate PR.
> 
> @lemald at least from my local testing it seemed to be the latter (clearing cookies needed), unless building in backwards compatibility for the old token format. Maintaining that backwards compatibility seemed sufficiently complicated to me where it feels worth keeping split in a separate PR for separate testing & deploying.

Okay - it might be worth testing on a dev environment because the login flow is definitely one of those things that's quite different when running locally, but that definitely doesn't bode well. If there's any doubt let's maintain backwards compatibility.

firestack (approved):

firestack:
Looks good, not much to comment on!

Just want to note that #1788 introduced merge conflicts for this.

KaylaBrady:
Closing for now in light of conversation to resolve the token issues first. Will likely re-open a revised version of this PR once that is resolved

PR #1790, opened 2022-11-07
feat(AuthManager): Accept user_id as token subject

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203331355829267/f

Followed the basic setup instructions from https://hexdocs.pm/guardian/introduction-implementation.html#basic-setup.

This allows us to move away from the AD username as the subject of the token to Skate's internal user id PK. The old username format is supported for backwards compatibility and can be removed once all the outstanding old tokens are expired (expected after 1 hour of deploying, will confirm with logs)

### DIFF
--- a/lib/skate_web/auth_manager.ex
+++ b/lib/skate_web/auth_manager.ex
@@ -6,8 +6,16 @@ defmodule SkateWeb.AuthManager do
   @skate_admin_group "skate-admin"
   @skate_dispatcher_group "skate-dispatcher"
 
+  def subject_for_token(%{username: username, user_id: user_id}, _claims) do
+    {:ok, %{"username" => username, "user_id" => user_id}}
+  end
+
   def subject_for_token(resource, _claims) do
     {:ok, resource}
+  end
+
+  def resource_from_claims(%{"sub" => %{"username" => username, "user_id" => user_id}}) do
+    {:ok, %{username: username, user_id: user_id}}
   end
 
   def resource_from_claims(%{"sub" => username}) do
@@ -17,12 +25,20 @@ defmodule SkateWeb.AuthManager do
   def resource_from_claims(_), do: {:error, :invalid_claims}
 
   def username_from_socket!(socket) do
-    {:ok, username} =
+    {:ok, resource} =
       socket
       |> Guardian.Phoenix.Socket.current_token()
       |> decode_and_verify!()
       |> resource_from_claims()
 
+    username_from_resource(resource)
+  end
+
+  defp username_from_resource(%{username: username}) do
+    username
+  end
+
+  defp username_from_resource(username) when is_binary(username) do
     username
   end
 
--- a/lib/skate_web/auth_manager.ex
+++ b/lib/skate_web/auth_manager.ex
@@ -1,21 +1,30 @@
 defmodule SkateWeb.AuthManager do
   use Guardian, otp_app: :skate
+  alias Skate.Settings.Db.User, as: DbUser
+  alias Skate.Settings.User
+  require Logger
 
   @type access_level :: :none | :general | :admin
 
   @skate_admin_group "skate-admin"
   @skate_dispatcher_group "skate-dispatcher"
+  @v2_resource_prefix "v2:"
 
-  def subject_for_token(%{username: username, user_id: user_id}, _claims) do
-    {:ok, %{"username" => username, "user_id" => user_id}}
+  def v2_resource_prefix, do: @v2_resource_prefix
+
+  def subject_for_token(%DbUser{id: user_id}, _claims) do
+    {:ok, "#{@v2_resource_prefix}#{user_id}"}
   end
 
   def subject_for_token(resource, _claims) do
     {:ok, resource}
   end
 
-  def resource_from_claims(%{"sub" => %{"username" => username, "user_id" => user_id}}) do
-    {:ok, %{username: username, user_id: user_id}}
+  def resource_from_claims(%{"sub" => @v2_resource_prefix <> user_id}) do
+    case User.get_by_id(String.to_integer(user_id)) do
+      nil -> {:error, :user_not_found}
+      user -> {:ok, user}
+    end
   end
 
   def resource_from_claims(%{"sub" => username}) do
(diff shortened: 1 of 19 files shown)


lib/skate_web/controllers/notification_read_states_controller.ex
```diff
@@ -5,7 +5,10 @@ defmodule SkateWeb.NotificationReadStatesController do
   alias SkateWeb.AuthManager
 
   def update(conn, %{"new_state" => new_read_state, "notification_ids" => notification_ids}) do
-    username = AuthManager.Plug.current_resource(conn)
+    username =
+      conn
+      |> AuthManager.Plug.current_resource()
+      |> AuthManager.username_from_resource()
```
KaylaBrady, line 11: this duplicated chain is temporary: once all tokens in circulation use the new format, we can access the username using `%{username: username} = AuthManager.Plug.current_resource(conn)`

lib/skate_web/controllers/auth_controller.ex
```diff
@@ -14,16 +14,15 @@ defmodule SkateWeb.AuthController do
...
 
     conn
     |> Guardian.Plug.sign_in(
       AuthManager,
-      username,
+      "#{AuthManager.v2_resource_prefix()}#{user_id}",
       %{groups: credentials.other[:groups]},
       ttl: {expiration - current_time, :seconds}
     )
-    |> Plug.Conn.put_session(:username, username)
```
KaylaBrady, line 26: this seems to have never been used (No references to `get_session(:username)`, and no tests failing on removal

test/support/channel_case.ex
```diff
@@ -26,7 +26,14 @@ defmodule SkateWeb.ChannelCase do
     end
   end
 
-  setup _tags do
```
KaylaBrady, line 29: I ran into issues adding tests to `AuthManager`  since the database wasn't sandboxed - modifying here so all tests using `ChannelCase` use the sandbox by default.

lib/skate_web/auth_manager.ex
```diff
@@ -1,28 +1,56 @@
...
+    {:ok, "#{@v2_resource_prefix}#{user_id}"}
+  end
 
   def subject_for_token(resource, _claims) do
+    Logger.info("old pattern matched: subject_for_token")
     {:ok, resource}
   end
 
+  def resource_from_claims(%{"sub" => @v2_resource_prefix <> user_id}) do
+    case User.get_by_id(String.to_integer(user_id)) do
```
KaylaBrady, line 25: I'm having second thoughts about introducing a DB query here. It seems to be a conventional approach based on the guardian docs linked in the PR description, but based on initial logging, it seems that this query will be performed on every endpoint hit. 

I'm looking into a way to avoid that query (resource should be cached on the token - why do we need to re-call this?). If not, we could avoid the query by passing username as part of "sub", but that seems a bit like a smell to me.

KaylaBrady: After some research into 2 (doesn't seem viable) and discussion with Eddie, going to pursue option 1. 

Removing the query will help minimize the impact of this change to keep it focused on only changing token format rather than changing token & introducing new frequent DB request.

lib/skate_web/channels/vehicles_channel.ex
```diff
@@ -31,8 +31,17 @@ defmodule SkateWeb.VehiclesChannel do
   def join(
         "vehicles:search:" <> search_params,
         _message,
-        %{assigns: %{guardian_default_resource: username}} = socket
+        socket
       ) do
+    username_from_socket! =
```
KaylaBrady, line 36: following pattern from `notifications_channel`

lib/skate_web/controllers/auth_controller.ex
lemald: Sounds good - I can't think of anywhere it was used. Just watch out for any errors on dev I guess.

KaylaBrady:
Seeing `old_pattern_matched` logs in Splunk for a user logged in before this deployed, will continue to monitor to make sure it successfully cuts over to the new format on expiration

lib/skate_web/controllers/route_tabs_controller.ex
```diff
@@ -5,7 +5,10 @@ defmodule SkateWeb.RouteTabsController do
   alias Skate.Settings.RouteTab
 
   def update(conn, %{"route_tabs" => route_tabs} = _params) do
-    username = AuthManager.Plug.current_resource(conn)
+    username =
+      conn
+      |> AuthManager.Plug.current_resource()
+      |> AuthManager.username_from_resource()
```
lemald, line 11: **suggestion:** In this and a few other places where we do get username, it's then passed into functions that use it to get / update some relevant data from the database - route tabs in this instance. However, for most of these we should just be able to rework the underlying function to take a user ID instead and use that in the relevant queries. Making that change might be too much added complexity for this ticket but it would be good to create a followup tech cleanup ticket for it.

lemald:
@KaylaBrady my one high-level concern here is what would happen in the theoretical edge case where someone with a username like `v2:1234` existed. I suppose since that would only be a problem with old tokens we could check before deploying just to 100% cover our bases?

lib/skate_web/controllers/route_tabs_controller.ex
KaylaBrady: 💯 - I was worried about putting that all into this PR, but have a [separate ticket ](https://app.asana.com/0/1200180014510248/1203331355829273/f)on deck to address it!

lemald: Ah, perfect! Apologies if we had already talked about that and I just forgot.

KaylaBrady:
> @KaylaBrady my one high-level concern here is what would happen in the theoretical edge case where someone with a username like `v2:1234` existed. I suppose since that would only be a problem with old tokens we could check before deploying just to 100% cover our bases?

@lemald good point! Will confirm by running a user name report before deploying this. Right now the AD username format is at least ~mostly predictable and doesn't conflict with the `v2:` prefix.

KaylaBrady:
Update: Confirmed that the old dev-blue tokens have expired and successfully switched over to the new format (no more `old_pattern_matched` logs)

PR #1791, opened 2022-11-07
Test group admin page

lemald:
Asana ticket: [⚙️ Admin pages for test groups](https://app.asana.com/0/1200180014510248/1202459690955957/f)

I ended up deviating slightly from the exact acceptance criteria in terms of paths for things and how various requests worked, but it all follows the general spirit. I also realized that removing users from a test group was important so made sure to include that. I'm open in particular to feedback on how the various form flows work (including which things are included in the URL vs. in the body of POST requests).

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -34,7 +34,7 @@ config :skate, SkateWeb.AuthManager, secret_key: "dev key"
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher"]]}
+    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher", "skate-admin"]]}
   ]
 
 config :logger, level: :notice
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -22,12 +22,10 @@ defmodule Skate.Settings.TestGroup do
 
   @spec get(integer()) :: t() | nil
   def get(id) do
-    test_group = Skate.Repo.get(DbTestGroup, id)
+    test_group = DbTestGroup |> Skate.Repo.get(id) |> Skate.Repo.preload(:users)
 
     if test_group do
-      test_group
-      |> Skate.Repo.preload(:users)
-      |> convert_from_db_test_group
+      convert_from_db_test_group(test_group)
     else
       nil
     end
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -20,6 +20,19 @@ defmodule Skate.Settings.TestGroup do
     |> convert_from_db_test_group()
   end
 
+  @spec get(integer()) :: t() | nil
+  def get(id) do
+    test_group = Skate.Repo.get(DbTestGroup, id)
+
+    if test_group do
+      test_group
+      |> Skate.Repo.preload(:users)
+      |> convert_from_db_test_group
+    else
+      nil
+    end
+  end
+
   @spec get_all() :: [t()]
   def get_all() do
     DbTestGroup
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -20,6 +20,19 @@ defmodule Skate.Settings.User do
     Skate.Repo.get_by(DbUser, email: String.downcase(email))
   end
 
+  @spec get_by_id(integer()) :: DbUser.t() | nil
+  @doc """
+  Get a user by ID, if one exists
+  """
+  def get_by_id(id) do
+    Skate.Repo.get(DbUser, id)
+  end
+
+  @spec get_all() :: [DbUser.t()]
+  def get_all() do
+    Skate.Repo.all(DbUser)
+  end
+
   @spec upsert(username :: String.t(), email :: String.t()) :: DbUser.t()
   @doc """
   Update the user with the given username if one exists, otherwise insert a new one.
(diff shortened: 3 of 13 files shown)


lib/skate/settings/test_group.ex
```diff
@@ -20,6 +20,19 @@ defmodule Skate.Settings.TestGroup do
     |> convert_from_db_test_group()
   end
 
+  @spec get(integer()) :: t() | nil
+  def get(id) do
+    test_group = Skate.Repo.get(DbTestGroup, id)
+
+    if test_group do
+      test_group
+      |> Skate.Repo.preload(:users)
```
KaylaBrady, line 29: nitpick (non-blocking): I believe `preload` works with `nil`, so this could be called immediately after `Skate.Repo.get` and won't break if there is no test group found. I think it is nice to keep the data querying function calls chained together, but that is a personal preference.

config/dev.exs
```diff
@@ -34,7 +34,7 @@ config :skate, SkateWeb.AuthManager, secret_key: "dev key"
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher"]]}
+    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher", "skate-admin"]]}
```
KaylaBrady, line 37: 🙌 nice! Appreciate this change to make local testing easier

lib/skate/settings/user.ex
```diff
@@ -20,6 +20,19 @@ defmodule Skate.Settings.User do
     Skate.Repo.get_by(DbUser, email: String.downcase(email))
   end
 
+  @spec get_by_id(integer()) :: DbUser.t() | nil
```
KaylaBrady, line 23: 👍 assuming this will land before https://github.com/mbta/skate/pull/1789, I'll modify it to reduce duplication by using this new `get_by_id`

lib/skate_web/controllers/test_group_controller.ex
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def show(conn, params) do
+    test_group = TestGroup.get(params["id"])
+
+    if test_group do
+      test_group_users =
+        test_group.users
+        |> Enum.map(&{&1.id, String.downcase(&1.email)})
```
KaylaBrady, line 36: nitpick(non-blocking): Emails are down-cased on insert, so this step can be skipped here

lib/skate_web/controllers/test_group_controller.ex
```diff
@@ -0,0 +1,118 @@
...
+  def add_user(conn, params) do
+    test_group = TestGroup.get(params["id"])
+    user = User.get_by_id(params["user_id"])
+
+    cond do
+      is_nil(test_group) ->
+        send_resp(conn, :not_found, "no test group found")
+
+      is_nil(user) ->
+        send_resp(conn, :bad_request, "no user found")
```
KaylaBrady, line 85: question (non-blocking): Why `:not_found` for one case and `:bad_request` for the other?

test/skate_web/controllers/test_group_controller_test.exs
```diff
@@ -0,0 +1,221 @@
...
+      conn =
+        post(conn, SkateWeb.Router.Helpers.test_group_path(conn, :add_user, test_group.id), %{
+          "user_id" => 123
+        })
+
+      assert response(conn, 400) =~ "no user found"
+    end
+  end
+
+  describe "remove_server/2" do
```
KaylaBrady, line 168: nitpick - `remove_user`

lib/skate_web/templates/layout/test_groups.html.eex
```diff
@@ -0,0 +1,22 @@
...
+    %>
+    <head>
+        <meta charset="utf-8"/>
+
+        <title>Skate Test Groups</title>
+
+        <link rel="apple-touch-icon" href="<%= static_href.(@conn, "/images/mbta-logo-t-180.png") %>" type="image/png">
+        <link rel="icon" href="<%= static_href.(@conn, "/images/mbta-logo-t-favicon.png") %>" sizes="32x32" type="image/png">
+        <link rel="icon" href="<%= static_href.(@conn, "/favicon.ico") %>" sizes="16x16" type="image/vnd.microsoft.icon">
+    </head>
```
KaylaBrady, line 15: question (non-blocking): I don't see any logos when I run locally, is there something I'm missing about what this is doing?

KaylaBrady (approved):
Looks good! A series of non-blocking questions, but the functionality seems solid to me! Good call on adding removing users from a group, should removing a group be possible as well?

lib/skate_web/templates/layout/test_groups.html.eex
lemald: Hmm, I'm seeing the logo when I run it locally - the main spot you should see it is just the favicon in the browser tab.

lib/skate/settings/user.ex
lemald: Whatever works for you! I was just looking at #1789 and thinking about how I'd rebase this PR on top of it if I needed to anyway.

lib/skate/settings/test_group.ex
lemald: Yep, it appears to work with `nil`! And yeah agreed it's nicer to keep them all in one pipeline.

lib/skate_web/templates/layout/test_groups.html.eex
KaylaBrady: Ah okay yep I see it!

lib/skate_web/controllers/test_group_controller.ex
lemald: This is one of the parts where I was sort of unsure about the best way to handle edge cases. Conceptually, one felt like "I can't find the resource you want me to update," and the other felt like "I found the primary resource (test group) you wanted me to update but something else about the request was bad."

PR #1796, opened 2022-11-09
chore: update hash of AWS certificate bundle for RDS

lemald:
No ticket, just a quick fix to make the deploy work again.


firestack (approved):
Oh hey, a TOFU hash like nix!

PR #1798, opened 2022-11-10
feat: Query by user id instead of username in all queries

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203331355829273/f

This must be deployed after https://github.com/mbta/skate/pull/1797.

### DIFF
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -87,12 +87,9 @@ defmodule Skate.Settings.User do
     end
   end
 
+  @spec user_ids_for_route_ids(any) :: any
   def user_ids_for_route_ids(route_ids) do
     Skate.Repo.all(from(u in users_for_route_ids_query(route_ids), select: u.id))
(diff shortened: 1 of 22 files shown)


lib/skate/settings/user.ex
```diff
@@ -87,14 +79,11 @@ defmodule Skate.Settings.User do
     end
   end
 
+  @spec user_ids_for_route_ids(any) :: any
```
lemald, line 82: **question:** is it possible to give this a more specific typing?

lemald (approved):
One quick question about a type that I think could be improved, but otherwise looks good!

lib/skate/settings/user.ex
KaylaBrady: Good point! Added in [b16ab90](https://github.com/mbta/skate/pull/1798/commits/b16ab90166e2370dfcda1538668577de7c337bbf)

PR #1802, opened 2022-11-14
feat: setup for separate maps mode

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203329634351702
There are some minor comments in the thread for that ticket, but I don't think they should block an initial review.

My general approach for this ticket was to create a new `MapPage` component as a copy of the existing `SearchPage` so that it is easy to make changes to the component without affecting the old `SearchPage` that all users should currently still see. The `MapPage.test.tsx` file is a copy of `SearchPage.test.tsx` as well, with only the relevant click interaction tests changed.

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -212,7 +212,7 @@ const LeafletShape = ({ shape }: { shape: Shape }) => {
   ])
 
   return (
-    <>
+    <div data-testid="routeShape">
       <Polyline
         className="m-vehicle-map__route-shape"
         positions={positions}
@@ -228,7 +228,7 @@ const LeafletShape = ({ shape }: { shape: Shape }) => {
           <Popup className="m-vehicle-map__stop-tooltip">{stop.name}</Popup>
         </CircleMarker>
       ))}
-    </>
+    </div>
   )
 }
 
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -33,7 +33,6 @@ import vehicleLabelString from "../helpers/vehicleLabel"
 import { drawnStatus, statusClasses } from "../models/vehicleStatus"
 import { TrainVehicle, Vehicle, VehicleId } from "../realtime.d"
 import { Shape } from "../schedule"
-import { selectVehicle } from "../state"
 import { UserSettings } from "../userSettings"
 import { equalByElements } from "../helpers/array"
 import appData from "../appData"
@@ -48,6 +47,7 @@ export interface Props {
   // trainVehicles are white, don't get a label, and don't affect autocentering
   trainVehicles?: TrainVehicle[]
   reactLeafletRef?: MutableRefObject<LeafletMap | null>
+  onPrimaryVehicleSelect?: (vehicle: Vehicle) => void
 }
 
 interface RecenterControlProps extends ControlOptions {
@@ -125,15 +125,14 @@ const makeLeafletLabelIcon = (
 const LeafletVehicle = ({
   vehicle,
   isPrimary,
+  onSelect,
 }: {
   vehicle: Vehicle
   isPrimary: boolean
+  onSelect?: (vehicle: Vehicle) => void
 }) => {
-  const [appState, dispatch] = useContext(StateDispatchContext)
-  const select = isPrimary
-    ? () => dispatch(selectVehicle(vehicle))
-    : // eslint-disable-next-line @typescript-eslint/no-empty-function
-      () => {}
+  const [appState] = useContext(StateDispatchContext)
+  const eventHandlers = onSelect ? { click: () => onSelect(vehicle) } : {}
   const position: LatLngExpression = [vehicle.latitude, vehicle.longitude]
   const vehicleIcon: Leaflet.DivIcon = makeLeafletVehicleIcon(
     vehicle,
@@ -152,17 +151,13 @@ const LeafletVehicle = ({
       <Marker
         position={position}
         icon={vehicleIcon}
-        eventHandlers={{
-          click: select,
-        }}
+        eventHandlers={eventHandlers}
         zIndexOffset={zIndexOffset}
       />
       <Marker
         position={position}
         icon={labelIcon}
-        eventHandlers={{
-          click: select,
-        }}
+        eventHandlers={eventHandlers}
         zIndexOffset={zIndexOffset}
       />
     </>
@@ -430,7 +425,12 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
           attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
         />
         {props.vehicles.map((vehicle: Vehicle) => (
-          <LeafletVehicle key={vehicle.id} vehicle={vehicle} isPrimary={true} />
+          <LeafletVehicle
+            key={vehicle.id}
+            vehicle={vehicle}
+            isPrimary={true}
+            onSelect={props.onPrimaryVehicleSelect}
+          />
         ))}
         {(props.secondaryVehicles || []).map((vehicle: Vehicle) => (
           <LeafletVehicle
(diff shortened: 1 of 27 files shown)


assets/src/components/mapPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const { socket }: { socket: Socket | undefined } = useContext(SocketContext)
+  const vehicles: VehicleOrGhost[] | null = useSearchResults(
+    socket,
+    searchPageState.isActive ? searchPageState.query : null
+  )
+  const onlyVehicles: Vehicle[] = filterVehicles(vehicles)
+  const [mobileDisplay, setMobileDisplay] = useState(MobileDisplay.List)
+  const [selectedVehicle, setSelectedVehicle] = useState<Vehicle | null>(null)
+  const selectedVehicleShapes = useTripShape(selectedVehicle?.tripId || null)
```
KaylaBrady, line 64: This is the key logic change vs `searchPage`: maintaining new state for `selectedVehicle` & fetching the trip shape rather than displaying the VPP.

assets/src/userTestGroups.ts
```diff
@@ -4,6 +4,8 @@ import appData from "./appData"
 const TestGroups = array(string())
 type TestGroups = Infer<typeof TestGroups>
 
+export const MAP_BETA_GROUP_NAME = "map-beta"
```
firestack, line 7: comment: I'm really glad you 'symbolized' this! 
I figured it might just be bikeshedding to ask what we should name the test group.

assets/src/util/mapMode.tsx
```diff
@@ -0,0 +1,16 @@
...
+import SearchPage from "../components/searchPage"
+import inTestGroup, { MAP_BETA_GROUP_NAME } from "../userTestGroups"
+
+export interface NavMode {
+  path: string
+  title: string
+  element: ReactElement
+}
+
+export const mapModeForUser = (): any =>
```
firestack, line 13: question: should this return `any` if there's the `NavMode` interface above?

KaylaBrady: 💯, thanks!

assets/src/userTestGroups.ts
KaylaBrady: Yeah it is a fair question! I'm not sure what an appropriately descriptive name would be, but have a question for product in the ticket & will update if there is a change requested!

assets/src/components/map.tsx
```diff
@@ -217,7 +212,7 @@ const LeafletShape = ({ shape }: { shape: Shape }) => {
   ])
 
   return (
-    <>
+    <div data-testid="routeShape">
```
lemald, line 215: **question:** does wrapping this in a `<div>` change anything about the DOM such that styling or other functionality is affected? The [fragment](https://reactjs.org/docs/fragments.html) that was here previously looks kind of like a DOM node but doesn't actually create one.

assets/src/util/streetViewUrl.ts
```diff
@@ -0,0 +1,12 @@
...
+  latitude,
+  longitude,
+  bearing,
+}: {
+  latitude: number
+  longitude: number
+  bearing?: number
+}): string =>
+  `https://www.google.com/maps/@?api=1&map_action=pano&viewpoint=${latitude}%2C${longitude}&heading=${
+    bearing || 0
```
lemald, line 11: **suggestion (non-blocking):** it might make sense to leave this blank if no `bearing` is specified rather than defaulting to 0, since the [documentation](https://developers.google.com/maps/documentation/urls/get-started#street-view-action) indicates that it's optional.

assets/src/components/map.tsx
KaylaBrady: Good question, it didn't seem to based on my initial poking around, but I'd want to have more testing in place to have enough confidence in that. I'll revert to using classname testing

assets/tests/util/streetViewUrl.test.ts
```diff
@@ -0,0 +1,26 @@
...
+        latitude: 42.36014757,
+        longitude: -71.09505978,
+        bearing: 131,
+      })
+    ).toEqual(
+      "https://www.google.com/maps/@?api=1&map_action=pano&viewpoint=42.36014757%2C-71.09505978&heading=131&pitch=0&fov=80"
+    )
+  })
+
+  test("returns with default heading of 0 when no bearing given", () => {
```
lemald, line 16: **nitpick (non-blocking):** test name should be updated to reflect change to functionality.

lemald (approved):
👍

firestack (approved):
LGTM!

KaylaBrady:
@mbta/skate-developers - requesting re-review since there have been some functionality changes since approval so interacting with the search results now shows/hides the route shape.

assets/src/components/searchForm.tsx
```diff
@@ -25,6 +31,9 @@ const SearchForm = () => {
   const clearTextInput = (event: React.FormEvent<EventTarget>): void => {
     event.preventDefault()
     dispatch(setSearchText(""))
+    if (onClear) {
+      onClear()
+    }
```
lemald, line 36: **suggestion:** even if this and the `onSubmit` callback functionality are hit by test cases in other components, it would be good to add test cases to `searchForm.test.tsx` directly.

assets/src/components/mapPage.tsx
```diff
@@ -0,0 +1,126 @@
...
+  vehicles: VehicleOrGhost[] | null,
+  searchPageState: SearchPageState
+): boolean => vehicles !== null && searchPageState.isActive
+
+const filterVehicles = (
+  vehiclesOrGhosts: VehicleOrGhost[] | null
+): Vehicle[] => {
+  return vehiclesOrGhosts === null
+    ? []
+    : (vehiclesOrGhosts.filter((vog) => isVehicle(vog)) as Vehicle[])
```
lemald, line 30: **suggestion(non-blocking):** this is another place where it would be nice to get rid of the `as` if possible.

assets/src/hooks/useShapes.ts
```diff
@@ -55,6 +55,8 @@ export const useTripShape = (tripId: TripId | null): Shape[] => {
       fetchShapeForTrip(tripId).then((shapeResult: Shape | null) =>
         setShape(shapeResult)
       )
+    } else {
+      setShape(null)
```
lemald, line 59: **suggestion:** I think this could also use a test case

lemald (approved):
A couple of non-blocking comments but overall looks good!

assets/tests/components/searchForm.test.tsx
```diff
@@ -137,6 +160,30 @@ describe("SearchForm", () => {
...
+      ...initialState,
+      searchPageState: validSearch,
+    }
+    const result = render(
+      <StateDispatchProvider state={validSearchState} dispatch={testDispatch}>
+        <SearchForm onClear={onClear} />
+      </StateDispatchProvider>
+    )
+
+    await userEvent.click(result.getByTitle("Clear"))
```
firestack, line 181: note(non-blocking): based on this [paragraph about using `screen`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen); Maybe we should replace usage of `const result` with `screen`?

I didn't check this, but maybe it'd look like this?
```suggestion
    render(
      <StateDispatchProvider state={validSearchState} dispatch={testDispatch}>
        <SearchForm onClear={onClear} />
      </StateDispatchProvider>
    )

    await userEvent.click(screen.getByTitle("Clear"))
```

firestack: I'm sure this could apply to like, almost all our tests, and I'm not sure how much it's actually worth doing.

KaylaBrady: Nice yeah I've been stuck in the habit of using `result` but can convert to using screen on new/revisited test cases! Looks like the only other change needed is to import `screen` at the top.

lemald: Yeah it will apply to a lot of our tests (and I'm similarly used to just using `result`) but it sounds like that would be more in line with best practices.

PR #1805, opened 2022-11-16
feat(User): only use email for associating user on login

KaylaBrady:

### DIFF
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -35,15 +35,16 @@ defmodule Skate.Settings.User do
 
   @spec upsert(username :: String.t(), email :: String.t()) :: DbUser.t()
   @doc """
-  Update the user with the given username if one exists, otherwise insert a new one.
+  Update the user with the given email if one exists, otherwise insert a new one.
   """
   def upsert(username, email) do
     email = String.downcase(email)
-    user_matching_email = get_by_email(email)
 
     user =
-      cond do
-        is_nil(user_matching_email) ->
+      case get_by_email(email) do
+        nil ->
+          # If there isn't an existing user with this email address create a new one.
+          # If there is an existing user with the same username, set the email address for that user
           Skate.Repo.insert!(
             DbUser.changeset(%DbUser{}, %{
               username: username,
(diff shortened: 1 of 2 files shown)


lib/skate/settings/user.ex
```diff
@@ -54,22 +55,9 @@ defmodule Skate.Settings.User do
...
-              username: username
-            }),
-            returning: true,
-            conflict_target: [:username],
-            # update a row with no effect so the returning works
-            on_conflict: {:replace, [:username]}
-          )
+        existing_user_with_email ->
+          # If there is an existing user with this email, return it without modifying the username
+          existing_user_with_email
```
KaylaBrady, line 60: > without modifying the username

This is due to a limitation of the old case "username format has changed. Create a new record for this user without an associated email". For example, now for some users there are two records in the db:
`%{email: "kb@fake_email.com", username: "old_format\kb"}`
`%{email: nil, username: "new_format\kb"}`

user "new_format\kb" still has email `"kb@fake_email.com", but we didn't save it in order to enforce the unique constraint on email. So when "new_format\kb" signs in, this will find user `%{email: "kb@fake_email.com", username: "old_format\kb"}` and return it. We cannot update the username to "new_format\kb" without violating the unique constraint. 

Once all the old records without email addresses are removed from the DB, we can change this clause to update the username (if the username field still exists and is in use).

lib/skate/settings/user.ex
```diff
@@ -8,7 +8,7 @@ defmodule Skate.Settings.User do
   @doc """
   Get a user with the matching email, if one exists
   """
-  def get_by_email(email) do
+  def get_by_email(email) when is_binary(email) do
```
lemald, line 11: **question:** what's the reasoning behind adding the `when is_binary(email)` guards here and in `upsert/2`? Just to make sure that we raise an exception if we somehow get a user without email information trying to log in rather than creating a new DB entry with a null email?

KaylaBrady: Yes, exactly! An exception would otherwise be thrown from `String.downcase(email)` if email is nil, but I think it would be clearer to raise that kind of error from the top of the function.

lemald (approved):
👍

PR #1807, opened 2022-11-17
chore: Update AWS certificate bundle hash

KaylaBrady:
to fix failing deploys: https://github.com/mbta/skate/actions/runs/3489790242


lemald (approved):
Interesting, this changed just a few days ago. If it starts to change more often we may have to rethink pinning the hash like this.

PR #1809, opened 2022-11-18
feat(MapMode): Vehicle card with street view link

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203207962355544/f


assets/src/components/liveVehicleCard.tsx
```diff
@@ -0,0 +1,28 @@
+import { Socket } from "phoenix"
+import React, { useContext } from "react"
+import { ReactElement } from "react"
+import { SocketContext } from "../contexts/socketContext"
+import useVehicleForId from "../hooks/useVehicleForId"
+import { isVehicle } from "../models/vehicle"
+import { Vehicle } from "../realtime"
+import VehicleCard from "./vehicleCard"
+
+const LiveVehicleCard = ({
```
KaylaBrady, line 10: I roughly based this off https://github.com/mbta/skate/blob/master/assets/src/components/propertiesPanel.tsx

firestack:
👀 I'm excited! Can you or I put this on dev-green to see?!

KaylaBrady:
@firestack deploy now in progress 😃

assets/src/components/mapPage.tsx
```diff
@@ -101,13 +106,23 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
         </div>
       </div>
+      <div className="m-map-page__map-container">
+        {selectedVehicle && showVehicleCard && (
+          <div className="m-map-page__vehicle-card">
+            <LiveVehicleCard
+              vehicle={selectedVehicle}
+              onClose={() => setShowVehicleCard(false)}
+            />
+          </div>
```
KaylaBrady, line 116: One piece I'd particularly appreciate feedback on is having this VehicleCard as an element that is styled over the map, vs. passing it to the map as a control. Keeping it separate from the map seemed the lowest lift to implement, but it means that the card isn't visible when the map is put in fullscreen mode. I'm not sure that matters at this point, but can spend a little more time turning this into a map control if that seems worthwhile.

KaylaBrady: Update: I ended up looking into it a bit more and looks like `LiveVehicleCard` could be passed as a child to the map container without needing explicitly be made into a control. That is fairly straightforward and would resolve the fullscreen issue described above. I'd still be interested in whether that seems like a sensible approach, or if there are benefits to keeping the `LiveVehicleCard` separate from the map component that I'm missing.

assets/css/_vehicle_card.scss
```diff
@@ -0,0 +1,52 @@
+.m-vehicle-card {
+  display: flex;
+  background-color: white;
+  border-radius: 1rem;
+  border: 1px solid black;
+  padding: 0.75rem;
+}
```
lemald, line 7: **question:** this is probably as much of a design question as a dev question, but is there any overlap between this and our `<Card/>` component that was built for notificatiions but intended to be general-purpose?

assets/src/components/liveVehicleCard.tsx
lemald: Some of the complexity with the VPP code, including the way it separately uses the `useVehicle` hook independently of other vehicle state, is actually due to the fact that it needs to keep working if you navigate to a page like settings that doesn't usually cause any vehicle information to be loaded. I think in this case since the card will only ever be appearing in the context of a map that already has the vehicle on it you could just pass the vehicle in via props.

assets/src/components/mapPage.tsx
lemald: I'd say that unless you discover additional downsides, passing it in as a child of the map would probably be best.

assets/src/components/vehicleCard.tsx
```diff
@@ -0,0 +1,83 @@
...
+          variant={vehicle.viaVariant}
+          status={drawnStatus(vehicle)}
+          userSettings={userSettings}
+        />
+      </div>
+      <div>
+        <div className="m-vehicle-card__header">
+          <div className="m-vehicle-card__header-contents">
+            <div>
+              {directionName(vehicle, route)}{" "}
```
lemald, line 55: **nitpick:** I think the `{" "}` may be a weird artifact of `prettier` that you can get rid of.

lemald:
@KaylaBrady okay, I'm done with my first pass. Aside from the comments I left all looks good. I think it's worth it to try simplifying `LiveVehicleCard` and how it gets its vehicle data if possible.

assets/src/components/liveVehicleCard.tsx
KaylaBrady: Thanks for drawing attention to this weirdness! Passing `selectedVehicle` as a prop wasn't live updating the data, but I added a `useEffect` hook so that when the search results are updated, the `selectedVehicle` is updated too.

assets/css/_vehicle_card.scss
KaylaBrady: Yeah good question, I overlooked `<Card/>` in implementing. The only limitation right now with `Card` is that it has a required `Title` param which doesn't seem obvious to me looking at the low-fi designs but is maybe something that can be worked out with design as the fidelity increases.

assets/src/components/mapPage.tsx
```diff
@@ -65,6 +65,20 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
+        (v) => v.id === selectedVehicle.id
+      )
+      if (updatedSelectedVehicle && isVehicle(updatedSelectedVehicle)) {
+        setSelectedVehicle(updatedSelectedVehicle)
+      } else {
+        setSelectedVehicle(null)
+        setShowVehicleCard(false)
+      }
+    }
+  }, [vehicles])
```
lemald, line 80: **suggestion:** I think it should be possible, and preferable, to disentangle the various pieces of state here such that we don't need a `useEffect` hook to make this update, since it's all within React state and / or props. That said, I haven't fully thought through exactly how to do that. Other hooks like `useMemo` may help with preventing repetition of the computational part when `vehicles` doesn't change.

KaylaBrady: That's fair, I'm going to re-mark this as a draft for now and do some more thinking about that after addressing the requested changes in the base branch.

KaylaBrady: This hook has been removed in [381021d](https://github.com/mbta/skate/pull/1809/commits/381021df8e2ee873f098978875f6212cb4e9c9d3). This also modifies the state to only store the selectedVehicleId rather than the whole vehicle - that helped me as I was thinking about it to distinguish between what is static and what is live. 

I started to go down a separate rabbit hole of considering whether or not it makes sense for the map page to be preserving its own notion of a separate vehicle rather than using the `separateVehicleOrGhost` from state, but I think that consideration can be addressed outside of this PR.

assets/tests/components/map.test.tsx
```diff
@@ -116,13 +116,16 @@ describe("map", () => {
...
+    jest.spyOn(global, "scrollTo").mockImplementationOnce(jest.fn())
+
     const onClick = jest.fn()
     render(<Map vehicles={[vehicle]} onPrimaryVehicleSelect={onClick} />)
     await userEvent.click(screen.getByText(runIdToLabel(vehicle.runId!)))
     expect(onClick).toHaveBeenCalledWith(expect.objectContaining(vehicle))
   })
 
   test("does not perform onPrimaryVehicleSelected function when secondary vehicle selected", async () => {
+    jest.spyOn(global, "scrollTo").mockImplementationOnce(jest.fn())
```
firestack, line 128: question(non-blocking): would you mind explaining why these were added? 
Seems like it's nessicary, but I don't have a good sense of why.

firestack (commented):
Apparently I had some comments pending per my last review before redrafting, so I'm getting those out of my editor here.

assets/src/components/mapPage.tsx
```diff
@@ -58,13 +59,18 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
   )
   const onlyVehicles: Vehicle[] = filterVehicles(vehicles)
   const [mobileDisplay, setMobileDisplay] = useState(MobileDisplay.List)
-  const [selectedVehicle, setSelectedVehicle] = useState<VehicleOrGhost | null>(
+  const [selectedVehicleId, setSelectedVehicleId] = useState<string | null>(
```
firestack, line 62: type-suggestion: can we use `VehicleId` here instead of `string`?

assets/src/components/mapPage.tsx
```diff
@@ -102,21 +111,27 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
-      <div className="m-search-page__map">
+      <div className="m-map-page__map">
         <Map
           vehicles={onlyVehicles}
-          onPrimaryVehicleSelect={setSelectedVehicle}
+          onPrimaryVehicleSelect={selectVehicle}
           shapes={selectedVehicleShapes}
-        />
+        >
+          {liveVehicle && showVehicleCard ? (
```
firestack, line 128: comment(non-blocking): I've been reading `vehicle` so many times that I thought this was a typo at first 🤪

assets/css/_map_page.scss
```diff
@@ -0,0 +1,102 @@
...
+    .m-map-page__input-and-results {
+      flex: 1 1 auto;
+      height: 100%;
+    }
+
+    .m-map-page__map {
+      // Needs to be hidden instead of not displayed so the map still gets drawn properly
+      position: fixed;
+      top: 100vh;
+    }
```
firestack, line 60: thought: I'm curious about this comment, would `visibility: hidden;` work here? 
https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

I'm not really sure what this is doing though, so I guess either I'd need to see it or the comment could possibly be made more clear?

assets/tests/components/vehicleCard.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+import routeFactory from "../factories/route"
+
+import userEvent from "@testing-library/user-event"
+import VehicleCard from "../../src/components/vehicleCard"
+import { useNearestIntersection } from "../../src/hooks/useNearestIntersection"
+import { RoutesProvider } from "../../src/contexts/routesContext"
+jest
+  .spyOn(dateTime, "now")
+  .mockImplementation(() => new Date("2018-08-15T17:41:21.000Z"))
+
```
firestack, line 17: question(non-blocking): any particular reason for this timestamp?

I've been idly mulling over if switching to a unix timestamp would be better for testing (eg let's findout what skate does when it's back in 0 seconds, 1970)
but I think this is mostly bikeshedding.

firestack (approved):
LGTM!

assets/tests/components/map.test.tsx
KaylaBrady: These resolve console logs that were introduced in #1802 - see previous run [here](https://github.com/mbta/skate/actions/runs/3577274849/jobs/6016096725#step:8:33).

Clicking a leaflet marker calls `scrollTo` under the hood, which is why this is only added to tests with marker click interactions.

lemald: I think this is to address some error output from running the tests that I noticed were introduced in #1802.

assets/src/components/vehicleIcon.tsx
```diff
@@ -42,6 +49,34 @@ export interface TooltipProps {
...
+    const ladderDirection: LadderDirection = getLadderDirectionForRoute(
+      ladderDirections,
+      vehicle.routeId
+    )
+    const vehicleDirection: VehicleDirection = directionOnLadder(
+      vehicle.directionId,
+      ladderDirection
+    )
+
+    if (vehicle.routeStatus === "laying_over") {
```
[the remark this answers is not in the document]
KaylaBrady: It seems to! To test I changed this line to`(vehicle.routeStatus === "a")` and got the error `This condition will always return 'false' since the types 'RouteStatus' and '"a"' have no overlap.`

assets/css/_map_page.scss
KaylaBrady: There seems to be something else at play here - I tried changing this to hidden but the map displayed very zoomed in when using mobile dimensions. 

This (and most of the styles in _map_page) were copied from the original styles in [_search_page.scss](https://github.com/mbta/skate/blob/493b9ec20e7896c32224d5fde00c2792a161d7b4/assets/css/_search_page.scss#L56) - I tried to keep the changes minimal beyond that for the purposes of only getting the new card positioned correctly relative to the existing elements.

lemald (approved):
Looks good!

assets/tests/components/vehicleCard.test.tsx
KaylaBrady: Great question! I pulled this mock from the search results test file, but turns out it is not needed for this component. The mock is only needed if we were to test that the operator logon is shown in this card (the default last logon time defined in the vehicle factory is [2018-08-15T13:38:21.000Z](https://github.com/mbta/skate/blob/15df510fc794a8b9a8e5c4e3045b46e33aed322a/assets/tests/factories/vehicle.ts#L19), but since that data is not present in this new card (at least for now) the mock can be omitted.

PR #1810, opened 2022-11-18
chore(aws-cert-bundle): update pinned file hash

firestack:
Seems it updated again because of https://github.com/mbta/skate/actions/runs/3498816299

I'll open a tech debt ticket to look into automating this.


KaylaBrady (approved):
😢 3 times in a month! Thanks for fixing & ticketing!

firestack:
I also learned that we have a skate developers group on github and that you can assign groups as reviewers!

Though, I wonder if the 
> approved these changes on behalf of
means it yanks it out of people's github inboxes?

I guess we'll see!

PR #1817, opened 2022-11-22
Detour alerts backend

lemald:
Asana ticket: [🎨⚙️ Show existing diversions on route ladder](https://app.asana.com/0/1200180014510248/1203142766477809/f)

This is exclusively the back-end code for the ticket. The front-end will follow in another PR, and we'll want to release the back-end first so that the channel is available when the front-end code is deployed.

There were a few possible ways to pull this alert data in and make it available to the channel. It may have been advantageous to use [Server Sent Event-based streaming](https://www.mbta.com/developers/v3-api/streaming), both for reducing latency and conserving bandwidth. However, that would involve managing the alerts on a by-ID basis, whereas we prefer to keep them grouped by route ID for internal purposes. This wasn't an insurmountable barrier but it would have made the code somewhat messier. Furthermore, the size of the data isn't too large (the API query specifically only gets currently-active alerts and only for bus), and the latency isn't a huge issue for our use case.

Finally, there was a bit of a question as to whether to use `Realtime.Server` (like most of our real-time bus data) or the Phoenix PubSub (like the live train data). I opted for the former simply because that's how most of our real-time data works but I'd be curious as to whether or not `Realtime.Server` should just be migrated to use `Phoenix.Pubsub` someday.

### DIFF
--- a/lib/realtime/alerts_fetcher.ex
+++ b/lib/realtime/alerts_fetcher.ex
@@ -94,29 +94,11 @@ defmodule Realtime.AlertsFetcher do
             is_detour? = alert.attributes["effect"] == "DETOUR"
 
             if is_detour? do
-              route_ids =
-                Enum.reduce(alert.attributes["informed_entity"], [], fn informed_entity,
-                                                                        route_ids ->
-                  if informed_entity["route"] && informed_entity["route"] not in route_ids do
-                    [informed_entity["route"] | route_ids]
-                  else
-                    route_ids
-                  end
-                end)
+              route_ids = route_ids_from_informed_entities(alert.attributes["informed_entity"])
 
               alert_text = alert.attributes["service_effect"]
 
-              Enum.reduce(route_ids, acc, fn route_id, acc ->
-                {_old_value, acc} =
-                  Map.get_and_update(acc, route_id, fn alerts_for_route ->
-                    case alerts_for_route do
-                      nil -> {alerts_for_route, [alert_text]}
-                      alerts -> {alerts_for_route, [alert_text | alerts]}
-                    end
-                  end)
-
-                acc
-              end)
+              add_alerts_by_routes(route_ids, alert_text, acc)
             else
               acc
             end
@@ -129,5 +111,32 @@ defmodule Realtime.AlertsFetcher do
       {:error, error} ->
         {:error, error}
     end
+  end
+
+  @spec route_ids_from_informed_entities([map()]) :: [Route.id()]
+  defp route_ids_from_informed_entities(informed_entities) do
+    Enum.reduce(informed_entities, [], fn informed_entity, route_ids ->
+      if informed_entity["route"] && informed_entity["route"] not in route_ids do
+        [informed_entity["route"] | route_ids]
+      else
+        route_ids
+      end
+    end)
+  end
+
+  @spec add_alerts_by_routes([Route.id()], String.t(), Route.by_id([String.t()])) ::
+          Route.by_id([String.t()])
+  defp add_alerts_by_routes(route_ids, alert_text, alerts_by_routes) do
+    Enum.reduce(route_ids, alerts_by_routes, fn route_id, acc ->
+      {_old_value, acc} =
+        Map.get_and_update(acc, route_id, fn alerts_for_route ->
+          case alerts_for_route do
+            nil -> {alerts_for_route, [alert_text]}
+            alerts -> {alerts_for_route, [alert_text | alerts]}
+          end
+        end)
+
+      acc
+    end)
   end
 end
--- a/lib/realtime/alerts_fetcher.ex
+++ b/lib/realtime/alerts_fetcher.ex
@@ -1,0 +1,133 @@
+defmodule Realtime.AlertsFetcher do
+  use GenServer
+  require Logger
+  alias Schedule.Route
+
+  @default_poll_interval_ms 3 * 60 * 1_000
+
+  @type state :: %{
+          update_fn: Route.by_id([String.t()]),
+          poll_interval_ms: integer(),
+          api_url: String.t(),
+          api_key: String.t()
+        }
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    {name, opts} = Keyword.pop(opts, :name, __MODULE__)
+
+    GenServer.start_link(__MODULE__, opts, name: name)
+  end
+
+  @impl true
+  def init(opts) do
+    update_fn = Keyword.get(opts, :update_fn, &Realtime.Server.update_alerts/1)
+
+    poll_interval_ms = Keyword.get(opts, :poll_interval_ms, @default_poll_interval_ms)
+
+    {:ok, _} = :timer.send_interval(poll_interval_ms, self(), :query_api)
+
+    {:ok,
+     %{
+       update_fn: update_fn,
+       poll_interval_ms: poll_interval_ms,
+       api_url: Application.get_env(:skate, :api_url),
+       api_key: Application.get_env(:skate, :api_key)
+     }, {:continue, :initial_poll}}
+  end
+
+  @impl true
+  def handle_continue(:initial_poll, state) do
+    do_poll(state)
+  end
+
+  @impl true
+  def handle_info(:query_api, state) do
+    do_poll(state)
+  end
+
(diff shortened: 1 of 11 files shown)


lib/realtime/server.ex
```diff
@@ -342,12 +378,17 @@ defmodule Realtime.Server do
...
     registry_key = self()
 
     Registry.dispatch(Realtime.Supervisor.registry_name(), registry_key, fn entries ->
-      Enum.each(entries, &send_data(&1, state))
+      Enum.each(entries, fn {pid, subscripition_key} ->
+        if (data_type == :alerts and match?({:alerts, _}, subscripition_key)) or
+             (data_type == :vehicles and !match?({:alerts, _}, subscripition_key)) do
+          send_data({pid, subscripition_key}, state)
+        end
+      end)
```
KaylaBrady, line 391: > Finally, there was a bit of a question as to whether to use Realtime.Server (like most of our real-time bus data) or the Phoenix PubSub (like the live train data).

suggestion (non-blocking): I don't have a  have a strong understanding of the technical differences between using `server` or `PubSub` yet so this may be a fully uniformed thought, but I do tend to like how the train & status data are managed separately from this server. It might be nice to have one file focused only on alerts data, whether that uses `PubSub` or is structured in the same way as `server`, so that this kind of conditional check isn't needed.

lib/skate_web/channels/alerts_channel.ex
```diff
@@ -0,0 +1,27 @@
+defmodule SkateWeb.AlertsChannel do
+  use SkateWeb, :channel
+
+  alias Realtime.Server
+  alias Util.Duration
+
+  @impl Phoenix.Channel
+  def join("alerts:route:" <> route_id, _message, socket) do
+    alerts = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
```
KaylaBrady, line 9: 🙌 `log_duration` is cool - haven't come across this before!

test/realtime/alerts_fetcher_test.exs
```diff
@@ -0,0 +1,218 @@
...
+      {:ok, state, _} = AlertsFetcher.init(update_fn: update_fn)
+
+      Bypass.expect(bypass, fn conn ->
+        Plug.Conn.resp(conn, 500, "server error")
+      end)
+
+      log =
+        capture_log([level: :warn], fn ->
+          assert {:noreply, _state} = AlertsFetcher.handle_info(:query_api, state)
+        end)
```
KaylaBrady, line 159: thought(non-blocking): I remember coming across [with_log](https://hexdocs.pm/ex_unit/1.14.2/ExUnit.CaptureLog.html#with_log/2) but have never actually used it. This could be a case for it so that the assertions can be kept together at the end of the test.

KaylaBrady (commented):
Looks good! Have just a few minor comments

lib/realtime/server.ex
firestack: I was a bit confused by the filter here, personally, it took a while to make sense of it due to indentation.

I agree with Kayla that removing this conditional and making it it's own codepath would possibly result in cleaner logic. I'd otherwise vote for filtering `entries` before using `Enum.each` instead of `if`s within the each lambda.

---

(Also I'm commenting now because I want the slack github reminders to stop on vacation lol)

firestack (approved):
Looks good! 
I agreeed some of Kayla's points and +1'ed them to keep noise down.

lib/realtime/alerts_fetcher.ex
```diff
@@ -0,0 +1,133 @@
+defmodule Realtime.AlertsFetcher do
+  use GenServer
+  require Logger
+  alias Schedule.Route
+
+  @default_poll_interval_ms 3 * 60 * 1_000
```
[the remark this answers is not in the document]
lemald: Yes, that's a good point. Of course, if streaming had been a more reasonable approach for implementing this feature it would have provided better latency. In any event it should be a fairly inexpensive API query so I'm okay with bumping it to once / minute.

test/realtime/alerts_fetcher_test.exs
lemald: Oh neat, I hadn't seen that before, will look into using it.

lib/realtime/server.ex
lemald: So my thinking here is that we might actually want to use this as a jumping off point to do more filtering in the `Realtime.Server` module around subscriptions. Currently, any new vehicle data triggers all vehicles to be pushed to their respective channels, even if those vehicles weren't actually updated, so I saw it as a bonus that this was creating logic to only updates to certain subscribers. I can create a tech debt ticket to actually do that filtering.

lib/realtime/alerts_fetcher.ex
```diff
@@ -0,0 +1,133 @@
...
+    {:noreply, state}
+  end
+
+  @spec parse_and_update_alerts(binary(), (Route.by_id([String.t()]) -> term())) ::
+          :ok | {:error, any()}
+  defp parse_and_update_alerts(body, update_fn) do
+    case JsonApi.parse(body) do
+      %JsonApi{data: data} ->
+        alerts =
+          Enum.reduce(data, %{}, fn alert, acc ->
```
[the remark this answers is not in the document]
KaylaBrady: This is helpful, thank you!

KaylaBrady (approved):

lib/realtime/server.ex
lemald: Okay, gotcha. Yeah, I think my idea of only broadcasting for updated vehicles / alerts might be harder than I initially thought with the code as it is. I still think it's a worthy goal, though, since it might make it possible to cut down on how much data we need to push over websockets by a fair bit. It sounds like we basically need a rethink / rationalization of how realtime data like this is handled on the back-end. I'll write up some thoughts in a tech cleanup ticket.

lemald: Asana ticket: https://app.asana.com/0/1152340551558956/1203441390849641/f

test/realtime/alerts_fetcher_test.exs
```diff
@@ -0,0 +1,218 @@
...
+      update_fn = fn _ -> :ok end
+
+      assert {:ok,
+              %{
+                update_fn: ^update_fn,
+                poll_interval_ms: 500,
+                api_url: ^api_url,
+                api_key: ^api_key
+              },
+              {:continue, :initial_poll}} =
```
firestack, line 85: comment(non-blocking): I'm new to elixir but this indentation/wrapping is kinda funky?

Because I'm learning, I'm wondering if the pattern match or assert be factored out to make it easier to read?

This seems like fairly standard elixir syntax so I don't think this needs to change.

test/realtime/alerts_fetcher_test.exs
```diff
@@ -0,0 +1,218 @@
...
+
+      {:ok, state, _} = AlertsFetcher.init(update_fn: update_fn)
+
+      Bypass.expect(bypass, fn conn ->
+        Plug.Conn.resp(conn, 200, Jason.encode!(@api_response))
+      end)
+
+      assert {:noreply, _state} = AlertsFetcher.handle_continue(:initial_poll, state)
+
+      assert_receive {:"$gen_cast", %{"1" => ["Route 1 detour"]}}, 5000
```
firestack, line 110: question(non-blocking): What's this `$gen_cast` atom? I currently couldn't find docs on it where I knew to look.

firestack (approved):

test/realtime/alerts_fetcher_test.exs
lemald: Ah, right, this actually confused me at first as well - the actual message sent to the process by `GenServer.cast` uses this format to indicate that it's a cast as opposed to a call or other generic message: https://github.com/elixir-lang/elixir/blob/0909940b04a3e22c9ea4fedafa2aac349717011c/lib/elixir/lib/gen_server.ex#L1098

test/realtime/alerts_fetcher_test.exs
lemald: I see what you mean about how it's weird, but it's what the Elixir LSP server did for formatting and `mix format --check` likes it. I'm not sure what part you mean that we might be able to factor out - do you mean something like putting `{:continue, :initial_poll}` in its own variable and then pinning it in the match?

PR #1822, opened 2022-11-29
Display detour alerts in UI

lemald:
Asana ticket: [🎨⚙️ Show existing diversions on route ladder](https://app.asana.com/0/1200180014510248/1203142766477809/f)

This will hopefully be somewhat easier to review than #1817. It's worth paying a bit of attention to the `useAlerts` hook since I think that the way I put it together could serve as a basis for refactoring `useVehicles` and some other hooks that still have issues. There's [an Asana ticket for this refactor](https://app.asana.com/0/1152340551558956/1202965256382465/f).


assets/src/hooks/useAlerts.ts
```diff
@@ -0,0 +1,89 @@
...
+  return channel
+}
+
+const useAlerts = (
+  socket: Socket | undefined,
+  routeIds: RouteId[]
+): ByRouteId<string[]> => {
+  const [alertsByRoute, setAlertsByRoute] = useState<ByRouteId<string[]>>({})
+  // eslint-disable-next-line react/hook-use-state
+  const [, setChannelsByRouteId] = useState<ByRouteId<Channel>>({})
```
lemald, line 54: Since `channelsByRouteId` is only ever referenced using the update function form of `setChannelsByRouteId`, we don't actually need to pull that variable out of the tuple. As you can see, though, this triggered a hooks ESLint rule. I'm not sure if there's a better way to handle this sort of case where we don't need to reference the state directly (but we do need it so that we know which channels need to be added / removed), but I'm open to suggestions.

assets/tests/components/ladderPage.test.tsx
```diff
@@ -423,9 +427,9 @@ describe("LadderPage", () => {
     ;(useVehicles as jest.Mock).mockImplementationOnce(() => ({
       ["1"]: [vehicle],
     }))
-    ;(useTimepoints as jest.Mock).mockImplementationOnce(
-      () => timepointsByRouteId
-    )
+    ;(useTimepoints as jest.Mock)
+      .mockImplementationOnce(() => timepointsByRouteId)
+      .mockImplementationOnce(() => timepointsByRouteId)
```
firestack, line 432: question: Why use `mockImplementationOnce` twice instead of `mockImplementation`?

assets/src/components/ladderPage.tsx
```diff
@@ -154,6 +156,16 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
...
+  const { socket } = useSocket()
+  const alerts = useAlerts(socket, selectedRouteIds)
+  const routesWithAlerts = []
+
+  for (const routeId in alerts) {
+    if (alerts[routeId].length > 0) {
+      routesWithAlerts.push(routeId)
+    }
+  }
+
```
firestack, line 168: question: why map/collect alerts here instead of passing `alerts[routeId]` inside of `routeLadders`?

To me it seems that having the actual alert objects within the route ladder could be useful in the future if we want to provide more context then the `"Active detour"` tooltip. We could also provide the number of alerts on a route if we passed the alerts array to the route ladder.

Additionally, if the routes with alerts default to an empty array, and the `.length > 0` check is moved into the route ladder component, then the property doesn't need to be set for all the tests that didn't have it before?

But maybe there's something else or something I'm missing?

assets/tests/hooks/useAlerts.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const mockSocket = makeMockSocket()
+    const mockChannel = makeMockChannel("timeout")
+    mockSocket.channel.mockImplementationOnce(() => mockChannel)
+
+    renderHook(() => useAlerts(mockSocket, ["1"]))
+
+    expect(reloadSpy).toHaveBeenCalled()
+    reloadSpy.mockRestore()
+  })
+})
```
firestack, line 92: nit: I think duplicate calls to `mockImplementationOnce` like `mockSocket.channel.mockImplementationOnce(() => mockChannel)` should be grouped into one `.mockImplementation(() => mockChannel)` if possible?

assets/tests/components/ladderPage.test.tsx
KaylaBrady: [This comment](https://github.com/mbta/skate/pull/1822#discussion_r1036064847) _may_ be related - I've previously run into wanting to call `mockImplementationOnce` twice due to multiple rerenders when calling `useSocket` instead of getting the socket from context.

assets/src/hooks/useAlerts.ts
```diff
@@ -0,0 +1,89 @@
...
+        Object.entries(oldChannelsByRouteId).forEach(([routeId, channel]) => {
+          if (currentRouteIds.includes(routeId)) {
+            newChannelsByRouteId[routeId] = channel
+          } else {
+            channel.leave()
+          }
+        })
+
+        currentRouteIds.forEach((routeId) => {
+          const channel = subscribe(socket, routeId, setAlertsByRoute)
```
KaylaBrady, line 76: question: if a currentRouteId was in oldChannelsByRouteId, would they be subscribed to that channel twice?

assets/src/components/ladderPage.tsx
```diff
@@ -154,6 +156,16 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
   const timepointsByRouteId: TimepointsByRouteId =
     useTimepoints(selectedRouteIds)
 
+  const { socket } = useSocket()
```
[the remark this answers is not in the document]
lemald: Not really. I had never really considered it, but doing some searching through the codebase it sadly appears that we're very inconsistent about using the socket from the context vs. creating a new one via `useSocket`. For this feature I don't think there's any reason we would need a new socket so I'll change it to use the context approach.

assets/src/hooks/useAlerts.ts
lemald: Right, I meant to do a new subscription only if there wasn't a corresponding channel in `oldChannelsByRouteId`, thanks for catching that.

assets/tests/components/ladderPage.test.tsx
lemald: I wasn't fully sure about this actually but @KaylaBrady was right - since I'm making the change to use the socket from the context I can go back to just a single call to `mockImplementationOnce`.

assets/src/components/ladderPage.tsx
lemald: I was taking a sort of conservative "pass only the data that the component really needs" approach, but you're definitely right that we might want to access some of that information in the future, so I'm open to the idea of changing it.

> if the routes with alerts default to an empty array, and the .length > 0 check is moved into the route ladder component, then the property doesn't need to be set for all the tests that didn't have it before?

I think I'm missing something here, could you elaborate?

firestack: > I was taking a sort of conservative "pass only the data that the component really needs" approach
Okay cool that's what I figured 👍🏻. 

> > if the routes with alerts default to an empty array, and the .length > 0 check is moved into the route ladder component, then the property doesn't need to be set for all the tests that didn't have it before?
> 
> I think I'm missing something here, could you elaborate?

Yeah it's strangely phrased and I don't think it was necsisary anyway.
I was trying to communicate that the codeblock this comment is attached to (and _maybe_ some of the supporting infrastructure) could disapear if the `hasAlert` property could be defined as `bool | null` and the route ladder could treat the null/undefined case also as `false`, which would reduce the extra changes needed to add `hasAlert` to all relavent tests, but that comes down to preference and shouldn't hold up merging.

KaylaBrady (approved):
Looks good!

PR #1824, opened 2022-12-01
feat(Map): Display garages for users in test group

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203329634351693/f


assets/src/components/map.tsx
```diff
@@ -39,6 +40,12 @@ import appData from "../appData"
 import { createControlComponent } from "@react-leaflet/core"
 import "leaflet.fullscreen"
 
+import garages, { Garage } from "../data/garages"
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
KaylaBrady, line 45: same ignores as https://github.com/mbta/skate/blob/f1633ba3f09b1aefaa8544c9ac0a5519003963d7/assets/src/helpers/icon.tsx#L1

firestack: question(non-blocking): do you know why we import SVG with javascript imports? 

maybe this is this something that gets inlined by webpack?
but if so, why do we want it inlined?

assets/tests/components/map.test.tsx
```diff
@@ -70,6 +72,14 @@ const vehicle: Vehicle = vehicleFactory.build({
   crowding: null,
 })
 
+jest.mock("appData")
+
+beforeEach(() => {
+  ;(appData as jest.Mock).mockImplementation(() => ({
+    userTestGroups: JSON.stringify([]),
+  }))
+})
```
firestack, line 81: nit(non-blocking): what if instead of using appData we mocked the `inTestGroup` function to return `true`?

If that's not possible, or if we want to be testing the actual test group string we're using, then I think we should mock `getTestGroups` to return an array, saving us a JSON "serde"(serialize-deserialize)?

I'm conflicted though because that feels like depending on the test groups utility API too specifically.

assets/src/components/map.tsx
```diff
@@ -371,6 +378,56 @@ const Autocenterer = ({
...
+    )}
+  </>
+)
+
+const Garages = () => {
+  const startingZoomLevel = useMap().getZoom()
+  const [zoomLevel, setZoomLevel] = useState(startingZoomLevel)
+  const map = useMapEvent("zoomend", () => {
+    setZoomLevel(map.getZoom())
+  })
```
firestack, line 420: question: what is setting the zoom level after zoom finishes acomplishing?

assets/src/components/map.tsx
```diff
@@ -371,6 +378,56 @@ const Autocenterer = ({
...
+    />
+    {zoomLevel >= 16 && (
+      <Marker
+        interactive={false}
+        position={[garage.lat, garage.lon]}
+        icon={Leaflet.divIcon({
+          className: "m-garage-icon__label",
+          html: `<svg height="30" width="200">
+                    <text x="22" y="15">${garage.name}</text>
+                  </svg>`,
```
firestack, line 408: question: why use inline svg here?

KaylaBrady: I was following the pattern of the existing [vehicle icon label](https://github.com/mbta/skate/blob/9addac1563f9fa0ea646b1efcd60d2ecbcaa0df1/assets/src/components/map.tsx#L94). Though I haven't done. much with SVGs before so if there is a alternative preferred way I handling I'm interested in learning!

assets/src/components/map.tsx
KaylaBrady: I'm not sure why - maybe @lemald knows more?

lemald: Yeah, that was already established practice when I joined and I'm not sure entirely why it's done the way it is. It probably does have something to do with making sure that it gets added to the Webpack build but it's not clear to me if that's actually the best way to handle things. I'm open to suggestions on changes we could make.

assets/tests/components/map.test.tsx
lemald: I'd probably lean slightly in the direction of mocking the `getTestGroups` function as well.

KaylaBrady: Thanks for the suggestion! Addressed in [cadf169](https://github.com/mbta/skate/pull/1824/commits/cadf169e9b18137c2f2b3c287bbd008ae008d48a)

KaylaBrady: Ah I tried mocking `getTestGroups` but it ran into issues since it is called by `inTestGroups` -- found this [post](https://medium.com/@DavideRama/mock-spy-exported-functions-within-a-single-module-in-jest-cdf2b61af642) about possible solutions (the better one to me seems like putting `getTestGroups` into its own file, but I'm not sure if that is overkill and only opted to mock `inTestGroup` for now

assets/src/components/map.tsx
lemald: @firestack thanks for digging that old PR up!

firestack: https://leafletjs.com/reference.html#divicon-html says it's `String|HTMLElement` so I'd assume that we could return an element fragment from JSX to this?

lemald: Oh, interesting, I saw an older page that just said that it had to be a string. Yeah if passing an `HTMLElement` works that would be great.

assets/tests/components/map.test.tsx
firestack: Thank you for finding that out for me AND linking relavent info @KaylaBrady!

I'm agreeing with Eddie that mocking `getTestGroups` is better, especially if we ever needed to test code where one test group is on and the other is off. If to achieve that we have to move it into it's own file?... I don't mind, but I agree it feels a little overkill. 
When originally writing this comment I was hoping jest would have some easy way to make it work :c

assets/src/components/map.tsx
KaylaBrady: This may be my own inexperience around `HTMLElement` - my first thought was that I would only need to remove the quotations around the SVG and it would just work:tm:, though it is in fact a ReactElement & doesn't overlap with `HTMLElement`. From how I've seen `HTMLElement`s created before, it involves calling `document.createElement` and then eventually `.innerHtml`, which takes a string and therefore doesn't seem that different from the current approach. 

Is there a terser way to create this svg element as an `HTMLElement` that would be preferable to the current approach?

assets/tests/components/map.test.tsx
KaylaBrady: Yeah I'm with you both that `getTestGroups` would be better here, and might as well do it now so it is easier for testing test groups in the future! I'll add a commit for it.

assets/src/components/map.tsx
firestack: Wow I didn't expect JSX to make it so difficult to go from `FunctionComponentElement` to `HTMLElement` lol!

KaylaBrady: Same here 😄

assets/tests/components/map.test.tsx
KaylaBrady: updated in [9afa4d0](https://github.com/mbta/skate/pull/1824/commits/9afa4d08600ce7dd57a4ab833eff022bddc0e12f)!

firestack: I love it!! this is great thank you!

assets/tests/userInTestGroup.test.ts
```diff
@@ -0,0 +1,26 @@
+import inTestGroup from "../src/userInTestGroup"
+import getTestGroups from "../src/userTestGroups"
+
+jest.mock("userTestGroups", () => ({
+  __esModule: true,
+  default: jest.fn(() => []),
+}))
+
+describe("User Test Groups API", () => {
```
lemald, line 9: **nitpick:** usually the string we use in `describe` is the name of the function / component being tested.

assets/tests/components/map.test.tsx
```diff
@@ -115,6 +127,35 @@ describe("map", () => {
     expect(result.container.innerHTML).toContain("m-vehicle-map__stop")
   })
 
+  test("draws garage icons only at zoom levels < 16", () => {
+    ;(getTestGroups as jest.Mock).mockReturnValue([MAP_BETA_GROUP_NAME])
+    const { container } = render(<Map vehicles={[vehicle]} />)
+    expect(container.innerHTML).toContain("m-garage-icon")
+    expect(screen.queryByText("Albany")).toBeNull()
+  })
```
lemald, line 135: **nitpick:** add blank line below to visually separate `test`s.

assets/src/components/map.tsx
firestack: I just want to say that I think this is fine as is, especially given this is something we do elsewhere, I'm a bit sad that JSX can't handle this case for us, but maybe we can address that somehow in some future work.

firestack (approved):
Thanks again for splitting up the test groups functions and tests!

lemald (approved):
Latest changes look good. 👍

PR #1834, opened 2022-12-08
Dedicated Street View

lemald:
Asana Ticket: [⚙️ Better Maps | Integrations | Dedicated Street View Entrypoint](https://app.asana.com/0/1200180014510248/1203329634351686/f)

The basic idea of implementing the control with a portal was inspired by the [react-leaflet-custom-control](https://github.com/chris-m92/react-leaflet-custom-control) repository, but that has [an issue](https://github.com/chris-m92/react-leaflet-custom-control/issues/5) with creating new `<div>` containers on every render because it has no way of cleaning up the old `<div>`. My approach uses `useEffect` and, because it isn't positioned in one of the corners like normal Leaflet controls, just lives directly inside of the `leaflet-control-container` `<div>`.

One other thing worth noting is that I made use of `useId` to link the switch control to the text that labels it via accessibility properties. The default behavior of `useId` in tests seems deterministic so far, but if it isn't deterministic for whatever reason this could create problems for snapshot tests. I'm open to suggestions on improving that part if anyone has feedback.


assets/package.json
```diff
@@ -27,6 +27,7 @@
     "@sentry/react": "^7.15.0",
     "@tippyjs/react": "^4.2.6",
     "@types/uuid": "^8.3.4",
+    "bootstrap": "^5.2.3",
```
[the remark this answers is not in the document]
lemald: That's a good point. For this one component it was easier to just work with Bootstrap directly, but something more React-native could be nice. My one concern would be whether or not react-bootstrap allows including only part of the Bootstrap CSS in your project, which I think is worth considering given that our static content bundle is already kind of big.

assets/src/components/map.tsx
```diff
@@ -265,6 +268,51 @@ export const autoCenter = (
   }
 }
 
+const StreetViewControl = ({
```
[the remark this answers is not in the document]
lemald: I'm a little uncertain how much could be made truly reusable - in particular, this component is sort of unique in that it's put directly under the `leaflet-control-container` element. I think I'm leaning towards leaving it as-is for now but if we add more custom controls in the future it's definitely something to revisit.

assets/package.json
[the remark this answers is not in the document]
KaylaBrady: My impression is we could still only include the essential pieces of Bootstrap CSS:
https://react-bootstrap.github.io/getting-started/introduction/#stylesheets

> Because React-Bootstrap doesn't depend on a very precise version of Bootstrap, we don't ship with any included CSS

assets/tests/components/map.test.tsx
```diff
@@ -234,6 +234,62 @@ describe("map", () => {
...
+    await userEvent.click(screen.getByRole("switch", { name: "Street View" }))
+
+    await userEvent.keyboard("{Escape}")
+
+    const switchInput = screen.getByRole("switch", {
+      name: "Street View",
+    }) as HTMLInputElement
+
+    expect(switchInput.checked).toBeFalsy()
+  })
```
firestack, line 284: suggestion: would using `checked: false` instead of `toBeFalsy` be acceptable?

https://testing-library.com/docs/queries/byrole/#checked

assets/src/components/map.tsx
```diff
@@ -342,16 +394,36 @@ const EventAdder = ({
...
+            )
+          },
+
+          keydown: (e) => {
+            if (e.originalEvent.key === "Escape") {
+              setStreetViewMode(false)
+            }
+          },
+        }
+      : {}),
```
firestack, line 426: questions(non-blocking):
- is there any particular reasoning for using `key: arrow function`?
- should `zoomend` be changed as well?

I'm personally of the preference of avoiding arrow functions if it's served by some other syntax, but I don't really know why lol! so I'm not strongly opposed to the change.

lemald: Re: your first question, I'm just way more used to the arrow function syntax and honestly kind of forgot that you could also write a function as a value associated with an object key in the way that `zoomend` does it. I agree that one way or another we should make it consistent though. @KaylaBrady preferences one way or another?

KaylaBrady: I slightly prefer this arrow function way - similar to Eddie, that is now how I'm most accustomed to reading/writing functions. If that is the predominant existing syntax of Skate then I'm for it!

assets/tests/components/map.test.tsx
```diff
@@ -209,7 +209,7 @@ describe("map", () => {
     expect(onClick).not.toHaveBeenCalled()
   })
 
-  test("renders street view link if in maps test group", async () => {
+  test("renders street view link from stop if in maps test group", async () => {
```
firestack, line 212: praise: thanks for adding this! would you mind adding it to the test below this one as well?

assets/tests/components/map.test.tsx
lemald: Do you mean something like this?

```
    expect(
      screen.queryByRole("checked", { name: "Street View" })
    ).not.toBeInTheDocument()
```

If so I'm all for it.

firestack: my understanding of the docs is it would be something like this?
```suggestion
    const switchInput = screen.getByRole("switch", {
      name: "Street View",
      checked: false,
    }) as HTMLInputElement

    expect(switchInput).toBeInTheDocument()
  })
```

which could compress to
```suggestion
    expect(screen.getByRole("switch", {
      name: "Street View",
      checked: false,
    })).toBeInTheDocument()
  })
```

lemald: Ah, yes, that works. Sorry, just wasn't putting all of the pieces together quite right.

assets/src/components/map.tsx
```diff
@@ -265,6 +268,51 @@ export const autoCenter = (
...
+      Leaflet.DomEvent.disableClickPropagation(portalElement)
+    }
+
+    return () => portalElement?.remove()
+  }, [portalElement, portalParent, setStreetViewEnabled])
+
+  const control = (
+    <>
+      <span id={id}>Street View</span>
+      <div className="form-check form-switch">
```
firestack, line 300: suggestion/question: can a [`<label for=[id]>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) be used here instead of `<span>`/`aria-labelledby={id}`?

I also think that the ID should be declaritive/a function input, but I'm not opposed to `useId`.

firestack (commented):
Looks good! a few questions/suggestions but otherwise seems ready.

assets/src/components/map.tsx
firestack: > but I still don't want to do things that will cause multiple instantiations of a single component to step on each other in weird ways.

Agreed.
I think what I'd prefer is generating an id but prefixing it with the component name or something, but this is because I don't like opaque web-apps that try to do "security through obscurity" and I think that it would be a better ergonomics to include some human readable information in there (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id there's a warning callout on the page that sorta touches on this idea)

This is also assuming that prefixing the result of `useId` doesn't break anything.

lemald: Ah, I see. Yes, I like the idea of prefixing it with something and it doesn't seem to break anything, I'll go ahead and do that.

lemald: One thing to flag re: using `<label>` is that the contents of the label tag become clickable and effectively fire a click for the element they are labeling. I think that's probably fine if it's standard behavior but it's worth noting.

assets/css/app.scss
```diff
@@ -86,6 +86,19 @@ $z-properties-panel-context: (
 @import "vehicle_map";
 @import "view_header";
 
+// Bootstrap variable customization
+$component-active-bg: $color-eggplant-700;
```
KaylaBrady, line 90: question (non-blocking): I may be missing something - wouldn't this get overridden by the bootstrap imports below?

KaylaBrady (approved):
Looks great! Just one tiny question

assets/css/app.scss
KaylaBrady: Oh very cool! That section makes it seem like the `/functions` import should come first, but I can't tell if that actually is required for the color override to work.

> Variable overrides must come after our functions are imported, but before the rest of the imports.

assets/tests/components/map.test.tsx
```diff
@@ -234,6 +234,64 @@ describe("map", () => {
       screen.queryByRole("link", { name: /Go to Street View/ })
     ).not.toBeInTheDocument()
   })
+
+  test("can turn on street view and click on the map", async () => {
+    const openSpy = jest.spyOn(window, "open").mockImplementationOnce(jest.fn())
```
firestack, line 239: suggestion/question: I'm having trouble coming up with a name, but I think this should describe that street view should disable itself when a point is clicked.

lemald: Yeah I wanted to write a better name but was struggling to come up with something that isn't an absurd run-on sentence.

firestack: I'd also be cool with just adding a comment near the end of the test that explains that constraint. 

> struggling to come up with something that isn't an absurd run-on sentence.

Me too! 😭

firestack: > struggling to come up with something that isn't an absurd run-on sentence.

It's like the test framework is trying to get us to write two separate (basically identical) tests lol

assets/src/components/map.tsx
```diff
@@ -450,18 +524,23 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
...
-  const autoCenteringClass = shouldAutoCenter
-    ? "m-vehicle-map-state--auto-centering"
-    : ""
+  const stateClasses = [
+    "m-vehicle-map-state",
+    shouldAutoCenter ? "m-vehicle-map-state--auto-centering" : null,
+    streetViewEnabled ? "m-vehicle-map-state--street-view-enabled" : null,
+  ]
+    .filter((c) => c !== null)
+    .join(" ")
```
firestack, line 539: suggestion: I noticed that this code exists already in `dom.ts#className`, can we use that instead?

firestack (approved):

PR #1835, opened 2022-12-08
fix(skate-web/channels): ensure token is authenticated when joining channels

firestack:
# Summary
When initiating a connection to the server the token is validated, but when joining a channel the socket is not checked to be authenticated.

[Phoenix.Channel Documentation on Authorization](https://hexdocs.pm/phoenix/Phoenix.Channel.html#module-authorization) says that 
> Your channels must implement a join/3 callback that authorizes the socket for the given topic.

So this PR modifies all channel `join/3` callbacks to instead be `join_authenticated/3` which is wrapped by `SkateWeb.AuthenticatedChannel` (name and file location not final) so that authentication code does not need to be manually written and managed for new topics, and should make implementing new channels safer.

## Implementation Details
Use of `SkateWeb.AuthenticatedChannel` shims a `Phoenix.Channel` `@callback join/3` into the module, which firsts checks if the socket is authenticated, which delegates to `join_authenticated/3` if the socket token is valid; and returns `{:error, :not_authenticated}` if the token is invalid.
`join_authenticated/3` is defined as a callback matching the `Phoenix.Channel.join/3` spec so that calls made to `join/3` can be transparently passed to `join_authenticated/3`.

## Notes
I've decided to add a macro which inserts a `join/3` callback for `Phoenix.Channel` which wraps calls to `join_authenticated`, the idea being that if you're getting a callback at `join_authenticated` then you should be able to expect that the request has been authenticated already for you.

I'm unsure if this is the way we should do it or if there's a better way I could've approached this, so I'm really interested in feedback.

Adding `SkateWeb.AuthenticatedChannel` to other channels in SkateWeb showed that some tests are also returning false from `valid_token?/1` too soon, but this may need further review.

In addition, I also moved `:valid_token_fn` and `:valid_token?` into the `ChannelAuth` module and consolidated them to a single atom, which can be updated if we prefer a certain atom name.

I've also realized that I could possibly get rid of `AuthenticatedChannel` and instead do the same work within `lib/skate_web.ex`'s `__using__` macro.

# Todo
- [x] Every new thing needs docs
- [ ] Add frontend tests and "response"(refresh)
- [ ] add `AuthenticatedChannel` specific tests
- [ ] Find edge cases to add more elixir tests for
- [ ] (hopefully) find ways to make tests for `join_authenticated` automatically find all the topics that need to be tested.
- [x] Change elixir test names to be more descriptive 

---

Asana Ticket: https://app.asana.com/0/0/1203412139193187

### DIFF
--- a/lib/skate_web/channels/alerts_channel.ex
+++ b/lib/skate_web/channels/alerts_channel.ex
@@ -12,10 +12,7 @@ defmodule SkateWeb.AlertsChannel do
 
   @impl Phoenix.Channel
   def handle_info({:new_realtime_data, lookup_args}, socket) do
-    valid_token_fn =
-      Application.get_env(:skate, :valid_token_fn, &SkateWeb.ChannelAuth.valid_token?/1)
-
-    if valid_token_fn.(socket) do
+    if SkateWeb.ChannelAuth.valid_token?(socket) do
       data = Server.lookup(lookup_args)
       :ok = push(socket, "alerts", %{data: data})
       {:noreply, socket}
--- a/lib/skate_web/channels/alerts_channel.ex
+++ b/lib/skate_web/channels/alerts_channel.ex
@@ -4,8 +4,9 @@ defmodule SkateWeb.AlertsChannel do
   alias Realtime.Server
   alias Util.Duration
 
-  @impl Phoenix.Channel
-  def join("alerts:route:" <> route_id, _message, socket) do
+  use SkateWeb.AuthenticatedChannel
+  @impl SkateWeb.AuthenticatedChannel
+  def join_authenticated("alerts:route:" <> route_id, _message, socket) do
     alerts = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
     {:ok, %{data: alerts}, socket}
   end
--- a/lib/skate_web/channels/alerts_channel.ex
+++ b/lib/skate_web/channels/alerts_channel.ex
@@ -1,10 +1,9 @@
 defmodule SkateWeb.AlertsChannel do
   use SkateWeb, :channel
+  use SkateWeb.AuthenticatedChannel
 
   alias Realtime.Server
   alias Util.Duration
-
-  use SkateWeb.AuthenticatedChannel
 
   @impl SkateWeb.AuthenticatedChannel
   def join_authenticated("alerts:route:" <> route_id, _message, socket) do
--- a/lib/skate_web/channels/alerts_channel.ex
+++ b/lib/skate_web/channels/alerts_channel.ex
@@ -5,6 +5,7 @@ defmodule SkateWeb.AlertsChannel do
   alias Util.Duration
 
   use SkateWeb.AuthenticatedChannel
+
   @impl SkateWeb.AuthenticatedChannel
   def join_authenticated("alerts:route:" <> route_id, _message, socket) do
     alerts = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
--- a/lib/skate_web/channels/channel_auth.ex
+++ b/lib/skate_web/channels/channel_auth.ex
@@ -14,12 +14,6 @@ defmodule SkateWeb.ChannelAuth do
   defp valid_socket_token?(socket) do
     token = Guardian.Phoenix.Socket.current_token(socket)
 
-    case AuthManager.decode_and_verify(token) do
-      {:ok, _claims} ->
-        true
-
-      _ ->
-        false
-    end
+    match?({:ok, _claims}, AuthManager.decode_and_verify(token))
   end
 end
--- a/lib/skate_web/channels/channel_auth.ex
+++ b/lib/skate_web/channels/channel_auth.ex
@@ -6,6 +6,12 @@ defmodule SkateWeb.ChannelAuth do
   """
   @spec valid_token?(Phoenix.Socket.t()) :: boolean()
   def valid_token?(socket) do
+    token_fn = Application.get_env(:skate, :valid_token_fn, &_valid_token?/1)
+    token_fn.(socket)
+  end
+
+  @spec _valid_token?(Phoenix.Socket.t()) :: boolean()
+  defp _valid_token?(socket) do
     token = Guardian.Phoenix.Socket.current_token(socket)
 
     case AuthManager.decode_and_verify(token) do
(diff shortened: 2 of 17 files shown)


lib/skate_web/channels/alerts_channel.ex
```diff
@@ -4,18 +4,16 @@ defmodule SkateWeb.AlertsChannel do
...
-  def join("alerts:route:" <> route_id, _message, socket) do
+  use SkateWeb.AuthenticatedChannel
+  @impl SkateWeb.AuthenticatedChannel
+  def join_authenticated("alerts:route:" <> route_id, _message, socket) do
     alerts = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
     {:ok, %{data: alerts}, socket}
   end
 
   @impl Phoenix.Channel
   def handle_info({:new_realtime_data, lookup_args}, socket) do
```
lemald, line 15: **thought:** I wonder if we could similarly wrap `handle_info` and also get rid of some code duplication there...

lib/skate_web/channels/channel_auth.ex
```diff
@@ -6,6 +6,12 @@ defmodule SkateWeb.ChannelAuth do
   """
   @spec valid_token?(Phoenix.Socket.t()) :: boolean()
   def valid_token?(socket) do
+    token_fn = Application.get_env(:skate, :valid_token_fn, &_valid_token?/1)
+    token_fn.(socket)
+  end
+
+  @spec _valid_token?(Phoenix.Socket.t()) :: boolean()
+  defp _valid_token?(socket) do
```
lemald, line 14: **nitpick:** I know in Elixir, prefixing a variable with `_` causes the compiler to not throw a warning if it isn't used. I'm trying to make up my mind as to whether using it in this context is confusing or not, but I also don't have a better idea. Maybe `default_valid_token?` or something?

lib/skate_web/channels/notifications_channel.ex
```diff
@@ -15,8 +12,10 @@ defmodule SkateWeb.NotificationsChannel do
     end
   end
 
-  @impl true
-  def join("notifications", _message, socket) do
+  use SkateWeb.AuthenticatedChannel
```
lemald, line 15: **nitpick:** I guess this isn't a strict requirement since it passes CI, but I think it would be easier to read this file if all of the `use` / `import` / `alias` statements are grouped together up top.

lib/skate_web/channels/channel_auth.ex
firestack: Oh right, I don't wanna keep the leading underscore, I'll think on it!

test/skate_web/channels/alerts_channel_test.exs
```diff
@@ -24,6 +24,18 @@ defmodule SkateWeb.AlertsChannelTest do
...
+      reassign_env(:skate, :valid_token_fn, fn _socket -> false end)
+
+      for %{result: result, route: route} <- [
+            %{result: {:error, :not_authenticated}, route: "alerts:route:"},
+            %{result: {:error, :not_authenticated}, route: "vehicles:route:"},
+            %{result: {:error, :not_authenticated}, route: "random:topic:1"},
+            %{result: {:error, :not_authenticated}, route: "random:topic:2"}
+          ],
+          do: assert(^result = subscribe_and_join(socket, AlertsChannel, route))
+    end
```
lemald, line 38: **suggestion:** it's a little confusing to have this specifically in the `AlertsChannel` test since it tests several different channels. One option is to simply make this a bunch of different tests in different channel modules, or maybe there's some good unified place to put this one test that makes sense.

lib/skate_web/channels/channel_auth.ex
firestack: @lemald what do you think of `valid_socket_token?`?

test/skate_web/channels/alerts_channel_test.exs
firestack: Oh that reminds me I need to go over and change all the test names, since I just copy pasted.

lib/skate_web/channels/channel_auth.ex
lemald: Sure, that works.

test/skate_web/channels/alerts_channel_test.exs
firestack: also this is only testing `AlertsChannel`, I was leaving random topics in there when testing to ensure I was hitting strange edge cases should they appear.

lemald:
@firestack looks like there are a few things with the tests that you're still working on, but overall I like the direction this is headed!

lib/skate_web/channels/channel_auth.ex
firestack: interesting note
https://hexdocs.pm/elixir/writing-documentation.html#hiding-internal-modules-and-functions
> Start the function name with one or two underscores, for example, `__add__/2`. Functions starting with underscore are automatically treated as hidden, although you can also be explicit and add `@doc false`. The compiler does not import functions with leading underscores and they hint to anyone reading the code of their intended private usage.

lemald: Interesting - usually we just use `defp` for private functions and leave it at that, but it's good to know that at least some parts of the Elixir documentation recommend the underscore approach.

firestack (commented):
I added some ideas I had that felt like bikeshedding, I left them as review comments, let me know if you like them and I'll integrate them.

I still haven't figured out how to let the frontend know that the channel validity has expired, if @mbta/skate-developers have any ideas on how to implement/handle that let me know, I'm not even sure how to test this all the way through the frontend to the backend.

I think the tests I added are a bit to repetitiive, and they might benifit from being consolodated into a `AuthenticatedChannel.test.exs` instead of `describe("join/3")` since we're just enumerating topics. But I wanted the tests to stay local so they're easy to find. Any feedback welcome.

I had an idea that maybe we could scrape the `@spec`s and generate tests from the topics but that's probably it's own ticket.

lib/skate_web/channels/channel_auth.ex
```diff
@@ -6,14 +6,14 @@ defmodule SkateWeb.ChannelAuth do
...
-      {:ok, _claims} ->
-        true
+  @spec valid_socket_token?(Phoenix.Socket.t()) :: boolean()
+  defp valid_socket_token?(socket) do
+    token = Guardian.Phoenix.Socket.current_token(socket)
 
-      _ ->
-        false
-    end
+    match?({:ok, _claims}, AuthManager.decode_and_verify(token))
```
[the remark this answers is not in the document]
lemald: I always like rewriting things as pipelines so I'd say to go ahead and integrate this change. 👍

firestack:
I'm gonna merge this and see if we get any logs on dev.

PR #1843, opened 2022-12-14
fix: Resolve mid-route swing => deadhead issues in minischedule & swing view

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1203278600620513/f

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -392,14 +392,12 @@ const Piece = ({
           const previousTrip: Trip | AsDirected | null =
             piece.trips[tripIndex - 1] || piece.startMidRoute?.trip
 
-          // KB: Was struggling to use getNonRevenueStatus at this level -
-          // I think we'd want to call it at this level since one level lower, <Trip/> doesn't know its positioning relative to the other trips.
-          // However, not every trip is a nonRevenue trip, so I'm not sure it makes sense to change the sequence field to nonRevenueStatus field on Trip. Unless maybe we expand it to
-          // include revenue trip types too?
-          const sequence = getSequence(
-            startsWithMidRouteTrip ? tripIndex + 1 : tripIndex,
-            startsWithMidRouteTrip ? piece.trips.length + 1 : piece.trips.length
-          )
+          const sequence = startsWithMidRouteTrip
+            ? // tripIndex and trips.length don't include the startWithMidRoute trip. Add one to
+              // account for the startsWIthMidRouteTrip as the actual first trip of the piece.
+              getSequence(tripIndex + 1, piece.trips.length + 1)
+            : getSequence(tripIndex, piece.trips.length)
+
           return (
             <Trip
               trip={trip}
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -323,10 +323,13 @@ const Piece = ({
   pieceIndex: number
   activeIndex: [number, number] | null
 }) => {
+  const startsWithMidRouteTrip = piece.startMidRoute ? true : false
   const isSwingOn: boolean =
     piece.trips.length > 0 &&
     isTrip(piece.trips[0]) &&
-    !isDeadhead(piece.trips[0])
+    !isDeadhead(
+      startsWithMidRouteTrip ? piece.startMidRoute!.trip : piece.trips[0]
+    )
   const isSwingOff: boolean =
     piece.trips.length > 0 &&
     isTrip(piece.trips[piece.trips.length - 1]) &&
@@ -396,7 +399,11 @@ const Piece = ({
                   ? previousTrip.endTime
                   : null
               }
-              sequence={getSequence(tripIndex, piece.trips)}
+              sequence={getSequence(
+                tripIndex,
+                piece.trips.length,
+                startsWithMidRouteTrip
+              )}
               tripTimeBasedStyle={tripTimeBasedStyle}
               vehicleOrGhost={vehicleOrGhost}
               view={view}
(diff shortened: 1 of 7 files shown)


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -461,16 +471,30 @@ const MidRouteSwingOnSecondHalf = ({
   />
 )
 
-const getSequence = (tripIndex: number, pieceTrips: (Trip | AsDirected)[]) => {
+type Sequence = "first" | "middle" | "last"
+type NonRevenueStatus = "pull-out" | "pull-back" | "deadhead"
```
lemald, line 475: **praise:** I like splitting these out into separate types even if it's a simple mapping, it makes things a bit clearer.

assets/tests/components/propertiesPanel/minischedule.test.tsx
```diff
@@ -178,6 +178,33 @@ const midRouteSwingPiece2: Piece = {
...
+    },
+  ],
+  endTime: 480,
+  endPlace: "terminal1",
+  startMidRoute: {
+    time: 180,
+    trip: midRouteSwingTrip1,
+  },
+  endMidRoute: false,
+}
```
lemald, line 206: **suggestion:** consider making a factory for this - we use Fishery and there are some existing factories located in `assets/tests/factories`.

lib/schedule/swing.ex
```diff
@@ -116,4 +116,28 @@ defmodule Schedule.Swing do
...
+      |> Enum.take(1)
+      |> List.first()
+
+    case {first_full_trip, piece.start_mid_route?} do
+      {nil, %{trip: mid_route_trip}} ->
+        mid_route_trip |> trip_or_trip_id_to_trip(trips_by_id)
+
+      {_, %{trip: mid_route_trip}} ->
+        with %{} = first_trip <- mid_route_trip |> trip_or_trip_id_to_trip(trips_by_id) do
+          %{first_trip | :run_id => first_full_trip.run_id}
```
lemald, line 136: You can use the slightly more compact map syntax when the key is an atom.

```suggestion
          %{first_trip | run_id: first_full_trip.run_id}
```

lib/schedule/swing.ex
```diff
@@ -116,4 +116,28 @@ defmodule Schedule.Swing do
...
+    first_full_trip =
+      piece.trips
+      |> Stream.map(&trip_or_trip_id_to_trip(&1, trips_by_id))
+      |> Stream.filter(fn trip -> !is_nil(trip) end)
+      |> Enum.take(1)
+      |> List.first()
+
+    case {first_full_trip, piece.start_mid_route?} do
+      {nil, %{trip: mid_route_trip}} ->
+        mid_route_trip |> trip_or_trip_id_to_trip(trips_by_id)
```
lemald, line 132: **question:** is this one of the data edge cases you were referring to? If I understand correctly, this is the case in which the piece starts with a swing but has no other trips. I'd say to make the same update you do in the other case to make sure the `run_id` is correct juuuust in case this ever comes up, but I doubt it's likely to ever happen.

KaylaBrady: Yeah this is that edge case! In this case though, what would the correct `run_id` be? Since there are no other full trips, would it be just the `run_id` of the swing?

lemald: I was thinking you could use the `run_id` from the `piece`, provided that that's actually populated and correct.

KaylaBrady: Ah! Do you have a sense of how reliable that field is? Wondering if that could be used for the `run_id` so we no longer have to look for `first_full_trip` to get the run_id when there is a mid route swing-on

test/schedule/swing_test.exs
```diff
@@ -333,41 +322,141 @@ defmodule Schedule.SwingTest do
...
+                   to_route_id: ^route_id_during_swing,
+                   to_trip_id: ^trip_id_during_swing,
+                   to_run_id: ^run_id_after_swing
                  }
                ]
-             }
+             } = Swing.from_blocks(blocks, trips_by_id)
+    end
+
+    @tag :skip
```
lemald, line 363: **question:** is this intentional? Looks like it may have inadvertently been left in.

lib/schedule/swing.ex
lemald: I think it should be pretty reliable. If I'm not mistaken that value would get populated [here](https://github.com/mbta/skate/blob/a837fa82f60652464deeae640c8152b04cf6bdff/lib/schedule/hastus/activity.ex#L135), and it appears that it's ultimately using the value of `run_id` from the relevant row in `activities.csv`.

test/schedule/swing_test.exs
firestack: that test is failing due to the data edge case, but with your other comments we're able to enable it.

assets/tests/components/propertiesPanel/minischedule.test.tsx
firestack: Experimenting/learning fishery here https://github.com/mbta/skate/tree/kb-kf-minischedule-fix-/fishery-factories

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -323,10 +323,13 @@ const Piece = ({
...
   activeIndex: [number, number] | null
 }) => {
+  const startsWithMidRouteTrip = piece.startMidRoute ? true : false
   const isSwingOn: boolean =
     piece.trips.length > 0 &&
     isTrip(piece.trips[0]) &&
-    !isDeadhead(piece.trips[0])
+    !isDeadhead(
+      startsWithMidRouteTrip ? piece.startMidRoute!.trip : piece.trips[0]
+    )
```
firestack, line 332: this is causing a warning about 
```
  331:32  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```
So I'm considering reverting the change and going back to `piece.startMidRoute ? <a> : <b>`. 

That okay with you @KaylaBrady ?

lemald (approved):
Nice to see a bunch of new Fishery factories added!

assets/src/components/propertiesPanel/minischedule.tsx
firestack: addressed in [mbta/skate@`52985e1` (#1843)](https://github.com/mbta/skate/pull/1843/commits/52985e175b13f54c74c74b33fb1bd7f3edd2e87b)

PR #1846, opened 2022-12-15
refactor(Data.parse_files/2): Only filter to bus data when needed to support accessing subway data

KaylaBrady:
Refactor to support https://app.asana.com/0/1200180014510248/1203520428951853/f.

Currently, for many gtfs files, non-bus data is immediately filtered during the parsing process. This PR pushes that filtering to a later stage for the files we'll need in order to calculate all route connections for bus stops: 
* route_patterns
* trips

I attempted to keep this refactor as small as possible to reduce the risk of unintended regressions for this core flow. 
Added:
There may be additional refactoring in `parse_files` to come, but this is the core chunk of it. If it would be preferable to review this refactor in the context of all changes to expose stop connection data, I can close this PR.

### DIFF
--- a/lib/schedule/gtfs/route.ex
+++ b/lib/schedule/gtfs/route.ex
@@ -11,6 +11,8 @@ defmodule Schedule.Gtfs.Route do
           garages: MapSet.t(Garage.id())
         }
 
+  @type route_type :: number()
+
   @type direction_names :: %{
           0 => String.t(),
           1 => String.t()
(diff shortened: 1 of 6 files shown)


lib/schedule/gtfs/route.ex
```diff
@@ -61,26 +75,25 @@ defmodule Schedule.Gtfs.Route do
...
+    true
   end
 
-  @spec bus_route_mbta?(Csv.row()) :: boolean
-  defp bus_route_mbta?(row) do
-    # Verify that route number is not one of the private carriers: 710, 712, 713, 714, 716
-    row["route_id"] not in ["710", "712", "713", "714", "716"]
+  @spec bus_route?(t()) :: boolean
+  def bus_route?(route) do
+    route.type == 3
```
lemald, line 90: **nitpick:** we currently classify the SL1/2/3 as route-type 3 (regular bus), but since we do still operate those routes that run at least partially as trolleybuses we should also include `route_type` 11 in case something about our GTFS data ever changes.

KaylaBrady: I'm inclined to keep this as 3 to match the previous row-level [bus filtering logic here](https://github.com/mbta/skate/pull/1846/files#diff-d659be3f7889c470bc13601bacd56a935912ad2857bc9996bdf60a168ec3c3b9L71) in order to keep this PR a pure refactor with no functionality change. Does that sound alright?

lemald: Oh, that's so strange, I wonder why that didn't break with the 71/73 back when they were actually trolleybuses.

PR #1872, opened 2022-12-27
feat(Map): Icons for stations

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203442408293386/f
Using data exposed in https://github.com/mbta/skate/pull/1871, adds icons to the map for all stations.


assets/src/components/map.tsx
```diff
@@ -597,9 +682,24 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
...
         {(props.shapes || []).map((shape) => (
-          <Shape key={shape.id} shape={shape} />
+          <RouteShape key={shape.id} shape={shape} />
         ))}
+        {
+          <RouteStops
+            stops={(props.shapes || []).flatMap((shape) => shape.stops || [])}
+            iconSize={stationIconSize}
+          />
+        }
```
lemald, line 694: **question:** does this need to be wrapped in curly braces if it's just a plain JSX element?

assets/tests/components/map.test.tsx
```diff
@@ -194,6 +203,75 @@ describe("map", () => {
...
+    const { container } = render(
+      <Map vehicles={[vehicle]} reactLeafletRef={mapRef} stations={[station]} />
+    )
+
+    // Manual zoom
+    act(() => {
+      mapRef.current!.setZoom(15)
+    })
+    await animationFramePromise()
+    expect(container.innerHTML).toContain("m-station-icon")
```
lemald, line 250: **nitpick:** you can probably drop this assertion from this test since it's handled by the previous test case.

assets/tests/components/map.test.tsx
```diff
@@ -194,6 +203,75 @@ describe("map", () => {
...
```
firestack, line 234: suggestion: would this work instead?
```suggestion
    expect(
      container.querySelector(".m-station-icon")
    ).toBeVisible()
```

assets/tests/hooks/useStations.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const { result } = renderHook(() => {
+      return useStations()
+    })
+
+    expect(fetchStations).toHaveBeenCalledTimes(1)
+    expect(result.current).toEqual(stations)
+  })
+
+  test("returns null when data not fetched ", () => {
```
lemald, line 35: **nitpick:** trailing space in test name

lemald:
**question:** since you're updating the various `stops[Line].tsx` files, is it possible to get those to somehow get the shuttle map to use the API endpoint instead? You'd need to somehow provide the lines associated with each station so that it would know which subset of the stations to show for each selected line. For that matter, how should the existing station icons on the shuttle map play along with these new icons (probably more of a design question)?

assets/src/components/map.tsx
```diff
@@ -217,41 +223,108 @@ export const strokeOptions = ({ color }: Shape): object =>
...
+    point.lon,
+  ])
+  return (
+    <Polyline
+      className="m-vehicle-map__route-shape"
+      positions={positions}
+      {...strokeOptions(shape)}
+    />
+  )
+})
```
lemald, line 326: **praise:** I like separating out the shape and stops components, makes things a little easier to read in my opinion.

assets/src/components/map.tsx
```diff
@@ -217,41 +223,108 @@ export const strokeOptions = ({ color }: Shape): object =>
...
+              longitude={stop.lon}
+            ></StreetViewButton>
+          )}
+        </Popup>
+      </CircleMarker>
+    )
+  }
+)
+
+const StationMarker = ({
```
lemald, line 258: **suggestion (non-blocking):** since `map.tsx` is getting kind of big, I wonder at what point we should split components like this out into other files. For instance we have separate `vehicleCard.tsx` and (soon) `stopCard.tsx` files, though it could be argued that those are larger components better suited for being separated out.

assets/src/components/map.tsx
```diff
@@ -217,41 +223,108 @@ export const strokeOptions = ({ color }: Shape): object =>
...
+  ({
+    stop,
+    iconSize,
+  }: {
+    stop: Stop
+    iconSize: StationIconSize
+  }): JSX.Element => {
+    if (stop.locationType === LocationType.Station) {
+      return <StationMarker station={stop} iconSize={iconSize} />
+    }
```
lemald, line 236: **suggestion:** I think it might be a little easier to follow the logic if there were an umbrella component (`StopOrStationMarker`?) with logic that decides whether to render `StopMarker` or `StationMarker` rather than having that logic here.

assets/src/components/map.tsx
KaylaBrady: Yeah I like that idea! Maybe as a starting point I'll move out all the markers to a single `mapMarkers` file.

KaylaBrady:
Note to self: Once https://github.com/mbta/skate/pull/1867 is merged, follow through on VPP stop icon formatting based on [this comment](https://github.com/mbta/skate/pull/1867#discussion_r1059098087)

assets/src/components/mapMarkers.tsx
```diff
@@ -0,0 +1,326 @@
...
+    iconSize,
+  }: {
+    stop: Stop
+    iconSize: StationIconSize
+  }): JSX.Element => {
+    if (stop.locationType === LocationType.Station) {
+      return <StationMarker station={stop} iconSize={iconSize} />
+    }
+
+    return <StopMarker stop={stop} />
```
lemald, line 198: **nitpick:** might be slightly clearer if rewritten in `if (...) { ... } else { ... }` format.

assets/src/components/mapMarkers.tsx
```diff
@@ -0,0 +1,326 @@
...
+        opacity: 1.0,
+        weight: 4,
+      }
+    : {
+        color: "#4db6ac",
+        opacity: 0.6,
+        weight: 6,
+      }
+
+const StopMarker = ({ stop }: { stop: Stop }) => (
```
lemald, line 169: **nitpick:** it probably makes more sense for this to go after `StopOrStationMarker` but before `StationMarker`.

assets/src/components/mapMarkers.tsx
```diff
@@ -0,0 +1,311 @@
...
+})
+
+const stationLeafletIcon = ({ size }: { size: number }): Leaflet.DivIcon => {
+  return Leaflet.divIcon({
+    html: stationIcon,
+    className: "m-station-icon",
+    iconSize: [size, size],
+    iconAnchor: [size / 2, size / 2],
+  })
+}
```
lemald, line 270: **nitpick (non-blocking):** it might be better to order this with the other stop / station icon stuff rather than in between garage-related components.

assets/src/components/mapMarkers.tsx
```diff
@@ -0,0 +1,311 @@
...
+}) => {
+  const position: LatLngExpression = [
+    trainVehicle.latitude,
+    trainVehicle.longitude,
+  ]
+  const icon: Leaflet.DivIcon = makeTrainVehicleIcon(trainVehicle)
+  return <Marker position={position} icon={icon} />
+}
+
+export const strokeOptions = ({ color }: Shape): object =>
```
lemald, line 156: **nitpick (non-blocking):** I know this code was just moved by this PR, not introduced, but `shapeStrokeOptions` may be a slightly clearer name.

lemald (approved):
A couple of final nitpick comments but looks good overall!

assets/tests/components/mapMarkers.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+          runId: "101",
+        })}
+        isPrimary={true}
+      />
+    )
+    expect(container.querySelector(".m-vehicle-map__icon")).toBeInTheDocument()
+    expect(screen.getByText("101")).toBeInTheDocument()
+  })
+})
+describe("TrainVehicleMarker", () => {
```
lemald, line 47: **nitpick:** missing blank line between this and previous `describe` block.

assets/src/hooks/useDeviceSupportsHover.ts
```diff
@@ -0,0 +1,7 @@
+import { useMediaQuery } from "@react-hook/media-query"
```
KaylaBrady, line 1: I ended up using the media query hook instead of a purely css solution for 2 reasons:
1. I was running into the pesky "Invalid LatLng" testing user click events for markers when they had both a tooltip and popup, one of which would have been hidden by css. Having only one popup or one tooltip per marker seemed to resolve this issue.
2. It was easier for testing to be able to mock `useDeviceSupportsHover`

assets/src/hooks/useDeviceSupportsHover.ts
```diff
@@ -0,0 +1,7 @@
...
+
+const useDeviceSupportsHover = (): boolean => {
+  return !useMediaQuery("(hover: none)")
```
lemald, line 4: **question(non-blocking):** is there a particular reason to do this instead of `return useMediaQuery("(hover: hover)")`?

lemald (approved):
One small question and one minor nitpick from earlier outstanding, but overall looks good!

assets/tests/api.test.ts
```diff
@@ -438,6 +453,60 @@ describe("fetchShuttleRoutes", () => {
...
+        },
+      ])
+      done()
+    })
+  })
+
+  test("returns empty list on error", (done) => {
+    mockFetch(500, {
+      data: null,
+    })
```
firestack, line 501: question(non-blocking): could these be done with fishery factories?

assets/tests/components/mapMarkers.test.tsx
```diff
@@ -0,0 +1,198 @@
...
+    <MapContainer
+      center={[0, 0]}
+      zoom={options?.zoom || 13}
+      maxBounds={[
+        [41.2, -72],
+        [43, -69.8],
+      ]}
+    >
+      {component}
+    </MapContainer>
```
firestack, line 197: praise: not highlighting the entire file but, all these tests are great!

assets/tests/components/mapPage.test.tsx
```diff
@@ -90,6 +97,31 @@ describe("MapPage", () => {
...
+        <BrowserRouter>
+          <MapPage />
+        </BrowserRouter>
+      </StateDispatchProvider>
+    )
+
+    await userEvent.click(screen.getByRole("button", { name: "Zoom in" }))
+    await userEvent.click(screen.getByRole("button", { name: "Zoom in" }))
+
+    expect(container.innerHTML).toContain("m-station-icon")
```
firestack, line 122: suggestion: how do you feel about using this instead?

```suggestion
    expect(container.querySelector(".m-station-icon")).toBeVisible()
```

assets/src/data/stopsRed.ts
```diff
@@ -1,33 +1,161 @@
...
+    lon: -71.103802,
+    locationType: LocationType.Station,
+  },
+  {
+    id: "place-davis",
+    name: "Davis",
+    lat: 42.39674,
+    lon: -71.121815,
+    locationType: LocationType.Station,
+  },
```
firestack, line 60: thought(non-blocking): I'm not sure this is a good idea, but while working on the vpc I started wondering if we could share our fishery factories with the frontend code.

I'm curious what @lemald thinks of that as well

what I was seeing was that the `place-` prefix and the `locationType` both seem like things that'd do well being specified as higher order functions around a base constructor.

and then also leave off a `id` if it was okay to just `toLower()` the `name` 🤣

assets/src/components/map.tsx
```diff
@@ -323,7 +127,7 @@ const StreetViewControl = ({
   return portalElement ? ReactDOM.createPortal(control, portalElement) : null
 }
 
-class RecenterControl extends Control {
+class LeafletRecenterControl extends Control {
```
firestack, line 130: question(non-blocking): why was `Leaflet` added here?`

firestack (approved):
🎉

assets/src/data/stopsRed.ts
KaylaBrady: An interesting thought! FWIW this data is hardcoding what would be returned from the API. For example, `place-davis` is the id of the Davis square station in GTFS (https://www.mbta.com/stops/place-davis is the URL for davis stop information). These hardcoded stop + shape files in Skate will be removed as part of [this ticket ](https://app.asana.com/0/1152340551558956/1203629824340404). I'd be interested to hear more about the other cases you've come across in your VPC work though!

KaylaBrady:
@mbta/skate-developers  [83789c4](https://github.com/mbta/skate/pull/1872/commits/83789c4b06e5a0523d8c002d8bce9a3be49b1803) to undo the map beta group gating - would appreciate one last look at that before I go ahead and merge!

lemald (approved):
Looks good to me!

firestack (approved):

PR #1831, opened 2022-12-07
feat: reinstall fullstory

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203385011755356/f

This largely reverts changes from https://github.com/mbta/skate/pull/1624, with a few adjustments
* username now pulled from the existing meta tag rather than re-adding to window
* helper `fullStoryIdentify` in separate file
* Adds `fs-mask` classes to sensitive data

PR #1873, opened 2022-12-30
feat: map vehicle properties card + mapPage vehicle selection

firestack:
## Summary
I've gotten most of this implemented, there's a few things left to do but I'm looking for any technical review that I'm going in the wrong direction.

## Todo
- [ ] Style for Vehicle properties card
- [ ] Rename css classes
- [ ] ensure tests are correctly specified
- [ ] finish cleanup of new code

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1203502925530262/f


assets/css/_map_page.scss
```diff
@@ -87,12 +87,23 @@
   }
 }
 
-.m-map-page__map .m-vehicle-card {
+.m-map-page__map .m-vpc {
```
firestack, line 90: to be renamed/expanded

assets/src/components/routeVariantName.tsx
```diff
@@ -22,3 +22,26 @@ export const RouteVariantName = ({ vehicle }: { vehicle: VehicleOrGhost }) => {
...
+  const route = useRoute(vehicle.routeId)
+  if (isVehicle(vehicle) && vehicle.isShuttle) {
+    return <div className="m-route-variant-name">Shuttle</div>
+  }
+
+  const { routeId, viaVariant, headsign } = vehicle
+  const viaVariantFormatted = viaVariant && viaVariant !== "_" ? viaVariant : ""
+
+  return (
+    <output title="Route Variant Name" className="m-route-variant-name">
```
lemald, line 36: **question:** what's the rationale for using `output` here? I'm not the most familiar with it but from looking at the MDN docs I'm not really sure I see how it fits.

firestack: `div` and `span` both are not semantically specific, where `output` is specifically for programmatic output from the web application, although I'm using `output` outside of html forms, I thought this was the most suitable element for conveying that data is outputted by the application, rather than static. 
As well as it made writing tests nicer because I could do `getByRole("status"`

While react does manage and efficiently update the DOM by diffing (although a implementation detail), I've been taking the approach I feel is cleaner and produces a better document by scoping changing or application produced values to an `output` element and leaving static text as apart of the document structure itself.

firestack: Open to changing it, but I'm looking for a semantic element for the purpose of testing + accessibility that conveys properly that this is the output/rendering of data.

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -0,0 +1,291 @@
...
+const VisualSeparator = (): React.ReactElement => (
+  // Visual accent to provide separation between elements
+  // This object is strictly for visual presentation
+  <object
+    className="m-vehicle-route-summary__separator"
+    role="separator"
+    aria-hidden={true}
+    aria-label="presentation separator"
+  />
+)
```
lemald, line 89: **praise:** I really like having this stuff separated out into smaller components, it definitely helps a bit with the readability of some otherwise-complex code.

assets/src/components/routeVariantName.tsx
lemald: Okay, makes sense! I think I have a tendency to interpret the meaning of various HTML elements a bit overly restrictively and reach for a `div` because something seems to be not _quite_ right about the more specialized elements I consider using.

lemald:
@firestack okay, I've completed an initial pass of review. I think I'll have an easier time digesting the whole thing once some of the code is a bit more cleaned up / finalized, but I don't see any major high-level issues so far. I'll keep thinking it over and add any comments that come to me.

assets/src/components/routeVariantName.tsx
firestack: agreed!
On a similar note, I wanted to use a `<dl></dl>` instead of a table for key value pairs, but apparently even though `<dl>` is _the_ element for key value pairs essentially, `<table>` support is better on screen readers than for `<dl>`

(though, I don't know how many screen readers we're supporting if any)

assets/css/_ui_kit.scss
```diff
@@ -116,6 +116,22 @@ $color-vehicle-ghost-border: $color-component-medium;
 $color-vehicle-nonrevenue: #8f7ed6;
 $color-vehicle-off-course: $color-component-dark;
 
+:root {
+  --color-vehicle-off-course: #{$color-vehicle-off-course};
+  --color-vehicle-ontime: #{$color-vehicle-ontime};
+}
+
+:root,
```
lemald, line 124: **question:** Love this change! One quick thing: why does this get applied to `:root`? Just to make it the default if nothing else is selected?

KaylaBrady (commented):
Very cool to see the VPC coming alive! Admittedly I had a hard time wrapping my head around the larger ui kit/svg changes and wonder if it might be helpful to split those out into a different branch for two smaller diffs, but I wouldn't block approval over it!

assets/css/_ui_kit.scss
firestack: I used `:root: based on observed common CSS practice, and that `:root` makes it available to the entire application. 
I considered adding it to the skate `<main>` tag, but I felt that it would be confusing if this "global" style behaviour changed based on where it was placed (we don't normally place things outside of `<main>` but just in case).

Also `:root` I believe is an alias for `html`, but is more clear what you're targeting in my opinion.

assets/css/app.scss
```diff
@@ -102,6 +105,20 @@ $component-active-bg: $color-eggplant-700;
...
+html {
+  box-sizing: border-box;
+}
+
+*:before,
+*:after,
+* {
+  box-sizing: inherit;
+}
+
```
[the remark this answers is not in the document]
firestack: I (perosnally) think the box model is easier to reason about with `box-sizing: border-box;` and I think we'd benifit applicaiton wide from using this instead of `content-box` by default. 
You're correct that it does break some things, and I had to patch those to make them continue to work as they used to (vehicle properties panel is an example of this), but I might not have caught everything :/.
I'd be open to moving it into it's own PR/Ticket for 'modernizing' our CSS.

This comes from things like css normalize and css reset type projects([css remedy](https://github.com/jensimmons/cssremedy) being one of my prefered favorites last I checked), which aim to make CSS/HTML styles respond more like people's internal models expect, as well as create a strong baseline for having all browsers render in the same style.

There's a bit of background on this on css-tricks
- https://css-tricks.com/box-sizing/
- https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/

assets/tests/components/map.test.tsx
```diff
@@ -106,14 +51,16 @@ afterAll(() => {
   global.scrollTo = originalScrollTo
 })
 
-describe("map", () => {
+describe("<Map />", () => {
```
firestack, line 54: I'm considering reverting this and the describe in vehiclePropertiesCard, given it doesn't match style of the rest of the application tests.

assets/src/components/mapPage.tsx
```diff
@@ -129,9 +139,12 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
           includeStopCard={true}
         >
           {liveVehicle && showVehicleCard ? (
-            <VehicleCard
+            <VehiclePropertiesCard
               vehicle={liveVehicle}
-              onClose={() => setShowVehicleCard(false)}
+              onClose={() => {
+                dispatch(setSelectedVehicle(null))
+                setShowVehicleCard(false)
```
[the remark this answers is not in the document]
firestack: Thank you! gonna make sure there's a test for this.

assets/src/models/searchQuery.ts
```diff
@@ -1,6 +1,9 @@
+export type SearchQueryType = "all" | "run" | "vehicle" | "operator"
```
lemald, line 1: **praise:** hooray for better types!

assets/css/_ui_kit.scss
firestack: also, not something I did, but we should be able to move the `early-red`/`early-blue` styles into the `<html>` tag and then the styles will be application wide, I got lucky in not needing to refactor a bunch of code because we specify those classes at all points needed, but we should be able to move to using only the variables or just the `early`/`late` classes which pull the variables in themselves.

lemald: Okay, makes sense. And agreed that `:root` is clearer for this purpose than `html`.

assets/src/models/searchQuery.ts
firestack: Yeah, I had to add `| string` to it's definition within the state because otherwise every usage required a `"all" as SearchQueryType`.

assets/css/app.scss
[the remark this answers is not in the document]
lemald: I agree that `border-box` is an easier mental model to work with most of the time, and the more I learn about CSS the more I wish Skate had started out using one of the available CSS reset libraries. That said, I am a little concerned about how we'll catch all of the possible cases where this might break - VPP is an obvious one, it seems like the sort of thing that will need a lot of QA. I actually wonder if this is a potential use case for automated visual tests.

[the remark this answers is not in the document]
firestack: I'm starting to notice more places it broke things (swings view). It's relatively easy to figure out what breaks it by switching between `content-box` and `border-box` repeatedly in dev tools (though time consuming and not an exact science), and I like the idea of having a list of things to fix by having to annotate where we use `content-box` explicitly, but I think I'm gonna localize this style to just the mapSearch page for now, unless y'all can think of a better common denominator.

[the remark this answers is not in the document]
lemald: Unfortunately I can't think of a better common denominator, though if you do end up limiting the scope of the change in this PR definitely create a followup tech debt PR to make the change for the whole application.

assets/css/_skate_ui.scss
```diff
@@ -0,0 +1,115 @@
+// #region Text Styles [v2]
+
+// https://www.notion.so/mbta-downtown-crossing/2022-12-19-Type-Styles-6f6c90eb449248b7a954e234b718137f
```
lemald, line 3: **suggestion:** I think it would be worth running this by design to see if this matches their general approach to things. Overall I really like these changes, though.

lemald: Ideally we could eventually get rid of some of the type weight / size stuff in `_ui_kit.scss` as well.

assets/css/app.scss
```diff
@@ -68,24 +70,25 @@ $z-properties-panel-context: (
...
-@import "shuttle_picker";
 @import "shuttle_picker_container";
+@import "shuttle_picker";
 @import "stop_card";
 @import "street_view_button";
 @import "swings_view";
-@import "vehicle_card";
 @import "vehicle_icon";
 @import "vehicle_map";
+@import "vehicle_properties_card";
```
lemald, line 91: **praise:** not going to select the thole thing, but thank you for sorting this!

assets/css/_skate_ui.scss
firestack: I can't tell if you're referring to the implementation (separating out font-weight/size from color) or the document(which was made by design :D), but I'll definitely make sure I sync with design before merging.

assets/css/app.scss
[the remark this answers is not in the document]
KaylaBrady: TIL about "CSS resets"! Thanks for the resources on box model - I agree that it seems like an easier mental model to work with.

assets/css/_skate_ui.scss
lemald: 🤦 I totally didn't realize that design created that Notion document. In that case I'd say this particular change is good to go!

assets/src/components/routeVariantName.tsx
```diff
@@ -22,3 +22,43 @@ export const RouteVariantName = ({ vehicle }: { vehicle: VehicleOrGhost }) => {
...
+        <>
+          <output
+            title="Route & Variant"
+            className="m-route-variant-name__route-id"
+          >
+            {`${route?.name || routeId}_${viaVariantFormatted}`}
+          </output>
+          &nbsp;
+          <output
+            title="Route Headsign"
```
lemald, line 55: **nitpick:** I think this could be shortened to just "headsign" (I think of headsign as something associated with a particular vehicle or trip, not route).

firestack: fixed in a6320b2309dd91740959877f5481ec449e577528

assets/src/components/closeButton.tsx
```diff
@@ -42,4 +42,22 @@ const CloseButton: React.FC<Props> = ({ onClick, closeButtonType }) => {
...
+  closeButtonType,
+  ...rest
+}: Props & React.HTMLAttributes<HTMLElement>) => (
+  <button
+    title="Close"
+    onClick={onClick}
+    className={[
+      "m-close-button",
+      ...buttonTypeClassNames[closeButtonType],
+    ].join(" ")}
```
lemald, line 56: **suggestion:** I think before you used the `className` function here, it might have gotten lost in the rebase?

assets/tests/components/vehiclePropertiesCard.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+      expect(onClose).toHaveBeenCalled()
+    })
+
+    test("when rerendered, should display latest intersection", () => {
+      const intersection = "Massachusetts Ave @ 1"
+      const intersection2 = "Massachusetts Ave @ 2"
+      ;(useNearestIntersection as jest.Mock)
+        .mockReturnValueOnce(intersection)
+        // .mockReturnValueOnce(null)
```
lemald, line 35: **nitpick:** commented out line that seems to be unnecessary

assets/src/components/closeButton.tsx
firestack: I was getting really annoyed with the default name of the `className` function colliding with the prop/function parameters heh, so I dropped it for the same syntax as the other.
I'm fine to pull it back in under an alias.

lemald: Ah, makes sense. As long as it was intentional I'm fine keeping it that way. Maybe we could just find a better name for the `className` function in the future.

assets/tests/components/map.test.tsx
firestack: I'm reconsidering this given that it's helpful seeing when something being tested is a component, especially helpful for nested component testing.

assets/src/components/mapPage.tsx
```diff
@@ -60,14 +60,16 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
   const onlyVehicles: Vehicle[] = filterVehicles(vehicles)
   const [mobileDisplay, setMobileDisplay] = useState(MobileDisplay.List)
   const [selectedVehicleId, setSelectedVehicleId] = useState<VehicleId | null>(
-    null
+    searchPageState.selectedVehicleId ?? null
   )
 
   const liveVehicle: Vehicle | null = selectedVehicleId
     ? onlyVehicles.find((v) => v.id === selectedVehicleId) || null
```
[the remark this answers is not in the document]
firestack: I still haven't figured out how to address this, given all tests break when changing where the state is stored :/

assets/css/_vehicle_properties_card.scss
```diff
@@ -0,0 +1,121 @@
...
+  flex-direction: row-reverse;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.m-vehicle-properties-card__body {
+  display: flex;
+  flex-flow: column;
+  gap: 10px;
+}
```
firestack, line 45:
```suggestion
.m-vehicle-properties-card__body > :not(:first_child) {
	margin-top: 10px
}
```

Could be simplified

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
KaylaBrady: I have a thought about a possible way to resolve the test breaking issue by using a `StateDispatchProvider` that doesn't have a mock `dispatch` function. I still consider this a non-blocking thought though and it is something I've never tried before -- happy to talk more or pair on it for this PR if you want, or after this PR is merged.

assets/css/_vehicle_properties_card.scss
firestack: Really struggling between what design has specified in figma vs what's "better"?

I _feel_ like the box model is faster than flex in this context, but this is an assumption.

Design used flex when building in figma, so it feels natural to reach for flex as if it's a specification from design, but design also says they won't recommend technical solutions, so this is up to us in the end?

assets/tests/components/mapPage.test.tsx
```diff
@@ -1,72 +1,63 @@
...
 }))
 
 jest.mock("../../src/hooks/useVehicleForId", () => ({
   __esModule: true,
   default: jest.fn(),
 }))
 
+beforeEach(() => {
+  expect.hasAssertions()
+})
```
firestack, line 50: Oh I left this in from when building the tests, does anyone want it removed?

assets/src/components/map.tsx
```diff
@@ -534,6 +539,11 @@ const Garages = ({ zoomLevel }: { zoomLevel: number }) => (
 )
 
 const Map = (props: Props): ReactElement<HTMLDivElement> => {
+  const [
+    {
+      searchPageState: { selectedVehicleId },
+    },
+  ] = useContext(StateDispatchContext)
```
KaylaBrady, line 546: suggestion: This may be an issue on the mini-map, where the selected vehicle is stored as `state.selectedVehicleOrGhost?.id`

Consider passing the selectedVehicleId as a prop to the map so that it can be set on a case-by-case basis (`selectedVehicleId` for the mini map and old search map, `searchPageState.selectedVehicleId for the new map page)

assets/src/components/scheduleAdherence.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { ComponentPropsWithoutRef, useContext } from "react"
```
KaylaBrady, line 1: question (non-blocking): Apologies for not noticing this before! It seems that there is overlap between this component and what is in `propertiesPanel/header.tsx` - is there enough overlap that `header` can be updated to use this component for consistency?

assets/tests/components/map.test.tsx
```diff
@@ -106,14 +51,16 @@ afterAll(() => {
...
   test("draws vehicles", () => {
+    const vehicle = vehicleFactory.build({})
```
KaylaBrady, line 56: 🙌 yay for using the default vehicle!

assets/tests/components/mapPage.test.tsx
KaylaBrady: Oh cool! I'd love to see this in `assets/tests/setup.tsx` (if in fact all our existing tests have assertions and it doesn't create more work for you!)

assets/tests/components/mapPage.test.tsx
```diff
@@ -193,73 +211,241 @@ describe("MapPage", () => {
...
+
+      test("after vehicle on the map is clicked", async () => {
+        const vehicle = vehicleFactory.build()
+        jest.spyOn(global, "scrollTo").mockImplementationOnce(jest.fn())
+        ;(useVehicleForId as jest.Mock).mockReturnValueOnce(vehicle)
+        const runId = "clickMe"
+        ;(useSearchResults as jest.Mock).mockReturnValue(
+          vehicleFactory.buildList(1, { runId: runId })
+        )
+        // const searchResults: VehicleOrGhost[] = vehicleFactory.buildList(1, { runId: runId })
```
KaylaBrady, line 401: nitpick: Can this comment be removed?

assets/tests/factories/applicationState.ts
```diff
@@ -0,0 +1,35 @@
+import { Factory } from "fishery"
+import { OpenView, State } from "../../src/state"
+import { searchPageStateFactory } from "./searchPageState"
+import { userSettingsFactory } from "./userSettings"
+
+const stateFactory = Factory.define<State>(() => ({
```
KaylaBrady, line 6: glad to see this as a factory! Wonder if it can use the `initialState` from `../../src/state` so that if the initialState changes for some reason, the factory changes with it?

KaylaBrady (commented):
Looking good! Only blocking comment is about identifying the selected vehicle in the map component.

assets/tests/factories/applicationState.ts
firestack: I think I copied this from `initalState` and I didn't understand what `initalState` was when I made this, so yes!

assets/src/components/scheduleAdherence.tsx
firestack: Nevermind! when trying to do this it broke in a lot of ways(spacing, flex container, alignment), I think we're better off waiting for design to have us replace the header in some form.

assets/css/_vehicle_properties_card.scss
lemald: I'd reach out to design, but I think in general design's choices about the method of layout in Figma aren't super important, it's more a question of getting the ultimate visual output.

assets/tests/components/mapPage.test.tsx
```diff
@@ -1,72 +1,59 @@
...
 }))
 
+afterEach(() => {
+  jest.clearAllMocks()
+})
+
 describe("MapPage", () => {
   test("renders the empty state", () => {
-    ;(useSearchResults as jest.Mock).mockImplementationOnce(() => null)
+    ;(useSearchResults as jest.Mock).mockReturnValue([])
```
lemald, line 54: **suggestion:** I think there should be test cases for both `null` and `[]` because they're both values that `useSearchResults` can return. Apologies if there is still a test for `null` and I'm just not seeing it.

assets/src/components/vehicleRouteSummary.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { ScheduleAdherence } from "./scheduleAdherence"
+import { Size, VehicleIcon, vehicleOrientation } from "./vehicleIcon"
+// import { VehicleProp } from "./vehiclePropertiesCard";
+interface VehicleOrGhostProp {
+  vehicle: VehicleOrGhost
+}
+
+// #endregion
+// #region Vehicle Summary
+const VehicleIcon1 = ({
```
lemald, line 20: **question:** can this be renamed to something like `WrappedVehicleIcon` that's more descriptive than the `1` suffix?

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <TrNameValue name="vehicle">{vehicle.label ?? "N/A"}</TrNameValue>
+        <TrNameValue name="operator">
+          {[
+            vehicle.operatorFirstName,
+            vehicle.operatorLastName,
+            vehicle.operatorId ? `#${vehicle.operatorId}` : null,
+          ]
+            .filter((e) => e !== null)
+            .join(" ") || "N/A"}
+        </TrNameValue>
```
lemald, line 81: **todo:** make sure that Fullstory is set to mask this value (@KaylaBrady set that up and may know more about how to do this).

KaylaBrady: adding a `fs-mask` class should do the trick - that is what I used in https://github.com/mbta/skate/pull/1831

firestack: Thanks for catching this!

assets/src/components/vehicleRouteSummary.tsx
firestack: Let me know what you think of 2f0893bca125c73f272c90852a0779f3de1d4349 ?

lemald: I like it!

assets/src/components/vehiclePropertiesCard.tsx
firestack: Is there any documentation on this?

I'm trying to figure out if I should put the `fs-mask` on the table cell that contains the sensitive value or on the table row itself

lemald: @KaylaBrady correct me if I'm wrong, but I think it masks the element with the `fs-mask` class and all of its children, right? So either should be fine.

firestack:
@mbta/skate-developers does this look okay to you? 
Can you think of any other ways to mock promise pending states?
c606f1dc4f58097e0103033fca33a1b3f91b716c

assets/tests/components/vehiclePropertiesCard.test.tsx
```diff
@@ -157,6 +163,36 @@ describe("<VehiclePropertiesCard/>", () => {
...
+        resolveFn!(intersection)
+
+        waitFor(
+          () => {
+            expect(
+              screen.getByRole("status", { name: "Current Location" })
+            ).toHaveTextContent(intersection)
+          },
+          { timeout: 2 }
+        )
```
firestack, line 194: @mbta/skate-developers does c606f1dc4f58097e0103033fca33a1b3f91b716c look okay to you? 
Can you think of any other ways to mock promise pending states?

lemald: Could you mock `useNearestIntersection` instead, since that's what the component invokes directly? That might be simpler.

firestack: Oh I guess what I forgot was that I'm changing the API of `useNearestIntersection`, so I don't need to mock the promise I just need to mock what I expect it should return in each state 👍🏻 
I was digging myself into a hole trying to figure out how to mock intermediate states heh

Thanks Eddie!

firestack: [`688eef3` (#1873)](https://github.com/mbta/skate/pull/1873/commits/688eef3397594eae331dbf5ed3a3a3ff15bc8260)
and
[`b6fad54` (#1873)](https://github.com/mbta/skate/pull/1873/commits/b6fad54cd7da76701ad163626f68a02805fe9da6)

assets/tests/hooks/useNearestIntersection.test.ts
```diff
@@ -40,3 +47,81 @@ describe("useNearestIntersection", () => {
...
+    const { result, rerender } = renderUseNearestIntersectionResult(location)
+
+    expect(result.current).toEqual(Loading())
+
+    await act(async () => rerender(location))
+
+    expect(result.current).toEqual(Ok(intersection))
+  })
+
+  // Should it retry if the last result was an error?
```
lemald, line 84: **todo:** if we do want it to behave this way, let's create a followup ticket and remove this comment, just because we'll probably lose track if the comment is the only thing documenting that we wanted to do that.

firestack: I'm not sure we should, the old code doesn't, and I'm not quite sure what edge cases cause the geocoding api to return errors. 
Do you think we should be retrying on error?

assets/src/hooks/useNearestIntersection.ts
```diff
@@ -1,13 +1,60 @@
...
+
+export const useNearestIntersection = (
+  latitude: number,
+  longitude: number
+): string | null => {
+  const statusResult = useNearestIntersectionResult(latitude, longitude)
+  if (!isLoadingState(statusResult) && isOk(statusResult)) {
+    return statusResult.ok
+  }
+  return null
```
lemald, line 59: **question:** if I understand correctly, this means that once the latitude and longitude change but before the new result comes in the hook will return `null`, correct? What will this do to existing consumers (mainly the VPP)?

firestack: oh! I meant to preserve old functionality, but I was going off only tests. I wasn't expecting to impact other areas because I don't have the scope to test/write tests for it.

I'm thinking of caching it and returning the old value till the next state, unless you think this is better and the usages should be updated instead.

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -105,42 +107,53 @@ const VehicleWorkInfo = ({ vehicle }: VehicleProp): React.ReactElement => (
...
+  vehicle,
+}: VehicleProp): React.ReactElement => {
+  const lastGoodResult = useRef<{
+    intersection: string
+    vehicleId: VehicleId
+  } | null>(null)
+  const statusResult = useNearestIntersectionResult(
+    vehicle.latitude,
+    vehicle.longitude
+  )
```
lemald, line 120: **question:** is it possible to use `useState` rather than `useRef`?

lemald: If so, I think `useState` would be preferable and a bit more idiomatic, `useRef` is mainly useful for imperatively accessing things in the DOM.

firestack: IIRC I switched to `useRef` because it would cause infinite state update loops and the component does not need to be rerendered when this updates.

I can double check though.

I wanted to extract this into a wrapper around a hook that returns `LoadingStatus<Result<T, E>>` which would instead return `Cached<LoadingStatus<Result<T,E>>>` but left that up to the future rather than spending more time on it.

firestack: https://reactjs.org/docs/hooks-reference.html#useref
> However, `useRef()` is useful for more than the `ref` attribute. It’s [handy for keeping any mutable value around](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) similar to how you’d use instance fields in classes.
>
> [...] Mutating the `.current` property doesn’t cause a re-render. [...]

assets/src/hooks/useNearestIntersection.ts
```diff
@@ -1,13 +1,60 @@
...
   }, [latitude, longitude])
-  return result
+
+  if (
+    lastLatLong.current.latitude !== latLong.latitude ||
+    lastLatLong.current.longitude !== latLong.longitude
+  ) {
+    lastLatLong.current = latLong
+    setLoadingState(Loading()) // redundant but needed to keep state synced?...
+    return Loading()
```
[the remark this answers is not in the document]
KaylaBrady: It does seem to me that this is redundant and that `lastLatLong`doesn't need to be tracked. By having `[latitude, longitude]` in the dependency array of the `useEffect` hook, that hook is going to be called any time either latitude or longitude changes, so this seems to duplicate the call to `setLoadingState(Loading())` on line 23.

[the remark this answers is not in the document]
firestack: The biggest issue is that when `useNearestIntersectionResult`'s input changes, it is required to invalidate the last result it got and return the `loading` state instead.

The `useState` here https://github.com/mbta/skate/pull/1873/files/a4de71efb0a7d81b53450660f8b137ae638bd5e3..1342a75301f478e8c70a8b0dc0fe5b22dc4f6286#diff-26b206fffc63b957a949bb9050dc9b9931d22ca0a6607bd710662c5f71ecdd83R18 will return the last successful value if not checked that the input has changed, which means the return will return the last loading state value, which when a new vehicle is selected, will be the last vehicles request status/result.

`useEffect` doesn't seem to run synchronously, as well as `setLoadingState` won't update the variable until next render.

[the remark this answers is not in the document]
KaylaBrady: I'm struggling with the boundaries between `useNearestIntersection` and where it is ultimately used in  `VehicleNearestIntersection`. One possible proposal for an alternate way to handle invalidating the intersection when a new vehicle is selected - what if `VehicleNearestIntersection` has a key for the vehicleId? 

Check out this blog post on forcing a reset on key change [here](https://medium.com/@albertogasparin/forcing-state-reset-on-a-react-component-by-using-the-key-prop-14b36cd7448e) - I haven't personally tried that approach before outside of lists, but I'm thinking it could maybe help us to manage less state if we don't have to account for the case where the vehicle changes. This seems like it could be relevant to @lemald's thread [here](https://github.com/mbta/skate/pull/1873#discussion_r1064738877)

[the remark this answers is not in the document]
firestack: THIS IS SO MUCH EASIER LOL THANK YOU! 🙇🏻‍♀️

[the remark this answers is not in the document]
lemald: Interesting, I hadn't seen that approach before but it's neat!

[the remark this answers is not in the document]
firestack: I had something working on the UI side but I wasn't happy with the tests, I was hitting issues similar to this https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning where I couldn't get the promise that returns from `fetchNearestIntersection` to resolve before the test completed, which meant that the promise was firing after expect had finished for whatever reason. 

I'm instead going to walk back these changes and add a few more tests and then we should be able to merge? 
mainly trying to prioritize unblocking people by finishing this PR.

I've still got them, so we could do this in a separate PR, unsure if I should open a draft PR with the changes or just push them up as a branch.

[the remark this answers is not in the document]
firestack: Knowing I can use `key` to force reset component state (e.g. to make sure it resets with it's parent component) is very helpful though!!!

assets/tests/hooks/useNearestIntersection.test.ts
```diff
@@ -39,4 +42,39 @@ describe("useNearestIntersection", () => {
     rerender()
     expect(mockNearestIntersection).toHaveBeenCalledTimes(1)
   })
+  const renderUseNearestIntersection = (location: GeographicCoordinate) =>
```
lemald: **nitpick:** this will scan slightly easier with a space separating it from the `test` block above.

lemald: Never mind, looks like this is changed in a future commit anyway.

assets/src/models/searchQuery.ts
```diff
@@ -1,6 +1,9 @@
...
+
 export interface SearchQuery {
   text: string
-  property: string
+  // property: SearchQueryType
```
lemald, line 5: **nitpick:** leftover commented out line

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+        Current Location
+      </label>
+      <output className="m-current-location__value label font-s-semi" id={id}>
+        <CurrentLocation vehicle={vehicle} />
+      </output>
+    </div>
+  )
+}
+
+// #endregion
```
firestack, line 134: What I did find from my work on implementing loading state was that the component with the `key` value will be discarded and rerendered, which means for react testing library, you need to do `screen.getByRole` again to get the updated value. 

But this actually is a test working correctly, which is that the DOM element should not be discarded, and instead the `key` value should go on the `<CurrentLocation>` element and not it's parent `<output>` element which is the target of `screen.getByRole("status", { name: "Current Location"})`.

lemald (approved):
Okay, whew, I think this is actually ready!

PR #1877, opened 2023-01-05
feat: No views display over map page

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203565462590421/f

This PR makes use of [Outlets](https://reactrouter.com/en/main/components/outlet) and nested routes so that the `RightPanel` and `LateView` components are rendered on all routes except for the new search map page.


assets/src/components/viewHeader.tsx
```diff
@@ -45,7 +45,9 @@ const ViewHeader: ViewHeaderType = ({
           {backlinkTitle(backlinkToView)}
         </button>
       ) : null}
-      <div className="m-view-header__title">{title}</div>
+      <div className="m-view-header__title" role="heading" aria-level={2}>
```
KaylaBrady, line 48: adding this role made it easier to look for in tests

assets/tests/components/app.test.tsx
```diff
@@ -78,34 +79,126 @@ describe("App", () => {
...
+      test("VPP ", () => {
+        render(
+          <StateDispatchProvider
+            state={{
+              ...initialState,
+              selectedVehicleOrGhost: vehicle,
+            }}
+            dispatch={mockDispatch}
+          >
+            <MemoryRouter initialEntries={[path]}>
```
KaylaBrady, line 108: Using MemoryRouter + AppRoutes here based on testing guidance:
from [reactrouter testing guide:](https://v5.reactrouter.com/web/guides/testing)

> Using MemoryRouter or a custom history is recommended in order to be able to reset the router between tests.
...
> You can also use BrowserRouter if your test environment has the browser globals window.location and window.history (which is the default in Jest through JSDOM, but you cannot reset the history between tests)

assets/src/components/viewHeader.tsx
firestack: can we style a h2 to match and use it instead?

firestack: I guess we'd also need a <h1> for the page header as well, but we don't have any places for that I can think of off the top of my head :/

KaylaBrady: Oh yeah, I was wary of messing with the styling but looks like the styles for `m-view-header__title` are already pretty specific. So I can change to `<h2>` if that is more conventional even without an `<h1>` present

assets/tests/components/app.test.tsx
```diff
@@ -49,7 +50,7 @@ describe("App", () => {
   })
 
   test("shows data outage banner if there's a data outage", () => {
-    ;(useDataStatus as jest.Mock).mockImplementation(() => "outage")
+    ;(useDataStatus as jest.Mock).mockImplementationOnce(() => "outage")
```
firestack, line 53: **suggestion**(non-blocking): what do you think of using `mockReturnValueOnce` here?

firestack (approved):

PR #1882, opened 2023-01-11
fix: do not display adherence info for shuttles

firestack:
# Summary
This slipped through #1873 

This is the simplest fix for this, but I think this logic should be pushed down further into the components used so that they can be used in more contexts. 
So looking for thoughts on if it's worth perusing allowing `ScheduleAdherence` to handle `Ghosts` and `Shuttles` by returning `null` to disable rendering instead of handling that within `VehicleRouteSummary`.

I kinda feel that the `output` element returned by `VehicleRouteSummary` should always be returned (and this thought currently applies to sub elements which create elements in the accessibility tree) and the text content within the accessibility nodes should be removed instead.

---

Related: #1873 
Asana Ticket: https://app.asana.com/0/1200180014510248/1203502925530262/f

### DIFF
--- a/assets/src/components/vehicleRouteSummary.tsx
+++ b/assets/src/components/vehicleRouteSummary.tsx
@@ -87,7 +87,7 @@ export const VehicleRouteSummary = ({
       className="m-vehicle-route-summary__route-variant headsign font-m-semi"
     />
 
-    {isVehicle(vehicle) && (
+    {isVehicle(vehicle) && !vehicle.isShuttle && (
       <ScheduleAdherence
         vehicle={vehicle}
         title="Vehicle Schedule Adherence"
--- a/assets/src/components/vehicleRouteSummary.tsx
+++ b/assets/src/components/vehicleRouteSummary.tsx
@@ -4,7 +4,7 @@ import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import vehicleLabel from "../helpers/vehicleLabel"
 import { emptyLadderDirectionsByRouteId } from "../models/ladderDirection"
 import { currentRouteTab } from "../models/routeTab"
-import { directionName, isVehicle } from "../models/vehicle"
+import { directionName } from "../models/vehicle"
 import { drawnStatus } from "../models/vehicleStatus"
 import { VehicleOrGhost } from "../realtime"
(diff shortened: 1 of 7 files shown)


assets/src/components/vehicleRouteSummary.tsx
```diff
@@ -87,7 +87,7 @@ export const VehicleRouteSummary = ({
       className="m-vehicle-route-summary__route-variant headsign font-m-semi"
     />
 
-    {isVehicle(vehicle) && (
+    {isVehicle(vehicle) && !vehicle.isShuttle && (
```
KaylaBrady, line 90: suggestion: Could you add a test for this shuttle-specific behavior?

firestack: In the process 👍🏻 
Should have something in a few.

KaylaBrady (approved):
Looks good once a test is added! I personally tend to like the approach you've taken here where the conditional is outside the `<ScheduleAdherence/>` and `<ScheduleAdherence/>` itself never returns null. 

I think that preference comes from the [react conditional rendering doc](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering) where it describes a component hiding itself as a "rare case". But I'd be open to a different view on tha!

firestack:
> I personally tend to like the approach you've taken here where the conditional is outside the <ScheduleAdherence/> and <ScheduleAdherence/> itself never returns null.

I changed this sorry 😭
I felt that it was better that the component handle the data integrity in case that this component was used elsewhere.

I did this for two reasons
- it made tests easier to write lol
	(This doesn't work unless the element with this role is present)
	```
      expect(
        screen.getByRole("status", { name: /Vehicle Schedule Adherence/i })
      ).toBeEmptyDOMElement()
	```
- (rephrasing the above) If this component was reused somewhere, I think this contract should be upheld there as well

KaylaBrady:
> I changed this sorry 😭

Ah @firestack I get now what you mean now from the PR description

> I kinda feel that the output element returned by VehicleRouteSummary should always be returned

Since `ScheduleAdherence` still always returns the `output` element, that "rare case" comment above doesn't actually apply here since the element isn't really hiding itself. The points in your comment both make sense to me!

lemald (approved):

PR #1884, opened 2023-01-12
feat(VPP): Link from VPP to map page

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203526554725734/f

~~I tried to do the minimal possible refactor of `MapPage` required to introduce the `useVehicleForId` hook. The `MapPage` component could potentially be broken down into smaller sub-components, but I wasn't sure that this PR is the right time to do it.~~

No changes are made to the MapPage - handling the selected vehicle on the MapPage will be addressed separately from this PR.


assets/css/properties_panel/_vehicle_properties_panel.scss
```diff
@@ -70,6 +70,25 @@
...
+  z-index: 401;
+  top: 10px;
+  // Adjacent to controls - control width is 40px
+  right: calc(env(safe-area-inset-right) + 40px + 0.5rem);
+  text-decoration: none;
+  background-color: $white;
+  padding: 0.5rem;
+  border-radius: 6px;
+  color: $color-gray-700 !important;
+  font-weight: 500;
```
lemald, line 85: **suggestion:** we now have a standard set of font weights defined in `_skate_ui.scss`. I know 500 isn't one of them, but it might be worth asking design if they want to change this to either 400 or 600 to be more standardized - I ran into something similar with the search pane and design decided to change the weight in question to one of the standard ones.

assets/tests/components/mapPage.test.tsx
```diff
@@ -329,8 +360,11 @@ describe("<MapPage />", () => {
     await userEvent.click(screen.getByRole("button", { name: /close/i }))
     expect(vehiclePropertiesCard).not.toBeInTheDocument()
     expect(routeShape).not.toBeInTheDocument()
-
-    expect(mapSearchPanel).toBeVisible()
+    expect(
+      screen.getByRole("generic", {
+        name: /map search panel/i,
+      })
+    ).toBeVisible()
```
firestack, line 367: question: This is having to be reaquired because of rerendering, right? do you know why it's being invalidated/recreated?

I'm noticing on dev blue that the map page blinks and resets a bunch of state when I select a new vehicle. If possible, I think it'd be best if we prevented it from destroying and recreating these objects when rendering?

assets/tests/components/mapPage.test.tsx
```diff
@@ -347,15 +381,46 @@ describe("<MapPage />", () => {
       tripId === vehicle.tripId ? shapes : null
     )
 
+    const { container } = render(<MapPage />, {
+      wrapper: (props) => (
+        <RealDispatchWrapper
+          initialState={stateFactory.build({ searchPageState: activeSearch })}
+          {...props}
+        />
```
firestack, line 389: thought(non-blocking): what do you think of deduplicating function passed to the the wrapper in render by either wraping or changing `RealDispatchWrapper` to be `(initialState) => (props) => ReactElement`?

assets/src/components/propertiesPanel/miniMap.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { useStations } from "../../hooks/useStations"
+import { Vehicle, VehicleId } from "../../realtime"
+import { Shape, Stop } from "../../schedule"
+import { setSelectedVehicle } from "../../state/searchPageState"
+import inTestGroup, { MAP_BETA_GROUP_NAME } from "../../userInTestGroup"
+import { mapModeForUser } from "../../util/mapMode"
+import Map from "../map"
+
+const SearchMapLink = ({ vehicleId }: { vehicleId: VehicleId }) => {
+  const [_state, dispatch] = useContext(StateDispatchContext)
```
lemald, line 14: **suggestion:** I think you can also do `const [, dispatch] = useContext(StateDispatchContext)` if you don't need to get the `state`.

assets/tests/components/mapPage.test.tsx
firestack: I personally like that we get to continue to write JSX with the form you have now. But having to duplicate `(props) {...props}` seems like it could be pulled out

assets/tests/components/propertiesPanel/miniMap.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      expect(screen.queryByText("Open Map")).not.toBeInTheDocument()
+    })
+
+    test("Map does include fullscreen button", () => {
+      render(
+        <MemoryRouter initialEntries={["/"]}>
+          <MiniMap vehicle={vehicle} routeVehicles={[]} shapes={[]} />{" "}
+        </MemoryRouter>
+      )
+      expect(screen.getByTitle("Full Screen")).toBeInTheDocument()
```
firestack, line 80: nitpick: could the queries in here be changed to `getByRole`?

assets/src/components/propertiesPanel/miniMap.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const ref = React.useRef(null)
+
+  useEffect(() => {
+    DomEvent.disableClickPropagation(ref.current!)
+  }, [])
```
lemald, line 19: **question:** thinking through React component lifecycle considerations here - is there any situation in which `ref.current` would change and we would want to fire the effect again?

KaylaBrady: Thanks for asking, I hadn't thought about where this fit into the lifecycle. 

I wouldn't expect `ref.current` to change, but maybe that is an assumption about the inner-workings of `Link`, which may not be a safe assumption. 

I can opt to follow the guidance of this [react doc](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) and [useCallback may be what you meant by useRef & useEffect](https://medium.com/welldone-software/usecallback-might-be-what-you-meant-by-useref-useeffect-773bc0278ae) and switch to useCallback instead.

lemald: Ah okay, I wasn't actually away of callback refs - after looking at those links, though, I like the approach.

assets/css/properties_panel/_vehicle_properties_panel.scss
KaylaBrady: Discussed with design and updated to use semi-bold

assets/tests/components/mapPage.test.tsx
KaylaBrady: Yeah I see what you mean! I'm thinking of switching this back to the form <Wrapper><ElementUnderTest/><Wrapper> instead of using the `wrapper` render option -- the `props` here is only passing the child element (`ElementUnderTest`). 

I also personally like getting to write/read these tests as JSX, and this seems like an easy change to reduce that duplication. Any opposition?

firestack: Works for me 👍🏻

KaylaBrady:
@mbta/skate-developers when you have the chance, this is ready for re-review with all MapPage changes removed.

assets/tests/components/propertiesPanel/miniMap.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      )
+      expect(
+        screen.queryByRole("link", { name: "Open Map" })
+      ).not.toBeInTheDocument()
+    })
+
+    test("Map does include fullscreen button", () => {
+      render(
+        <MemoryRouter initialEntries={["/"]}>
+          <MiniMap vehicle={vehicle} routeVehicles={[]} shapes={[]} />{" "}
```
firestack, line 86: question(non-blocking): what is the `{" "}` needed for?

firestack (approved):

firestack (approved):

assets/tests/components/propertiesPanel/miniMap.test.tsx
KaylaBrady: Thanks for noticing - that was needlessly added by the formatter. Removed!

assets/src/components/propertiesPanel/miniMap.tsx
```diff
@@ -0,0 +1,69 @@
...
+  routeVehicles: Vehicle[]
+}) => {
+  const stations: Stop[] | null = useStations()
+
+  return inTestGroup(MAP_BETA_GROUP_NAME) ? (
+    <Map
+      selectedVehicleId={vehicle.id}
+      vehicles={[vehicle]}
+      shapes={shapes}
+      secondaryVehicles={routeVehicles}
```
firestack, line 52: From the asana ticket:
> Any other vehicle currently on the map can be selected by clicking on it. This will display that vehicle’s vehicle properties card.

I don't think secondaryVehicles can be selected, is this critiera still relavant?

lemald (approved):
Looks like all of my comments have been addressed though I see @firestack still has a few outstanding.

assets/src/components/propertiesPanel/miniMap.tsx
KaylaBrady: My understanding is that criteria was referring to the main map page, not the mini map (and therefore not really in scope for this PR). I'll confirm that with design though before merging!

firestack: Oh I misread this, I read `vehicle properties panel` somehow 😵‍💫

firestack (approved):

PR #1886, opened 2023-01-18
refactor(ts/hooks/useVehicleForId): make vehicleId nullable 

firestack:
make vehicleId allowed to be nullable to avoid rendering conditional hooks

Shared between #1884 
 
---

Asana Ticket: https://app.asana.com/0/1200180014510248/1203641439240315


KaylaBrady (approved):
Cool solution!

lemald (approved):
+1, I like it!

PR #1890, opened 2023-01-19
refactor(ts/components/map): separate Autocenterer from map to allow configuration through composition

firestack:
This does an initial refactor to <Map/> needed to be able to have a separate implementation and more control over the autocentering function needed to complete the asana ticket.
Looking for feedback related to feasibility and if this seems like the right direction to go in to allow autocenterer to be wielded with more specificity.

This does not change any functionality of the autocenterer or the map and does it's best to preserve existing tests and features. Tests have been used to ensure that functionality remains the same as it did before this PR so that it may be iterated on in the future.

If I were to spend more time on this, I would convert the RecenterButton control to be a `input[checkbox] role=switch` and change tests to use `toBeChecked()` instead of the existing tests which test for if a class is present on the map status classes div.

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1203641439240315

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -300,10 +300,12 @@ interface AutoCenterMapOnProps {
   setShouldAutoCenter: Dispatch<SetStateAction<boolean>>
 }
 
-const AutoCenterMapOn = (props: AutoCenterMapOnProps) => {
-  const { shouldAutoCenter, setShouldAutoCenter, isAutoCentering, latLngs } =
-    props
-
+const AutoCenterMapOn = ({
+  shouldAutoCenter,
+  setShouldAutoCenter,
+  isAutoCentering,
+  latLngs,
+}: AutoCenterMapOnProps) => {
   useMapEvents({
     // If the user drags or zooms, they want manual control of the map.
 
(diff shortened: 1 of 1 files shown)


assets/src/components/map.tsx
```diff
@@ -312,25 +283,88 @@ const Autocenterer = ({
...
+interface AutoCenterMapOnProps {
+  latLngs: LatLng[]
+  isAutoCentering: MutableRefObject<boolean>
+  shouldAutoCenter: boolean
+  setShouldAutoCenter: Dispatch<SetStateAction<boolean>>
+}
+
+const AutoCenterMapOn = (props: AutoCenterMapOnProps) => {
+  const { shouldAutoCenter, setShouldAutoCenter, isAutoCentering, latLngs } =
+    props
```
lemald, line 306: **suggestion (non-blocking):** could you destructure the `props` in the function argument on the line above instead?

lemald (approved):
One minor comment that might make the code slightly less verbose, but over all looks good!

assets/src/components/map.tsx
firestack: [`5c1bd47` (#1890)](https://github.com/mbta/skate/pull/1890/commits/5c1bd476541682377de75c1dd05397a9bf21a80e)

PR #1891, opened 2023-01-20
feat(mapPage): selection logic for map page

firestack:
I split out the autocenterer for this so that the offset for the mapPage can be configured independently of the other maps in skate. 

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1203641439240315/f

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -364,17 +364,6 @@ export const AutoCenterMapOn = ({
   )
 }
 
-type PaddingArgument = {
-  padding: [number, number, number, number]
-}
-
-export const AutoCenterMapWithPadding = ({
-  padding,
-  ...props
-}: AutoCenterMapOnProps & PaddingArgument) => {
-  return AutoCenterMapOn(props)
-}
-
 export const ContainedAutoCenterMapOn = (props: AutoCenterMapArgs) => {
   return AutoCenterMapOn({ ...props, ...useAutoCenterState() })
 }
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -345,11 +345,6 @@ export const InterruptibleFollower = ({
         onUpdate={onUpdate}
       />
       <RecenterControl position="topright" recenter={turnOnFollowing} />
-      {/* <ReactLeafletRecenterControl
-        position="topright"
-        recenter={turnOnFollowing}
-        shouldFollow={shouldFollow}
-      /> */}
     </>
   )
 }
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -283,7 +283,7 @@ const AutoCenter = ({
   return <></>
 }
 
-const useAutoCenterState = () => {
+export const useAutoCenterState = () => {
   const [shouldAutoCenter, setShouldAutoCenter] = useState<boolean>(true)
   const isAutoCentering: MutableRefObject<boolean> = useRef(false)
 
@@ -294,22 +294,24 @@ const useAutoCenterState = () => {
   }
 }
 
-interface AutoCenterMapOnProps {
-  latLngs: LatLng[]
+export interface AutoCenterMapState {
   isAutoCentering: MutableRefObject<boolean>
   shouldAutoCenter: boolean
   setShouldAutoCenter: Dispatch<SetStateAction<boolean>>
 }
 
-export const AutoCenterMapOn = ({
-  shouldAutoCenter,
-  setShouldAutoCenter,
-  isAutoCentering,
-  latLngs,
-}: AutoCenterMapOnProps) => {
+export interface AutoCenterMapArgs {
+  positions: LatLng[]
+}
+
+export type AutoCenterMapOnProps = AutoCenterMapState & AutoCenterMapArgs
+
+function useDisablingAutoCenterMapEvents(
+  isAutoCentering: React.MutableRefObject<boolean>,
+  setShouldAutoCenter: React.Dispatch<React.SetStateAction<boolean>>
+) {
   useMapEvents({
     // If the user drags or zooms, they want manual control of the map.
-
     // `zoomstart` is fired when the map changes zoom levels
     // this can be because of animating the zoom change or user input
     zoomstart: () => {
@@ -337,13 +339,22 @@ export const AutoCenterMapOn = ({
     // `autopanstart` is invoked when opening a popup causes the map to pan to fit it
     autopanstart: () => setShouldAutoCenter(false),
   })
+}
+
+export const AutoCenterMapOn = ({
+  shouldAutoCenter,
+  setShouldAutoCenter,
+  isAutoCentering,
+  positions,
+}: AutoCenterMapOnProps) => {
+  useDisablingAutoCenterMapEvents(isAutoCentering, setShouldAutoCenter)
 
   return (
     <>
       <AutoCenter
         shouldAutoCenter={shouldAutoCenter}
         isAutoCentering={isAutoCentering}
-        latLngs={latLngs}
+        latLngs={positions}
       />
       <RecenterControl
         position="topright"
@@ -351,6 +362,21 @@ export const AutoCenterMapOn = ({
       />
     </>
   )
+}
+
+type PaddingArgument = {
+  padding: [number, number, number, number]
+}
+
+export const AutoCenterMapWithPadding = ({
+  padding,
+  ...props
+}: AutoCenterMapOnProps & PaddingArgument) => {
+  return AutoCenterMapOn(props)
+}
+
+export const ContainedAutoCenterMapOn = (props: AutoCenterMapArgs) => {
+  return AutoCenterMapOn({ ...props, ...useAutoCenterState() })
 }
 // #endregion
 
@@ -457,16 +483,19 @@ export const BaseMap = (props: Props): ReactElement<HTMLDivElement> => {
   )
 }
 
+export const vehicleToLeafletLatLng = ({
+  latitude,
+  longitude,
+}: Vehicle): Leaflet.LatLng => Leaflet.latLng(latitude, longitude)
+
 export const AutoCenteringMap = (props: Props) => {
   const state = useAutoCenterState(),
     { shouldAutoCenter } = state
 
-  const latLngs: LatLng[] = props.vehicles.map(({ latitude, longitude }) =>
-    Leaflet.latLng(latitude, longitude)
-  )
+  const positions: LatLng[] = props.vehicles.map(vehicleToLeafletLatLng)
   const centerOnProps: AutoCenterMapOnProps = {
     ...state,
-    latLngs,
+    positions,
   }
 
   return (
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -1,4 +1,5 @@
 import Leaflet, {
+  Bounds,
   Control,
   ControlOptions,
   DomUtil,
@@ -6,6 +7,7 @@ import Leaflet, {
   LatLngExpression,
   LatLngLiteral,
   Map as LeafletMap,
+  Point,
 } from "leaflet"
 
 import "leaflet-defaulticon-compatibility" // see https://github.com/Leaflet/Leaflet/issues/4968#issuecomment-483402699
@@ -233,11 +235,27 @@ export const autoCenter = (
   latLngs: LatLngExpression[],
   pickerContainerIsVisible: boolean
 ): void => {
-  // console.debug({ latLngs })
   if (latLngs.length === 0) {
     map.setView(defaultCenter, 13, { animate: false })
   } else if (latLngs.length === 1) {
-    map.setView(latLngs[0], 16)
+    const bounds = map.getContainer().getBoundingClientRect()
+    const containerBounds = new Bounds([0, 0], [bounds.width, bounds.height])
+
+    const topLeft = new Point(445, 0)
+    const innerBounds = new Bounds(
+      topLeft,
+      containerBounds.getBottomRight() //.subtract([42, 15])
+    )
+
+    const targetZoom = 16
+    const targetPoint = map
+        .project(latLngs[0], targetZoom)
+        .subtract(
+          innerBounds.getCenter().subtract(containerBounds.getCenter())
+        ),
+      targetLatLng = map.unproject(targetPoint, targetZoom)
+    // const targetLatLng = latLngs[0]
+    map.setView(targetLatLng, targetZoom)
   } else if (latLngs.length > 1) {
     map.fitBounds(Leaflet.latLngBounds(latLngs), {
       paddingBottomRight: [20, 50],
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -1,5 +1,4 @@
 import Leaflet, {
-  Bounds,
   Control,
   ControlOptions,
   DomUtil,
@@ -7,7 +6,6 @@ import Leaflet, {
   LatLngExpression,
   LatLngLiteral,
   Map as LeafletMap,
-  Point,
 } from "leaflet"
 
 import "leaflet-defaulticon-compatibility" // see https://github.com/Leaflet/Leaflet/issues/4968#issuecomment-483402699
@@ -17,6 +15,7 @@ import React, {
   MutableRefObject,
   ReactElement,
   SetStateAction,
+  useCallback,
   useContext,
   useEffect,
   useId,
@@ -24,7 +23,6 @@ import React, {
   useState,
 } from "react"
 import ReactDOM from "react-dom"
-
 import {
   AttributionControl,
   MapContainer,
@@ -34,6 +32,7 @@ import {
   ZoomControl,
 } from "react-leaflet"
 import { createControlComponent } from "@react-leaflet/core"
+
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import { className } from "../helpers/dom"
 import { TrainVehicle, Vehicle, VehicleId } from "../realtime.d"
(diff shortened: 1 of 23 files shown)


assets/tests/testHelpers/leafletMapWidth.ts
```diff
@@ -0,0 +1,45 @@
...
+  // failures from floating point math resulting in `Float.NaN` which propagates
+  // down and causes issues elsewhere
+
+  // the max padding we currently request is
+  // 20 + 220, 50+20, so we at minimum need 220x70 from the padding.
+  // setHtmlDefaultWidthHeight(240, 70)
+
+  // jsdom defaults to 1024x768 and usually has the window sizes populated
+  // setHtmlDefaultWidthHeight(1024, 768)
+  setHtmlDefaultWidthHeight(window.innerWidth, window.innerHeight)
```
firestack, line 38: I couldn't figure out a good default, some tests have been built around 0,0 without knowing it, I think it's best to follow what jsdom is telling us, but I could be convinced otherwise.

Advice for where to put this documentation would be welcome.

assets/tests/testHelpers/leafletMapWidth.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+afterEach(() => {
+  clientHeight.mockClear()
+  clientWidth.mockClear()
+})
```
firestack, line 44: not super happy that this was a side-effect of importing, but it made tests a lot more stable IIRC. Should probably revisit?

assets/tests/testHelpers/wrappers.tsx
```diff
@@ -0,0 +1,24 @@
...
+    reducer,
+    undefined,
+    () => initialState || stateFactory.build()
+  )
+  return (
+    <StateDispatchProvider state={state} dispatch={dispatch}>
+      {children}
+    </StateDispatchProvider>
+  )
+}
```
firestack, line 24: Still iffy on this but tests that interact with the global state needed this to reduce boilerplate and complexity in tests.

As a behavioral testing tool, it's very useful.

assets/src/components/mapPage.tsx
```diff
@@ -116,27 +127,32 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
     <div
-      className={`m-map-page ${mobileDisplayClass} ${mobileMenuClass}`}
+      className={`m-map-page ${mobileDisplayClass} ${mobileMenuClass} inherit-box border-box`}
       aria-label="Search Map Page"
     >
       <div
         className="m-map-page__input-and-results"
         aria-label="Map Search Panel"
-        {...(vpcEnabled ? { hidden: true } : {})}
+        hidden={vpcEnabled === true}
```
lemald, line 136: **suggestion:** I think you can remove the `=== true` here because `vpcEnabled` was created from the `Boolean` constructor already.

firestack: I know I _can_ but I really prefer the explicitness of how it reads. I'm happy to remove it, but I do think this is better for future readers of the code.

assets/src/components/mapPage.tsx
```diff
@@ -150,22 +166,28 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
-            <VehiclePropertiesCard
-              vehicle={liveVehicle}
-              onClose={() => {
-                dispatch(setSelectedVehicle(null))
-                setShowVehicleCard(false)
-              }}
-            />
-          ) : undefined}
+          <>
+            {selectedVehicle && isVehicle(selectedVehicle) && <></>}
```
lemald, line 177: **question:** is this fragment a placeholder for something? Right now I don't think it actually changes anything about the DOM if `selectedVehicle` is present and also a `Vehicle`.

firestack: yeah that's where the autocenterer is going

assets/src/components/mapPage.tsx
lemald: I guess the way I think of it is that the name `vpcEnabled` already connotes that it's a boolean, although sadly JS / TS lacks an actual convention for naming boolean values like putting a `?` at the end in Elixir so it's not explicit as it could be. Since it's purely subjective I'm fine with leaving it the way you prefer.

assets/tests/components/map.test.tsx
```diff
@@ -330,9 +331,25 @@ describe("<Map />", () => {
...
+    const { container } = render(
+      <Map vehicles={[vehicle]} selectedVehicleId={vehicle.id} />
+    )
+
+    expect(
+      container.querySelector(".m-vehicle-map__icon .selected")
+    ).toBeInTheDocument()
+    expect(
+      container.querySelector(".m-vehicle-map__label.selected")
+    ).toBeInTheDocument()
```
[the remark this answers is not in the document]
lemald: If you have time and feel like it, I'd be happy to look at a PR for that refactor! Otherwise I agree that these assertions aren't _great_ but I don't think they're dealbreakers.

assets/tests/testHelpers/leafletMapWidth.ts
lemald: I like the approach of going with what jsdom tells us. Screen dimensions are one of those things I've never really considered very much when writing tests since we're usually so focused on the raw DOM structure, so thank you for doing some work to make it handled somewhat more correctly!

assets/src/components/map.tsx
```diff
@@ -227,88 +229,66 @@ const EventAdder = ({
...
-const AutoCenter = ({
-  latLngs,
-  shouldAutoCenter,
-  isAutoCentering,
-}: {
-  shouldAutoCenter: boolean
-  isAutoCentering: MutableRefObject<boolean>
-  latLngs: LatLng[]
-}) => {
+export const Follower = ({
```
KaylaBrady, line 239: Appreciate this name change!

assets/src/components/map.tsx
```diff
@@ -227,88 +229,66 @@ const EventAdder = ({
...
-
-  const [{ pickerContainerIsVisible }] = useContext(StateDispatchContext)
-  const [currentLatLngs, setCurrentLatLngs] = useState<LatLng[]>(latLngs)
+  const [currentLatLngs, setCurrentLatLngs] = useState<LatLng[]>(positions)
 
   if (
-    !equalByElements(latLngs, currentLatLngs, (latLng1, latLng2) =>
-      latLng1.equals(latLng2)
-    )
+    !equalByElements(positions, currentLatLngs, (lhs, rhs) => lhs.equals(rhs))
```
KaylaBrady, line 249: question (non-blocking): I realize this mostly from the existing logic so doesn't really need to change, but is there a case when these would not be equal since `currentLatLngs` is initialized to `positions`?

assets/src/components/map.tsx
```diff
@@ -227,88 +229,66 @@ const EventAdder = ({
...
-    isAutoCentering,
-    currentLatLngs,
-    pickerContainerIsVisible,
-  ])
+  }, [map, shouldFollow, isUpdating, currentLatLngs, onUpdate])
 
   return <></>
 }
+export interface InteractiveFollowState {
+  isUpdating: MutableRefObject<boolean>
```
KaylaBrady, line 266: nitpick (non-blocking): The name `isUpdating` was a bit confusing to me at first with the combination of the name `InteractiveFollowState` - I didn't quite get what it is that is being updated, in combination with it being set to `isFollowing` on line 280 below.

assets/src/components/map.tsx
```diff
@@ -227,88 +229,66 @@ const EventAdder = ({
...
-const AutoCenterMapOn = ({
-  shouldAutoCenter,
-  setShouldAutoCenter,
-  isAutoCentering,
-  latLngs,
-}: AutoCenterMapOnProps) => {
+// Sets up map events to get a callback when the user interacts with the map
+// which should override the follower on state
+function useStopFollowingOnInteraction(
+  isAutoCentering: React.MutableRefObject<boolean>,
```
KaylaBrady, line 287: question: should this be `isFollowing` or `isUpdating` for consistency?

assets/src/components/map.tsx
```diff
@@ -336,24 +316,79 @@ const AutoCenterMapOn = ({
...
+  isUpdating,
+  positions,
+  onUpdate,
+}: InterruptibleFollowerProps) => {
+  useStopFollowingOnInteraction(isUpdating, setShouldFollow)
+
+  const turnOnFollowing = useCallback(
+    () => setShouldFollow(true),
+    [setShouldFollow]
+  )
```
KaylaBrady, line 337: question (non-blocking): Why `useCallback` here? I'm guessing it prevents re-renders, but I'm not too familiar with this hook

assets/src/components/mapPage.tsx
```diff
@@ -57,70 +65,298 @@ const ToggleMobileDisplayButton = ({
...
+}) => {
+  const { socket } = useSocket()
+  const vehicles = useVehiclesForRoute(socket, selectedVehicleRoute)
+  return (
+    <>
+      {filterVehicles(vehicles).map((vehicle: Vehicle) => (
+        <VehicleMarker
+          key={vehicle.id}
+          vehicle={vehicle}
+          isPrimary={true}
```
KaylaBrady, line 164: question: would these vehicles not be primary, similar to the non-selected vehicles in the VPP?

assets/src/components/mapPage.tsx
```diff
@@ -57,70 +65,298 @@ const ToggleMobileDisplayButton = ({
...
+      allowStreetView={true}
+      stopCardDirection={selectedVehicleRef?.directionId}
+      includeStopCard={true}
+      stations={stations}
+      shapes={selectedVehicleRef?.isShuttle ? [] : shapes}
+      stateClasses={FollowerStatusClasses(state.shouldFollow)}
+    >
+      <>
+        {showVpc && selectedVehicleRef && isVehicle(selectedVehicleRef) && (
+          <>
```
KaylaBrady, line 280: question (non-blocking): Are these empty tags needed?

KaylaBrady (commented):
This is looking good! Still wrapping my head around some of the following/selecting logic which may take a few more reads on my part, but I think some of the discussion from these questions will help me get there!

assets/src/components/map.tsx
lemald, line 239: **praise:** I like the updated terminology better than "autocenter."

assets/src/components/mapPage.tsx
firestack: nope! I actually cleaned this up when implementing the VPC ghost card https://github.com/mbta/skate/pull/1895
I was considering cherry picking that commit into this PR

assets/src/components/mapPage.tsx
```diff
@@ -57,70 +65,298 @@ const ToggleMobileDisplayButton = ({
...
-  const [searchOpen, setSearchOpen] = useState<boolean>(true)
+}
 
-  const liveVehicle: Vehicle | null = selectedVehicleId
-    ? onlyVehicles.find((v) => v.id === selectedVehicleId) || null
-    : null
-  const [showVehicleCard, setShowVehicleCard] = useState<boolean>(
-    searchPageState.selectedVehicleId ? true : false
+// #region Map Display
+const useCachedVehicleById = (
```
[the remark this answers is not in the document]
firestack: when `useVehicleForId` changes which topic it's subscribed to, it initially returns `null` which makes the VPC and route shape, route vehicles, and selected vehicle, blip in and out of existance as it rapidly rerenders.

I cache the last successful result before `selectedVehicleId` changes so that we don't blip the UI while loading the next vehicle.
I had this working with a reducer/state, but because what triggers the render is `useVehicleForId` updating, we only need to hold onto the previous vehicle till the next one is available/not null WHILE the `selectedVehicleId` value tells us that we didn't deselect a vehicle by not being `null`.

Maybe naming this `useDiferredVehicleById` would be better?... though this doesn't match what `useDiferredValue` does exactly. (I also tried to get this to work with `useDiferredValue` but that didn't work because `useSelectedVehicleForId` may return `null` multiple times before the next value/vehicle has resolved)

assets/src/components/mapPage.tsx
firestack: I dunno! I can check with design on that, from the mockups I'd seen it seemed better to go with the primary vehicle style.

assets/src/components/mapPage.tsx
```diff
@@ -57,70 +65,298 @@ const ToggleMobileDisplayButton = ({
...
+  const innerBounds = new Bounds(topLeft, mapContainerBounds.getBottomRight())
+  // The "new center" is the offset between the two bounding boxes centers
+  const offset = innerBounds
+    .getCenter()
+    .subtract(mapContainerBounds.getCenter())
+
+  const targetZoom = 16
+  const targetPoint = map
+      // Project the target point into screenspace for the target zoom
+      .project(points[0], targetZoom)
```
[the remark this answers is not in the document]
firestack: Because it was easier to implement for MVP heh.
The logic to fit a bounds would be similar but need a bit more thinking/work, so I left that to the future.

I wasn't sure how to allow the follower to change the type of it's `positions` prop based on the `onUpdate` function at the time of implementaiton, but that could probably be done?(unsure how long it'd take)
But then the component should be able to take a `LatLng | null` instead of `LatLng[]` if the component could change the type of the `positions` prop based on the second argument type of the `onUpdate` prop.

assets/src/components/map.tsx
firestack: I did it to prevent rerenders, but I don't actually remember if it was needed. It seemed like it was causing issues but it might have just been something I did while debugging and didn't reevaluate.

assets/src/components/map.tsx
firestack: Yep! Thank you!

assets/src/components/map.tsx
firestack: I didn't really question this becuase it was the existing logic. I think this is doing a value check rather than using the `Object.is` check that `useEffect` does to keep them updated only when the actual values of `positions` change?

Otherwise, because the `Object.is` function would return `null` for two different instances of arrays of the same data, I think the `useEffect` would fire constantly if it was dependant on `positions`? 
This should probably be checked.

assets/src/components/map.tsx
firestack: thank you!

firestack: let me know what you think of 7be135c4013b264282c75fe8618aca76dff3e0d7 ?

assets/src/components/map.tsx
firestack: https://github.com/mbta/skate/pull/1891#discussion_r1085875987

assets/src/components/map.tsx
lemald: @firestack yes, that's all correct as to why it works this way currently. @KaylaBrady if I'm understanding your question correctly, basically on the first render they will be equal, but on subsequent renders the `positions` prop will be updated but `currentLatLngs` will have the old value. Then, the state is only updated if there's an actual change in the contents.

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
lemald: Just spitballing here, how about `useMostRecentVehicleById`?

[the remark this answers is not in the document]
firestack: @lemald 👍🏻👍🏻

lemald (dismissed):
Okay, I've finally done a start-to-finish read over and I think this looks good, with the only issues I'm aware of being those already in flight. 👍

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
firestack: I just had it come to me that I was wrong about implementation difficulty, I'll make a patch for review!

assets/src/components/map.tsx
KaylaBrady: Thank you - `isAnimatingFollowUpdate` is so clear!!

assets/src/components/mapPage.tsx
KaylaBrady: Ah yeah I see what you mean! A separate question in this same area - is there an advantage to creating the markers in the `MapPage` component rather than delegating that to the existing `BaseMap` functionality using the `vehicles` and `selectedVehicleId` props?

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
KaylaBrady: Ohh maybe I missed something about this the first time around - is points is only ever a list of size one? I had it in my head that it could be a list including all the visible vehicles

assets/src/components/mapPage.tsx
firestack: The biggest issue was the autocenterer used all "primary" vehicles by default, and not wanting to add more props/functionality to the base component to support the map page configuration, and "idiomatic" react seemed to be pushing me in the direction of composition over "inheritance" (adding more props..?) to make the map easier to configure for all the different use cases.

I've actually been thinking that we might want to slim down the props that `BaseMap` takes and provide more functionality via composition instead (and preconfigued wrapper components, like the MiniMap). As well as maybe grouping our markers into `pane`s so that it's easier to look through in the react dev tools.

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
firestack: Yep! https://github.com/mbta/skate/pull/1891/files#diff-a6a64d5b0e8199bb4a7ea39da345d5247083943b92fd078e359cb4d1d9ab60ffR259

assets/src/components/map.tsx
firestack: So this `useCallback` doesn't actually seem to do anything? so I'll remove it!

assets/src/components/mapPage.tsx
firestack: Design said yes to the non selected vehicles not being primary 👍🏻

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
firestack: @mbta/skate-developers please take a look at [d90c0e53](https://github.com/mbta/skate/pull/1891/commits/d90c0e53016f508ce9a39a7d24ebc82fff14a66a) 🙏🏻

assets/src/components/mapPage.tsx
firestack: I think I covered everything in this thread, so I'm gonna resolve this, but I'll pay attention to this thread if anyone wants to point anything else out!

assets/src/components/map.tsx
firestack: [`efb9a92` (#1891)](https://github.com/mbta/skate/pull/1891/commits/efb9a92e7648a720e61c66030b73133176ffe0d2)

assets/src/components/mapPage.tsx
firestack: Cherry picked into [`c536580` (#1891)](https://github.com/mbta/skate/pull/1891/commits/c53658007572b47cc69963b14d988ae50cde0875)

firestack:
The only open question for code review I'm aware of is https://github.com/mbta/skate/pull/1891#discussion_r1086864229

Let me know if there's anything else, otherwise once design & product approve I'll merge 👍🏻

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
KaylaBrady: This looks good to me! I'm still working on understanding the benefits of `useRef` over `useState` in this case, but maybe I just need to revisit the `useRef` docs & don't want to block merging over it!

assets/src/components/mapPage.tsx
```diff
@@ -247,9 +242,8 @@ const MapDisplay = ({
 
   const stations = useStations()
 
-  const { socket } = useContext(SocketContext)
-
-  const selectedVehicleRef = useVehicleForIdCached(socket, selectedVehicleId),
+  const _selectedVehicleRef = useMostRecentVehicleById(selectedVehicleId),
+    selectedVehicleRef = _selectedVehicleRef && isVehicle(_selectedVehicleRef) && _selectedVehicleRef || null,
```
lemald: **comment (non-blocking):** I realize the intent of splitting this into two separate assignments is to avoid calling hooks conditionally, but usually we use the convention from Elixir (and maybe other languages?) where a variable name prefixed with an underscore isn't used later on, so this feels _slightly_ at odds with that convention, although its one use is in the came `const` statement... so I'm conflicted about how I feel about it.

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
lemald: One nitpick-y comment about variable names but otherwise I like it!

assets/src/components/mapPage.tsx
firestack: maybe we can setup a lint rule so that we get warned when variables with a leading `_` are used?

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
firestack: > benefits of useRef over useState in this case

Here's my understanding on what would happen if we used `useState` instead of `useRef`. 

When the `useVehicleForId` hook updates asynchronously, the state update causes `<MapDisplay/>` to rerender.

On that next render, `MapDisplay` can save the result of `useVehicleForId` using either `useState` or `useRef`.

When `useVehicleForId` updates, we need to store the value in case the next render returns `null` (which should only happen when `selectedVehicleId` changes and is not `null`). [^1]


__If `useVehicleForId` is cached in a `useState`, then updating the cache will cause a rerender, which would mean for every `useVehicleForId` update we would render the `<MapDisplay/>` twice.__



[^1]: Actually, maybe we can catch the last return value of `useVehicleForId` when `selectedVehicleId` changes? But unless the lambda in a `useEffect(..., [selectedVehicleId])` can capture the `VehicleOrGhost` before/while it changes to `null`, I still think we'd need a ref or state to track `lastSelectedVeihcleId`?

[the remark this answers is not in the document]
firestack: I'm still learning react so maybe there's something I missed! but 'TLDR;' I picked `useRef` because I didn't want to cause a rerender when updating it.

firestack:
I'm going to put this into draft until I confirm about final Acceptance Criteria, will reopen when ready for review.

firestack:
After design/product review I'm a little unsure of what the AC is around test group users, as it's a little unclear if these changes should also be done for users not in the test group. This PR should only affect users with access to the MapPage, and I've opened https://github.com/mbta/skate/pull/1897 in the case that we do want some of these changes pushed to users.

I'd like to be able merge this to unblock any PR's waiting on it, but we need final approval from product first, so I've split this out into two PR's.

lemald (approved):

PR #1894, opened 2023-01-24
feat(nav): Updated search map title & icon

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203548727085588

A few notes on this PR: 
* I thought about generalizing the other modes (ladder page, settings, shuttle map) to also use the `NavMode` interface so that searchMap isn't a special case in `pageOrTabName`. It seems doable and could also potentially simplify the router with something like a [route-config file](https://v5.reactrouter.com/web/example/route-config). That seems to be pushing the scope of this ticket. If reviewers are interested I can timebox incorporating that into this PR or create a tech-debt ticket for it. 

* Commits ~4c37a3265ec11ab4d42daa99b8a5b8af2b0f6d90~ 212519fb56fb2a1d2a88716e53908fa724faaab2 and 4c37a3265ec11ab4d42daa99b8a5b8af2b0f6d90 both have more detailed commits messages that provide some extra context for those changes.


assets/src/helpers/icon.tsx
```diff
@@ -215,6 +217,8 @@ export const SaveIcon = svgIcon(saveIconSvg)
 
 export const SearchIcon = svgIcon(searchIconSvg)
 
+export const SearchMapIcon = svgIcon(searchMapIconSvg)
```
lemald, line 220: **suggestion:** it's probably worth adding this icon to the list in `icon.test.tsx`, even if I have my doubts about how useful those particular tests are.

lemald (approved):
One quick comment about an opportunity to add a test, but otherwise looks good.

firestack (approved):
Oh neat! I wonder if [`212519f` (#1894)](https://github.com/mbta/skate/pull/1894/commits/212519fb56fb2a1d2a88716e53908fa724faaab2) might fix the [logo icon root scaling issue](https://app.asana.com/0/1148853526253437/1203216404241022) I encountered.

It seems like it should, but when I tried it locally somehow the skate logo is still getting it's viewbox stripped (after I exported the logo from the component library which had a viewbox on it, where our asset in the repo does not).

PR #1895, opened 2023-01-24
fix(ts/components/mapPage): ghost card implementation and fixes

firestack:
This does the final work allowing the VPC to also handle ghost types. Additionally, there were a few easy style changes that I ran by design and added as well. 

I see the biggest issue being around where to put type boundaries for components. I felt it was best to have a component handle as many relevant types as possible (within reason) until a final verdict is reached where only a single type can pass. 
This is why the current location component can accept a `VehicleOrGhost` argument, despite always returning the same string for a ghost, this is made type safe by the component which is used which can only display a location if the type is a `Vehicle`.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1203641819945189/f

### DIFF
--- a/assets/css/_vehicle_route_summary.scss
+++ b/assets/css/_vehicle_route_summary.scss
@@ -72,7 +72,7 @@
   .m-vehicle-route-summary__separator {
     justify-self: center;
     align-self: center;
-    height: 80%;
+    height: 100%;
     background-color: $color-gray-300;
     width: 1px;
   }
--- a/assets/css/_vehicle_route_summary.scss
+++ b/assets/css/_vehicle_route_summary.scss
@@ -38,6 +38,7 @@
     justify-self: end;
   }
 
+  .m-vehicle-route-summary__icon,
   .m-vehicle-route-summary__separator {
     justify-self: center;
     align-self: center;
--- a/assets/css/_vehicle_route_summary.scss
+++ b/assets/css/_vehicle_route_summary.scss
@@ -2,7 +2,7 @@
 // #region Layout
 .m-vehicle-route-summary {
   display: grid;
-  grid-template-rows: 1fr 2fr;
+  grid-template-rows: 3fr 8fr;
   grid-template-columns: min-content 1fr 6fr 12fr;
   grid-template-areas:
     "icon separator direction  adherence"
--- a/assets/css/_vehicle_route_summary.scss
+++ b/assets/css/_vehicle_route_summary.scss
@@ -8,7 +8,7 @@
     "icon separator direction  adherence"
     "icon separator route-name route-name";
 
-  align-items: center; // center element vertically within cell
+  align-items: start; // align element vertically within cell
   justify-items: start; // align element horizontally within the cell
 
   // assign element grid locations
@@ -57,9 +57,7 @@
   }
 
   // route variant name should not move when text gets longer
-  // so also align
   .m-route-variant-name {
-    align-self: start;
 
     --line-height: 1.5;
     line-height: var(--line-height);
--- a/assets/css/_vehicle_route_summary.scss
+++ b/assets/css/_vehicle_route_summary.scss
@@ -38,13 +38,15 @@
     justify-self: end;
   }
 
-  .m-schedule-adherence * {
-    display: inline-block;
-  }
-
   .m-vehicle-route-summary__separator {
     justify-self: center;
     align-self: center;
+  }
+
+  .m-schedule-adherence {
+    & > * {
+      display: inline-block;
+    }
   }
 
   // route variant name should not move when text gets longer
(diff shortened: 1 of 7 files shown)


assets/css/_vehicle_route_summary.scss
```diff
@@ -69,7 +80,7 @@
   .m-vehicle-route-summary__separator {
     justify-self: center;
     align-self: center;
-    height: 80%;
+    height: 100%;
     background-color: $color-gray-300;
     width: 1px;
   }
```
firestack, line 86: Changes in here were reviewed with design, individual commits labelled with `feedback(design)` have _slightly_ more details.

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -84,20 +92,24 @@ const TrNameValue = ({
...
+            joinTruthy([
               vehicle.operatorFirstName,
               vehicle.operatorLastName,
               vehicle.operatorId && `#${vehicle.operatorId}`,
-            ]).toLowerCase() || "N/A"}
-          </span>
+            ])) ||
+            "N/A"}
         </TrNameValue>
       </tbody>
```
firestack, line 114: I considered leaving the span and changing the class, but it felt like it was a bit useless to leave it (aside that it changed snapshots)

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -127,12 +137,24 @@ const VehicleNearestIntersection = ({
         Current Location
       </label>
       <output className="m-current-location__value label font-s-semi" id={id}>
-        <CurrentLocation vehicle={vehicle} />
+        {isVehicle(vehicle) ? (
+          <CurrentLocation vehicle={vehicle} />
+        ) : (
+          "Exact location cannot be determined"
+        )}
```
lemald, line 144: **question:** is handling this case actually necessary? It looks like `VehicleNearestIntersection` is only actually rendered if the `vehicleOrGhost` passes `isVehicle`, which should allow TypeScript to do type narrowing if I'm following this all correctly.

firestack: This case is not _necessary_, but I liked it because it allowed me to use `not.toBeVisible` instead of `not.toBeInTheDocument`. I don't like `not.toBeInTheDocument` because it accepts `null`, and the functions which are compatible with that, like `queryByRole` won't help you if you typed your query wrong (e.g. it could still be in the document, you just typed the `name` wrong or some other query property) 

So by keeping it in the DOM but not visible, I'm able to test and be certain that it's not visible.

Design didn't ask for a location component which could accept a ghost, but it was easy enough that I felt it'd be okay to keep, but if it's unwanted I can change it and edit the tests and https://github.com/mbta/skate/pull/1895/files/b5e6a29dbe85586ad761c5191afd6d82bd002e91#diff-0d5230dc7a702ef8c07b6ea444bff96e31f173e04e5b36f45cff2ef60db98543R206.

lemald: Hmm, okay. So this is an interesting tradeoff between test elegance and actual prod code elegance, in a way. I think my own decision here would probably be to not render the hidden stuff in the DOM in the ghost case, but as a reviewer I don't feel strongly about it so I trust your judgement. 👍

lemald (approved):

assets/src/components/vehiclePropertiesCard.tsx
```diff
@@ -7,16 +7,18 @@ import React, {
...
-  GeographicCoordinateBearing,
-} from "./streetViewButton"
+import StreetViewButton from "./streetViewButton"
 import { VehicleRouteSummary } from "./vehicleRouteSummary"
 
 interface VehicleProp {
   vehicle: Vehicle
 }
+interface VehicleOrGhostProp {
+  vehicle: VehicleOrGhost
```
KaylaBrady, line 20: thought (non-blocking): I like how `mapPage` now uses the name `selectedVehicleOrGhost` wherever type is `VehicleOrGhost`. I wonder if it is worth following that same naming convention in more places?

assets/src/components/vehiclePropertiesCard.tsx
firestack: I also see this as taking a component and styling it to look right, rather than generating a component to have the dom look exactly as the visual presentation looks. I'd like to/am going to keep it, because I think styling is the way to handle this, but I could be wrong!

assets/src/components/vehiclePropertiesCard.tsx
firestack: Thanks! PTALA fc35c035 🙏🏻

firestack:
I'm gonna merge at the end of retros if there's no comments before then, thanks all!

PR #1902, opened 2023-01-27
feat: Expose data needed for route properties card

KaylaBrady:

This surfaces all data needed for the route properties card via the API. Development work for the UI component has been started in https://github.com/mbta/skate/tree/kb-rpc and is at the point where I feel pretty good that this PR includes all the data we will need. 

Note: I added `route_pattern_id` to the vehicle instead of using a combination of `route_id`, `via_variant` and `direction` because the MBTA GTFS [doc](https://github.com/mbta/skate/tree/kb-rpc) doesn't guarantee that the ID will be a combination of those fields.


lib/realtime/vehicle.ex
```diff
@@ -188,6 +190,7 @@ defmodule Realtime.Vehicle do
       longitude: VehiclePosition.longitude(vehicle_position),
       direction_id: direction_id,
       route_id: route_id,
+      route_pattern_id: trip && trip.route_pattern_id,
```
lemald, line 193: **suggestion (non-blocking):** since the other fields like `headsign` that pull data from `trip` do the lookup logic above, it may be easier to read if this is instead grouped at the beginning of the function with them.

lemald (approved):

lemald (approved):
Glad the compression change was easy!

PR #1904, opened 2023-01-30
refactor(MapPage): Support selected vehicles + route patterns

KaylaBrady:

This refactor introduces a few changes to make adding the `RoutePropertiesCard` easier. The new card is not included in this branch to keep this limited to the refactor. 
* SearchPageState now has `selectedEntity`, as described by [this doc](https://www.notion.so/Route-Picker-Implementation-4284217f9f2f4648935ed9ddf898a42e)
* `MapDisplay` is split into a separate file. The tests have largely been copied over from `MapPage` tests.

This also includes a small fix for ensuring that vehicles are fully cleared between selections when the VPC is closed - see https://github.com/mbta/skate/pull/1904/commits/f74b246a0600fdd17d774b6d3f3cef9e93d8822f


assets/src/state/searchPageState.ts
```diff
@@ -6,12 +6,32 @@ import {
   SearchQuery,
 } from "../models/searchQuery"
 import { VehicleId } from "../realtime"
+import { RouteId, RoutePatternId } from "../schedule"
+
+export enum SelectedEntityType {
+  VEHICLE = 1,
+  ROUTE,
+}
```
lemald, line 14: **suggestion:** usually for enum types we've used title case for the values. See [`ConnectionStatus`](https://github.com/mbta/skate/blob/f49857bd10e68fdabd34f4a04cc6c86065ad1949/assets/src/hooks/useSocket.ts#L6) for instance.

assets/src/components/mapPage/vehiclePropertiesCard.tsx
```diff
@@ -4,14 +4,14 @@ import React, {
...
-import StreetViewButton from "./streetViewButton"
-import { VehicleRouteSummary } from "./vehicleRouteSummary"
+import { joinTruthy, className as classNames } from "../../helpers/dom"
+import { useCurrentTimeSeconds } from "../../hooks/useCurrentTime"
+import { useNearestIntersection } from "../../hooks/useNearestIntersection"
+import { isGhost, isVehicle } from "../../models/vehicle"
+import { Vehicle, VehicleOrGhost } from "../../realtime"
+import { CloseButton } from "../closeButton"
+import StreetViewButton from "../streetViewButton"
+import { VehicleRouteSummary } from "../vehicleRouteSummary"
```
lemald, line 14: **praise:** I like moving this into its own directory and creating a little bit more structure.

assets/src/models/routePatternData.ts
```diff
@@ -0,0 +1,40 @@
+import { define, Infer, nullable, number, string, type } from "superstruct"
+import { RoutePattern } from "../schedule"
+import { ShapeData, shapeFromData } from "./shapeData"
+
+const DirectionId = define<0 | 1>(
+  "DirectionId",
+  (value) => value === 1 || value === 0
+)
```
lemald, line 8: **thought (non-blocking):** I feel like this is going to be common to a lot of data types that we receive and handle via Superstruct, so it might be nice to start a specific module for those somewhere.

assets/src/hooks/useRoutePatterns.ts
```diff
@@ -0,0 +1,19 @@
...
+import { fetchRoutePatterns } from "../api"
+import { RouteId, RoutePattern } from "../schedule"
+
+const useRoutePatterns = (routeId: RouteId | null): RoutePattern[] | null => {
+  const [routePatterns, setRoutePatterns] = useState<RoutePattern[] | null>(
+    null
+  )
+  useEffect(() => {
+    if (routeId === null) {
+      setRoutePatterns([])
```
lemald, line 11: **question:** should this be `null` instead? Also, what about the race condition where you call `useRoutePatterns("some route")` and kick off a fetch, then immediately call `useRoutePatterns(null)`, but then the promise from the first call resolves and sets the route patterns to the value returned for `"some route"`?

KaylaBrady: Thanks for catching this race condition! On whether it should be null in this case, I've been thinking about null as representing data that is loading, but in this instance, there is no data to load. 

Both of these questions make me think about the `useFetch` library @firestack explored adding in a previous PR - looking at the docs it seems to handle the race conditions and clearer loading state. I may look into using that if it seems appropriate / reasonably low effort

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -0,0 +1,391 @@
...
+          }
+        : null
+    default:
+      return null
+  }
+}
+type LiveSelectedEntity =
+  | SelectedRoute
+  | { type: SelectedEntityType.Vehicle; vehicleOrGhost: VehicleOrGhost | null }
+  | null
```
lemald, line 143: **suggestion:** I got a little confused between this type and `SelectedVehicle` in `searchPageState.ts`, since here the value contains an actual vehicle or ghost but there it just has an ID. Similarly for `SelectedRoute` the payload is actually a combination of IDs, not a route data structure. Maybe calling the types in `searchPageState.ts` something like `SelectedVehicleId` and so on would make it a little bit clearer?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+  const routePatternOfSelectedEntity =
+    (routePatterns &&
+      selectedEntityRouteData?.routePatternId &&
+      routePatterns?.find(
+        (rp) => rp.id === selectedEntityRouteData.routePatternId
+      )) ||
+    null
+
+  return [routePatterns, routePatternOfSelectedEntity]
```
lemald, line 333: **question:** are there any places where we will actually use the full list of `routePatterns`?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -0,0 +1,391 @@
...
+    if (selectedVehicleId !== null && vehicleIdDiffers === false) {
+      setShouldFollow(true)
+    }
+  }, [vehicleIdDiffers, setShouldFollow, selectedVehicleId])
+  return state
+}
+
+const selectedEntityRoutePattern = (
+  selectedEntity: LiveSelectedEntity | null
+): { routeId: RouteId; routePatternId: RoutePatternId | null } | null => {
```
lemald, line 122: **suggestion:** it sort of confused me that the return value of this function isn't actually of type `RoutePattern`. Calling it `selectedEntityRoutePatternId` might be a little bit clearer but it's still not actually a route pattern ID. `selectedEntityRoutePatternIdentifier` might better?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -0,0 +1,391 @@
...
+  }
+}
+
+const useSelectedEntityRoutePatterns = (
+  liveSelectedEntity: LiveSelectedEntity | null
+): [RoutePattern[] | null, RoutePattern | null] => {
+  const selectedEntityRouteData = selectedEntityRoutePattern(liveSelectedEntity)
+  const routePatterns: RoutePattern[] | null = useRoutePatterns(
+    selectedEntityRouteData?.routeId || null
+  )
```
lemald, line 323: **suggestion:** if I understand correctly, regardless of whether we're dealing with a vehicle or route we will have an actual route pattern ID in `selectedEntityRouteData`, right? Setting aside shuttles, I suppose, but those don't have a route ID either. If that's the case, could we rework this so that instead there's a `useRoutePattern` hook that just takes the route pattern ID directly and queries the backend with that ID rather than route ID? That way we wouldn't have to `.find(...)` the right route pattern below based on its ID.

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Yeah I struggled a bit with the naming of that - `selectedEntityRoutePatternIdentifier` does seem more accurate

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Yeah that would be doable, and probably clearer! I was working with maybe a false assumption that fetching each route pattern shape could lead to some challenges keeping the selected state of the route pattern properties card in sync with the shape that is showing on the map. Though there is likely a more straight forward way to address that problem - I'll do some thinking and maybe set up some sync time after lunch!

assets/src/hooks/useRoutePatterns.ts
KaylaBrady: Opted for a lower-effort approach in [5add81d](https://github.com/mbta/skate/pull/1904/commits/5add81d752c260d82b89bb0632e3905ef40161b3)

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: I tried out the suggestion of `SelectedVehicleId` - let me know what you think!

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Closing the loop here from our sync discussion - these will be used in the upcoming RPC.

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Tried some new renaming here and simplified some of the nested calls, hopefully this helps but can make further changes if needed!

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Based on our sync discussion, updated this to use a map of route patterns instead of a list.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -0,0 +1,392 @@
...
+  const [stateClasses, setStateClasses] = useState<string | undefined>(
+    undefined
+  )
+
+  const routePatternShape =
+    allPatternsForRoute &&
+    routePatternIdentifier &&
+    shouldShowShape(liveSelectedEntity)
+      ? allPatternsForRoute[routePatternIdentifier.routePatternId]?.shape
+      : null
```
KaylaBrady, line 359: I was **really** hoping to do away with the `?` in `allPatternsForRoute[routePatternIdentifier.routePatternId]?.shape` here. I want it to be guaranteed that allPatternsForRoute contains the routePatternId of the selected vehicle/route. 

The only time it doesn't is when switching between vehicles: the routePatternIdentifier may be of the new vehicle, but allPatternsForRoute may lag behind with the patterns for the old vehicle. I was considering making this more transactional, but I'm thinking that may be out of scope based on our sync discussion on handling intermediate loading states.

assets/src/hooks/useRoutePatternsById.ts
```diff
@@ -9,10 +9,18 @@ const useRoutePatternsById = (
...
     } else {
-      fetchRoutePatterns(routeId).then(setRoutePatterns)
+      fetchRoutePatterns(routeId).then((routePatterns) => {
+        if (!canceled) {
+          setRoutePatterns(routePatterns)
+        }
+      })
+    }
+    return () => {
+      canceled = true
```
lemald: **suggestion:** might be worth adding a test for the race condition as well if it isn't too difficult.

lemald (approved):
One quick comment about an opportunity for more testing but overall this looks good!

assets/src/hooks/useRoutePatternsById.ts
KaylaBrady: Test added in [192c2ad](https://github.com/mbta/skate/pull/1904/commits/192c2ad935593841c3e2a862f9bd487f16342383) - I don't love the way it relies on the `patternsFor66DidResolve` variable though.

assets/src/components/ZoomLevelWrapper.tsx
```diff
@@ -0,0 +1,14 @@
+import { ReactElement, useState } from "react"
+import { useMap } from "react-leaflet"
+
+const ZoomLevelWrapper = ({
+  render,
+}: {
+  render: (zoomLevel: number) => ReactElement
+}) => {
```
firestack, line 8: nit(non-blocking): I'd personally like the `render` the prop to be named `children` instead, but I'm more curious as to if there was a style or personal reason for using `render` over any other name.

assets/css/_map_page.scss
```diff
@@ -141,7 +141,7 @@ $map-page-search-panel-width: 24rem;
 
 $map-page-search-stub-size: 10px;
 
-.m-map-page__map .m-vehicle-properties-card {
+.m-map-page__map .m-map-display__selected-entity-card-container {
```
firestack, line 144: praise: thank you!! I'm glad that this cleaned up easily on the css side

assets/tests/hooks/useVehiclesForRunIds.test.ts
```diff
@@ -1,68 +1,10 @@
...
-  end_of_trip_type: "another_trip",
-  block_waivers: [],
-  crowding: {
-    load: 1,
-    capacity: 57,
-    occupancy_status: "MANY_SEATS_AVAILABLE",
-    occupancy_percentage: 0.018,
-  },
-})
+const vehicleData: VehicleData = vehicleDataFactory.build()
```
firestack, line 7: suggestion(non-blocking): since there's only one test, are you okay with moving this vehicle into the test "arrange" section?

assets/tests/hooks/useVehicleForNotification.test.tsx
```diff
@@ -20,85 +24,11 @@ jest.mock("../../src/helpers/googleTagManager", () => ({
...
-    timepoint_id: "shunt",
-    fraction_until_timepoint: 0.3342478185618462,
-  },
-  scheduled_logon: null,
-  route_status: "on_route",
-  block_waivers: [],
-  current_piece_first_route: null,
-  current_piece_start_place: null,
-})
+const runId = "run_id"
```
firestack, line 27: suggestion(non-blocking): I added a `runIdFactory` that might help standardize runId formats across tests?

assets/tests/components/zoomLevelWrapper.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+afterAll(() => {
+  global.scrollTo = originalScrollTo
+})
+
+const renderProps = (zoomLevel: number) => <p>{zoomLevel}</p>
+
+const getMapZoomInButton = () => {
+  return screen.getByRole("button", { name: "Zoom in" })
+}
```
firestack, line 20: praise+thought(non-blocking): I'm glad you pulled this out into a funciton!
I was really liking writing my tests with functions to get the elements we work with or expect to exist in a certain form. 

Since we now have this specific function twice, do y'all want to consolidate common elements we use into a helper file somewhere in tests? 
I couldn't figure out a good place to put them for sharing.

firestack: @mbta/skate-developers

assets/tests/components/mapPage.test.tsx
```diff
@@ -826,14 +832,10 @@ describe("<MapPage />", () => {
...
 
-          const shape = shapeFactory.build({
-            stops: stopFactory.buildList(8),
-          })
-
           setHtmlWidthHeightForLeafletMap()
           mockUseVehiclesForRouteMap({})
           mockUseVehicleForId([selectedVehicle])
-          mockUseTripShape({ [selectedVehicle.tripId!]: shape })
+          mockUserRoutePatternsByIdForVehicles([]) // no route pattern for shuttle vehicle
```
firestack, line 838: note: btw, I think I wrote this, and if I remember correctly, I was providing a real route pattern for this test so that if the map rendered the route pattern for some reason, the assertion would fail[^1] if the map invalidated this. My understanding of the map component before was that if there's no shape to render, even if the map tried to render a shape, we wouldn't know or be able to fail the test.


[^1]: this assertion
```
expect(
    container.querySelector(".m-vehicle-map__route-shape")
).not.toBeInTheDocument()
```

firestack (approved):
🎉

assets/tests/components/zoomLevelWrapper.test.tsx
KaylaBrady: I definitely lifted this from your earlier tests - it makes writing/reading tests much easier! Agree it would be nice to consolidate them, though I'm not sure at what level. Seems like just one file could cause naming collisions and confusion, but I'm not sure what the appropriate line to draw would be

firestack: I originally was thinking that maybe keeping the relevant functions with their tests would be a good idea, but I think that means it would evaluate all the root level function calls (mock, describe, test) when trying to import a single function. So maybe it'd be better to have a file tree of matching "get component definitions" grouped by which `asset/components/*.tsx` it's from when it becomes relevant for each component we write tests for?

KaylaBrady: Yeah I like that idea! I'll plan on adding that in the [following PR](https://github.com/mbta/skate/pull/1906) which introduces more query functions.

firestack: Oh that's cool! I was wondering how we might make it more mailable/configurable! I'd say add it heh.

It seems like it hasn't been updated in a year or two, which is likely fine, but we should probably be aware that we might need to do security updates manually if it's unable to be maintained upstream.

I still wish there was a way to add extra search props, like `checked` `selected` `hidden`, but it's probably fine without it.

assets/src/components/map.tsx
```diff
@@ -461,27 +447,35 @@ export const BaseMap = (props: Props): ReactElement<HTMLDivElement> => {
...
+                  <StationMarker
+                    key={station.id}
+                    station={station}
+                    zoomLevel={zoomLevel}
+                  />
+                ))}
+              {zoomLevel >= 15 && <GarageMarkers zoomLevel={zoomLevel} />}
+            </>
+          )}
+        </ZoomLevelWrapper>
```
firestack, line 478: I rebased myself off of this and I'm having trouble with `RouteStopMarkers` being provided always instead of if there's shapes available, as well as we're always going to have two `ZoomLevelWrappers` if this stays in `BaseMap`, should we maybe move this into a `DefaultMap`?.. I'm struggling to figure out the right layers to allow a good default map dev experience but also allow escaping out of default elements we provide that we don't always use. But we're getting closer and closer to `BaseMap` providing a few defaults that we like and otherwise being declaratively specified through `children`, which then could be wrapped to provide the existing functionality.

firestack: I've been considering using leaflet-react's form of allowing disabling of default controls via props like `zoomControl={false}` but I've been trying to remove props from `BaseMap` lol

firestack: like, we could add a `zoomElements: (zoomLevel) => ReactElement` and call that within the `ZoomLevelWrapper` provided in `BaseMap` but I think it'd be better to do that in a wrapper of `BaseMap` instead.

KaylaBrady: Fully agree on your first point - I'll update to conditionally render <RouteStopMarkers/> only if there are actually stops to render. 

I don't totally mind the two separate ZoomLevelWrappers - it seems advantageous to me to have the garage rendering done by one of the lower level maps so that it is done consistently across all maps.

PR #1905, opened 2023-01-31
Disable views on search map

lemald:
Asana ticket: [⚙️ Better Maps | Disabled Left Nav View Buttons](https://app.asana.com/0/1200180014510248/1203641819945181/f)

This should address all of the acceptance criteria aside from the tooltip. The issue with the tooltip is that, if we set `disabled` on the button, the various events that might trigger the tooltip, like mousing over it, don't fire. It would probably be possible to work around this somehow, for instance by not actually using the `disabled` attribute, that would result in less standard HTML and presumably break web conventions. We'll revisit this and decide whether it's worth somehow including. A couple of other minor things:

1. I may revisit some of the styling of the SVG icons themselves in the disabled state based on design feedback.
2. After discussion on Slack we agreed to disable the notifications drawer in addition to swings view and late view.
3. One of the acceptance criteria is to disable notifications on the map page, but it looks like we already don't render the `Notifications` component there, so it should be taken care of.
4. For closing the open view when going to the map page, I opted to fire a dispatch to the state reducer inside of a `useEffect` hook from the page component. This is perhaps the part that I'm technically least sure about, so I'm open to feedback on better ways to do it.

### DIFF
--- a/assets/src/components/mapPage.tsx
+++ b/assets/src/components/mapPage.tsx
@@ -19,7 +19,7 @@ import useVehiclesForRoute from "../hooks/useVehiclesForRoute"
 import { filterVehicles, isGhost, isVehicle } from "../models/vehicle"
 import { Vehicle, VehicleId, VehicleOrGhost } from "../realtime"
 import { RouteId } from "../schedule"
-import { closeView } from "../state"
+import { closeView, OpenView } from "../state"
 import { SearchPageState, setSelectedVehicle } from "../state/searchPageState"
 import DrawerTab from "./drawerTab"
 import {
@@ -320,13 +320,16 @@ const MapDisplay = ({
 // #endregion
 
 const MapPage = (): ReactElement<HTMLDivElement> => {
-  const [{ searchPageState, mobileMenuIsOpen }, dispatch] =
+  const [{ searchPageState, mobileMenuIsOpen, openView }, dispatch] =
       useContext(StateDispatchContext),
     { selectedVehicleId = null } = searchPageState
 
   useEffect(() => {
-    dispatch(closeView())
-  }, [dispatch])
+    // don't dispatch closeView if the VPP is open
+    if (openView !== OpenView.None) {
+      dispatch(closeView())
+    }
+  }, [dispatch, openView])
 
   // #region Search Drawer Logic
   const [searchOpen, setSearchOpen] = useState<boolean>(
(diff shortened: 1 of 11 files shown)


assets/src/components/mapPage.tsx
```diff
@@ -323,6 +324,10 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
       useContext(StateDispatchContext),
     { selectedVehicleId = null } = searchPageState
 
+  useEffect(() => {
+    dispatch(closeView())
+  }, [dispatch])
+
```
KaylaBrady, line 330: question: I may have outdated memory on this, but my expectation was that if you open the VPP on the ladder page, go to map page, then back to the ladder page, the VPP would still be open. Is that no longer the expected behavior?

lemald: Ah good catch, I forgot that `closeView` also closes the VPP. I'll have to revisit that logic.

lemald: Okay should be fixed in 2774ae4c.

KaylaBrady: Appreciate the comment added too!

assets/src/components/nav.tsx
```diff
@@ -29,6 +28,7 @@ const Nav: React.FC<Props> = ({ children }) => {
               defaultToCollapsed={true}
               dispatcherFlag={readDispatcherFlag()}
               closePickerOnViewOpen={true}
+              allowViews={allowViews}
```
KaylaBrady, line 31: question: Is this change needed for the MobilePortrait nav too?

assets/src/components/nav/leftNav.tsx
```diff
@@ -235,26 +241,40 @@ const ViewToggle = ({
...
+      aria-disabled={disabled}
     >
       {icon}
       {collapsed ? null : name}
     </button>
   )
+
+  if (disabled) {
+    return (
+      <Tippy content="Not available in Search Map" placement="right">
```
KaylaBrady, line 271: thought (non-blocking): this seems to bake in an assumption about the reason for why the button is disabled. Seems okay for now while there is only one reason a button would be disabled, but maybe something that we'd want to think about parameterizing down the line.

KaylaBrady (approved):
Looks good, just one question about mobile!

assets/src/components/nav/leftNav.tsx
lemald: Yeah, that's a good point. Maybe instead of just taking a `disabled` prop of type boolean it could take an optional `disabledReason` prop that, if specified, would both disable the button and also add the string given as the contents of the tooltip.

PR #1906, opened 2023-02-01
feat(MapPage): Route Properties Card

KaylaBrady:


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -371,13 +476,23 @@ const MapDisplay = ({
       includeStopCard={true}
       stations={stations}
       shapes={routePatternShape ? [routePatternShape] : []}
+      // TODO: try rendering shapes as a child instead
```
KaylaBrady, line 479: I want to avoid adding `onShapeSelect` as a base map prop. I think it would be ideal to just pass the `shapes` with their `onClick` as children to the `BaseMap`. @firestack I can't find the thread, but I remember you previously advocating for an approach passing children to the `BaseMap` over adding props. 

I'm interested to try it at this point since it may provide other opportunities to simplify some of the `liveSelectedEntity` complexity in MapDisplay. However,  the `RouteStopMarkers` component is currently dependent on zoom level, and I'm concerned that it would conflict with your current stop size work.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -312,6 +338,71 @@ const SelectedVehicleDataLayers = ({
...
+  setStateClasses: (classes: string | undefined) => void
+}) => {
+  const selectedRoutePattern: RoutePattern | undefined = routePatterns
+    ? routePatterns[routePatternIdentifier.routePatternId]
+    : undefined
+  const position = selectedRoutePattern
+    ? selectedRoutePattern.shape?.points?.map((p) =>
+        Leaflet.latLng(p.lat, p.lon)
+      ) || []
+    : []
```
lemald, line 366: **nitpick:** it would probably be slightly clearer to name this `positions` since it's a list.

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+            </label>
+          </div>
+        ))}
+      </>
+    )
+  }
+}
+
+const RoutePropertiesCard = ({
+  routeId,
```
lemald, line 105: **suggestion:** if you want to minimize the number of props, you could get `routeId` directly from `selectedRoutePattern` once that's known.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -312,6 +338,71 @@ const SelectedVehicleDataLayers = ({
...
+  const followerState = useInteractiveFollowerState()
+
+  useEffect(() => {
+    setStateClasses(FollowerStatusClasses(followerState.shouldFollow))
+  }, [followerState.shouldFollow, setStateClasses])
```
lemald, line 371: **suggestion:** I'm not actually sure that you need a `useEffect` here in order to prevent excessive re-renders: according to the documentation, [state updates are ignored if the new value equals the old](https://beta.reactjs.org/reference/react/useState#ive-updated-the-state-but-the-screen-doesnt-update).

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+      }
+      {[0, 1].map((directionId) => (
+        <div key={directionId}>
+          <input
+            type="radio"
+            id={`direction-radio-${directionId}`}
+            name="direction"
+            defaultChecked={selectedRoutePattern.directionId === directionId}
+            onChange={() => {
+              const targetDirectionId = directionId === 0 ? 1 : 0
```
lemald, line 45: **question:** How necessary is this line? I thought it might be to get TypeScript to do type narrowing on `directionId`, but I tried locally removing `targetDirectionId` and using `directionId` directly in the `find` below and it didn't complain.

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    routePattern.directionId === targetDirectionId
+                ) || null
+
+              if (targetRoutePattern) {
+                selectRoutePattern(targetRoutePattern)
+              }
+            }}
+          />
+          <label htmlFor={`direction-radio-${directionId}`}>
+            {directionNames[directionId as DirectionId]}
```
lemald, line 58: **suggestion:** I like avoiding `as` wherever possible since that causes us to loose out on TypeScript's typechecking. I was able to rewrite things slightly as following and get it to pass:

```
  const directionIds: DirectionId[] = [0, 1]

  return (
    <div className="m-route-properties-card__direction-picker">
      {
        // TODO: Should we always show both buttons? Some routes only go one direction?
      }
      {directionIds.map((directionId: DirectionId) => (
        <div key={directionId}>
          <input
            type="radio"
            id={`direction-radio-${directionId}`}
            name="direction"
            defaultChecked={selectedRoutePattern.directionId === directionId}
            onChange={() => {
              const targetDirectionId = directionId === 0 ? 1 : 0
              const targetRoutePattern =
                sortRoutePatterns(Object.values(routePatterns)).find(
                  (routePattern) =>
                    routePattern.directionId === targetDirectionId
                ) || null

              if (targetRoutePattern) {
                selectRoutePattern(targetRoutePattern)
              }
            }}
          />
          <label htmlFor={`direction-radio-${directionId}`}>
            {directionNames[directionId]}
          </label>
        </div>
      ))}
    </div>
  )
```

It's maybe a bit more verbose but I wasn't able to figure out a syntactically-correct way of typing `[0, 1]` inline with the call to `.map`.

assets/tests/components/mapPage.test.tsx
```diff
@@ -606,10 +598,103 @@ describe("<MapPage />", () => {
...
+              ladderVehicleLabel: VehicleLabelSetting.VehicleNumber,
+            },
+          })}
+        >
+          <RoutesProvider routes={[route]}>
+            <MapPage />
+          </RoutesProvider>
+        </RealDispatchWrapper>
+      )
+      expect(vehiclePropertiesCard.get()).toBeVisible()
```
lemald, line 644: **question:** how necessary is this assertion for this test?

assets/tests/components/mapPage/mapDisplay.test.tsx
```diff
@@ -322,7 +388,111 @@ describe("<MapPage />", () => {
...
+                }}
+                setSelection={jest.fn()}
+                showSelectionCard={true}
+              />
+            </RoutesProvider>
+          )
+
+          expect(routePropertiesCard.query()).not.toBeInTheDocument()
+        })
+        test("clicking vehicle route dispatches setSelection event", async () => {
```
lemald, line 452: **suggestion:** the "clicking vehicle route..." part of the title is a little confusing compared to what the test actually does, maybe "selecting pattern in RPC dispatches setselection event" or something like that would be better?

lemald:
@KaylaBrady okay I'm through my first round of review, added a few comments but I think the rest looks good!

assets/src/components/mapPage/routePropertiesCard.tsx
KaylaBrady: Not entirely necessary - I found it helpful in thinking through the logic to name it, but there is actually a bug here so maybe it wasn't that helpful after all! Since onChange is only called when the radio button becomes checked, this should be targeting the directionId represented by the input, not switching the direction.

assets/tests/components/mapPage.test.tsx
KaylaBrady: I'll reword this test - I think it is useful for testing that the VPC and RPC are mutually exclusive, but that isn't totally captured by what is here.

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: It seems that the `useEffect` is needed here, not for preventing excessive re-renders, but rather for performing the state update of `stateClasses` as a side effect. When removing this, I get the following errors in tests:
```
console.error
      Warning: Cannot update a component (`MapDisplay`) while rendering a different component (`SelectedRouteDataLayers`). To locate the bad setState() call inside `SelectedRouteDataLayers`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

```

lemald: Ahh, okay, my bad.

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  isOpen: boolean
+  toggleOpen: () => void
+  className: string
+  children: JSX.Element
+}) => {
+  return (
+    <details
+      className={`m-route-properties-card__details-section ${className}`}
+      {...(isOpen ? { open: true } : {})}
+    >
```
firestack, line 42: question/suggestion: what is the reasoning behind the object spread here?

Does this maybe work here?

```suggestion
open={isOpen}
```

This might be closer to the current code
```suggestion
open={isOpen || undefined}
```

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+      <summary
+        onClick={(event) => {
+          // preventDefault so that open/closed state is managed entirely by react
+          // see https://github.com/facebook/react/issues/15486
+          event.preventDefault()
+          toggleOpen()
+        }}
+      >
+        {`${isOpen ? "Hide" : "Show"} ${title}`}
```
firestack, line 51: I dunno if I'm changing the logic here with this suggestion

```suggestion
        {isOpen ? "Hide" : "Show"} {title}
```

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+  return (
+    <div className="m-route-properties-card" aria-label="route properties card">
+      <Card
+        style={"white"}
+        noFocusOrHover={true}
+        title={
+          <>
+            <div className="route-title">
+              <RoutePill routeName={routeId} />
+              <h4>{displayName(selectedRoutePattern)}</h4>
```
firestack, line 204: question+suggestion: why `<h4>` here?

I think `<h2>` would be more approprate, but that's because I think of the RPC as is it's own section which restarts the header index to 2.

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,259 @@
...
+      <>
+        {sortRoutePatterns(routePatterns).map((routePattern) => (
+          <VariantOption
+            key={routePattern.id}
+            routePattern={routePattern}
+            isSelected={routePattern.id === selectedRoutePatternId}
+            selectRoutePattern={selectRoutePattern}
+          />
+        ))}
+      </>
```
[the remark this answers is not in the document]
KaylaBrady: Yes, thanks for suggesting!

assets/src/components/mapPage/routePropertiesCard.tsx
KaylaBrady: Yes, this is much nicer, thanks! I was thinking we had to leave off the `open` attribute entirely when it is not open, but setting to `undefined` is better!

KaylaBrady: The reason  `open={isOpen}` doesn't work is because the [attribute is a boolean ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attr-open), which tripped me up for a bit!

firestack: I learned[^1] that jsx properly handles the `false` state on [boolean attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean) , so you might try leaving off the `|| undefined` part even.

[^1]: this attribute worked with `boolean` even though it was supposed to be left off if `hidden=false` https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden so I've learned that default html props in jsx are not always 1 to 1 with the actual resulting html output

firestack: > The reason `open={isOpen}` doesn't work is because the [attribute is a boolean ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attr-open), which tripped me up for a bit!

Oh interesting, maybe it's inconsistent or requires jsx to implement extra handling on top of how it renders html. `hidden` isn't a `boolean` attribute but I expected that `boolean` attributes might work.

KaylaBrady: Ah you know what there must have been something else at play when this first wasn't working for me - trying again with `open={isOpen}` works!

KaylaBrady:
@mbta/skate-developers thanks for your initial reviews! I believe all your comments have been addressed, plus some minor styling/functionality tweaks. Marking ready for re-review!

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -0,0 +1,302 @@
...
+        }}
+      >
+        {isOpen ? "Hide" : "Show"} {title}
+      </summary>
+      {children}
+    </details>
+  )
+}
+
+const routePatternForTargetDirection = (
```
KaylaBrady, line 70: logic pulled from https://github.com/mbta/dotcom/blob/d22bec8822852063671626ba973c36573af44065/apps/site/assets/ts/schedule/components/direction/reducer.ts#L72

assets/tests/components/mapPage/routePropertiesCard.test.tsx
```diff
@@ -0,0 +1,356 @@
...
+  test("When not formatted correctly with ' - '", () => {
+    const routePattern = routePatternFactory.build({
+      name: "BAdFormatted RP",
+    })
+
+    expect(patternDisplayName(routePattern)).toEqual({
+      name: routePattern.name,
+      description: "",
+    })
+  })
```
lemald, line 27: **suggestion:** I guess this is more a comment on the application code than the test but this got me thinking - I think ideally we would want to at least display something in this case, even if it's not exactly what we want. I worry that displaying nothing when the route pattern name isn't formatted to match our expectations introduces too much of a dependency on the details of the GTFS data.

assets/css/_route_properties_card.scss
```diff
@@ -0,0 +1,158 @@
...
+  padding: 0.5rem 0.5rem;
+  justify-content: space-between;
+  margin: 0;
+}
+
+.m-route-properties-card .route-title {
+  display: flex;
+  flex-direction: row;
+  gap: 0.5rem;
+  & h2 {
```
lemald, line 18: **suggestion:** I don't think you need the `&` here or in the selector immediately below, since the descendant combinator is the default for nested SCSS selectors.

lemald: Looks like this applies to a few other selectors throughout the file as well.

assets/tests/components/mapPage/routePropertiesCard.test.tsx
lemald: Ahh, okay, makes sense. It's clear, I just didn't read carefully enough. 🙂

assets/css/_route_properties_card.scss
KaylaBrady: Great, thank you! Also doing some reading up on BEM and will likely change some class names to follow the conventions more closely

lemald (approved):
I think that covers all of my feedback!

assets/css/_route_properties_card.scss
```diff
@@ -0,0 +1,160 @@
...
+
+.m-route-properties-card hr {
+  height: 1px;
+  background-color: $color-gray-400;
+  border: none;
+}
+
+.m-route-properties-card__details-section.variant-stop-list {
+  max-height: 490px;
+  overflow: scroll;
```
firestack, line 156: I get weird scrollbar artifacts on firefox and chrome around the outbound stops, setting to auto seems to ensure that scroll is used for overflow, but doesn't show scrollbars unless an overflow happens.


```suggestion
  overflow: auto;
```

firestack (commented):
I'd like [this comment](https://github.com/mbta/skate/pull/1906#discussion_r1098646857) addressed but otherwise LGTM!

firestack (approved):
I'd like [this comment](https://github.com/mbta/skate/pull/1906#discussion_r1098646857) addressed but otherwise LGTM!

assets/css/_route_properties_card.scss
KaylaBrady: Thanks for the suggestion! I also accidentally removed the `ol` selector here - we only want the scroll to affect the list of stops itself, not the summary too. Will add that back!

KaylaBrady:
@mbta/skate-developers mind taking one last look for the FS event? https://github.com/mbta/skate/pull/1906/commits/05555303bfe556eb017038f20e9bdbfcabb21812

firestack (approved):

firestack (approved):

PR #1907, opened 2023-02-02
Map page-related FullStory events

lemald:
Asana ticket: [⚙️ Better Maps | Custom event for maps](https://app.asana.com/0/1200180014510248/1203500886824771/f)

This should implement all of the events aside from the one for the route properties card, as the RPC is still under development in a different branch. We can coordinate on how to get the RPC event added.


assets/src/components/map.tsx
```diff
@@ -125,7 +125,13 @@ const StreetViewControl = ({
           type="checkbox"
           role="switch"
           checked={streetViewEnabled}
-          onChange={() => setStreetViewEnabled((enabled) => !enabled)}
+          onChange={() => {
+            if (!streetViewEnabled && window.FS) {
+              window.FS.event("Dedicated street view enabled")
+            }
```
firestack, line 131: I dunno how this would look in the fullstory UI, but given that it accepts structured data as well, maybe we could use metadata attributes to communicate state instead?

Also, I think the `?.` operator would remove the need for checking if it exists first?

```suggestion
            window.FS?.event("Dedicated street view", { enabled: streetViewEnabled})
```

lemald: I guess when I read this acceptance criteria:

> Custom event that tracks when any dispatcher toggles the dedicated Street View entrypoint

I mentally filled that in to mean tracking when a dispatcher toggles dedicated Street View _on_, which may not be quite correct. I'll ask product + design.

lemald: I did just change all of the calls to use `window.FS?.event` rather than wrapping them in an `if` statement in c1d2f7fd.

lemald: ...and changes to the event itself are in 72d348af!

assets/src/components/nav/leftNav.tsx
```diff
@@ -112,6 +112,9 @@ const LeftNav = ({
               }
               title={mapMode.title}
               to={mapMode.path}
+              onClick={() => {
+                mapMode.navEventText && window.FS?.event(mapMode.navEventText)
```
firestack, line 116: nit(non-blocking): this seems like it's not part of the requirements product asked for, but I think having dynamic variables be structured event data instead of the actual message is a more stable interface.

placeholder event text and key are just examples, I'm not settled on those as final, but only for demonstration.

```suggestion
                mapMode.navEventText && window.FS?.event("clicked navigation tab", { mode: mapMode.navEventText })
```

assets/src/components/streetViewButton.tsx
```diff
@@ -35,6 +35,9 @@ const StreetViewButton = ({
     href={streetViewUrl(worldPosition)}
     target="_blank"
     rel="noreferrer"
+    onClick={() => {
+      window.FS?.event("Street view link followed")
```
firestack, line 39: thought(non-blocking): would it be beneficial to save("save" = structured event data 😉 ) the street view link so that we can review later if the links are useful?

assets/tests/components/searchForm.test.tsx
```diff
@@ -114,6 +115,31 @@ describe("SearchForm", () => {
...
+    const onSubmit = jest.fn()
+    const validSearch: SearchPageState = {
+      query: { text: "12", property: "run" },
+      isActive: false,
+      savedQueries: [],
+    }
+    const validSearchState = {
+      ...initialState,
+      searchPageState: validSearch,
+    }
```
firestack, line 129: nit: we have factories for both these states that might be easier to use?

firestack (approved):
A few comments/thoughts but nothing stands out and LGTM 🎉

assets/tests/components/map.test.tsx
```diff
@@ -277,6 +278,33 @@ describe("<Map />", () => {
...
+      <Map vehicles={[]} allowStreetView={true} reactLeafletRef={mapRef} />
+    )
+
+    await userEvent.click(screen.getByRole("switch", { name: /Street View/ }))
+
+    await userEvent.click(screen.getByRole("switch", { name: /Street View/ }))
+
+    expect(window.FS!.event).toHaveBeenCalledWith(
+      "Dedicated street view toggled",
+      { streetViewEnabled: false }
```
firestack, line 306: oh hey btw, the docs seem to say that they need a type suffix https://help.fullstory.com/hc/en-us/articles/360020623234

I've not seen the fullstory UI so I dunno what this does or looks like when you comply or not.
They also say that you can use camel case or underscores, and seem to suggest underscores on their [event api docs](https://help.fullstory.com/hc/en-us/articles/360020623274-FS-event-API-Sending-custom-event-data-into-FullStory#Examples), so we should probably standardize on camelCase or underscores.

Because the type suffix seems to need to start with an underscore, I think we should standardize on underscores.

lemald: Ah, thanks for catching this! I'll update it to match their patterns.

lemald: The docs at the first link you sent seem to have a mix of camelcase and snakecase, since camelcase is more in line with our variable naming on the rest of the Skate frontend I think I'll stick with that.

lemald: Fixed in 0ddbf285!

PR #1908, opened 2023-02-03
feat(map-stops): tap targets selection state and style+size updates

firestack:
I grouped vehicles into panes to make debugging in devtools easier as well as making z-index/stacking context a little easier to manage, we could do this manually by setting z-index per item if that'd be preferred, but I liked the way that panes helped keep things grouped. A difficulty here is that I don't think `BaseMap` should have as many props as it does and provide as much default functionality as it does, mainly because we'd have to add more props to allow users of `BaseMap` to put things into the pane's that it reserves for itself.

I also dislike that pane's specify their z-index via the `style` attribute, but there wasn't a `zindexOffset` like there is for `Markers`. Leaflet seems to set z-index according via css which matches the [pane `z-index` values provided in the leaflet documentation](https://leafletjs.com/reference.html#map-tilepane)

### Implemented AC:
- [x] Deploy to all users 
  - [x] Implement hover states + styling outlined in the Figma file 
  - [x] Change the tap target size to 16x16  
  - [x] The z-index of the tap target should place it above some of the other non-interactive elements on the map (route shapes, other buses on the same route).
	  - This was applied to only the `BaseMap` for things like the `Search Page`, `MiniMap`, and `Shuttle Page`, because the `MapPage` manages it's own stacking context/order and does not have the ability to insert elements into the specific panes otherwise.

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1203222524521880/f

### DIFF
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -79,12 +79,27 @@
 .m-vehicle-map__stop {
   fill: $color-stop-fill;
   fill-opacity: 1;
-  stroke: $color-stop-stroke;
-  stroke-width: 1;
-}
 
-.m-vehicle-map__stop:focus {
-  outline: 0;
+  stroke: $color-stop-stroke;
+  stroke-width: 2px;
+
+  &:hover {
+    stroke-width: 4px;
+  }
+
+  &:focus {
+    stroke-width: 4px;
+    box-shadow: 0px 0px 0px 1px #514f55;
+    filter: drop-shadow(0 0 2px #514f55);
+    outline: 0;
+  }
+
+  &.selected {
+    stroke-width: 4px;
+
+    fill: $color-eggplant-200;
+    stroke: $color-eggplant-700;
+  }
 }
 
 .leaflet-tooltip.m-vehicle-map__mobile-friendly-tooltip {
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -89,7 +89,6 @@
 
   &:focus {
     stroke-width: 4px;
-    box-shadow: 0px 0px 0px 1px #514f55;
     filter: drop-shadow(0 0 2px #514f55);
     outline: 0;
   }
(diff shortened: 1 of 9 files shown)


assets/css/_vehicle_map.scss
```diff
@@ -79,12 +79,27 @@
...
+
+  &:hover {
+    stroke-width: 4px;
+  }
 
-.m-vehicle-map__stop:focus {
-  outline: 0;
+  &:focus {
+    stroke-width: 4px;
+    box-shadow: 0px 0px 0px 1px #514f55;
```
firestack, line 92: Figma designs used this box-shadow property, but it doesn't seem to actually be applying for me, so I added the filter below, but I've got to run that by design first.

Ideally design can help me decide what to do, but this line should likely be removed.

assets/src/components/mapMarkers.tsx
```diff
@@ -217,6 +228,38 @@ const MobileFriendlyTooltip = ({
...
+  selected = false,
+  ...props
+}: StopIconProps) => {
+  const ref = useRef<null | LeafletCircleMarker>(null)
+  const element = ref.current?.getElement()
+  useEffect(() => {
+    if (element) {
+      element.classList.toggle("selected", selected)
+    }
+  }, [selected, element])
```
lemald, line 249: **question:** since `element` comes from a `ref`, that means changes to it won't necessarily trigger a re-render, correct, so this `useEffect` might not actually run when the `element` changes, right? Or am I missing something?

firestack: Yeah you're right! 

I don't have a good understanding of leaflet's lifecycle and when `getElement()` can return `undefined`, and I was hoping the `useEffect` would help here, but it actually introduced a rendering error where it only respected `selected` on rerender.
I'm not quite sure what the right path forward is here, I've gotten it working a couple of different ways, but I've settled on a `useState` method that I'll have up in a bit for re-review.

firestack: Please take a look at d7cbf91d87136cd2ba7ccb42a8a7434d23e4cf3c 🙏🏻

lemald: I always forget that you can pass setters into `ref`! Yes, that looks good.

lemald (approved):

assets/css/_vehicle_map.scss
firestack: [`6100fb4` (#1908)](https://github.com/mbta/skate/pull/1908/commits/6100fb4d1d7e6a0d6cdfa616aa8ac0a6e933c37b)

PR #1910, opened 2023-02-06
fix: position children in the same spot under Nav in all variants

lemald:
Asana ticket: [⚙️🐞 Better Maps | Investigate how to not re-render components when breakpoint changes ](https://app.asana.com/0/1148853526253437/1203180633835795/f)

This just reorders the children of `<Nav />` so that, regardless of which variant of the nav is rendered, the children end up under a `<div>` which is the first child of the main `<div>` of the overall nav component. This keeps the children in the same place in the UI tree and therefore causes React to preserve their state. The React documentation on [Preserving and Resetting State](https://beta.reactjs.org/learn/preserving-and-resetting-state) was a useful reference here (kudos to @firestack for pointing me to it).

### DIFF
--- a/assets/src/components/nav/mobilePortraitNav.tsx
+++ b/assets/src/components/nav/mobilePortraitNav.tsx
@@ -9,11 +9,7 @@ import {
 import BottomNavMobile from "./bottomNavMobile"
 import TopNavMobile from "./topNavMobile"
 
-const MobilePortraitNav = ({
-  children,
-}: {
-  children?: React.ReactNode | undefined
-}): JSX.Element => {
+const MobilePortraitNav = (): JSX.Element => {
   const [state, dispatch] = useContext(StateDispatchContext)
 
   const { mobileMenuIsOpen, routeTabs, openView, selectedVehicleOrGhost } =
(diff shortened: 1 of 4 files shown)


assets/src/components/nav/mobilePortraitNav.tsx
```diff
@@ -25,6 +25,8 @@ const MobilePortraitNav = ({
 
   return (
     <div className="m-nav--narrow">
+      <div className="m-nav__app-content">{children}</div>
```
KaylaBrady, line 28: question (non-blocking): There may be an issue with how I'm testing this since chrome doesn't seem to let you resize down to "mobile" sizes, but when I switch to responsive mode in the inspector tools and go from desktop => mobile, the state is lost. Is this an artifact of how I'm testing? Though maybe not an issue if it is impossible for users to resize their desktop windows that small anyway!

KaylaBrady (approved):

assets/src/components/nav/mobilePortraitNav.tsx
lemald: Interesting - in Firefox you can resize the window narrow enough to hit the mobile breakpoint and I still see the state resetting. I think there must be some combination of the React element tree and the DOM at play here, but I think I can refactor things a bit more in order to get it all to work.

lemald: I think that f0bcf458 should fix this, though I'm still not sure if the process you described in Chrome dev tools will preserve the state or not.

KaylaBrady: Yes, I'm not seeing the state reset anymore in chrome either! 🎉

firestack (approved):

PR #1913, opened 2023-02-08
feat(skate/fullstory): add Skate GA custom events

firestack:
I implemented most of these by following how the Google Analytics tags were done, a few might need review to refine where the event is triggered to be more accurate to the AC.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1203449129917720/f


assets/tests/hooks/useVehicleForNotification.test.tsx
```diff
@@ -131,12 +137,16 @@ describe("useVehicleForNotification", () => {
     expect(tagManagerEvent).toHaveBeenCalledWith(
       "notification_linked_to_inactive_modal"
     )
+    expect(window.FS!.event).toBeCalledWith(
+      "User clicked Notification but it Failed"
```
firestack, line 141: Thinking of renaming this event

assets/src/components/inputModals/createPresetModal.tsx
```diff
@@ -24,8 +24,12 @@ const CreatePresetModal = ({
           event.preventDefault()
           const existingPreset = findPresetByName(routeTabs, presetName)
           if (existingPreset) {
+            window.FS?.event(
+              "User tried to Save new Preset under existing Name"
```
KaylaBrady, line 28: question (non-blocking): Is this necessary if the save/cancel overwrite events are separately tracked? If so, the wording "User tried" makes me think they tried but didn't succeed. Maybe something like "User trying to..." since there are still confirmation steps to take?

assets/src/components/nav/bottomNavMobile.tsx
```diff
@@ -66,6 +66,7 @@ const BottomNavMobile: React.FC<Props> = ({
             className="m-bottom-nav-mobile__button"
             onClick={() => {
               tagManagerEvent("swings_view_toggled")
+              window.FS?.event("User opened Swings View")
```
KaylaBrady, line 69: question (non-blocking): Is it possible to repeatedly click this button, similar to the ViewToggle in leftNav?

assets/src/components/swingsView.tsx
```diff
@@ -280,7 +280,10 @@ const SwingRow = ({
           <SwingCellContent
             vehicleOrGhost={swingOnVehicleOrGhost}
             runId={swing.toRunId}
-            tagManagerEventText="clicked_swing_on"
```
KaylaBrady, line 283: praise: I like this being removed in favor of onClick!

assets/src/util/isChelseaBridgeNotification.ts
```diff
@@ -0,0 +1,13 @@
...
+  ChelseaBridgeNotificationReason,
+} from "../realtime"
+
+export function isChelseaBridgeNotification(
+  reason: NotificationReason
+): reason is ChelseaBridgeNotificationReason {
+  return (
+    reason === "chelsea_st_bridge_lowered" ||
+    reason === "chelsea_st_bridge_raised"
+  )
```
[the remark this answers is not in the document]
KaylaBrady: There is also a `/src/helpers` directory - not suer which is more appropriate in this case. But maybe generically named `notifications` file in either directory so it is easy to add other notifications-related utils down the line!

assets/tests/components/ladderPage.test.tsx
```diff
@@ -199,6 +201,7 @@ describe("LadderPage", () => {
   })
 
   test("can save a route tab as a preset from the save icon", async () => {
+    mockFullStoryEvent()
```
KaylaBrady, line 204: question: Would we want this mocked for all tests? Maybe worth pulling into `setup.tsx`

KaylaBrady (approved):
Looks good! A few minor comments.

assets/src/util/isChelseaBridgeNotification.ts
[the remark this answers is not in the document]
lemald: Yeah I'm honestly not sure why we have both `util` and `helpers`. It might be best to just consolidate them.

assets/tests/components/ladderPage.test.tsx
firestack: I like the declarative approach I've been taking by only doing it per test that needs it, and nothing should fail because all `FS.event` calls use typescript's null operator `window.FS?.event`.
But I'm open to putting it into `setup.tsx`

assets/src/components/propertiesPanel/crowdingDiagram.tsx
```diff
@@ -22,20 +22,34 @@ const CrowdingDiagram = ({ crowding }: { crowding: Crowding | null }) => {
...
     <div className="m-crowding-diagram">
       <div className="m-crowding-diagram__properties">
-        <span className="m-properties-list__property-label">
+        <label
+          className="m-properties-list__property-label"
+          htmlFor="riders-onboard"
+        >
           Riders onboard
-        </span>
+        </label>
```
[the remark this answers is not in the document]
firestack: product approved 🎉

assets/src/components/lateView.tsx
```diff
@@ -512,6 +516,9 @@ const LateBusRow = ({
           className="m-late-view__run-link"
           onClick={() => {
             tagManagerEvent("selected_late_view_run_number")
+            window.FS?.event("User clicked Late View Run Number", {
+              isGhost_bool: false,
+            })
```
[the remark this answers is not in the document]
firestack: product approved 🎉

assets/src/hooks/useGarageFilter.tsx
```diff
@@ -89,6 +89,9 @@ export const GarageFilter = ({
                 onClick={() => {
                   if (!filteredGarages.includes(garage)) {
                     tagManagerEvent("filtered_routes_by_garage")
+                    window.FS?.event("User filtered Route Selector by Garage", {
+                      garageName_str: garage,
```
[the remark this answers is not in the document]
firestack: product approved 🎉

assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -218,6 +219,11 @@ export const useNotificationsReducer = (
 
       if (latestMessage.type === "new") {
         tagManagerEvent("notification_delivered")
+        if (isChelseaBridgeNotification(latestMessage.payload.reason)) {
+          window.FS?.event("User was Delivered a Chelsea Bridge Notification")
+        } else {
+          window.FS?.event("User was Delivered a Notification")
+        }
```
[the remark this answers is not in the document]
firestack: product doesn't care either way, lighter lift was the target.

assets/src/components/inputModals/createPresetModal.tsx
firestack: I don't think it's nescisary, and it wasn't in AC, so I'm open to changing it. I had a feeling that this might be useful as a metric which is why I added it (and it was easy with how tests were configured).

assets/tests/components/propertiesPanel/crowdingDiagram.tippy.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from "react"
+import CrowdingDiagram from "../../../src/components/propertiesPanel/crowdingDiagram"
+import { Crowding } from "../../../src/models/crowding"
+import crowdingFactory from "../../factories/crowdingFactory"
+import { mockFullStoryEvent } from "../../testHelpers/mockHelpers"
+
+// Tippy sets some accessability attributes that is used for testing
+// which means it's better to not mock Tippy
+// Due to this unmock interacting badly with other tests
+jest.unmock("@tippyjs/react")
```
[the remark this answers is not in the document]
firestack: I'm not having any luck with that. I've read some stuff that says jest uses hoisting to make it's mock calls happen before imports, and doing any unmocking seems to require hoisting, which doesn't work(for me, so far) if it's invoked within a describe/test/or other function.

because tippy is mocked for the entire application, my guess/observation is that the unmock needs to happen at the root of the module so it gets hoisted properly.

I could possibly unmock at the root, and then mock it for the regular tests using beforeAll/afterAll?
Any preferences? I'm leaning toward leaving it because it's easier.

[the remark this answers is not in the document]
KaylaBrady: Yeah that is more of a pain than I thought it would be - leaving it seems good to me 😃

assets/src/components/inputModals/createPresetModal.tsx
firestack: Maybe this?
> User manually typed existing Saved Preset name

Trying to think about what would be most descriptive and helpful when looking at this on the fullstory timeline or search has been helpful but I'm still unsure what is the best descriptor for what is happening.

firestack: Heh I've come up with a few

> User manually typed existing Saved Preset name
> User trying to save over existing Saved Preset
> User submitted name of existing preset when creating a new Saved Preset

firestack: PTALA 🙏🏻 6e5975e9f57bddb9da40e8afc4a811e27f067dc3

KaylaBrady: LGTM!

PR #1914, opened 2023-02-08
Loading state for current intersection in VPC

lemald:
Asana ticket: [⚙️🐛 Show Loading state for current location when VPC changes selected vehicle](https://app.asana.com/0/1200180014510248/1203674636294153/f)

I opted to create some fairly generic types for representing various states in the lifecycle of fetching data from an external source. After some initial attempts at making the result more composable (h/t to @firestack for the initial template I started with), I opted for an approach that doesn't result in as many composable / generically useful types, but is at least fairly readable in my opinion as the overall type is a sum type of each of the various possible states.

### DIFF
--- a/assets/src/components/mapPage/vehiclePropertiesCard.tsx
+++ b/assets/src/components/mapPage/vehiclePropertiesCard.tsx
@@ -9,6 +9,7 @@ import { useCurrentTimeSeconds } from "../../hooks/useCurrentTime"
 import { useNearestIntersection } from "../../hooks/useNearestIntersection"
 import { isGhost, isVehicle } from "../../models/vehicle"
 import { Vehicle, VehicleOrGhost } from "../../realtime"
+import { isOk } from "../../util/fetchResult"
 import { CloseButton } from "../closeButton"
 import StreetViewButton from "../streetViewButton"
 import { VehicleRouteSummary } from "../vehicleRouteSummary"
@@ -121,7 +122,12 @@ const CurrentLocation = ({ vehicle }: VehicleProp): React.ReactElement => {
     vehicle.latitude,
     vehicle.longitude
   )
-  return <>{intersection ?? "Exact location cannot be determined"}</>
+
+  if (isOk(intersection)) {
+    return <>{intersection.ok}</>
+  }
+
+  return <>Exact location cannot be determined</>
 }
 
 const VehicleNearestIntersection = ({
--- a/assets/src/components/mapPage/vehiclePropertiesCard.tsx
+++ b/assets/src/components/mapPage/vehiclePropertiesCard.tsx
@@ -9,8 +9,9 @@ import { useCurrentTimeSeconds } from "../../hooks/useCurrentTime"
 import { useNearestIntersection } from "../../hooks/useNearestIntersection"
 import { isGhost, isVehicle } from "../../models/vehicle"
 import { Vehicle, VehicleOrGhost } from "../../realtime"
-import { isOk } from "../../util/fetchResult"
+import { isLoading, isOk } from "../../util/fetchResult"
 import { CloseButton } from "../closeButton"
+import Loading from "../loading"
 import StreetViewButton from "../streetViewButton"
 import { VehicleRouteSummary } from "../vehicleRouteSummary"
 
@@ -123,7 +124,9 @@ const CurrentLocation = ({ vehicle }: VehicleProp): React.ReactElement => {
     vehicle.longitude
   )
 
-  if (isOk(intersection)) {
+  if (isLoading(intersection) && !isOk(intersection)) {
+    return <Loading />
+  } else if (isOk(intersection)) {
     return <>{intersection.ok}</>
   }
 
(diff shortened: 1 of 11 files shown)


assets/tests/components/mapPage/vehiclePropertiesCard.test.tsx
```diff
@@ -176,8 +197,8 @@ describe("<VehiclePropertiesCard/>", () => {
         const vehicle = vehicleFactory.build()
         const intersection = "intersection ave @ street"
         ;(useNearestIntersection as jest.Mock)
-          .mockReturnValueOnce(undefined)
-          .mockReturnValueOnce(intersection)
+          .mockReturnValueOnce({ is_loading: true })
```
KaylaBrady, line 200: question(non-blocking): Non-blocking since this seems to replicate the existing setup of the test, but looks like the part where it tests the loading state text is commented out

KaylaBrady (approved):
Looks good to me! A few minor comments, and would otherwise defer to @firestack's review since she has thought more about the result types!

assets/src/util/fetchResult.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+interface FetchError {
+  is_error: true
+}
+
+export type FetchResult<T> = Loading | Ok<T> | (Ok<T> & Loading) | FetchError
+
+export const isLoading = <T>(r: FetchResult<T>): r is Loading =>
+  "is_loading" in r && r.is_loading === true
```
firestack, line 16: question/suggestion: would this work?
If it does, the `isFetchError` could also be changed

```suggestion
  r?.is_loading === true
```

assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
```diff
@@ -73,10 +74,12 @@ const Location = ({ vehicle }: { vehicle: Vehicle }) => {
...
   const { isOffCourse, latitude, longitude, stopStatus } = vehicle
-  const nearestIntersection: string | null = useNearestIntersection(
-    latitude,
-    longitude
-  )
+
+  const nearestIntersectionResult = useNearestIntersection(latitude, longitude)
+  const nearestIntersection = isOk(nearestIntersectionResult)
+    ? nearestIntersectionResult.ok
+    : null
```
firestack, line 81: note: when working on this idea before, I was struggling deciding whether to implement `Result` myself or use a library, one of the benifits of using a library is that this kind of transformation(and others) would be provided by default via `unwrapOr` if we used something like https://github.com/vultix/ts-results

I'm unsure what the best "Rust `Result<T,E>` type but in Typescript" Library is, and there's a few, but it's also not necessary, just QOL.

firestack (approved):
## QA
- Checked out branch and launched phoenix
- Went to map page and selected a search result
- [x] Loading state initially shown, then resolves 
- [x] When the same vehicle updates, loading state is not shown
- [x] When a new vehicle is selected, loading state is shown on first load
- [x] Checked that it works from a few different selection sources
	- [x] Clicking vehicles on the map
	- [x] Selecting vehicles from search results
	- [x] Selecting vehicles via VPP -> Go to map button

assets/tests/components/mapPage/vehiclePropertiesCard.test.tsx
lemald: Thanks for catching! I'll uncomment it.

assets/src/util/fetchResult.ts
lemald: Unfortunately it doesn't - the `r?` only ensures that `r` is defined, so it still doesn't know if `is_loading` is a valid key and trying to use it in a comparison throws an error.

firestack: Hmm, I was thinking that if `r = {}` then `r.is_loading` should be `undefined` and then when you do `undefined === true` should return `false` without errors, but if there's an error no worries.

firestack: I guess `(r?.is_loading || undefined) === true` might work, but you're good to leave it as is lol.

assets/src/components/mapPage/vehiclePropertiesCard.tsx
```diff
@@ -121,7 +123,14 @@ const CurrentLocation = ({ vehicle }: VehicleProp): React.ReactElement => {
     vehicle.latitude,
     vehicle.longitude
   )
-  return <>{intersection ?? "Exact location cannot be determined"}</>
+
+  if (isLoading(intersection) && !isOk(intersection)) {
```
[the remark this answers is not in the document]
lemald: Hmm that's an interesting idea. I just implemented helper functions that were needed as type guards, and I'm not sure if I could write that in a way that corresponds nicely to a type guard because of how TypeScript's object types permit you to have extra keys beyond what is specified in the type. I think it may become useful as we use these types in more places but I'd rather wait and see how we end up using this and what sorts of helpers end up being useful.

assets/src/util/fetchResult.ts
lemald: When trying that I still get:

```
src/util/fetchResult.ts:16:7 - error TS2339: Property 'is_loading' does not exist on type 'FetchResult<T>'.
  Property 'is_loading' does not exist on type 'FetchError'.
```

...so I think I'm just going to leave it as is and merge.

PR #1915, opened 2023-02-08
feat(RouteShape): Tooltip + updated selected, hover states

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203766403597552/f

**Note:  userEvent => fireEvent for some tests**
Adding a tooltip to shapes caused existing tests to fail where a route shape is clicked. Kudos to @firestack for helping me debug and pointing me to the fact that the `clientX` and `clientY` were undefined [here](https://github.com/Leaflet/Leaflet/blob/9c58fcdd0a61167bafc4496528aaece1e4fe44af/src/dom/DomEvent.js#L259), leading to an `Invalid LatLng object: (NaN, NaN)` error. 

Stepping into [here](https://github.com/Leaflet/Leaflet/blob/3f5ea5a271b41ec09147ea774fea90bc960a1ab2/src/map/Map.js#L1419), `clientX/Y` was not `in src` [here](https://github.com/Leaflet/Leaflet/blob/3f5ea5a271b41ec09147ea774fea90bc960a1ab2/src/core/Util.js#L14). This was not the case when stepping through the browser. I suspect it is because of the way the `MouseEvent` is created when using `userEvent.click`: `clientX/Y` are set using [`defineProperty`](https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/src/event/createEvent.ts#L81). From the defineProperty [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#description):

> By default, properties added using Object.defineProperty() are not writable, not enumerable, and not configurable. 

Locally, I added `enumerable: true` to `defineProperty`  and confirmed the tests passed.

With the limitation of using the current rtl library (I'll likely open up a PR to address), I tried using `fireEvent.click` instead, which worked. Stepping through, I confirmed that `clientX/Y` were hit as expected.

### DIFF
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -272,9 +272,8 @@ const RoutePatternLayers = ({
                         sticky={true}
                         direction="top"
                       >
-                        <>
-                          {`Click to select route ${routePattern.routeId}${variantFormatted}.`}
-                        </>
+                        Click to select route {routePattern.routeId}
+                        {variantFormatted}.
                       </Tooltip>
                     ) : undefined}
                   </RouteShape>
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -252,6 +252,11 @@ const RoutePatternLayers = ({
     <>
       <ZoomLevelWrapper>
         {(zoomLevel) => {
+          const routePatternIdSplit = routePattern.id.split("-")
+          const variantFormatted =
+            routePatternIdSplit.length === 3 && routePatternIdSplit[1] !== "_"
+              ? "_" + routePatternIdSplit[1]
+              : ""
           return (
             <>
               {routePattern.shape && (
@@ -267,7 +272,9 @@ const RoutePatternLayers = ({
                         sticky={true}
                         direction="top"
                       >
-                        <>Click to select route {routePattern.id}.</>
+                        <>
+                          {`Click to select route ${routePattern.routeId}${variantFormatted}.`}
+                        </>
                       </Tooltip>
                     ) : undefined}
                   </RouteShape>
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -266,7 +266,7 @@ const RoutePatternLayers = ({
                     isSelected={isSelected}
                     onClick={onShapeClick}
                   >
-                    {zoomLevel >= 16 && !isSelected ? (
+                    {(zoomLevel >= 16 && !isSelected && (
                       <Tooltip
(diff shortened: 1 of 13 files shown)


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -240,25 +241,50 @@ const SelectionCardContainer = ({
...
+                    onClick={onShapeClick}
+                  >
+                    {zoomLevel >= 16 && !isSelected ? (
+                      <Tooltip
+                        className="route-shape__tooltip"
+                        sticky={true}
+                        direction="top"
+                      >
+                        <>
+                          {`Click to select route ${routePattern.routeId}${variantFormatted}.`}
```
lemald, line 276: **suggestion:** I think you might be able to do this without string interpolation or even a fragment

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -87,18 +97,26 @@ const onFollowerUpdate: UpdateMapFromPointsFn = (map, points) => {
...
-      .project(points[0], targetZoom)
-      // Offset the target point in screenspace to move the center of the map
-      // to apply the padding to the center
-      .subtract(offset),
-    // convert the target point to worldspace from screenspace
-    targetLatLng = map.unproject(targetPoint, targetZoom)
-
-  // Zoom/Pan center of map to offset location in worldspace
-  map.setView(targetLatLng, targetZoom)
+  if (points.length === 0) {
```
firestack, line 100: just a note that this brought in/still has the `=== 0`

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -199,26 +217,99 @@ const SelectionCardContainer = ({
...
+                    {zoomLevel >= 16 && !isSelected ? (
+                      <Tooltip
+                        className="route-shape__tooltip"
+                        sticky={true}
+                        direction="top"
+                      >
+                        Click to select route {routePattern.routeId}
+                        {variantFormatted}.
+                      </Tooltip>
+                    ) : undefined}
```
firestack, line 278: suggestion: instead of the ternary operator, I think you could do this with the `&&` operator?

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Thanks, took care of it with some merge conflict cleanup!

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: It doesn't like the type Element | false, but `&& !isSelected ... || undefined` does the trick!

firestack (approved):

PR #1916, opened 2023-02-13
feat(ts/components/stopCard): prevent stop card from being obscured

firestack:
# Summary

This PR implements a simple way of getting the "safe zone" passed down into the component tree to use leaflet's `autoPanPadding` features[^1]. We could probably also use the new context provider to provide the safe zone bounds for the follower logic too.

The two largest problems were
 - Should this be done via Context or Prop Drilling? (or something else?)
 - What should these new components/contexts/types be named. (All the names are composite names, which has been difficult to find something which feels "right")

I'm still considering the names used and would appreciate any input's y'all may have.

## Todo
 - Design and product approved this solution as ok for MVP fidelity
	 - [x] Design 🆗
	 - [x] Product 🆗

---

Asana Ticket: https://app.asana.com/0/inbox/1203014709797732/1203674636294159/1203947212316076

[^1]: https://leafletjs.com/reference.html#popup-autopanpaddingtopleft

### DIFF
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -553,8 +553,8 @@ const MapDisplay = ({
     >
       <MapSafeArea.Provider
         value={{
-          paddingTopLeft: [445, 50],
-          paddingBottomRight: [20, 50],
+          paddingTopLeft: [445, 54],
+          paddingBottomRight: [50, 50],
         }}
       >
         <SelectionDataLayers
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -115,7 +115,7 @@ const onFollowerUpdate: UpdateMapFromPointsFn = (map, points) => {
   } else {
     const pointsBounds = Leaflet.latLngBounds(points)
     map.fitBounds(pointsBounds, {
-      paddingBottomRight: [20, 50],
+      paddingBottomRight: [50, 20],
       paddingTopLeft: topLeft,
     })
   }
@@ -554,7 +554,7 @@ const MapDisplay = ({
       <MapSafeArea.Provider
         value={{
           paddingTopLeft: [445, 54],
-          paddingBottomRight: [50, 50],
+          paddingBottomRight: [50, 20],
         }}
       >
         <SelectionDataLayers
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -38,7 +38,7 @@ import {
   useInteractiveFollowerState,
 } from "../map"
 import { RouteShape, RouteStopMarkers, VehicleMarker } from "../mapMarkers"
-import { MapSafeArea } from "../stopCard"
+import { MapSafeAreaContext } from "../../contexts/mapSafeAreaContext"
 import ZoomLevelWrapper from "../ZoomLevelWrapper"
 import RoutePropertiesCard from "./routePropertiesCard"
 import VehiclePropertiesCard from "./vehiclePropertiesCard"
@@ -551,7 +551,7 @@ const MapDisplay = ({
       shapes={[]}
       stateClasses={stateClasses}
     >
-      <MapSafeArea.Provider
+      <MapSafeAreaContext.Provider
         value={{
           paddingTopLeft: [445, 54],
           paddingBottomRight: [50, 20],
@@ -564,7 +564,7 @@ const MapDisplay = ({
           setSelection={setSelection}
           setStateClasses={setStateClasses}
         />
-      </MapSafeArea.Provider>
+      </MapSafeAreaContext.Provider>
     </BaseMap>
   )
 }
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -92,14 +92,15 @@ const onFollowerUpdate: UpdateMapFromPointsFn = (map, points) => {
   // with the distance from the right side of the VPC to the left side of the
   // map container
   const topLeft = new Point(445, 0)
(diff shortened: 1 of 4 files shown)


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -549,13 +551,20 @@ const MapDisplay = ({
...
-        selectedEntity={selectedEntity}
-        showSelectionCard={showSelectionCard}
-        deleteSelection={deleteSelection}
-        setSelection={setSelection}
-        setStateClasses={setStateClasses}
-      />
+      <AutoPanPaddingConfig.Provider
+        value={{
+          autoPanPaddingTopLeft: [445, 50],
+          autoPanPaddingBottomRight: [50, 20],
```
KaylaBrady, line 557: question: Is `autoPanPaddingBottomRight` meant to match the `paddingBottomRight` on 118?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -549,13 +551,20 @@ const MapDisplay = ({
...
       stateClasses={stateClasses}
     >
-      <SelectionDataLayers
-        selectedEntity={selectedEntity}
-        showSelectionCard={showSelectionCard}
-        deleteSelection={deleteSelection}
-        setSelection={setSelection}
-        setStateClasses={setStateClasses}
-      />
+      <AutoPanPaddingConfig.Provider
```
KaylaBrady, line 554: question (non-blocking): Does this need to wrap `<SelectionDataLayers/>`, or only the relevant `<RouteStopMarkers/>` element? I lean towards having it wrap at the lowest level it is needed and pulling it up to a higher level as needed, so it is easier to see how it is used, but that is just my preference!

assets/src/components/stopCard.tsx
```diff
@@ -91,4 +106,15 @@ const StopCard = ({
   )
 }
 
+export const AutoPanPaddingConfig = createContext<LeafletAutoPanPadding>({})
```
KaylaBrady, line 109: suggestion: Towards your question about names, I wonder if it would be helpful to make this more generic than autoPan by making it something like `SafeZoneConfig`. Maybe then that SafeZoneConfig could easily be shared with other functions like `onFollowerUpdate`

assets/src/components/stopCard.tsx
```diff
@@ -91,4 +106,15 @@ const StopCard = ({
...
+
+export const StopCardWithSafeZone = (props: StopCardProps) => {
+  const autoPanPaddingConfig = useContext(AutoPanPaddingConfig)
```
KaylaBrady, line 112: Praise: I like your approach of using contexts here!

assets/src/components/stopCard.tsx
firestack: Agreed!
One issue with `onFollowerUpdate` is that becuase it's called within `useEffect`, we would need to `useContext` and reconstruct/curry `onFollowerUpdate` with that value before passing to the `onUpdate` prop.

assets/src/components/stopCard.tsx
```diff
@@ -91,4 +108,16 @@ const StopCard = ({
...
+
+/**
+ * A <{@link StopCard}/> which provides it's `autoPanPadding` parameter via the
+ * nearest {@link MapSafeArea} Context
+ */
+export const SafeAreaContextStopCard = (props: StopCardProps) => {
+  const safeArea = useContext(MapSafeArea)
+
+  return <StopCard {...props} autoPanPadding={safeArea} />
+}
```
firestack, line 121: I also considered doing 
> `<StopCardWithSafeArea/>`
> and `<StopCard.WithSafeArea/>`

I liked the dot notion for accessing a specialized version of `<StopCard/>` but we don't have any existing components in skate which utilize this to use as a reference for naming convention/style.

assets/src/components/stopCard.tsx
firestack: renamed in [`b8a40ca` (#1916)](https://github.com/mbta/skate/pull/1916/commits/b8a40ca3a3cd3b7393ebb87c440893ecd86d4a1f)

lemald: Followup question: since `MapSafeArea` seems like it could generically be used for other components, and it's already exported and referenced in `mapMarkers.tsx`, would it make sense to move it to `map.tsx` or `mapPage.tsx` or some other higher-level map file, rather than staying here in `stopCard.tsx`?

firestack: > would it make sense to move it to `map.tsx` or `mapPage.tsx` or some other higher-level map file, rather than staying here in `stopCard.tsx`?

I just found/remembered `src/contexts`, moved 👍🏻 8841cd3fd366c6eaaa3623ecc30e184b1c8408df

assets/src/components/stopCard.tsx
KaylaBrady: I like either of the two alternatives you present here - should there be other SafeAreaContext aware components, it seems nice for scanability to have the "important"/differentiating part of the name first!

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: That makes sense to me!

KaylaBrady (approved):
Looks good to me! Would be nice to have a test for this, but it seems like that would be challenging. Maybe a candidate for visual testing once we get to [automated visual tests](https://app.asana.com/0/1152340551558956/1203152611716214/f)!

lemald (approved):

assets/src/components/stopCard.tsx
firestack: I decided to go with the dot notation! [mbta/skate@`b6869c9` (#1916)](https://github.com/mbta/skate/pull/1916/commits/b6869c9bfee5c1e93e158649775b9a9157560a79)

PR #1917, opened 2023-02-13
Map page narrow screen behavior

lemald:
Asana ticket: [⚙️ Better Maps | Search Map behavior at <=800px](https://app.asana.com/0/1200180014510248/1203888345713854/f)

Simply removing the unneeded code seems to have worked as expected. I did also have to play around with the z-index of the search drawer a little bit because on narrow screens it becomes apparent that the Leaflet controls actually have a higher z-index than what we were using, so I replaced `$leaflet-pane-z-index` with a new variable that represents the maximum z-index out of the entire Leaflet CSS file.

### DIFF
--- a/assets/css/_leaflet.scss
+++ b/assets/css/_leaflet.scss
@@ -1,5 +1,5 @@
-// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css#L87
-$leaflet-pane-z-index: 400;
+// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css
+$max-leaflet-z-index: 1000;
 
 .leaflet-container {
   width: 100%;
--- a/assets/css/_leaflet.scss
+++ b/assets/css/_leaflet.scss
@@ -1,4 +1,4 @@
-// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css
+// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css#L110
 $max-leaflet-z-index: 1000;
 
 .leaflet-container {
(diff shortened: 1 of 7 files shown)


assets/css/_leaflet.scss
```diff
@@ -1,5 +1,5 @@
-// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css#L87
-$leaflet-pane-z-index: 400;
+// From here: https://github.com/Leaflet/Leaflet/blob/5c29d615563835d373a51b91fc3513a17907da96/dist/leaflet.css
```
firestack, line 1: question(non-blocking): just curious why you removed the line info? is it worth linking to the new line which has the z-index value we're referencing here?

assets/src/components/mapPage.tsx
```diff
@@ -152,26 +125,9 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
-      ? "m-map-page--show-list"
-      : "m-map-page--show-map"
-
-  const mobileMenuClass = mobileMenuIsOpen ? "blurred-mobile" : ""
-  // #endregion
-
   return (
     <div
-      className={`m-map-page ${mobileDisplayClass} ${mobileMenuClass} inherit-box border-box`}
+      className={`m-map-page inherit-box border-box`}
```
firestack, line 130: nit(non-blocking): this doesn't need to be a interpolated string anymore

firestack (approved):
Couple questions but I don't see any real issues 👍🏻

assets/css/_leaflet.scss
lemald: For some reason I thought there were a couple of places in the file with `z-index: 1000`, but it looks like there's just one, so I'll link to it!

assets/css/_leaflet.scss
```diff
@@ -1,5 +1,5 @@
...
+$max-leaflet-z-index: 1000;
```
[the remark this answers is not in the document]
lemald: Good question! It looks like that code is in reference to the `VehicleMarker` component, and experimenting with the search drawer and vehicle markers it looks like the vehicle markers are still behind the search drawer as I would expect, so I think we're good unless there's some other possible interaction I'm missing.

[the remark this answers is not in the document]
firestack: Nope that's all, I figured it would be okay due to how they've got stacking contexts setup but I just wanted to double check.

PR #1918, opened 2023-02-14
feat(RPC): keep height < window height and scroll within details sections

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1203921165676874/f

Notes: 
* RPC no longer uses the universal card; it was easier to adjust the styling without working in the card component tree
* Removed use of details/summary: to get the details to shrink to the height of the parent component I wanted to make it `display: flex`, but that does not work for `<details/>`. Using an alternative accessible approach suggested by @firestack instead


assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -47,25 +46,25 @@ const DetailSection = ({
...
-          toggleOpen()
-        }}
+    <div className="m-route-properties-card__details-section">
+      <button
+        className={`m-route-properties-card__toggle`}
+        aria-expanded={isOpen}
+        aria-controls={"details" + className}
+        onClick={toggleOpen}
+      >
+        <span id={"summary" + className}>
```
lemald, line 56: **suggestion (non-blocking):** Using `useId` as part of the ID could help us avoid collisions in the future if we ever want to have multiple routes open at once, though I realize that's probably unlikely with our designs.

lemald (approved):
One minor suggestion but overall looks good!

PR #1920, opened 2023-02-16
feat(VPP): Block waivers & invalid banners as cards 

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1203449129917737/f


assets/static/images/icon-ghost.svg
```diff
@@ -1 +1,6 @@
....297.402zm-.407-.795a.975.975 0 00.41.798l.29-.408a.475.475 0 01-.2-.388l-.5-.002zm0-5.137v5.138h.5V5.067h-.5zm8.497 5.118c0-.044.022-.085.056-.11l.302.398a.359.359 0 00.142-.283l-.5-.005zm0-5.118v5.121h.5v-5.12h-.5zM5.45 1.892c2.064 0 3.68 1.454 3.68 3.175h.5c0-2.06-1.907-3.675-4.18-3.675v.5zM1.77 5.067c0-1.716 1.618-3.175 3.681-3.175v-.5c-2.273 0-4.181 1.62-4.181 3.675h.5zm0 5.127V5.067h-.5v5.127h.5zm-.056-.113a.141.141 0 01.056.11l-.5.006a.359.359 0 00.143.283l.301-.4zm-.092-.031a.134.134 0 01.08.022l-.277.416a.366.366 0 00.219.061l-.022-.5zm1.309-.821l-1.427.856.257.429 1.427-.857-.257-.428zm.292-.08a.579.579 0 00-.29.078l.253.432a.078.078 0 01.04-.01l-.003-.5zm.282.072a.568.568 0 00-.283-.073l.004.5c.012 0 .024.003.034.009l.245-.436zm2.07 1.169L3.503 9.22l-.245.435 2.068 1.17.246-.436zm1.82-1.17l-2.067 1.17.246.435 2.067-1.17-.247-.436zm.571 0a.572.572 0 00-.574 0l.251.433a.072.072 0 01.072 0l.251-.433zm1.43.858l-1.427-.856-.257.429 1.427.856.257-.428zm-.198-.012a.135.135 0 01.08-.023l-.021.5a.365.365 0 00.218-.061l-.277-.416z"/><ellipse cx="4.481" cy="5.881" rx=".674" ry=".649" stroke-width="0.25"/><ellipse cx="7.855" cy="5.881" rx=".674" ry=".649" stroke-width="0.25"/></svg>
```
KaylaBrady, line 1: <GhostIcon/> was unused, so updating file to use ghost specified in the designs, which is filled in white.

assets/tests/components/propertiesPanel/vehiclePropertiesPanel.test.tsx
```diff
@@ -134,6 +134,13 @@ describe("VehiclePropertiesPanel", () => {
     expect(tree).toMatchSnapshot()
   })
 
+  test("Includes invalid bus banner when vehicle is off course", () => {
+    render(
+      <VehiclePropertiesPanel selectedVehicle={invalidVehicleFactory.build()} />
+    )
+    expect(screen.getByText("Invalid Bus")).toBeInTheDocument()
```
firestack, line 141: suggestion(non-blocking): I'd like it if we could change this to be more semantic and use a `getByRole` to guide adjusting the content if it's possible.

```suggestion
    expect(screen.getByRole("heading", { name: "Invalid Bus" })).toBeVisible()
```

I could be wrong about "heading" being the correct role here, but I feel like there's some role that should represent this part of the card?

assets/src/components/propertiesPanel/blockWaiverBanner.tsx
```diff
@@ -22,77 +24,120 @@ const currentFuturePastClass = (blockWaiver: BlockWaiver): string => {
...
 
+const formattedReason = (blockWaiver: BlockWaiver) =>
+  [
+    blockWaiver.causeDescription,
+    (blockWaiver.remark || "")
+      .replace(blockWaiver.causeDescription, "")
+      .replace(/:$/, ""),
+  ]
+    .filter((s) => s !== null && s !== "")
+    .join(" ")
```
lemald, line 85: **praise:** I appreciate that you also factored logic like this out while making your changes

assets/tests/components/propertiesPanel/vehiclePropertiesPanel.test.tsx
KaylaBrady: Thanks for the suggestion! I added as `<h3>` since there is already a `<h2>` as the VPP header. This required some h3 styling overrides for "Invalid Bus" & block waiver card titles, open to suggestions for a more consistent way to do that! I was thinking of making the card title a `<h3>` always, but I think the level would need to depend on where the card is.

assets/tests/components/card.test.tsx
```diff
@@ -18,24 +18,24 @@ describe("Card", () => {
     expect(
       result.queryByText(/Foo/)?.parentElement?.parentElement?.parentElement
         ?.className
-    ).toMatch(/m-card--read/)
+    ).toMatch(/m-card--inactive/)
```
firestack, line 21: suggestion(non-blocking): maybe while updating these, we could use `toHaveClass`?

KaylaBrady: Addressed this and other test cleanup for this file in [aaed28f](https://github.com/mbta/skate/pull/1920/commits/aaed28fe45f4f0e55bbf09a5cfeb5a92f6c2b03c)

firestack: 🙌🏻 Thank you 😻

firestack (approved):

PR #1921, opened 2023-02-17
feat(ts/components/map): make street view switch fit on smaller screen sizes

firestack:
A few notes

I added a prefix to the BEM class name when I moved it into it's own file since this feels like it's own component rather than just a subcomponent of the map.

Also, I was able to stop fighting with leaflet and bootstrap css specificity by importing them first before any of our styles, I don't think I missed anything or caused any problems and I didn't see anything but that might be a good thing to mull over a bit.

Also bootstrap's component was making it somewhat difficult to get layout to look nice, so I ended up overriding padding and margins to shrinkwrap the div to the input, I don't like having had to override bootstrap and worry I may have misunderstood why the padding/negative margins were used in the first place.

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1203888345713851

### DIFF
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -177,38 +177,6 @@
   border-top-color: $color-tooltip-background;
 }
 
-// Adding .leaflet-control-container increases specificity to make sure we override
-// base Leaflet control positioning
-.leaflet-control-container .m-vehicle-map__street-view-control {
-  position: absolute;
-  // Leaflet uses px for positioning of their controls so we use px here to match
-  top: 10px;
-  right: 55px;
-
-  padding: 0.5rem 0.5rem 0.5rem 1rem;
-
-  background-color: $white;
-
-  border: 1px solid $color-gray-400;
-  border-radius: 7px;
-
-  color: $color-eggplant-700;
-  font-size: 1rem;
-  line-height: 1.5rem;
-  user-select: none;
-  display: flex;
-
-  label {
-    span svg {
-      height: 1rem;
-      padding-right: 0.5rem;
-      fill: currentColor;
-    }
-
-    padding-right: 0.5rem;
-  }
-}
-
 .m-vehicle-map__recenter-button {
   path {
     fill: none;
--- a/assets/css/map/controls/_street_view_switch.scss
+++ b/assets/css/map/controls/_street_view_switch.scss
@@ -1,0 +1,31 @@
+// Adding .leaflet-control-container increases specificity to make sure we override
+// base Leaflet control positioning
+.leaflet-control-container .m-vehicle-map__street-view-control {
+  position: absolute;
+  // Leaflet uses px for positioning of their controls so we use px here to match
+  top: 10px;
+  right: 55px;
+
+  padding: 0.5rem 0.5rem 0.5rem 1rem;
+
+  background-color: $white;
+
+  border: 1px solid $color-gray-400;
+  border-radius: 7px;
+
+  color: $color-eggplant-700;
+  font-size: 1rem;
+  line-height: 1.5rem;
+  user-select: none;
+  display: flex;
+
+  label {
+    span svg {
+      height: 1rem;
+      padding-right: 0.5rem;
+      fill: currentColor;
+    }
+
+    padding-right: 0.5rem;
+  }
+}
--- a/assets/css/map/controls/_street_view_switch.scss
+++ b/assets/css/map/controls/_street_view_switch.scss
@@ -1,6 +1,3 @@
-// Adding .leaflet-control-container increases specificity to make sure we override
-// base Leaflet control positioning
-.m-vehicle-map__street-view-control,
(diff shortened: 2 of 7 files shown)


assets/css/_vehicle_map.scss
```diff
@@ -177,38 +177,6 @@
   border-top-color: $color-tooltip-background;
 }
 
-// Adding .leaflet-control-container increases specificity to make sure we override
-// base Leaflet control positioning
-.leaflet-control-container .m-vehicle-map__street-view-control {
```
firestack, line 182: btw, [this commit `f303566` (#1921)](https://github.com/mbta/skate/pull/1921/commits/f3035669520d6300c328dd2c483fe27cf0111b64) allows us to drop the `.leaflet-control-container` to override specificity.

assets/css/map/controls/_street_view_switch.scss
```diff
@@ -0,0 +1,72 @@
...
+
+    // Remove Negative Margins so shrinkwrap works
+    .form-check-input {
+      margin-left: 0;
+    }
+  }
+  // #endregion
+
+  // #region Media Queries
+  @media screen and (max-width: 800px) {
```
firestack, line 62: I couldn't figure out a good way to factor this out into a variable. Suggestions welcome!

assets/src/components/map.tsx
```diff
@@ -82,66 +80,6 @@ interface RecenterControlProps extends ControlOptions {
...
-
-const StreetViewControl = ({
-  streetViewEnabled: streetViewEnabled,
-  setStreetViewEnabled: setStreetViewEnabled,
-}: StreetViewControlProps): JSX.Element | null => {
-  const map = useMap()
-  const portalParent = map
-    .getContainer()
-    .querySelector(".leaflet-control-container")
-  const [portalElement, setPortalElement] = useState<HTMLElement | null>(null)
```
firestack, line 98: I just wanted to know if it'd be a good pattern to break it out into it's own file. I can move other controls to their own files in another PR if @mbta/skate-developers are interested.

assets/css/map/controls/_street_view_switch.scss
```diff
@@ -0,0 +1,72 @@
+// BEM-PREFIX: LC = Leaflet Control
```
firestack, line 1: TODO: discuss what the prefix should be (if any) or if we use another prefix instead?

assets/css/map/controls/_street_view_switch.scss
lemald: I think ideally we'd do something like a map in SCSS of the various breakpoint sizes to replace the outdated variables like `$mobile-max-width`, although I don't have strong opinions on things like where it would live.

assets/src/components/map.tsx
lemald: Moving it out is probably a good call given how much code it is.

assets/css/map/controls/_street_view_switch.scss
lemald: Is this `BEM-PREFIX` comment some kind of standardized thing? I can't seem to find any references to it with a quick Google search.

firestack: Nope, just wanted to make a note of it to discuss

firestack: I typed BEM in all uppercase and I think it just happened kinda like `// TODO:`

assets/src/components/map.tsx
KaylaBrady: I am in favor of breaking it out too! If we want to consider standardizing to only one (or one stateful) component per file, I've come across [this eslint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md) we could consider adding

lemald (approved):

assets/css/map/controls/_street_view_switch.scss
firestack: Maybe we do something along these lines? https://medium.com/free-code-camp/the-100-correct-way-to-do-css-breakpoints-88d6a5ba1862

firestack: Opened a ticket for this to be addressed later 👍🏻 https://app.asana.com/0/1152340551558956/1204009922904919/f

assets/src/components/map.tsx
firestack: Oooh I like the idea of using that rule!
I worry that that would have a large impact on our codebase but it'd probably be positive in the long run!

assets/css/map/controls/_street_view_switch.scss
firestack: I noted it on the [BEM Naming Ticket](https://app.asana.com/0/1152340551558956/1203760641228313/f) that this prefix was added in this PR so it can get addressed then.

PR #1922, opened 2023-02-17
fix: don't position search page contents over views

lemald:
Non-ticketed hotfix.

We updated the `z-index` of the search panel to make sure it appears above all of the contents of the Leaflet map, however it turns out that on mobile the search panel is now staying on top of the view - see attached screenshot.

IMG_2098

This update keeps the navbar above the views, and also integrates views into the `$z-page-layout-context` map we have. It doesn't integrate those with the `$max-leaflet-z-index` constant unfortunately as I suspect that might be a larger lift (at least on the QA side) and I'd like to get this fix out.

### DIFF
--- a/assets/css/_late_view.scss
+++ b/assets/css/_late_view.scss
@@ -8,7 +8,7 @@ $late-view-background-color: $color-bg-light;
   position: absolute;
   right: 0;
   top: 0;
-  z-index: 1001;
+  z-index: map-get($z-page-layout-context, "view");
 
   @media screen and (max-width: 700px) {
     width: 100%;
--- a/assets/css/_notification_drawer.scss
+++ b/assets/css/_notification_drawer.scss
@@ -8,7 +8,7 @@
   right: 0;
   top: 0;
   width: 23.5rem;
-  z-index: 1001;
+  z-index: map-get($z-page-layout-context, "view");
 
   @media screen and (max-width: $mobile-max-width) {
     width: 100%;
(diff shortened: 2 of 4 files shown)


firestack (approved):
I'm curious if somehow we can use stacking contexts to address this issue?
z-index works, but because we didn't expect this to happen, maybe we should make it so it can't happen? 

we could open a tech debt ticket for this if devs agree?

I otherwise have no issues 👍🏻

lemald:
> I'm curious if somehow we can use stacking contexts to address this issue? z-index works, but because we didn't expect this to happen, maybe we should make it so it can't happen?
> 
> we could open a tech debt ticket for this if devs agree?
> 
> I otherwise have no issues 👍🏻

Definitely agreed that we should look into that. For the purposes of getting a fix put together I figured I'd go with the most straightforward approach, but stacking contexts are probably the more appropriate fix long-term.

KaylaBrady:
Non-blocking, but maybe a related issue looking at the old maps page on screens bigger than mobile - should the search results list be underneath the "modal-overlay"?

lemald:
> Non-blocking, but maybe a related issue looking at the old maps page on screens bigger than mobile - should the search results list be underneath the "modal-overlay"?

Good question! After some experimenting, I think the answer is "no." The modal overlay in question is currently only used for the overlay background that shows up on the rest of the page when you open the vehicle properties panel or ghost properties panel and prevents direct interaction with them. I don't think it makes sense to have this overlay also obscure the search results panel, though it might be worth asking product / design.

PR #1942, opened 2023-02-24
feat(Realtime.Server): periodically log ghost stats

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204005660133255/f

### DIFF
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -253,7 +253,7 @@ defmodule Realtime.Server do
     |> Enum.filter(&match?(%Ghost{}, &1))
     |> Enum.each(fn ghost ->
       Logger.info(
-        "ghost: id=#{ghost.id} block_waiver_causes=#{inspect(Enum.map(ghost.block_waivers, & &1.cause_id))}"
+        "ghost: id=#{ghost.id} route=#{ghost.route_id} run_id=#{ghost.run_id} block_waiver_causes=#{inspect(Enum.map(ghost.block_waivers, & &1.cause_id))}"
       )
     end)
   end
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -247,7 +247,7 @@ defmodule Realtime.Server do
     {:noreply, state, :hibernate}
   end
 
-  def handle_info(:ghost_stats, %__MODULE__{ets: ets}) do
+  def handle_info(:ghost_stats, %__MODULE__{ets: ets} = state) do
     {explained_count, unexplained_count} =
       {ets, :all_vehicles}
       |> lookup()
@@ -265,6 +265,8 @@ defmodule Realtime.Server do
     Logger.info(
       "ghost_stats: explained_count=#{explained_count} unexplained_count=#{unexplained_count}"
     )
+
+    {:noreply, state, :hibernate}
   end
 
   @impl true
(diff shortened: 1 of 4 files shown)


lib/realtime/server.ex
```diff
@@ -246,6 +247,17 @@ defmodule Realtime.Server do
     {:noreply, state, :hibernate}
   end
 
+  def handle_info(:ghost_stats, %__MODULE__{ets: ets}) do
+    {ets, :all_vehicles}
+    |> lookup()
+    |> Enum.filter(&match?(%Ghost{}, &1))
+    |> Enum.each(fn ghost ->
+      Logger.info(
+        "ghost: id=#{ghost.id} block_waiver_causes=#{inspect(Enum.map(ghost.block_waivers, & &1.cause_id))}"
```
lemald, line 256: **suggestion:** this will provide very verbose logging on all ghosts, but I wonder if it might be simpler to just write a single log message that contains the count of explained and unexplained ghosts - if I'm not mistaken Splunk should make it easy to take the sum of fields from your query to then get the metrics we're looking for.

KaylaBrady: I was thinking about that, but wondering if it is worth it to log a greater level of detail for the metrics in splunk so that we can accurately aggregate the metrics over time to satisfy the acceptance criteria.
Trying to think through the possible difference in calculations with the following scenario, please let me know whether this seems on the right track!

At 4:00 there are 10 ghost buses numbered 1-10, the first 7 have waivers.

At 4:01 there are 10 ghost buses numbered 3-12. (Ghosts 1&2 disappeared - assumption on my part: ghosts go away when their scheduled trip is over).  3-7 still have waivers, a waiver was added for 8, and new buses 11 & 12 have waivers.

If we only log the totals, we would see
4:00 - 7/10 waived
4:01 - 8/10 waived

If we aggregate by summing in splunk, we would get 75%. 

However, there were actually 12 buses total over this period, and 10 of them had waivers in this time period, so the actual % of buses with waivers would be 83%. 

Additionally, logging each ghost would also allow us to calculate the amount of time between a ghost bus is detected & a waiver created, which was a question posed in the context section of the ticket (though not acceptance criteria).

If we're in agreement that there is value in logging individual ghosts for flexibility in calculating metrics, perhaps it would be worth adding some logic to only log when there is a change to a ghost bus.

KaylaBrady: Update: The above comment only applies for getting a metric about % of explained ghost buses over a period. Since we are instead doing metric about the % of time explained, I will make the suggested change.

lemald (approved):
Code looks generally good! One question, did you settle on a way to deduplicate the logging across multiple instances in prod?

KaylaBrady:
@lemald 

> One question, did you settle on a way to deduplicate the logging across multiple instances in prod?

My plan is to dedup records within the same minute of the day, which will result in the last record that appears within each minute being kept. This could potentially result in oddities during deploys though

lemald:
> @lemald
> 
> > One question, did you settle on a way to deduplicate the logging across multiple instances in prod?
> 
> My plan is to dedup records within the same minute of the day, which will result in the last record that appears within each minute being kept. This could potentially result in oddities during deploys though

Makes sense, and I think over a long period of time (a day or more) it probably won't have any noticeable effect.

lemald (approved):
Good catch on the return value, sorry I missed that!

PR #1951, opened 2023-03-02
fix(Hastus.Trip): expand_through_routed_trips/2 returns trips in order

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1203976763779491/f

### DIFF
--- a/lib/schedule/hastus/trip.ex
+++ b/lib/schedule/hastus/trip.ex
@@ -91,22 +91,19 @@ defmodule Schedule.Hastus.Trip do
   # to the trip ID. It does this because Schedules has to merge these
   # through-routed trips into continuous trips that change route number,
   # because they can't depend on the HASTUS "fixed link" feature that kept
-  # the separate trips attached. But we wanted to keep presenting these to 
+  # the separate trips attached. But we wanted to keep presenting these to
   # passengers as separate trips with multiple route numbers. In theory the
   # HASTUS problem will eventually get fixed, but it's been several years
   # already.
 
   @spec expand_through_routed_trips([t()], MapSet.t()) :: [t()]
   def expand_through_routed_trips(trips, gtfs_trip_ids) do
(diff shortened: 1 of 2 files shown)


lib/schedule/hastus/trip.ex
```diff
@@ -91,22 +91,19 @@ defmodule Schedule.Hastus.Trip do
...
   @spec expand_through_routed_trips([t()], MapSet.t()) :: [t()]
   def expand_through_routed_trips(trips, gtfs_trip_ids) do
-    through_routed_trip_ids = Enum.filter(gtfs_trip_ids, &Regex.match?(~r/_\d+$/, &1))
+    through_routed_suffix_regex = ~r/_\d+$/
 
     original_id_to_through_routed_trip_ids =
-      Enum.reduce(through_routed_trip_ids, %{}, fn through_routed_trip_id, result ->
-        original_id = String.replace(through_routed_trip_id, ~r/_\d+$/, "")
-        through_routed_trip_ids = Map.get(result, original_id, [])
-        new_through_routed_trip_ids = [through_routed_trip_id | through_routed_trip_ids]
```
KaylaBrady, line 107: the reason the trips were out of order is because they were accumulated in reverse in `new_through_routed_trip_ids`.  `Enum.group_by` [reverses the enumerable first](https://github.com/elixir-lang/elixir/blob/a64d42f5d3cb6c32752af9d3312897e8cd5bb7ec/lib/elixir/lib/enum.ex#L1326) so that the results are accumulated in order.

firestack (approved):
this clean-up and fix is great!

lib/schedule/hastus/trip.ex
lemald: Furthermore, this behavior is [specified in the docs](https://hexdocs.pm/elixir/1.12/Enum.html#group_by/3), so I think we're good to rely on it. 👍

lemald (approved):

PR #1952, opened 2023-03-02
fix(ts/components/mapPage): put stop markers on top of station markers

firestack:
The stop markers were under the station markers because the stop markers used the default `z-index` of the markers pane (`400`) but the station markers were moved into their own pane of `z-index` `410`.

This makes another pane similar to the `routeStopMarkers` pane at `zIndex` `450` as this was the easiest way to address this bug.

I tried to reuse the existing `routeStopMarkers` pane by setting `pane="routeStopMarkers"` on the `<StopMarker>` component, but this breaks both encapsulation and tests. While I think that would be one of the nicer solutions, it required a larger refactor than would be useful to make it work.

Similarly, I've been considering pulling out some self contained components of the `<BaseMap>`, such as converting the `stations` prop into a `includeStations` prop similar to the leaflet `attributionControl={false}` props. I made a [stub of a ticket](https://app.asana.com/0/1152340551558956/1204081206701126/f) for getting to this in the future.
If we pulled out the stations we could sort them under the stop markers via HTML order instead of `z-index`. If devs prefer that solution I can pursue that for this ticket instead.

## Todo
- [ ] [Rename the Pane](https://github.com/mbta/skate/pull/1952#discussion_r1123595928)

---

Asana Ticket: https://app.asana.com/0/1148853526253437/1203941643219723/f

### DIFF
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -7,7 +7,7 @@ import React, {
   useEffect,
   useState,
 } from "react"
-import { Tooltip } from "react-leaflet"
+import { Pane, Tooltip } from "react-leaflet"
 import { SocketContext } from "../../contexts/socketContext"
 import useMostRecentVehicleById from "../../hooks/useMosRecentVehicleById"
 import usePatternsByIdForRoute from "../../hooks/usePatternsByIdForRoute"
@@ -280,12 +280,18 @@ const RoutePatternLayers = ({
                     )) ||
                       undefined}
                   </RouteShape>
-                  <RouteStopMarkers
-                    stops={routePattern.shape.stops || []}
-                    includeStopCard={true}
-                    direction={routePattern.directionId}
-                    zoomLevel={zoomLevel}
-                  />
+                  <Pane
+                    name="routeStopMarkers2"
+                    pane="markerPane"
+                    style={{ zIndex: 450 }} // should be above other non-interactive elements
+                  >
+                    <RouteStopMarkers
+                      stops={routePattern.shape.stops || []}
+                      includeStopCard={true}
+                      direction={routePattern.directionId}
+                      zoomLevel={zoomLevel}
+                    />
+                  </Pane>
                 </>
               )}
             </>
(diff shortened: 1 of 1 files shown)


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -280,12 +280,18 @@ const RoutePatternLayers = ({
...
                       undefined}
                   </RouteShape>
-                  <RouteStopMarkers
-                    stops={routePattern.shape.stops || []}
-                    includeStopCard={true}
-                    direction={routePattern.directionId}
-                    zoomLevel={zoomLevel}
-                  />
+                  <Pane
+                    name="routeStopMarkers2"
```
firestack, line 284: Help me come up with a better name please 😭

lemald: Maybe `selectedRoutePatternStops` is slightly more specific, if I understand the role of this specific component correctly?

lemald:
> If we pulled out the stations we could sort them under the stop markers via HTML order instead of z-index. If devs prefer that solution I can pursue that for this ticket instead.

I actually think I prefer this solution because it's a bit more explicit about what we're doing.

lemald (approved):

firestack:
> Is there an advantage to adding this new pane above stations rather than giving the stations a lower z-index via the existing `notableLocationMarkers` pane? I'm thinking that suggestion may not be accounting for the full hierarchy of map elements, but figured I'd ask!

I decided to test this! I couldn't get the pane configured to have a low enough z-index that it would properly sort as this ticket expects it to. Probably partially to do with stacking contexts, because stop markers usually go into the `overlayPane` and the station markers are in the `markerPane` which has a higher stacking context. When I put the stop markers in a pane in this ticket and moved it into the same stacking context, the z-index then applies.

assets/src/components/mapPage/mapDisplay.tsx
firestack: From what I've gathered, panes are simply ways to group elements via DOM nodes, which also allow you to manipulate stacking context and cascade z-index to it's children. So the name is pretty arbitrary, but there's a requirement that new panes have unique names.

firestack: addressed in [`6bd32b2` (#1952)](https://github.com/mbta/skate/pull/1952/commits/6bd32b2f341845cb0067b5081b20dfcba01e8f88)

PR #1954, opened 2023-03-03
feat(RoutePropertiesCard): Display headsign of selected route pattern

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1203948983628890/f

To deploy after https://github.com/mbta/skate/pull/1955

PR #1955, opened 2023-03-03
feat(RoutePatternController): Include headsign of representative trip

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1203948983628890/f

New headsign field used in RPC in https://github.com/mbta/skate/pull/1954

### DIFF
--- a/lib/skate_web/controllers/route_pattern_controller.ex
+++ b/lib/skate_web/controllers/route_pattern_controller.ex
@@ -13,12 +13,15 @@ defmodule SkateWeb.RoutePatternController do
         &Schedule.shape_with_stops_for_trip/1
       )
 
+    trip_fn = Application.get_env(:skate_web, :trip_fn, &Schedule.trip/1)
+
     route_patterns =
       route_id
       |> route_patterns_fn.()
       |> Enum.map(fn route_pattern ->
         Map.merge(Map.from_struct(route_pattern), %{
-          shape: shape_for_trip_fn.(route_pattern.representative_trip_id)
+          shape: shape_for_trip_fn.(route_pattern.representative_trip_id),
+          headsign: trip_fn.(route_pattern.representative_trip_id).headsign
         })
       end)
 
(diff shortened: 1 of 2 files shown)


lib/skate_web/controllers/route_pattern_controller.ex
```diff
@@ -13,12 +13,15 @@ defmodule SkateWeb.RoutePatternController do
         &Schedule.shape_with_stops_for_trip/1
       )
 
+    trip_fn = Application.get_env(:skate_web, :trip_fn, &Schedule.trip/1)
```
lemald, line 16: **suggestion non-blocking):** I'm not sure about the performance implications either way (we'd probably have to do some benchmarking), but one alternative to speed up the API call would be to actually have the route pattern returned by the `Schedule` server include the representative trip headsign (just fill it in during schedule parsing).

KaylaBrady: Thanks for the suggestion! I ran benchee comparison using the following function, where `route` is this proposed change without passing and returning a`conn`, and `route_no_headsign` is the status quo before this PR without `conn`.
```
  def compare_route_pattern_fns() do
    route_ids = Enum.map(Schedule.all_routes(), & &1.id)

    Benchee.run(
      %{
        "headsign_added_after" => fn -> Enum.map(route_ids, &route(&1)) end,
        "no_headsign" => fn -> Enum.map(route_ids, &route_no_headsign(&1)) end
      },
      time: 20
    )
    :ok
  end
  ```
  
Here are the results - the difference is small, especially considering a single iteration is running the function for all routes.
```
Name                           ips        average  deviation         median         99th %
headsign_added_after         59.95       16.68 ms    ±17.88%       15.84 ms       25.76 ms
no_headsign                  56.27       17.77 ms    ±21.20%       17.22 ms       26.08 ms

Comparison: 
headsign_added_after         59.95
no_headsign                  56.27 - 1.07x slower +1.09 ms

```

My proposal from this would be to keep this suggested change as-is. If it still seems worthwhile to do this upfront in the `Schedule.Data` for the purposes of representing this data type, my suggestion would be rather than adding `headsign` to the existing `GTFS.RoutePattern` type, we create a new `Schedule.RoutePattern` type for adding headsign and other non-standard fields.

lemald (approved):
Oops, meant to approve this last week as my suggestion was tagged "non-blocking."

PR #1956, opened 2023-03-03
Make shuttle map re-enable autocenter when selected shuttle changes

lemald:
Asana ticket: [⚙️🐛 Subsequent clicks on shuttle map entry don't start auto-centering](https://app.asana.com/0/1148853526253437/1203884615154359/f)

### DIFF
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -500,6 +500,3 @@ export const MapFollowingPrimaryVehicles = (props: Props) => {
     </BaseMap>
   )
 }
-
-const Map = MapFollowingPrimaryVehicles
-export default Map
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -69,8 +69,6 @@ export interface Props {
   stopCardDirection?: DirectionId
   includeStopCard?: boolean
   stations?: Stop[] | null
-  // if set, changes to this value will cause the autofollower to reset to following the new selection
-  followerResetKey?: string
 }
 
 export const defaultCenter: LatLngLiteral = {
@@ -331,7 +329,7 @@ export const usePickerContainerFollowerFn = () => {
 // #endregion
 // #endregion
 
-export const BaseMap = (props: Props): ReactElement<HTMLDivElement> => {
+const Map = (props: Props): ReactElement<HTMLDivElement> => {
   const mapRef: MutableRefObject<LeafletMap | null> =
     // this prop is only for tests, and is consistent between renders, so the hook call is consistent
     // eslint-disable-next-line react-hooks/rules-of-hooks
@@ -478,17 +476,12 @@ export const FollowerStatusClasses = (
 
 export const MapFollowingPrimaryVehicles = (props: Props) => {
   const state = useInteractiveFollowerState(),
-    { shouldFollow, setShouldFollow } = state
+    { shouldFollow } = state
 
   const positions: LatLng[] = props.vehicles.map(vehicleToLeafletLatLng)
 
-  useEffect(
-    () => setShouldFollow(true),
-    [props.followerResetKey, setShouldFollow]
-  )
-
   return (
-    <BaseMap {...props} stateClasses={FollowerStatusClasses(shouldFollow)}>
+    <Map {...props} stateClasses={FollowerStatusClasses(shouldFollow)}>
       <>
         <InterruptibleFollower
           positions={positions}
@@ -497,6 +490,32 @@ export const MapFollowingPrimaryVehicles = (props: Props) => {
         />
         {props.children}
       </>
-    </BaseMap>
+    </Map>
   )
 }
+
+export const MapFollowingSelectionKey = (
+  props: Props & { selectionKey?: string }
+) => {
+  const state = useInteractiveFollowerState(),
+    { shouldFollow, setShouldFollow } = state
+
+  const positions: LatLng[] = props.vehicles.map(vehicleToLeafletLatLng)
+
+  useEffect(() => setShouldFollow(true), [props.selectionKey, setShouldFollow])
+
+  return (
+    <Map {...props} stateClasses={FollowerStatusClasses(shouldFollow)}>
+      <>
+        <InterruptibleFollower
+          positions={positions}
+          {...state}
+          onUpdate={usePickerContainerFollowerFn()}
+        />
+        {props.children}
+      </>
+    </Map>
+  )
+}
+
+export default Map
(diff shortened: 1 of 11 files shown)


assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -130,6 +130,48 @@ describe("Shuttle Map Page", () => {
...
+    await animationFramePromise()
+
+    // We can't directly use the recenter button to disable recentering, but zooming works.
+    // However, for some reason a single change in zoom doesn't suffice to turn it off.
+    // There's probably something else we could await for to get the recentering disabled
+    // by firing just a single click
+    await userEvent.click(screen.getByRole("button", { name: "Zoom in" }))
+    await animationFramePromise()
+    await userEvent.click(screen.getByRole("button", { name: "Zoom out" }))
+    await animationFramePromise()
```
lemald, line 155: Open to suggestions for improving this! Since we don't have access to `mapRef` we can't use that, unfortunately.

assets/src/components/map.tsx
```diff
@@ -476,10 +478,15 @@ export const FollowerStatusClasses = (
...
 export const MapFollowingPrimaryVehicles = (props: Props) => {
   const state = useInteractiveFollowerState(),
-    { shouldFollow } = state
+    { shouldFollow, setShouldFollow } = state
 
   const positions: LatLng[] = props.vehicles.map(vehicleToLeafletLatLng)
 
+  useEffect(
+    () => setShouldFollow(true),
+    [props.followerResetKey, setShouldFollow]
```
KaylaBrady, line 487: I messed up by adding this as a commit-level comment rather than PR comment - copying here for easier conversation! Original discussion [here](https://github.com/mbta/skate/commit/4877dbec0456966aa20ada968e3e4258183cf793#r102950878):

> One thing I'm trying to understand is why this fix is needed for the shuttle page and not the search map page which has a similar sidebar selection mechanism. It looks like MapDisplay avoids this issue by using BaseMap with a custom child that uses some [custom selection state](https://github.com/mbta/skate/blob/e1bb5da335693db2ed0f38238f13a4861d2b5a93/assets/src/components/mapPage/mapDisplay.tsx#L331).
> 
> I'd be interested in seeing if a similar solution would work here (if-minor, which I suspect it might not be!) If it isn't minor, what do you think about separating the <BaseMap> props type from the <MapFollowingPrimaryVehicles/> props type so that this prop is only added for <MapFollowingPrimaryVehicles/>?

config/dev.exs
```diff
@@ -1,7 +1,7 @@
 import Config
 
 config :skate,
-  tileset_url: "https://mbta-map-tiles-dev.s3.amazonaws.com/osm_tiles",
+  tileset_url: "https://mbta-map-tiles-dev.s3.amazonaws.com/skate_osm_tiles",
```
KaylaBrady, line 4: praise: Thank you for changing this -- much nicer to see the tiles locally match what is in prod!

KaylaBrady, line 4: praise: Thank you for changing this -- much nicer to see the tiles locally match what is in prod!

assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -130,6 +130,47 @@ describe("Shuttle Map Page", () => {
...
+      </StateDispatchProvider>
+    )
+    await animationFramePromise()
+
+    // We can't directly use the recenter button to disable recentering, but zooming works.
+    // However, there is currently a bug causing us to not disable autocentering with a single
+    // click on one of the zoom buttons, instead requiring two.
+    await userEvent.click(screen.getByRole("button", { name: "Zoom in" }))
+    await animationFramePromise()
+    await userEvent.click(screen.getByRole("button", { name: "Zoom out" }))
```
firestack, line 153: Could you use the selector helpers from `assets/tests/testHelpers/selectors/components/map.ts`?

assets/src/components/map.tsx
firestack: I was having trouble wording it but I support what Kayla's getting at above. I think the way to go is using `<BaseMap>` on the shuttle page and keeping special logic there rather than adding more props to the `<BaseMap>`

assets/tests/components/shuttleMapPage.test.tsx
lemald: Updated in 84f7a5d8.

assets/src/components/map.tsx
lemald: Okay, I think I see the advantages but I'm just struggling to tease apart the different pieces of work that would be needed to do this neatly. One thing that jump out to me is that the `<Map />` component actually encapsulates the logic used by the minimap and search page, whereas `<BaseMap />` is the more generic version, and I think it would be good to reorganize and rename some components to make that clearer (so that `<Map />` is the most generic version that all other components build off of).

KaylaBrady: This seems OK to me! I'm struggling a bit with the name `MapFollowingSelectionKey` since the map is still following the primary vehicles, and only the focus is reset based on the selectionKey. Maybe something like `MapWithFocusResetKey`? It still seems ideal to not introduce a new special map if possible, but I consider that non-blocking unless @firestack sees it differently!

KaylaBrady (approved):

assets/src/components/map.tsx
lemald: Do you mean special map, or special prop? I could just use `<Map />` directly from the `<SearchPage />` component and put the logic there, but then it wouldn't be reusable by anything else that has similar functionality where changes in selection are supposed to reset the follower state. That means no new component and no new prop but it's also less modular, and this way the logic isn't added to the same map that other components use.

KaylaBrady: I meant special map, but your point about reusability is well taken and am fully persuaded by it! It would be nice for consistency if `<SearchPage/> ,`<ShuttleMapPage/>`, and any other future interactive map to use the same underlying reusable map component.

assets/src/components/map.tsx
```diff
@@ -478,17 +476,36 @@ export const FollowerStatusClasses = (
...
-    { shouldFollow, setShouldFollow } = state
+    { shouldFollow } = state
 
   const positions: LatLng[] = props.vehicles.map(vehicleToLeafletLatLng)
 
-  useEffect(
-    () => setShouldFollow(true),
-    [props.followerResetKey, setShouldFollow]
+  return (
+    <Map {...props} stateClasses={FollowerStatusClasses(shouldFollow)}>
```
firestack, line 484: comment(non-blocking): I just noticed that if `props` specifies `stateClasses` we're discarding those here. This isn't an issue _now_ but I wanted to make others aware of this logic (since I originally wrote it and I didn't mean to miss that).

firestack (approved):

PR #1958, opened 2023-03-06
feat(/admin): New admin tools index page

KaylaBrady:
Ticket: https://app.asana.com/0/1152340551558956/1203938664653490/f
Notes:
* `reports`, and `test_groups` had very similar layouts, with the important difference seeming to be their title. This condenses  these and the `admin` index page to all use a single `barebones` layout with a variable `title`.
* The health check page only makes an API request to `_health`, but doesn't display anything. Should this display anything? Is it still useful to include?


lemald:
> The health check page only makes an API request to _health, but doesn't display anything. Should this display anything? Is it still useful to include?

That endpoint mainly exists for ECS to determine if the instance is healthy and should have traffic routed to it - I'm actually kind of surprised it's even exposed outside of the load balancer. Given that the result should always be a 200 if you're able to reach the instance from the outside, I'm not sure if it's worth it - the best way to dig into that information if there's an issue is probably to look in the AWS console.

lemald (approved):

lib/skate_web/templates/admin/index.html.eex
```diff
@@ -0,0 +1,8 @@
+<h2>Admin Tools</h2>
```
firestack, line 1: nitpick(non-blocking): Are there other headers on the page "preventing" this from being `<h1>`?

firestack (approved):

lib/skate_web/templates/admin/index.html.eex
KaylaBrady: Thanks for noticing!

PR #1962, opened 2023-03-09
init(ts/app/demo-mode): demo mode 

firestack:
I wanted to try implementing the demo mode quickly. I was able to get it working with just this, but wasn't sure how to test it or if there should be a better way to toggle the demo mode than using the test groups feature.

---

Asana Ticket: https://app.asana.com/0/1152340551558956/1203938664653490/f
Related: https://github.com/mbta/skate/pull/1958#issuecomment-1458315134

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useContext } from "react"
+import React, { ReactElement, useContext, useEffect } from "react"
 import { Socket } from "phoenix"
 import {
   BrowserRouter,
@@ -26,6 +26,7 @@ import { allOpenRouteIds } from "../models/routeTab"
 import Nav from "./nav"
 import RightPanel from "./rightPanel"
 import { mapModeForUser } from "../util/mapMode"
+import inTestGroup from "../userInTestGroup"
 
 export const AppRoutes = () => {
   useAppcues()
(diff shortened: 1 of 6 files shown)


assets/src/components/app.tsx
```diff
@@ -96,6 +97,10 @@ export const AppRoutes = () => {
 }
 
 const App = (): ReactElement<HTMLDivElement> => {
+  useEffect(() => {
+    document.documentElement.dataset.demoMode =
+      inTestGroup("demo-mode").toString()
+  })
```
lemald, line 103: **question:** does this need to be in a `useEffect`? If so, should it have an empty dependency array since the return value of `inTestGroup` will only change on a page reload? Although maybe there's something I'm overlooking about `documentElement` here.

firestack: Thanks for catching this! I had meant for it to be a empty dependency array, I kidna thought it would do that by default and forgot to check my knowledge against the docs.


I don't know if a `useEffect` is the right tool here, I effectively wanted a way to run code once when we create the application and figured this was a good place to start. If you can think of a better place to put this I'm all for it.

firestack: I tried to address this in 068fd33267fc1302ace44398b07aa7c932bb58ac, I'm curious to see what you think!

lemald: I think that makes sense! Unfortunately we don't have tests for that `app.tsx` but we do have [an Asana ticket](https://app.asana.com/0/1152340551558956/1202219847571509/f) to add some.

lemald (approved):

PR #1969, opened 2023-03-10
fix: make drawers not visible after collapse

firestack:
# Summary
This animates the `visibility` property after the animation finishes when collapsing, and before the animation starts when expanding.

I was unable to update tests for this because jsdom doesn't have access to our scss files when testing.

I also opted to add `u-hide` to all elements which needed to be hidden, but the `drawer_hide` selector could also be inverted to select all children but the button. I think changing the document structure would be better but it resulted in a much larger changeset.

### DIFF
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -189,7 +189,7 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
 
       <PickerContainer>
         <>
-          <div className="m-ladder-page__routes-presets-toggle">
+          <div className="m-ladder-page__routes-presets-toggle u-hide">
             <button
               id="m-ladder-page__routes_picker_button"
               className={
--- a/assets/css/_drawer_hide.scss
+++ b/assets/css/_drawer_hide.scss
@@ -1,0 +1,11 @@
+.u-hide {
+  transition: visibility 0s 1.2s;
+
+  .visible & {
+    transition-delay: 0s;
+  }
+
+  .hidden & {
+    visibility: hidden;
+  }
+}
--- a/assets/css/_drawer_hide.scss
+++ b/assets/css/_drawer_hide.scss
@@ -1,11 +1,14 @@
 .u-hide {
-  transition: visibility 0s 1.2s;
+  // Delay visibility transition to end of slide transition
+  transition: visibility 0s $transition-slide-duration;
 
   .visible & {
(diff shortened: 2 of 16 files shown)


assets/src/components/ladderPage.tsx
```diff
@@ -189,7 +189,7 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
 
       <PickerContainer>
         <>
-          <div className="m-ladder-page__routes-presets-toggle">
+          <div className="m-ladder-page__routes-presets-toggle u-hide">
```
lemald, line 192: **question:** why does the `u-hide` class go here instead of on the main div of the `PickerContainer` component? It looks like this will only apply to the toggle buttons between routes and presets, not the whole picker.

firestack: > I also opted to add u-hide to all elements which needed to be hidden, but the drawer_hide selector could also be inverted to select all children but the button. I think changing the document structure would be better but it resulted in a much larger changeset.

I could invert the selector and then add `u-hide` to the parent, but the visibility value needs to go on the actual content, and because the content is a sibling of the drawer tab, setting visibility on the picker container would hide the drawer tab and make it impossible to expand it again.

lemald: Ah, sorry, I should have read the description more thoroughly.

By "changing the document structure," what exactly did you have in mind? Putting the `{children}` in the `RoutePicker` component in their own container that gets the `u-hide` class? Or something else?

firestack: > Ah, sorry, I should have read the description more thoroughly.
No problem! I just wanted to point it out that it is an option I'm happy to try.

> By "changing the document structure," what exactly did you have in mind? Putting the `{children}` in the `RoutePicker` component in their own container that gets the `u-hide` class? Or something else?

Yeah, when I tried to make a "unified" drawer component it looked like
```
- (optional absolute position) drawer container
	- content container (optional, used for animation)
		- header (screen reader only by default)
			- button
			- div.children (just a container to address all children at once, even if not a single element) 
				- {children}
```

and I think that'd be

lemald: Ah, right, I forgot about those issues with CSS encapsulation. Okay, yeah, I think this makes sense as an incremental improvement over what we have now.

lemald (approved):

PR #1973, opened 2023-03-13
feat(ts/components/StreetViewSwitch): add FS event with source data

firestack:
I decided to also add this to the street view buttons, just in case we get any other weirdness with street view.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1204139772279191/f

### DIFF
--- a/assets/tests/components/map.test.tsx
+++ b/assets/tests/components/map.test.tsx
@@ -343,7 +343,8 @@ describe("<MapFollowingPrimaryVehicles />", () => {
 
     expect(openSpy).toHaveBeenCalled()
 
-    expect(streetViewModeSwitch.get()).toBeInTheDocument()
+    expect(streetViewModeSwitch.get()).toBeVisible()
+    expect(streetViewModeSwitch.get()).not.toBeChecked()
   })
 
   test("turning off street view also fires a FullStory event", async () => {
--- a/assets/tests/components/map.test.tsx
+++ b/assets/tests/components/map.test.tsx
@@ -325,6 +325,10 @@ describe("<MapFollowingPrimaryVehicles />", () => {
 
     await userEvent.click(mapRef.current!.getPane("mapPane")!)
 
+    /**
+     * These "magic numbers" correspond to the nearest floating point value of
+     * our map's {@link defaultCenter} (and also constrained by `maxBounds`).
+     */
     const latitude = 42.360700296138525
     const longitude = -71.0588836669922
     const url = streetViewUrl({ latitude, longitude })
--- a/assets/tests/components/map.test.tsx
+++ b/assets/tests/components/map.test.tsx
@@ -333,7 +333,7 @@ describe("<MapFollowingPrimaryVehicles />", () => {
       2,
       "User clicked map to open street view",
       {
-        streetViewUrl_str: `${url}`,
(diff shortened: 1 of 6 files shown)


assets/tests/components/map.test.tsx
```diff
@@ -314,19 +316,35 @@ describe("<MapFollowingPrimaryVehicles />", () => {
...
+    )
 
     await userEvent.click(mapRef.current!.getPane("mapPane")!)
 
-    expect(openSpy).toHaveBeenCalled()
-    expect(window.FS!.event).toHaveBeenCalledWith(
-      "Dedicated street view toggled",
-      { streetViewEnabled_bool: true }
+    const latitude = 42.360700296138525
+    const longitude = -71.0588836669922
```
lemald, line 329: **question (non-blocking):** where do these exact values come from? Is it possible to somehow more explicitly link these to the coordinates where the `userEvent.click` happens?

lemald (approved):
One question but basically looks good!

assets/tests/components/map.test.tsx
firestack: I "tofu"-ed it, which is to say, I copied the "right" answer from a failing test to make the test pass :^)

I don't know _why_, but it seems like these values correspond to the map's default center. But, just plugging in defaultCenter for these values doesn't work. I could see this being for a few reasons, 
1. floating point rounding because our decimal representation of default center doesn't fall on a floating point value
2. map size + click position (the upper left or wherever the click happens could be anywhere, but it seems incredibly close to the default center, so maybe the map is of 1x1 size or something)
3. (and as always it could be something else)

firestack: jest matchers have "to be close to" for floats, but the string representation of the URL is trickier to match with.

I'd chose `(0, 0)` as the default center for tests if I could easily override that, but the `<Map>` doesn't expose that (which also means if any other agency decided to use skate, we'd be forcing them to look at Boston lol, I'll look for/write up with a tech debt ticket to get to someday)

firestack: If I remove `maxBounds` and set default center to `(0, 0)`, then the tests do pass, so I think this is floating point rounding.

lemald: > jest matchers have "to be close to" for floats, but the string representation of the URL is trickier to match with.

Ah, yeah, I was about to suggest looking at other jest matchers but that's a good point about the URL, unless you want to do some annoying parsing and processing of it before matching. Anyhow, I think it's good as-is, I was just curious!

firestack: I don't like it and would prefer to avoid this! I'll add some comment to the test so that if we can fix it(or if we break the test) in the future it is at least obvious what's happening

KaylaBrady: This seems helpful to me! May be worth also linking directly to this test from the jest upgrade ticket you added, or as a separate ticket dependent on the jest upgrade one.

lemald: Looks good to me as well!

firestack: Ticket opened 👍🏻 https://app.asana.com/0/1152340551558956/1204178291670771/f

PR #1978, opened 2023-03-17
Fail tests on unexpected console warning / error

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1203458460363782/f

See examples of what the failing test output will look like [here](https://github.com/mbta/skate/actions/runs/4441085829/jobs/7795718019).

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -72,7 +72,7 @@
     "file-loader": "^6.2.0",
     "fishery": "^2.2.2",
     "jest": "^26.6.3",
-    "jest-fail-on-console": "^3.0.2",
+    "jest-fail-on-console": "^3.1.0",
     "mini-css-extract-plugin": "^2.6.0",
     "node-sass": "^8.0.0",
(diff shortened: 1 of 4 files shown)


lemald (approved):
Awesome!

assets/package.json
```diff
@@ -72,6 +72,7 @@
     "file-loader": "^6.2.0",
     "fishery": "^2.2.2",
     "jest": "^26.6.3",
+    "jest-fail-on-console": "^3.0.2",
```
firestack, line 75: note(non-blocking): It looks like [`3.1.0`](https://www.npmjs.com/package/jest-fail-on-console/v/3.1.0) was posted 4 days ago! can we use that?

firestack (approved):
Yay! glad to have this!
one note about the version of the package but otherwise LGTM!

assets/package.json
KaylaBrady: Thanks for noticing, just updated!

PR #1997, opened 2023-03-29
fix(css/vehiclePropertiesCard): update vpc css prefix to `c-`

firestack:
Asana Ticket: https://app.asana.com/0/0/1204265743414890/f


lemald (approved):
Also did some quick QA and the styling still looks correct. 👍

PR #1998, opened 2023-03-29
refactor: css prefix cleanup chunk

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1203760641228313/f

5 class renames split by commit.

### DIFF
--- a/assets/src/components/settingsPage.tsx
+++ b/assets/src/components/settingsPage.tsx
@@ -27,13 +27,13 @@ const SettingsPage = (): ReactElement<HTMLDivElement> => {
   const mobileMenuClass = mobileMenuIsOpen ? "blurred-mobile" : ""
 
   return (
-    <div className={`l-page m-settings-page ${mobileMenuClass}`}>
+    <div className={`l-page c-settings-page ${mobileMenuClass}`}>
       <Notifications />
       <div className="c-page__container">
         <h1 className="c-page__title">Settings</h1>
 
         <div className="c-page__section">
-          <h2 className="m-settings-page__section-header">Vehicle Settings</h2>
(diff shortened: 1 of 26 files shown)


KaylaBrady:
Requested review too early - taking a pass to convert some classenames to layout

KaylaBrady:
Converted some of the internal card classes to layout classes, so this is now ready for review!

assets/src/components/settingsPage.tsx
```diff
@@ -27,7 +27,7 @@ const SettingsPage = (): ReactElement<HTMLDivElement> => {
   const mobileMenuClass = mobileMenuIsOpen ? "blurred-mobile" : ""
 
   return (
-    <div className={`c-page c-page--settings ${mobileMenuClass}`}>
+    <div className={`l-page m-settings-page ${mobileMenuClass}`}>
```
lemald, line 30: **question:** it looks like (if I'm using `git grep` correctly) there is no actual styling for the `c-page--settings` / `m-settings-page` classes. Does it make sense to keep this class if there's no actual styling associated with it? I don't have a strong opinion but I'd lean slightly in the direction of avoiding classnames that aren't used somewhere.

KaylaBrady: I'm glad you caught that! Happy to remove.

KaylaBrady: Actually changing my answer: I think it is still worth including since there are descendent like `m-settings-page__section-header` exist, so I think we should have a class name to define the Block. I'll rename to `c-settings-page` though

lemald:
Update: the changes are actually not hard to look over quickly it turns out, I ran through the various different pages and components and didn't see any issues. 👍

firestack (approved):

PR #2000, opened 2023-03-30
fix(css/route-pill): convert `m-route-pill` to `c-` component prefix

firestack:
I checked on my local dev env that the bus stop card still renders okay and it seemed fine.
image

---

Asana Ticket: https://app.asana.com/0/0/1204265743414868/f


KaylaBrady (approved):

lemald (approved):
I also did some quick QA in addition to looking at the code and it looks good! 👍

PR #2001, opened 2023-03-30
fix(css-prefixes): kf/chunk/1

firestack:
I liked the way @KaylaBrady was chunking CSS so I decided to try it, Thanks Kayla!

Each commit should be individually reviewable and a atomic update to the css classes.

Everything in this was changed over to component prefixes, I didn't see opportunities for other prefixes but I also wasn't looking super closely on the larger changesets.

---

<details>
<summary>
My change process
</summary>

I've been using `ripgrep` and `sd` (and `fish` shell) to update files using the form similar to the following snippet

```
❯ sd m-vehicle-icon c-vehicle-icon (rg m-vehicle-icon -l)
```

</details>

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1203760641228313

### DIFF
--- a/assets/css/_ladder.scss
+++ b/assets/css/_ladder.scss
@@ -29,29 +29,29 @@
   cursor: pointer;
 
   &:hover {
-    .m-vehicle-icon__triangle {
+    .c-vehicle-icon__triangle {
       stroke: $color-secondary-light;
     }
 
-    .m-vehicle-icon__ghost-highlight {
+    .c-vehicle-icon__ghost-highlight {
       stroke: $color-secondary-light;
     }
 
-    .m-vehicle-icon__label-background {
+    .c-vehicle-icon__label-background {
       fill: $color-secondary-light;
     }
   }
 
   &.selected {
-    .m-vehicle-icon__triangle {
+    .c-vehicle-icon__triangle {
       stroke: $color-secondary-medium;
     }
 
-    .m-vehicle-icon__ghost-highlight {
+    .c-vehicle-icon__ghost-highlight {
       stroke: $color-secondary-medium;
     }
 
-    .m-vehicle-icon__label-background {
+    .c-vehicle-icon__label-background {
       fill: $color-secondary-medium;
     }
   }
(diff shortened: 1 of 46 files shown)


assets/css/_ladder.scss
```diff
@@ -29,29 +29,29 @@
...
       stroke: $color-secondary-light;
     }
 
-    .m-vehicle-icon__label-background {
+    .c-vehicle-icon__label-background {
       fill: $color-secondary-light;
     }
   }
 
   &.selected {
```
lemald, line 45: **question:** in a few places we do have classnames like this that represent some sort of component state that don't fall into our overall naming scheme... should we do something with them or let that wait for another time?

lemald: One point of view is that this is what the "M" in "BEM" is for, so it should actually be `.m-ladder__vehicle--selected`.

lemald:
@firestack minor nitpick, but I the commit message for fee1915 mentions `c-vehicle-map-state` rather than `c-vehicle-map`, the actual class base name in question.

assets/src/components/routeVariantName.tsx
```diff
@@ -17,7 +17,7 @@ export const RouteVariantName = ({
   return (
     <output
       aria-label="Route Variant Name"
-      className={"m-route-variant-name " + className}
+      className={"c-route-variant-name " + className}
```
lemald, line 20: **issue (non-blocking):** I actually just noticed while QA-ing that `className` can actually be undefined, meaning that we put the literal string `"undefined"` here. If you want to fix it quickly that could be nice but if you feel like it's out of scope for this PR that's fine too.

lemald (dismissed):
Okay, I left some non-blocking comments but otherwise this looks good and I did a bit of QA on all of the changed classes and didn't notice any issues. 👍

assets/css/_ladder.scss
firestack: I don't think we should handle it in this PR but I think it's reasonable to update it to match, maybe we just update them as we catch them?
Either way, I'll open a PR for this one

cc: @KaylaBrady

firestack: https://github.com/mbta/skate/pull/2004

PR #2010, opened 2023-04-03
fix(css-prefix): kf/chunk/4

firestack:
Similar to #2006 and #2008 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204265743414857
  - https://app.asana.com/0/0/1204265743414858
  - https://app.asana.com/0/0/1204265743414860
  - https://app.asana.com/0/0/1204265742473559

PR #2012, opened 2023-04-03
fix(css-prefixes): rename u-hide to u-hideable and fix `.visible`|`.hidden`

firestack:
`u-hide` was borrowing the `.hidden` class that the different panels were using to implement it's features, this splits them apart and makes each page a bit more encapsulated with BEM.

---
Merge After: #2010 

---
  - https://app.asana.com/0/0/1204265743414896
  - https://app.asana.com/0/0/1204321116044633
  - https://app.asana.com/0/0/1204321116044637


assets/css/_search_page.scss
```diff
@@ -42,14 +42,6 @@ $search-page-search-panel-width: 22.0625rem;
...
     margin: 0 1.5625rem;
   }
-
-  &.visible {
-    left: 0;
-  }
-
-  &.hidden {
-    left: calc(-1 * #{$search-page-search-panel-width} + 1rem);
-  }
```
lemald, line 52: I think you're right that this wasn't doing anything for the search page. It looks like I introduced it back in [this commit](https://github.com/mbta/skate/pull/1887/commits/9726fb39c5c73ed4aa751d878787e3e2de4bf3bf), I think in my zeal to make sure that all of the changes to the styling in the map page made it back to the search page I carried over some stuff about hiding the panel that isn't relevant.

lemald (approved):
Okay, I think this looks good! See my comment, I think I figured out what was up with those `visible` and `hidden` styles that you got rid of.

PR #2019, opened 2023-04-10
Stop selected vehicle from disappearing when no longer on its original route

lemald:
Asana ticket: [⚙️🐛 Selected vehicle can disappear from map in Search Map page](https://app.asana.com/0/1200180014510248/1204291981474926/f)

It turns out that we actually had the vehicle from `useVehicleForId`, but the actual display was happening via the `RouteVehicles` component which uses `useVehiclesForRoute`. This separates things out so that `SelectedVehicleDataLayers` always displays the selected vehicle directly, and `RouteVehicles` displays all of the vehicles on the route _except for_ the selected one.

Also extends one of the snapshot `MapPage` tests to display a non-primary vehicle on the route, to verify that both are rendered properly before and after this change.

### DIFF
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -55,18 +55,19 @@ const RouteVehicles = ({
   const vehicles = useVehiclesForRoute(socket, selectedVehicleRoute)
   return (
     <>
-      {filterVehicles(vehicles).map((vehicle: Vehicle) => {
-        const isSelected = vehicle.id === selectedVehicleId
-        return (
-          <VehicleMarker
-            key={vehicle.id}
-            vehicle={vehicle}
-            isPrimary={isSelected === true}
-            isSelected={isSelected}
-            onSelect={onPrimaryVehicleSelect}
-          />
-        )
-      })}
+      {filterVehicles(vehicles)
+        .filter((vehicle) => vehicle.id !== selectedVehicleId)
+        .map((vehicle: Vehicle) => {
+          return (
+            <VehicleMarker
+              key={vehicle.id}
+              vehicle={vehicle}
+              isPrimary={false}
+              isSelected={false}
+              onSelect={onPrimaryVehicleSelect}
+            />
+          )
+        })}
     </>
   )
 }
@@ -361,21 +362,22 @@ const SelectedVehicleDataLayers = ({
               />
             </SelectionCardContainer>
           )}
-          {isVehicle(selectedVehicleOrGhost) &&
-            (selectedVehicleOrGhost?.isShuttle ? (
+          {isVehicle(selectedVehicleOrGhost) && (
+            <>
               <VehicleMarker
                 key={selectedVehicleOrGhost.id}
                 vehicle={selectedVehicleOrGhost}
                 isPrimary={true}
                 isSelected={true}
               />
-            ) : (
+
               <RouteVehicles
                 selectedVehicleRoute={selectedVehicleOrGhost.routeId}
                 selectedVehicleId={selectedVehicleOrGhost.id}
                 onPrimaryVehicleSelect={selectVehicle}
               />
-            ))}
+            </>
+          )}
           {showShapeAndStops && routePatternForVehicle && (
             <RoutePatternLayers
               routePattern={routePatternForVehicle}
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -371,11 +371,13 @@ const SelectedVehicleDataLayers = ({
                 isSelected={true}
               />
 
-              <SecondaryRouteVehicles
-                selectedVehicleRoute={selectedVehicleOrGhost.routeId}
-                selectedVehicleId={selectedVehicleOrGhost.id}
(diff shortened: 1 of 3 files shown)


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -55,18 +55,19 @@ const RouteVehicles = ({
...
+      {filterVehicles(vehicles)
+        .filter((vehicle) => vehicle.id !== selectedVehicleId)
+        .map((vehicle: Vehicle) => {
+          return (
+            <VehicleMarker
+              key={vehicle.id}
+              vehicle={vehicle}
+              isPrimary={false}
+              isSelected={false}
+              onSelect={onPrimaryVehicleSelect}
```
KaylaBrady, line 67: suggestion: Maybe for clarity this component could be renamed `SecondaryRouteVehicles`? And similarly `onPrimaryVehicleSelect` `onVehicleSelect`

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -361,21 +362,22 @@ const SelectedVehicleDataLayers = ({
...
+            <>
               <VehicleMarker
                 key={selectedVehicleOrGhost.id}
                 vehicle={selectedVehicleOrGhost}
                 isPrimary={true}
                 isSelected={true}
               />
-            ) : (
+
               <RouteVehicles
```
KaylaBrady, line 374: question: Is it intentional for `<RouteVehicles/>` to be included for shuttles?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -361,21 +362,22 @@ const SelectedVehicleDataLayers = ({
...
                 selectedVehicleRoute={selectedVehicleOrGhost.routeId}
                 selectedVehicleId={selectedVehicleOrGhost.id}
                 onPrimaryVehicleSelect={selectVehicle}
               />
-            ))}
+            </>
```
KaylaBrady, line 379: question(non-blocking): Based on my understanding, if the selected vehicle changes route, this will automatically switch over to rendering the new route shape + secondary vehicles because `selectedVehicleRoute={selectedVehicleOrGhost.routeId}` will change. Is that correct? 

I'm trying to wrap my head around the bug video and am getting stuck on why the route shape didn't clear if the primary vehicle was no longer assigned to the 23. That may be separate from the focus of this ticket though!

lemald: That seems correct based on my observations.

Honestly I'm still trying to wrap my head around the _exact_ set of circumstances needed to reproduce this as well. The common cases I run through all seem to be handled more-or-less correctly in Skate already, and it's hard to find an example of a bus that's going to go off-route. Maybe I could do some poking around with bus routes that have detour alerts up.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -55,18 +55,19 @@ const RouteVehicles = ({
...
-            key={vehicle.id}
-            vehicle={vehicle}
-            isPrimary={isSelected === true}
-            isSelected={isSelected}
-            onSelect={onPrimaryVehicleSelect}
-          />
-        )
-      })}
+      {filterVehicles(vehicles)
+        .filter((vehicle) => vehicle.id !== selectedVehicleId)
```
firestack, line 59: is it possible that the selected vehicle from `useVehiclesForRoute` has newer data than the vehicle returned from `useVehicleById`?

assets/src/components/mapPage/mapDisplay.tsx
lemald: Not necessarily, I guess I need to think this part of the logic over a little bit more carefully.

assets/src/components/mapPage/mapDisplay.tsx
lemald: It's possible that one will be ever so slightly newer than the other, but basically the way it works on the backend is that once new vehicle data comes in and is processed, it pushes the update out on all applicable channels with vehicle subscriptions, so while there may be slight variations in the order in which different channels get them it's unlikely to be a noticeable difference.

assets/src/components/mapPage/mapDisplay.tsx
KaylaBrady: Yeah it seems tricky to reproduce! An assumption I made in the comment above about the video was wrong having watched at it again 
> am getting stuck on why the route shape didn't clear if the primary vehicle was no longer assigned to the 23.
I think the vehicle actually was still assigned to the 23 since the VPC still showed 23. So My concern about the route changing was not relevant! 

In any case, maybe it would be worth adding a test case where the selected vehicle is *not* returned by `useVehiclesForRoute` but is returned by `useVehicleForId`?

firestack: > In any case, maybe it would be worth adding a test case where the selected vehicle is not returned by `useVehiclesForRoute` but is returned by `useVehicleForId`?

Adding my support for adding this test case!

firestack (commented):
One [note](https://github.com/mbta/skate/pull/2019#discussion_r1162839017) about test cases but LGTM otherwise.

PR #2021, opened 2023-04-11
feat(NotificationsDrawer): New loading state

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1202584021125991/f
This adds `null` to the type of notifications.
A couple of other approaches I considered
* using `FetchResult` 
* incorporating the existing `isInitialLoad` variable which isn't currently part of the state returned by `useNotificationsReducer`

Ultimately I opted for this simpler change of only adding `null` to the type, but open to other suggestions!

### DIFF
--- a/assets/src/hooks/useNotificationsReducer.ts
+++ b/assets/src/hooks/useNotificationsReducer.ts
@@ -19,12 +19,12 @@ import {
 } from "./useNotifications"
 
 export interface State {
-  notifications: Notification[]
+  notifications: Notification[] | null
   showLatestNotification: boolean
 }
 
 export const initialState: State = {
-  notifications: [],
+  notifications: null,
   showLatestNotification: false,
 }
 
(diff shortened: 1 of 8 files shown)


assets/src/hooks/useNotificationsReducer.ts
```diff
@@ -107,41 +107,47 @@ export type Dispatch = ReactDispatch<Action>
...
+      return notifications
+        ? notifications.map((notification) =>
+            notification.id === notificationToToggle.id
+              ? {
+                  ...notification,
+                  state: otherNotificationReadState(notification.state),
+                }
+              : notification
+          )
+        : null
```
firestack, line 150: suggestion(non-blocking): because `notifications` is possibly `null`, if you use the `&&` operator instead of the ternary operator, I think it should return `null` if `notifications` is `null`?

But if you prefer to be explicit with the ternary operator I'm fine with that!

KaylaBrady: Thanks for the suggestion, updated in [4e15e57](https://github.com/mbta/skate/pull/2021/commits/4e15e5750a0dd621a7cbd89439398529983a88bb)

firestack (approved):
I agree with you that it'd be better to have some sort of type to describe the loading state, but! this works so maybe we create some backlog item for the future.
Thanks Kayla!

assets/tests/components/notificationDrawer.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react"
-import { render } from "@testing-library/react"
-import renderer from "react-test-renderer"
+import { render, screen } from "@testing-library/react"
+import "@testing-library/jest-dom"
```
firestack, line 3: praise: thank you for updating this!!

KaylaBrady:
Thank you both, added a rough ticket for that [here](https://app.asana.com/0/1148853526253420/1204383901911922/f)!

PR #2025, opened 2023-04-14
cleanup(ex/skate/database): remove users without email

firestack:
We confirmed locally that when a user is deleted in the DB the deletion cascades to records linked via foreign key. We considered writing a test for this but due to unknown complexity opted for testing manually.

We've deployed this on dev green and successfully removed 1 user which did not have an email associated. We also need to run this on `dev blue`, `dev`, and `prod` to clean up the DB records for each environment, before we remove this code in #2026.

DB Cleanup applied to
- [x] Dev Green
- [x] Dev Blue
- [ ] Dev (will happen after merging this)
- [ ] Prod (will happen on production release)

---
  - https://app.asana.com/0/0/1204389997693497

### DIFF
--- a/lib/skate/application.ex
+++ b/lib/skate/application.ex
@@ -32,6 +32,7 @@ defmodule Skate.Application do
     link = Supervisor.start_link(children, strategy: :one_for_all, name: Skate.Supervisor)
 
     Migrate.up()
+    Skate.RemoveUsersWithoutEmail.run()
 
     link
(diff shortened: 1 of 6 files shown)


lemald (approved):
I'm always sort of conflicted as to whether this sort of thing should be a migration, but it's not reversible like most of the other things we put in migrations and this approach works fine. 👍

PR #2030, opened 2023-04-19
feat(EnsureEnvironmentAccess): Restrict access to dev environments

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204400571478306/f

The environment variable `RESTRICT_ENVIRONMENT_ACCESS?` has not yet been added in terraform, but I started a branch for that [here](https://github.com/mbta/devops/compare/kb-skate-restrict-env-access-var). Hoping to get feedback on this first. Including on whether using the word "ENVIRONMENT" here is unnecessary or confusing. 

I largely followed the pattern of the existing `EnsureAdminGroup` plug.


config/config.exs
```diff
@@ -13,6 +13,7 @@ config :skate,
   # Default. Can be configured via environment variable, which is loaded in application.ex
   api_url: {:system, "API_URL"},
   api_key: {:system, "API_KEY"},
+  restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS?") == "true",
```
firestack, line 16: Based on the following links, I think the `?` should be removed
- https://stackoverflow.com/a/2821201
- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_10_02
- https://stackoverflow.com/a/2821183


> > Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.
> 
> So while the names may be valid, your shell might not support anything besides letters, numbers, and underscores.

config/config.exs
```diff
@@ -13,6 +13,7 @@ config :skate,
   # Default. Can be configured via environment variable, which is loaded in application.ex
   api_url: {:system, "API_URL"},
   api_key: {:system, "API_KEY"},
+  restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS") == "true",
```
lemald, line 16: **issue:** sorry I don't have a good explainer handy to link (I'm trying and failing to find the right spot in the docs to point to), but since the configuration is read in at compile time rather than runtime this won't reflect changes to this environment variable in different application environments. You'll want to use `config/releases.exs` (which is evaluated at runtime) instead.

lib/skate_web/templates/unauthorized/restricted_environment.html.eex
```diff
@@ -0,0 +1,6 @@
+
+<div>
+    Hi there!<br>
+    The URL you followed is actually for Skate's test environment, meant for testing purposes only. Please do not use this URL; use <a href="https://skate.mbta.com">skate.mbta.com</a> instead.<br> If you're unsure how this happened or are confused, please reach out to the Skate team at
+    <a href = "mailto: skateteam@mbta.com">skateteam@mbta.com</a>.
```
lemald, line 5: **nitpick:** Is it correct to have a space after the `:` and before the email address?

config/config.exs
KaylaBrady: Ah yes, I remember [this doc](https://elixir-lang.org/getting-started/mix-otp/config-and-releases.html#configuration) about that. Thanks for catching!

lib/skate_web/templates/unauthorized/restricted_environment.html.eex
KaylaBrady: Seems to work either way! I pulled it from a [w3 example](https://www.tutorialspoint.com/html/html_email_links.htm) but can change if it is more conventional to omit the space!

config/config.exs
KaylaBrady: Though looking again that doc specifies the file `config/runtime.exs`, which seems like an evolution of `releases.exs`. Looks like `releases.exs` maybe isn't automatically used when running locally? But it does seem like the appropriate place for this configuration.

lemald: From [the release notes for Elixir 1.11](https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/), it looks like `config/releases.exs` works for us because it's specifically run at runtime in the context of a release, but doesn't do anything when (say) run locally via Mix. It looks like their recommended course of action is to just rename the file if you're already using `releases.exs` (though keeping it as-is sounds like it's still supported).

lemald: (Though thanks for saying something, that led me to actually look up what the difference was - it has confused me before.)

lib/skate_web/templates/unauthorized/restricted_environment.html.eex
lemald: If it works I'd say it's fine either way.

config/config.exs
KaylaBrady: Thanks for pointing to that doc! I made the switch to `runtime` in [7b47e69](https://github.com/mbta/skate/pull/2030/commits/7b47e69ca23d9eca5e7669a74097fae5ada2c7cd) since it made this easier to test locally. I'll make a separate tech debt ticket for moving the [custom runtime configuration ](https://github.com/mbta/skate/blob/7053aed97e9281294b9f17c275bac7d485d5d3a8/lib/skate/application.ex#L12) to use `config/runtime.exs` insead

lemald (approved):

lib/skate_web/router.ex
```diff
@@ -112,7 +131,8 @@ defmodule SkateWeb.Router do
       :accepts_html,
       :browser,
       :auth,
-      :ensure_auth
+      :ensure_auth,
+      :ensure_environment_access
```
firestack, line 135: thought(non-blocking): I'm thinking of opening a ticket for this(if there isn't one), but I think maybe we should consider adding trailing comma's to our formatter config? It'd be nice to take off the mental overhead of having to realize that the change that occurred to `:ensure_auth` was adding a `,`. (Since github isn't highlighting the only character that changed on line `134`)

Thoughts @mbta/skate-developers ?

firestack (approved):

lib/skate_web/router.ex
lemald: I like that idea in general, but I don't believe it's actually configurable - the Elixir code formatter explicitly lists "provide as little configuration as possible" as one of its [design principles](https://hexdocs.pm/elixir/main/Code.html#format_string!/2-design-principles) and I don't see an option for trailing commas.

PR #2036, opened 2023-04-21
fix(frontend-template): put meta `charset` tag in first `1024` bytes

firestack:
I noticed that the `meta` `charset` tag wasn't the first entry in the `head` tag. We use UTF-8 which is the default `charset` so this shouldn't have any impact on browsers parsing this page.

[`<meta charset>` tag MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#charset)

### DIFF
--- a/lib/skate_web/templates/layout/app.html.eex
+++ b/lib/skate_web/templates/layout/app.html.eex
@@ -5,9 +5,7 @@
     static_href = fn conn, path -> apply(static_href_module, static_href_fn, [conn, path]) end
   %>
   <head>
-    <%= if @google_tag_manager_id do %>
-      <%= render "_google_tag_manager.html", assigns %>
-    <% end %>
+
     <meta charset="utf-8"/>
     <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
     <meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, viewport-fit=cover">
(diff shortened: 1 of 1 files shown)


lib/skate_web/templates/layout/app.html.eex
```diff
@@ -5,9 +5,7 @@
     static_href = fn conn, path -> apply(static_href_module, static_href_fn, [conn, path]) end
   %>
   <head>
-    <%= if @google_tag_manager_id do %>
-      <%= render "_google_tag_manager.html", assigns %>
-    <% end %>
+
```
lemald, line 8: **question:** is the newline necessary? Getting rid of it would look slightly nicer IMO and save us an extra byte while we're at it.

firestack: Oops I hadn't realized I left that in, thanks!

firestack: [`1479175` (#2036)](https://github.com/mbta/skate/pull/2036/commits/1479175801405439e9b637a0c56773f8f0573908)

lemald:
For what it's worth, I was poking around in [the Google Tag Manager documentation](https://developers.google.com/tag-platform/tag-manager/web#standard_web_page_installation) and they do recommend putting the tag manager code as high up in the `<head>` as possible. I think the `charset` tag clearly takes precedence over that, but maybe we could put it immediately below those first few `<meta>` tags or something?

lemald (approved):
One comment on the exact positioning of the Google Tag Manager code but I don't think it matters that much in practice.

PR #2043, opened 2023-04-25
cleanup(ex/skate/database): remove users without email

KaylaBrady:
This restores the previous function for deleting users from the DB. It includes an additional change to delete users one by one rather than all at onceDeleting users doesn't need to be done as a single transaction, so opting to split it up just in case. There are ~300 records to delete, so I do not expect the volume of queries to cause an issue.


lemald (approved):
Looks good to me! I don't think we don't have to run this at the same time as we add the index, we can create the index after-hours and then pick this up the next day.

PR #2046, opened 2023-04-27
feat: limit auth failure retry to 3 attempts

lemald:
Asana ticket: [⚙️ Prevent infinite redirects on auth failure](https://app.asana.com/0/1152340551558956/1201523184137417/f)

I opted to use the Plug session to store a retry count in cookies and allow at most 3 tries. This will likely be enough to still prevent most of the user-facing errors while also not causing people to hit a redirect loop. We still reset the retry count on either a successful login or when the limit is hit, so the user can always hit refresh to try again manually.

### DIFF
--- a/lib/skate_web/auth_manager/error_handler.ex
+++ b/lib/skate_web/auth_manager/error_handler.ex
@@ -1,10 +1,22 @@
 defmodule SkateWeb.AuthManager.ErrorHandler do
+  import Plug.Conn
+
   @behaviour Guardian.Plug.ErrorHandler
 
   @impl Guardian.Plug.ErrorHandler
   def auth_error(conn, {_type, _reason}, _opts) do
-    Phoenix.Controller.redirect(conn,
-      to: SkateWeb.Router.Helpers.auth_path(conn, :request, "cognito")
-    )
+    auth_retries = get_session(conn, :auth_retries) || 3
+
+    if auth_retries > 0 do
+      conn
+      |> put_session(:auth_retries, auth_retries - 1)
+      |> Phoenix.Controller.redirect(
+        to: SkateWeb.Router.Helpers.auth_path(conn, :request, "cognito")
+      )
+    else
+      conn
+      |> delete_session(:auth_retries)
+      |> send_resp(:unauthorized, "unauthorized")
+    end
   end
 end
--- a/lib/skate_web/controllers/auth_controller.ex
+++ b/lib/skate_web/controllers/auth_controller.ex
@@ -2,6 +2,7 @@ defmodule SkateWeb.AuthController do
   use SkateWeb, :controller
   plug(Ueberauth)
 
+  import Plug.Conn
   alias Skate.Settings.User
   alias SkateWeb.AuthManager
   alias SkateWeb.Router.Helpers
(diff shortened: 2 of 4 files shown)


lib/skate_web/controllers/auth_controller.ex
```diff
@@ -28,10 +30,10 @@ defmodule SkateWeb.AuthController do
...
   def callback(%{assigns: %{ueberauth_failure: _fails}} = conn, _params) do
     # Users are sometimes seeing unexpected Ueberauth failures of unknown provenance.
-    # Instead of sending a 403 unauthenticated response, we are signing them out and
-    # sending them to the home page to start the auth path over again.
-    # We should be on the lookout for users getting trapped in a loop because of this.
-    # If we observe that happening we should rethink this remedy. -- MSS 2019-07-03
+    # Instead of sending a 403 unauthenticated response immediately, we are signing them out and
+    # sending them to the home page to start the auth path over again. -- MSS 2019-07-03
+    # We are maintaining the retry logic but limiting it to only a set number of retries using
+    # logic in SkateWeb.AuthManager.ErrorHandler. -- LEM 2023-04-27
```
KaylaBrady, line 36: question: Should this comment & the retry logic be on the same function? I'm a bit confused as to how this `callback` function relates to the `error_handler`

lemald: Okay, so this is part of what's so confusing about this logic. 😅 

`AuthController` handles the HTTP request that is made in the callback phase of the auth flow handled by Ueberauth - this is the point where Cognito redirects the user's browser back to Skate with some information about the login flow, ideally a signed assertion that the logic was successful with further metadata. The reason that the comment is here is that usually a failure would mean that we would return an unauthorized response, and the part that's weird is that we're redirecting back to the homepage instead.

`ErrorHandler` is part of our usage of Guardian. This is the functionality that is invoked when a user attempts to access a page requiring authentication (i.e. most of Skate) but there is no current valid Guardian ticket. Since we use Ueberauth to handle the authentication step the usual way to handle this state is to redirect to the request phase of the auth flow. My initial impression had been that the retry logic would actually be easier to implement cleanly in `ErrorHandler` since that's the start of the whole auth flow rather than an intermediate step but now that I think about it it would probably be doable as well. I guess it's a tradeoff in terms of what we think is most readable.

lib/skate_web/auth_manager/error_handler.ex
```diff
@@ -1,10 +1,22 @@
...
+      conn
+      |> put_session(:auth_retries, auth_retries - 1)
+      |> Phoenix.Controller.redirect(
+        to: SkateWeb.Router.Helpers.auth_path(conn, :request, "cognito")
+      )
+    else
+      conn
+      |> delete_session(:auth_retries)
+      |> send_resp(:unauthorized, "unauthorized")
+    end
```
firestack, line 20: question: I don't see anything in splunk but I might not be searching for the right term, are there any logs for this so we can track this happening?

lemald: If you search for `path=/auth/cognito/callback error` you can find instances where the `AuthController` received an error from Cognito and therefore triggered a redirect back to the start of the flow, is that the sort of thing you're looking for?

If we want to find instances of us reaching the max retries and giving the unauthorized response you could probably search for `controller=SkateWeb.PageController status=401` or something like that.

KaylaBrady (approved):

PR #2066, opened 2023-05-31
feat: Pull VPC into map search drawer

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204524735598198/f

This includes a mix of minor styling tweaks and minimal component restructuring to display the VPC in the search drawer rather than on top of the map. I'm still mulling over some possible approaches to reduce redundant data fetching between the map and search drawer, but not yet sure whether it is worth doing so at this stage or once the RPC is also in the search drawer. Will start a thread on that for further discussion!

### DIFF
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -96,20 +96,36 @@ $map-page-search-panel-width: 24rem;
 $map-page-search-stub-size: 10px;
 
 .c-map-page__search-actions {
-  margin: 1em;
+  margin: 0 1em 1em 1em;
   display: flex;
-  justify-content: center;
   gap: 1em;
 }
 .c-map-page__back-button {
-  // TODO: where should this live
-  background-color: #6c757d;
-  padding: 0.5em;
+  @include button-icon(1rem);
+  display: flex;
+  align-items: center;
+  padding: 0.5rem;
+
+  span {
+    svg {
+      fill: none;
+      stroke-width: 2;
+    }
+  }
 }
 
 .c-map-page__new-search-button {
-  // TODO: where should this live
-  background-color: #165c96;
+  @include button-icon(1rem);
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  border-radius: 4px;
+
+  padding: 0.5rem;
+}
+
+.c-map-page__new-search-button:only-child {
+  justify-content: center;
   flex: 1;
 }
 
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -20,7 +20,7 @@ $map-page-search-panel-width: 24rem;
 }
 
 .c-map-page .c-vehicle-properties-card {
-  margin: 1em;
+  margin: 1rem;
 }
 
 .c-map-page__input-and-results {
@@ -96,9 +96,9 @@ $map-page-search-panel-width: 24rem;
 $map-page-search-stub-size: 10px;
 
 .c-map-page__search-actions {
-  margin: 0 1em 1em 1em;
+  margin: 0 1rem 1rem 1rem;
   display: flex;
-  gap: 1em;
+  gap: 1rem;
 }
 .c-map-page__back-button {
   @include button-secondary;
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -101,6 +101,7 @@ $map-page-search-stub-size: 10px;
   gap: 1em;
 }
 .c-map-page__back-button {
+  @include button-secondary;
   @include button-icon(1rem);
   display: flex;
   align-items: center;
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -19,6 +19,10 @@ $map-page-search-panel-width: 24rem;
   padding: 0 1.5625rem;
 }
 
+.c-map-page .c-vehicle-properties-card {
+  margin: 1em;
+}
+
 .c-map-page__input-and-results {
   background-color: $color-gray-300;
   display: flex;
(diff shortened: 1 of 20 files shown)


assets/css/_ui_kit.scss
```diff
@@ -255,12 +255,27 @@ $color-tooltip-background: #333333;
 }
 
 @mixin button-secondary {
```
KaylaBrady, line 257: This mixin was otherwise unused

assets/css/_vehicle_properties_card.scss
```diff
@@ -103,5 +102,6 @@
 .c-vehicle-work-info th {
   padding-right: 36px;
   text-align: start;
+  vertical-align: top;
```
KaylaBrady, line 105: I'm wondering if it would be worth updating `vehicle-work-info` to use the `PropertiesList` component for consistency with VPP, or if these styles should continue to further diverge as they are here

assets/src/components/mapPage.tsx
```diff
@@ -76,6 +74,67 @@ const SearchInputAndResults = ({
   )
 }
 
+const SelectedVehicle = ({ vehicleId }: { vehicleId: VehicleId }) => {
+  // TODO: When using socket from context, this doesn't work as-is
+  // Presumably because the useMostRecentVehicleById hook is being used twice, but
+  // haven't tracked down error yet
+  const { socket } = useSocket()
```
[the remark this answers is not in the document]
lemald: Our tech cleanup [Asana task about centralizing / managing front-end data flow](https://app.asana.com/0/1152340551558956/1203857155511245/f) is probably also relevant.

assets/css/_vehicle_properties_card.scss
firestack: Personally, I don't like the interface that `PropertiesList` provides, but I think making them use the same component is a good idea.

Maybe there's a way to keep the existing interface for `PropertiesList` and also integrate the interface that `TrNameValue` provides?

priv/repo/migrations/20230419134906_notification_users_index.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Skate.Repo.Migrations.Notification_users_index do
```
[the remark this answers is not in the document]
KaylaBrady: 🙀 thanks for noticing, will deploy a rollback of this

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
KaylaBrady: Good point! I'll plan on adding a comment to that existing ticket about this case and likely not make further changes at this point

assets/css/_vehicle_properties_card.scss
KaylaBrady: It seems like the child component`PropertyRow` of `PropertyList` could be updated to use `TrNameValue` with only a small modification to add a `className` to `TrNameValue`. I'm going to make a separate ticket to do so since I don't want to increase the scope of this ticket to include possible styling regressions for existing components using the `PropertiesList` component.

assets/css/_map_page.scss
```diff
@@ -19,6 +19,10 @@ $map-page-search-panel-width: 24rem;
   padding: 0 1.5625rem;
 }
 
+.c-map-page .c-vehicle-properties-card {
+  margin: 1em;
```
firestack, line 23: nit/suggestion(non-blocking): because it's a layout property, I think this should be done with `px` instead.

Open to reasons why not to though!

KaylaBrady: Can you remind me why layout properties should be done with `px` instead? I'm trying to remember from past conversations and found [this thread](https://mbta.slack.com/archives/C047ZT8ACP5/p1674152194007119). The linked comment includes the following, but I may be missing some important context or later discussion/resources

> 4. Yes. rem for paddings and margins. Unlike borders and box shadows, these DO affect the readability, and should grow with the font size.

firestack: Thanks for digging that up, I guess I misremembered about which unit!
Based on that, are you alright with setting it to `1rem`?
```suggestion
  margin: 1rem;
```

lemald: I think we have used px for some sizing / layout of map controls in the past, but the reason for that was specifically that px is what Leaflet used so it was useful to use common units for things that had to align with built-in Leaflet components.

assets/css/_vehicle_properties_card.scss
KaylaBrady: Added a ticket: https://app.asana.com/0/1148853526253420/1204762479194649/f

assets/src/components/mapPage.tsx
```diff
@@ -76,6 +74,67 @@ const SearchInputAndResults = ({
...
+        >
+          <SearchIcon />
+          New Search
+        </button>
+      </div>
+      <hr />
+      {selectedEntity.type === SelectedEntityType.Vehicle ? (
+        <SelectedVehicle vehicleId={selectedEntity.vehicleId} />
+      ) : (
+        <p>TODO: Route selection</p>
```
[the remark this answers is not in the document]
KaylaBrady: That is correct! From discussion with product and design during [review](https://app.asana.com/0/1200180014510248/1204729792351246/f), the plan is to not merge this PR until the RPC changes are also in place. Would it be preferable to create a new branch to merge this into rather than master that we can merge all the related changes into?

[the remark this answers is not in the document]
lemald: Yeah, I guess my concern is that, as discussed on Slack yesterday, there's actually a fair bit of interdependence between this work and the RPC work, so we probably need to either accept that one will go live slightly before the other or create a shared feature branch to merge these changes into.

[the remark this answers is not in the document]
KaylaBrady: 👍 I'd be down to try the shared feature branch approach so that it is clearer what work is considered "done" and ready to start building on. Will add one now!

firestack (approved):
One nitpick but otherwise LGTM 👍🏻

PR #2068, opened 2023-06-01
feat(SearchMap): Primary styling for search drawer tab

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204505872151578/f

### DIFF
--- a/assets/css/_drawer_tab.scss
+++ b/assets/css/_drawer_tab.scss
@@ -18,5 +18,5 @@
   font-size: 2rem;
   line-height: 1.5rem;
   margin: 0;
-  padding: 2rem 0;
+  padding: 2rem 0 !important;
 }
--- a/assets/css/_drawer_tab.scss
+++ b/assets/css/_drawer_tab.scss
@@ -1,7 +1,9 @@
 .c-drawer-tab {
   background-color: $color-bg-medium;
-  border-top-right-radius: 1rem;
-  border-bottom-right-radius: 1rem;
+  border-top-right-radius: 1rem !important;
+  border-bottom-right-radius: 1rem !important;
+  border-bottom-left-radius: 0rem !important;
+  border-top-left-radius: 0rem !important;
   position: absolute;
   top: 2.5rem;
   width: $drawer-tab-width;
--- a/assets/css/_ui_kit.scss
+++ b/assets/css/_ui_kit.scss
@@ -254,6 +254,26 @@ $color-tooltip-background: #333333;
   }
 }
 
+@mixin button-primary-v2 {
+  color: $white;
+  background-color: $color-eggplant-700;
+  border-radius: 4px;
+
+  &:hover {
+    background-color: $color-eggplant-500;
+  }
+
+  &:focus {
+    background-color: $color-eggplant-500;
+    outline: 2px solid $color-eggplant-700;
+    outline-offset: -2px;
+  }
+
+  &:disabled {
+    background-color: $color-gray-500;
+  }
+}
+
 @mixin button-secondary {
   @include button-base;
   background-color: $color-button-secondary;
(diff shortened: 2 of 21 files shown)


assets/css/_ui_kit.scss
```diff
@@ -254,6 +254,26 @@ $color-tooltip-background: #333333;
   }
 }
 
+@mixin button-primary-v2 {
```
KaylaBrady, line 257: There already is an existing button-primary button with different styles. I was inclined to introduce this as a separate mixin for now, but could alternatively update the existing primary buttons if desired.

assets/css/_ui_kit.scss
```diff
@@ -254,6 +254,26 @@ $color-tooltip-background: #333333;
...
+  border-radius: 4px;
+
+  &:hover {
+    background-color: $color-eggplant-500;
+  }
+
+  &:focus {
+    background-color: $color-eggplant-500;
+    outline: 2px solid $color-eggplant-700;
+    outline-offset: -2px;
```
KaylaBrady, line 269: This didn't affect the button size in the way that adding a border did (even with `box-sizing: border-box` - maybe something small I'm missing?), but open to other suggestions if there is a better approach!

This doesn't respect the border-radius for [firefox versions < 88](https://developer.mozilla.org/en-US/docs/Web/CSS/outline). Our docs say we support versions 78+, but in FS I see that the lowest version used is 112. If outline does seem like the right approach for this, wondering if it would be reasonable to update our docs to support version 88+

firestack: I believe `outline` does not affect layout so it makes sense that it does not do what border did to the sizing https://developer.mozilla.org/en-US/docs/Web/CSS/outline#description

firestack: If design didn't explicitly ask for the offset to be within the button, then I think it'd be better to leave it as default and remove the offset, but otherwise I think this is good!

assets/css/_ui_kit.scss
firestack: I think we should try and find a way to make this a single style if possible, or at least have a ticket for it?

KaylaBrady: Just checked and the old style is only actually used in one place, confirmed with design that we can make the switch to the new styles for that location. Udpate incoming!

KaylaBrady: Oops I missed one existing reference on the shuttle page, checking with design about that, but otherwise updated this new mixin to be `button-primary` and marked the old one as `button-primary-legacy`

KaylaBrady: Confirmed with design to leave the shuttle route picker using the `button-primary-legacy` styles for now

assets/css/_drawer_tab.scss
```diff
@@ -18,5 +18,5 @@
   font-size: 2rem;
   line-height: 1.5rem;
   margin: 0;
-  padding: 2rem 0;
+  padding: 2rem 0 !important;
```
firestack, line 21: note(non-blocking): while on my-charlie, we ended up using a lot of bootstrap/tailwind-like utilities, and bootstrap already has `px-2` and `py-0` ([padding x-axis value 2, and padding y axis value 0](https://getbootstrap.com/docs/5.3/utilities/spacing/))

So ideally we could standardize with design on what kind of value they want instead of hard coding it per component.

I'm on the fence as to whether this would be good for skate, but I do like the value bootstrap utilities bring to "DRYing" out CSS and removing repetition (like `!important`). I also don't know that we've _decided_ on using bootstrap as a design library, so I only bring this up as a learning and not anything we need to act on right now.

(if you wanted to try this out though, you'd need to import these files to `app.scss`)
```
@import "../node_modules/bootstrap/scss/utilities";
@import "../node_modules/bootstrap/scss/utilities/api";
```

PR #2072, opened 2023-06-05
feat(mix): add mix aliases for setup and reset

firestack:
I'd gotten used to the aliases that come in Phoenix 1.7 and I want them back! 😭

These are a few quality of life improvements that make setting up a new project folder or getting up to speed a bit faster and require less manual interaction.

I basically just copied the aliases from the my-charlie project that were generated by the phoenix application template and updated the readme a bit.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -99,6 +99,7 @@ defmodule Skate.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       "assets.setup": ["cmd npm --prefix=assets install"],
+      "assets.reset": ["cmd npm --prefix=assets ci"],
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]
     ]
   end
--- a/mix.exs
+++ b/mix.exs
@@ -9,6 +9,7 @@ defmodule Skate.MixProject do
       elixirc_paths: elixirc_paths(Mix.env()),
       compilers: [:phoenix] ++ Mix.compilers(),
       start_permanent: Mix.env() == :prod,
+      aliases: aliases(),
       deps: deps(),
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -85,4 +86,20 @@ defmodule Skate.MixProject do
...
+  #
+  #     $ mix setup
+  #
+  # See the documentation for `Mix` for more info on aliases.
+  defp aliases do
+    [
+      setup: ["deps.get", "ecto.setup", "assets.setup"],
+      "ecto.setup": ["ecto.create", "ecto.migrate"],
+      "ecto.reset": ["ecto.drop", "ecto.setup"],
+      "assets.setup": ["cmd npm --prefix=assets install"],
```
lemald, line 101: **suggestion (non-blocking):** I don't have a strong feeling on it, but it may be better to use `npm ci` since I think that's more likely to produce an end result that exactly matches your `package-lock.json` without changing anything.

lemald (approved):

mix.exs
lemald: Interesting, I just timed it on my old 2018 MacBook and `npm ci` for Skate took only 17 seconds. I wonder what's different - if you have an M1 maybe there are binaries that need to be compiled that come prepackaged for x86 or something like that?

In any event, either way I like the `assets.reset` idea!

lemald: > Just clarifying, with `assets.reset`, are you good with leaving `assets.setup` as is?

Yeah that sounds good to me. 👍

lemald: Totally random, but it was only yesterday that I learned that `npm ci` stood for "clean install" as opposed to "continuous integration," although a continuous integration environment is one of the contexts where you might want to use the clean install command.

lemald (approved):

PR #2075, opened 2023-06-07
fix(SearchPage): Resolve mobile z-index issues

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1204649005443955/f

This PR sets the z-index of the map and search drawer within the search page. This creates a new stacking context for each, so that they can be z-indexed relative to each other as a whole unit. In trying to better understand stacking contexts for this ticket, I found the folder and file analogy in[ this article ](https://www.lullabot.com/articles/understanding-debugging-stacking-contexts)very helpful! 

This PR makes the same change to the new map page too, which didn't have the same bug but was prone to the same issue of search drawer & map elements being within a single stacking context.

### DIFF
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -1,5 +1,10 @@
 $map-page-search-panel-width: 24rem;
 
+$z-map-page-context: (
+  "map": 100,
+  "search-drawer": 200,
+);
+
 .c-map-page {
   display: flex;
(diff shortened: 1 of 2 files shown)


firestack:
Hmm, I'm getting some strange interactions with the banner? is that within this scope to address? 
image

firestack (approved):

PR #2079, opened 2023-06-09
fix(Concentrate.VehiclePosition.merge/2): Prioritize swiftly block_id when there isn't a block overload

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1204382924233721/f

### DIFF
--- a/test/concentrate/vehicle_position_test.exs
+++ b/test/concentrate/vehicle_position_test.exs
@@ -374,6 +374,78 @@ defmodule Concentrate.VehiclePositionTest do
       assert Mergeable.merge(first, second) == expected
     end
 
+    test "merge/2 takes the latest non-nil value for crowding" do
+      crowding_1 = %{
+        load: 98,
+        capacity: 100,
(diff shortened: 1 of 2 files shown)


test/concentrate/vehicle_position_test.exs
```diff
@@ -374,6 +374,78 @@ defmodule Concentrate.VehiclePositionTest do
...
+      third_no_crowding =
+        new(
+          last_updated: 3,
+          latitude: 2,
+          longitude: 2,
+          crowding: nil
+        )
+
+      assert %{crowding: ^crowding_2} = Mergeable.merge(first, second)
+      assert %{crowding: ^crowding_2} = Mergeable.merge(second, third_no_crowding)
```
lemald, line 417: **suggestion:** for this and the other test, I think it might make sense to also test that you get the same result with the order of the arguments swapped, since our implementation of merge is supposed to base its priority on `last_updated` rather than just order the values are passed in.

lemald (approved):

PR #2080, opened 2023-06-09
feat(ts/components/StopMarker): Update bus stops for discoverability

firestack:
I ended up implementing this using a react abstraction around `divIcon` so that the SVG could be updated in place by react without a round trip through leaflet, as well as using bootstrap utility api's to generate styles for states.

The `ReactMarker` should be reusable for any interactive markers or JSX derived markers we want to do in the future, as well as our current `divIcon`s _could_ be updated in the future, to avoid generating HTML from a string at runtime.

(edit: flattened the CSS down into a single file https://github.com/mbta/skate/pull/2080/commits/93a3980e69fdb67e056b12f4dc844bd5c7fdbe67)

One other question I had was around the `useReactDivIcon` hook. It currently updates the `divIcon` whenever the variable passed in changes, but this is not how hooks like `useState` work, where they explicitly provide a `setState` action. I was unsure if I should additionally return a `setDivIconOptions` function or some way to affect the state without worrying about the parameters, but I'll leave that up to review.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1204553794065818/f


assets/src/components/map/markers/stopMarker.tsx
```diff
@@ -0,0 +1,305 @@
...
+          <circle
+            className={joinClasses([
+              ...circleClasses,
+              "c-stop-icon-hover_stroke-eggplant-700",
+            ])}
+            cx="7"
+            cy="7"
+            r="5"
+            fill="white"
+            stroke="#3C3F4C"
```
lemald, line 167: **suggestion:** it looks like this is actually the `$color-gray-700` from our design system, could you use a utility class instead of setting the value manually?

firestack: I left it because it was part of the figma export, but I'll add the classes(to all SVG's that match this same issue) 👍🏻.

If we have the classes, should we still have the fill and stroke as fallbacks because they were part of the export, or do you still feel it should be dropped?

assets/src/components/map/markers/stopMarker.tsx
```diff
@@ -0,0 +1,305 @@
...
+  StopIconSizeFromStopType(StopTypeFromZoomLevel(zoomLevel))
+
+/**
+ * Represents the box size of the icon at each `StopType`.
+ *
+ * For a `<circle>`, this means the size needs to include the `radius` AND
+ * `stroke-width`, or specify `overflow-visible`. This will directly impact
+ * the tap target on the leaflet map.
+ */
+export const StopIconSizeFromStopType = (stopType: StopType): PointTuple => {
```
lemald, line 50: **nitpick:** since this isn't a component, shouldn't it be `StopIconSizeFromStopType`?
```suggestion
export const stopIconSizeFromStopType = (stopType: StopType): PointTuple => {
```

assets/src/components/map/markers/stopMarker.tsx
firestack: Oh turns out they already have the classes specified. So, remove the attributes?

assets/src/components/map/markers/stopMarker.tsx
firestack: 👍🏻 19d50226388f15afff0111d6126c9fa07c194102

assets/src/components/map/markers/stopMarker.tsx
firestack: Turns out that we could also specify `clip-rule` and `fill-rule` via CSS

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule
> Note: As a presentation attribute, fill-rule can be used as a CSS property.

Thoughts on generating the CSS utilitiy classes and also specifing these properties via CSS?

lemald: > Turns out that we could also specify `clip-rule` and `fill-rule` via CSS
> 
> https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule
> 
> > Note: As a presentation attribute, fill-rule can be used as a CSS property.
> 
> Thoughts on generating the CSS utilitiy classes and also specifing these properties via CSS?

I can't immediately think of any downsides - if you're okay with it then so am I.

assets/tests/components/map/utilities/reactDivIcon.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  test("when given custom options, should return divIcon and icon element", () => {
+    const { result } = renderHookReactDivIcon({
+      className: "testClass",
+      iconSize: [20, 20],
+    })
+
+    expect(result.current.divIcon).toBeInstanceOf(DivIcon)
+    expect(result.current.iconContainer).toBeEmptyDOMElement()
+  })
```
lemald, line 33: **question:** is it possible to somehow assert that these options have their desired effect? Or do you think that would get too into the weeds of testing Leaflet's `divIcon` itself?

lemald:
Okay, this basically looks good but there's a lot going on here with the React / Leaflet interface logic. I think I like the basic ideas but I want to think it over a bit before hitting approve, if that's all right.

assets/tests/components/map/utilities/reactDivIcon.test.tsx
firestack: I had to think on this for a while, I didn't think that `createIcon`[^1] was in our `@types/leaflet` file but upon checking I did find it there. The actual icon isn't created until a `Marker` calls `createIcon` so we're having to simulate this being consumed by something to assert that it's actually taking on new options.
See 4d8c219bdc99449e11de9a7dc7133fead1e4f552.

Because `Marker` also passes in the previous `div` when updating, I've also tried to simulate that in c6f2aea995c4b4c9d394730aa829f65d971340ea, but maybe it's too specific to leaflet internals and I should drop that or move it into it's own test?

[^1]: `createIcon` is in their public documentation, so it seems to technically be in the public api?

lemald: Hmm, okay... I'm not sure I'm following 100% since this logic is all rather complicated but I think this is good enough as far as testing goes. 👍

KaylaBrady (approved):
Looks good to me, a few non-blocking comments!

assets/src/components/map/markers/stopMarker.tsx
```diff
@@ -0,0 +1,305 @@
...
+import { ReactMarker } from "../utilities/reactMarker"
+
+/**
+ * Specific variants that the `StopIcon` can render
+ */
+export enum StopType {
+  Small,
+  Medium,
+  Large,
+}
```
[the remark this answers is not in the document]
firestack: I feel that the `StopIcon` is a (independent) component which has a type paremter, although for lack of a better names I did give the values names based on sizes, I don't feel that this type directly relates to size. I'd be open to changing to `StopSize` but how do you feel about doing `StopIconType` instead?

assets/src/components/map/markers/stopMarker.tsx
```diff
@@ -0,0 +1,304 @@
...
+            fill="#3c3f4c"
+          />
+        </svg>
+      )}
+    </svg>
+  )
+}
+
+interface StopMarkerProps extends Partial<MarkerProps> {
```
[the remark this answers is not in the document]
firestack: Oh actually I meant to remove this[^1] as it actually hides that `posistion` is required, thanks for commenting on this!

Responding to the quesiton with my original intent though: I was thinking that there were some marker props that I was going to fill in that the user of this API didn't need to be concerned about and if there were any requirements of the react-leaflet Marker that were required, the user shouldn't get a type error on a prop that's already handled. Given that `posistion` is not handled by this component, `MarkerProps` should not be `Parital`

[^1]: I actually thought I had already removed this, but I must not have committed it or something.

assets/src/components/map/markers/stopMarker.tsx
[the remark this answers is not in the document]
KaylaBrady: I like `StopIconType`!

assets/src/components/map/markers/stopMarker.tsx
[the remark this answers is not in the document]
firestack: Oh actually I do provide `posistion`, it's `Omit<_, "posistion">` that I want, because the user cannot specify that property, but can specify basically everything else.

assets/src/components/map/markers/stopMarker.tsx
firestack: a1b4d00c aceb53c6

assets/src/components/map/markers/stopMarker.tsx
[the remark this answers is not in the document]
firestack: 8e7da8ec

assets/src/components/map/markers/stopMarker.tsx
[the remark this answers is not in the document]
firestack: 8df6351a31e5332ff3ad2b1c30d6f0d1f05fb9e9

assets/src/components/map/markers/stopMarker.tsx
```diff
@@ -0,0 +1,305 @@
...
+    return StopType.Small
+  } else if (15 <= zoomLevel && zoomLevel <= 16) {
+    return StopType.Medium
+  } else if (17 <= zoomLevel) {
+    return StopType.Large
+  } else {
+    throw new Error(
+      "internal error: entered unreachable code: invalid zoom level"
+    )
+  }
```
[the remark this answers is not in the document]
firestack: The issue I encountered here was that typescript cannot seem to tell that the entire range is covered by the `if, else if` statements above the `else` so the `else` block should never be reached. So while `17 <= zoomLevel` does cover level's `17` and `18`, I didn't want to decide what a "default" was and what to put in the `else` block because the `else` block should technically never be reached unless there's an improper type passed in or something.

assets/src/components/map/utilities/reactDivIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  useEffect(() => {
+    let element = iconContainer
+    if (element === null) {
+      element = createPortalElement()
+      setIconContainer(element)
+    }
+
+    // Leaflet by default doesn't support updating a `divIcon` in place.
+    // To ensure that the `divIcon` updates when `opts` change
+    // regenerate the `divIcon` with the portal element and provided `opts`
```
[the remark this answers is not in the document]
firestack: I wasn't quite sure of what to write for the doc comment, and I opted to leave the comment inline, let me know if you had something else in mind or want any changes 👍🏻
94b0f4b0613baa1cbdd9a85e6a2df8553e6de00a

assets/src/components/map/utilities/reactMarker.tsx
```diff
@@ -0,0 +1,61 @@
...
+  loadingState?: ReactNode
+  /**
+   * Options to pass to `Leaflet.DivIcon`
+   *
+   * Defaults to `{}`
+   */
+  divIconSettings?: Partial<DivIconOptions>
+}
+
+export const ReactMarker = ({
```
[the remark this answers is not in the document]
KaylaBrady: Looks good!

lemald (approved):

lemald (approved):
👍

PR #2087, opened 2023-06-13
feat: Expose logged out vehicle data in searches for users in test group

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204771228618690/f


lib/realtime/vehicles.ex
```diff
@@ -58,7 +58,6 @@ defmodule Realtime.Vehicles do
     incoming_from_another_route = incoming_from_another_route(incoming_trips, vehicles_and_ghosts)
 
     vehicles_and_ghosts
-    |> Enum.filter(fn vehicle_or_ghost -> vehicle_or_ghost.route_id != nil end)
```
KaylaBrady, line 61: This where logged out vehicles were originally being filtered out. Keeping them in results in a key `nil` being added to the result of `group_by_route_with_blocks`. Something about a nil key of a map is funky to me, but it seemed like the simplest way to keep these logged out vehicles for use elsewhere

lib/realtime/server.ex
```diff
@@ -379,14 +384,20 @@ defmodule Realtime.Server do
       _ = :ets.insert(ets, {{:route_id, route_id}, active_vehicles_and_ghosts})
     end
 
-    all_vehicles = Enum.uniq(all_vehicles(vehicles_by_route_id) ++ shuttles)
+    {logged_out_vehicles, logged_in_vehicles} =
+      vehicles_by_route_id
+      |> all_vehicles()
+      |> Enum.concat(shuttles)
+      |> Enum.uniq()
```
lemald, line 391: **question:** I know your code wasn't the first to introduce this, but I wonder if it would be more efficient to do `uniq_by/2` here and compare the vehicles by IDs, rather than the whole maps.

More broadly, I want to better understand why the `uniq` step is necessary - my understanding is that that's because `all_vehicles` takes a map of routes to vehicles and combines them, and due to incoming vehicles a vehicle may be associated with multiple routes, but let me know if I'm wrong!

KaylaBrady: I think your'e right that this is due to incoming vehicles which may be associated with multiple routes. If the vehicle structs for incoming vehicles are exactly equal to the regular vehicle than I agree `uniq_by` seems appropriate here! Everything I've seen so far in code suggests that they would be exactly equal. And if that were not the case, then searching for a vehicle that is in the incoming box would return two search results, which doesn't seem to be the case. 

There is [one test ](https://github.com/mbta/skate/blob/7d882b404c493ce651138451e7b7fdafedce2964/test/realtime/server_test.exs#L279)that looks for these values to be deduplicated that is testing with exactly duplicated shuttles. I haven't seen anything to point to there being a reason for duplicated shuttle vehicles with the same ID but different other properties.

With all that in mind, I can update to use `uniq_by` instead!

lib/realtime/server.ex
```diff
@@ -379,14 +384,20 @@ defmodule Realtime.Server do
       _ = :ets.insert(ets, {{:route_id, route_id}, active_vehicles_and_ghosts})
     end
 
-    all_vehicles = Enum.uniq(all_vehicles(vehicles_by_route_id) ++ shuttles)
+    {logged_out_vehicles, logged_in_vehicles} =
```
KaylaBrady, line 387: I'm considering moving this partitioning up a level to `handle_cast({:update_vehicles...`, to ensure that the `active...` and `new_active...` values don't include any data for logged out vehicles. They shouldn't because `route_id`, `run_id`, and `block_id` should all be null for logged out vehicles, but I think it may be helpful to it more explicit.

lemald: Yeah, I like that idea! I wonder if you could go a bit further and move it to `Concentrate.Consumer.VehiclePositions.handle_events` so that the logged out vehicles are passed in as a separate value to `Server.update_vehicles`, thereby also getting rid of the thing where logged out vehicles get included in the map of vehicles by route but under the `nil` key.

lib/realtime/server.ex
```diff
@@ -345,15 +344,13 @@ defmodule Realtime.Server do
     {:noreply, %{state | active_alert_route_ids: new_active_route_ids}}
   end
 
-  defp update_vehicle_positions(
+  defp remove_inactive_keys(
```
lemald, line 347: **praise:** I like having this separated out into a separate function from the rest of the update logic.

lib/realtime/server.ex
KaylaBrady: Oh nice yeah that would be even cleaner! I'll try that

PR #2096, opened 2023-06-16
feat(ts/components/mapPage): open RPC when VPC Route is clicked, remove route shape interaction

firestack:
# Summary
This drills the event prop from the `VehicleRouteSummary` component, up to the `VehiclePropertiesCard`, and then when the `VehiclePropertiesCard` is embedded in the `MapPage` via `MapDisplay`, adds the callback to set the selection when the event is fired.

The click route shape to set the selection to the route has also been removed.

# Reviewer Notes
As there are a few fixes made that weren't directly related to the ticket, this should be reviewable by commit to be able to ignore the fixes at the beginning of the history.

---
  - https://app.asana.com/0/0/1204680913839486


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -212,7 +212,7 @@ const RoutePatternLayers = ({
 }: {
   routePattern: RoutePattern
   isSelected: boolean
-  onShapeClick: () => void
+  onShapeClick?: () => void
```
lemald, line 215: **suggestion:** with these changes, am I correct that `onShapeClick` will no longer be used for anything? If so, it may make the code a bit cleaner if it's removed, along with downstream places it's passed, mainly `onClick` for `<RouteShape/>` assuming that's not used for anything else.

firestack: a7a23b828b0f8cf860cc8b5075cd9e05d22bd8d9

assets/src/components/mapPage.tsx
```diff
@@ -80,21 +80,42 @@ const SearchMode = ({
...
   // Presumably because the useMostRecentVehicleById hook is being used twice, but
   // haven't tracked down error yet
   const { socket } = useSocket()
 
   const selectedVehicleOrGhost = useMostRecentVehicleById(socket, vehicleId)
 
-  return selectedVehicleOrGhost ? (
+  if (selectedVehicleOrGhost === null) {
+    return <Loading />
+  }
```
KaylaBrady, line 99: This early return is nice!

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -208,42 +208,22 @@ const MapElementsNoSelection = ({
...
-                        className="route-shape__tooltip"
-                        sticky={true}
-                        direction="top"
-                      >
-                        Click to select route {routePattern.routeId}
-                        {variantFormatted}.
-                      </Tooltip>
-                    )) ||
-                      undefined}
-                  </RouteShape>
```
KaylaBrady, line 246: question: Is there styling that should be removed since this shape is no longer clickable? There is a pointer cursor when hovering over the shape

firestack: I believe this comes from leaflet adding the `leaflet-interactive` class and I'm not sure how to get rid of it yet, but I can look into that.

firestack: Thanks, that seems to work!

What do you think of c19ba3552532523bf5536e6c1c1eb55729efd0e9 ?
I was unsure if we should simply set it to `false` always or make it possible for the interactive state to be set based on if `onClick` is bound.

firestack: Personally, I don't see a reason to remove the functionality from the component api (similar reason to why I didn't remove it originally before https://github.com/mbta/skate/pull/2096#discussion_r1232329359), but as a part of the larger systems of maps, I do see that this could essentially be tech debt if we don't use it. Removed in 3d9142de9a95f9ad700d42e8b2d336db5bc78226

lemald (approved):

assets/css/_vehicle_route_summary.scss
```diff
@@ -84,6 +84,26 @@
     background-color: $color-gray-300;
     width: 1px;
   }
+  // #endregion
+  // #endregion
```
lemald, line 88: **question (non-blocking): is this a mistake or are there nested regions? I'm not really familiar with this syntax to be honest.

lemald (approved):
👍

PR #2102, opened 2023-06-21
fix:test(ex/): replace operator info with factories

firestack:
I mostly opted for find and replace with this, with a few interventions to make tests a little more parametrized. 

Also, because module attributes are essentially pasted into the calling locations, I tried to avoid instantiating module attributes for field accesses.

I only did a quick check, but I didn't see anything for generating names in Elixir, if we want more accurate names in the future it should be possible by editing the factories for these fields.
Other libraries (like faker), seem to have a [long list of names](https://github.com/faker-js/faker/blob/651d1a8cd11b2d562c7a6473b8a544b81d5fbb95/src/locales/en/person/first_name.ts) that they pick from.

This took a little longer because I went back and made sure that all the `operator_name` fields did end up matching their corresponding `operator_last_name` fields.

---

Asana Ticket: https://app.asana.com/0/1152340551558956/1204825065206827/f


test/notifications/notification_server_test.exs
```diff
@@ -300,16 +302,18 @@ defmodule Notifications.NotificationServerTest do
 
     test "broadcasts, saves to the DB, and logs new notifications for waivers with recognized reason for vehicles on selected routes",
          %{user: user} do
+      vehicle = @vehicle
```
KaylaBrady, line 305: question (non-blocking): Why is this assignment needed?

test/realtime/vehicle_or_ghost_test.exs
```diff
@@ -62,38 +66,50 @@ defmodule Realtime.VehicleOrGhostTest do
 
   describe "find_by/2" do
     test "matches on any of run, vehicle, or operator" do
```
KaylaBrady, line 68: suggestion (non-blocking): The test was already written this way so no need to update, but maybe it would be helpful to split this into three tests - run, vehicle, and operator since there are a lot of assertions here

KaylaBrady (approved):
Looks good to me - I like the factory approach you took!

KaylaBrady (commented):
Oh @firestack one last thing - I see the ticket includes the `swiftly_realtime_vehicles.json` file - could you add updates to that file or as a separate PR?

test/notifications/notification_server_test.exs
firestack: From how I understand it, any invocation of a module attribute is expanded inline at each point. So to [avoid expanding this somewhat large map all over the place](https://github.com/mbta/skate/pull/2102/files/27e69efac8647e966a57d8cb02f02a874fb81fe6#diff-5ca46ffdee1a00ab30040cca268ee561ffc4aa61a4533dcf066372f54842b50cR315-R316) I've bound it to a local variable for field access. I also did this in hopes that we could convert this to use factories in the future.

```suggestion
      vehicle = build(:vehicle)
```

firestack: https://hexdocs.pm/elixir/Kernel.html#@/1-attention-multiple-references-of-the-same-attribute

lemald (approved):
👍

firestack:
> Oh @firestack one last thing - I see the ticket includes the `swiftly_realtime_vehicles.json` file - could you add updates to that file or as a separate PR?

Thank you @KaylaBrady!
Please see these two commits
- [`b287eda` (#2102)](https://github.com/mbta/skate/pull/2102/commits/b287eda166cb357dbd3cc4fda2e7c01379085189)
- [`ef4bbe9` (#2102)](https://github.com/mbta/skate/pull/2102/commits/ef4bbe9f4b0400f6031a42fec84d09b4c03cd935)

KaylaBrady (approved):

PR #2103, opened 2023-06-21
feat(MapPage): Back button navigates through selection history

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204762161648730/f

### DIFF
--- a/assets/src/state/searchPageState.ts
+++ b/assets/src/state/searchPageState.ts
@@ -34,12 +34,14 @@ export interface SearchPageState {
   isActive: boolean
   savedQueries: SavedSearchQuery[]
   selectedEntity?: SelectedEntity | null
+  selectedEntityHistory: SelectedEntity[]
 }
 
 export const initialSearchPageState = {
   query: emptySearchQuery,
   isActive: false,
   savedQueries: [],
+  selectedEntityHistory: [],
 }
 
 interface SetSearchTextAction {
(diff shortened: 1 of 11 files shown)


assets/src/state/searchPageState.ts
```diff
@@ -76,28 +78,37 @@ export const submitSearch = (): SubmitSearchAction => ({
...
+): NewSearchSessionAction => ({
+  type: "NEW_SEARCH_SESSION",
+  payload: newSelection || null,
 })
 
-interface SelectVehicleAction {
-  type: "SELECT_SEARCH_VEHICLE"
-  payload: { vehicleId: VehicleId } | null
-}
-
```
KaylaBrady, line 90: The SELECT_SEARCH_VEHICLE action was a specific case of SELECT_SEARCH_ENTITY, which resulted in some duplication. I opted to keep the `setSelectedVehicle` helper function but have it map to the "SELECT_SEARCH_ENTITY" instead. Though now I'm realizing `setSelectedVehicle` is unused, so unless there is feedback to keep it I'll remove that entirely!

lemald: I'm fine with just getting rid of it!

assets/src/state/searchPageState.ts
```diff
@@ -76,28 +78,37 @@ export const submitSearch = (): SubmitSearchAction => ({
...
+  payload: SelectedEntity | null
 }
-export const clearSearch = (): ClearSearchAction => ({
-  type: "CLEAR_SEARCH",
+export const newSearchSession = (
+  newSelection?: SelectedEntity | null
+): NewSearchSessionAction => ({
+  type: "NEW_SEARCH_SESSION",
+  payload: newSelection || null,
 })
```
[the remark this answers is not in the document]
lemald: Hmm... I see what you mean about how it might be confusing, but unfortunately I can't think of anything else. Maybe `newSearchContext`? But that's also overloaded due to React contexts.

lemald (approved):
Looks good to me. 👍

assets/src/state/searchPageState.ts
```diff
@@ -149,27 +152,66 @@ export const reducer = (
...
+    }
   }
   return state
 }
+/*
+The last selection should only be added to the selection history if it is different from the new selection.
+For vehicles, this means having a different vehicleId. 
+For route patterns, this means having a different routeId.
+ */
+export const shouldAddLastSelectionToHistory = (
```
KaylaBrady, line 192: this feels a bit awkward to me

lemald (approved):
Latest changes look good. 👍

PR #2105, opened 2023-06-23
fix(Realtime.Vehicles): don't list interlining vehicles as incoming on their old route)

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1204758166006987/f

Since the only use of `active_trips` was within `Realtime.Vehicles`, I opted to remove it in favor of `trips_starting_in_range`. If you think it should be preserved, I'll add it back in!


lemald (approved):
Good catch!

PR #2108, opened 2023-06-27
feat(Application): test for connections before running migrator

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204865954744424/f

This adds logic to attempt concurrent queries to the DB prior to running the migrator to ensure that the connections have started. This is based on the discussion and recommendation in this issue: https://github.com/elixir-ecto/ecto/issues/3428

> Postgrex is asynchronous when it connects to the database. So when the repo starts, you may not have a connection yet. If your database is loaded with extensions, this bootstrap time can be high.
>
>My suggestion is to have a single process immediately after Ecto.Repo in the supervision tree that is responsible for checking for the repository availability:

Two test deployments to dev-green succeeded without DBConnection.ConnectionErrors or task thrashing in ECS:
image

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -108,7 +108,8 @@ config :skate, Skate.Repo,
 
 config :skate, Skate.WarmUp,
   minimum_percent_queries_to_succeed: 0.6,
-  max_attempts: 3
+  max_attempts: 3,
+  seconds_between_attempts: 3
 
 config :laboratory,
   features: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -106,6 +106,10 @@ config :skate, Skate.Repo,
   port: System.get_env("POSTGRES_PORT", "5432") |> String.to_integer(),
   show_sensitive_data_on_connection_error: true
 
+config :skate, Skate.WarmUp,
+  minimum_percent_queries_to_succeed: 0.6,
+  max_attempts: 3
+
 config :laboratory,
   features: [
     {:late_view, "Late View", "Grants access to experimental Late View"}
--- a/lib/skate/application.ex
+++ b/lib/skate/application.ex
@@ -13,7 +13,7 @@ defmodule Skate.Application do
 
     # List all child processes to be supervised
     children =
-      [{Skate.Repo, []}] ++
+      [{Skate.Repo, []}, Skate.WarmUp] ++
         if Application.get_env(:skate, :start_data_processes) do
           [
             Schedule.Supervisor,
@@ -29,11 +29,14 @@ defmodule Skate.Application do
           SkateWeb.Endpoint
         ]
 
-    link = Supervisor.start_link(children, strategy: :one_for_all, name: Skate.Supervisor)
+    case Supervisor.start_link(children, strategy: :one_for_all, name: Skate.Supervisor) do
+      {:ok, _pid} = link ->
+        Migrate.up()
+        link
 
-    Migrate.up()
-
-    link
+      error ->
+        error
+    end
   end
 
   # Tell Phoenix to update the endpoint configuration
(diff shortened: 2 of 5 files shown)


lib/skate/application.ex
```diff
@@ -29,11 +29,14 @@ defmodule Skate.Application do
           SkateWeb.Endpoint
         ]
 
-    link = Supervisor.start_link(children, strategy: :one_for_all, name: Skate.Supervisor)
+    case Supervisor.start_link(children, strategy: :one_for_all, name: Skate.Supervisor) do
+      {:ok, _pid} = link ->
+        Migrate.up()
+        link
```
KaylaBrady, line 35: This is based on @firestack's observation that the migrator hides information in the case that `Supervisor.start_link` results in a failure. I didn't add Migrator as a child yet to limit the changes for now

lib/skate/warm_up.ex
```diff
@@ -0,0 +1,92 @@
...
+         %{
+           pool_size: pool_size,
+           required_success_count: required_success_count,
+           max_attempts: max_attempts,
+           attempt: attempt,
+           check_fn: check_fn
+         } = config
+       ) do
+    count_success =
+      Enum.map(0..(pool_size - 1), fn index ->
```
KaylaBrady, line 51: running the same number of queries as there are connections in the pool isn't really an exact science, but gives some level of confidence that multiple concurrent queries should be able to succeed quickly

config/config.exs
```diff
@@ -106,6 +106,10 @@ config :skate, Skate.Repo,
   port: System.get_env("POSTGRES_PORT", "5432") |> String.to_integer(),
   show_sensitive_data_on_connection_error: true
 
+config :skate, Skate.WarmUp,
+  minimum_percent_queries_to_succeed: 0.6,
```
KaylaBrady, line 110: maybe this should be 100%? After testing two deploys, 100% of the queries succeeded on the first attempt.

lib/skate/warm_up.ex
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    if result.status == :success do
+      result
+    else
+      if max_attempts == attempt do
+        result
+      else
+        Logger.warn(format_result_message(result))
+        check_connections_warmed_up(%{config | attempt: attempt + 1})
```
lemald, line 79: **question:** is it worth pausing between attempts, even if just for a few seconds?

lib/skate/warm_up.ex
```diff
@@ -0,0 +1,92 @@
...
+    case check_connections_warmed_up(%{
+           pool_size: pool_size,
+           required_success_count: pool_size * minimum_percent_queries_to_succeed,
+           max_attempts: max_attempts,
+           attempt: 1,
+           check_fn: warm_up_check_fn
+         }) do
+      %{status: :success} = result ->
+        Logger.info(format_result_message(result))
+        {:ok, state}
```
lemald, line 32: **suggestion:** I think you could return `:ignore` here to cause the process to exit without affecting the rest of the supervision tree or causing it to attempt to restart, which might be more appropriate for something transient like this that's only needed at startup.

lib/skate/warm_up.ex
KaylaBrady: Added a configurable pause in https://github.com/mbta/skate/pull/2108/commits/ed2a8d51705c4bec37b382ddf3f06b83eb498991

lib/skate/warm_up.ex
KaylaBrady: Updated in https://github.com/mbta/skate/pull/2108/commits/72a0579f62f29e20e5f3f4e9866f677dfc9b2c5f

lemald (approved):
👍

PR #2112, opened 2023-06-29
feat(tilesetUrls): Expose satellite tileset url

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204573236791609/f
This ticket adds new tileset environment variables to distinguish between the base & satellite URLs. 
I'll add a corresponding terraform PR as well if these variable names look good. 

For the latency consideration of this ticket, testing with throttling for chrome's "fast 3G" setting, satellite tiles were taking ~2 seconds for the first load. Subsequent loads were faster due to browser caching. For the base tileset at the same network speed & location, tile requests took 1-1.5s. 

The ArcGIS tiles have permissive [terms of use](https://massgis.maps.arcgis.com/home/item.html?id=bd1ea555a71649508cb215edcc73feaa):

> No restrictions apply to these data

Order of operations:
* Merge terraform PR configuring these variables
* Merge this PR
* Start using `tilesetUrlForType`
* Remove references to the old tilesetUrl configuration

### DIFF
--- a/.envrc.template
+++ b/.envrc.template
@@ -40,6 +40,12 @@ export ERL_FLAGS="+MIscs 2048"
 ## Location of map tile images
 # export TILESET_URL=
 
+## URL for map tile images, including {x} {y} coordinate and {z} zoom level placeholders
+# export BASE_TILESET_URL=
+
+## URL for satellite map tile images, including {x} {y} coordinate and {z} zoom level placeholders
+# export SATELLITE_TILESET_URL=
+
 ## Used by Erlang (only required in production)
 # export RELEASE_COOKIE=
 
--- a/assets/src/components/map.tsx
+++ b/assets/src/components/map.tsx
@@ -131,6 +131,7 @@ export const FullscreenControl = createControlComponent(
   Leaflet.control.fullscreen
 )
 
+// TODO: use tilesetForType instead
 const tilesetUrl = (): string => appData()?.tilesetUrl || ""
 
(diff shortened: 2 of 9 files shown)


assets/src/components/map.tsx
```diff
@@ -131,6 +131,7 @@ export const FullscreenControl = createControlComponent(
   Leaflet.control.fullscreen
 )
 
+// TODO: use tilesetForType instead
 const tilesetUrl = (): string => appData()?.tilesetUrl || ""
```
KaylaBrady, line 135: For the order of operations of deploys, I **think** we'd want to hold off on making this change until after doing a first deploy with the new env vars configured

lemald: Makes sense!

.envrc.template
```diff
@@ -40,6 +40,12 @@ export ERL_FLAGS="+MIscs 2048"
 ## Location of map tile images
 # export TILESET_URL=
 
+## URL for map tile images, including {x} {y} coordinate and {z} zoom level placeholders
+# export BASE_TILESET_URL=
+
+## URL for satellite map tile images, including {x} {y} coordinate and {z} zoom level placeholders
+# export BASE_TILESET_URL=
```
lemald, line 47: **nitpick:** shouldn't this be `SATELLITE_TILESET_URL`?

lemald (approved):

PR #2113, opened 2023-06-30
Display logged out vehicles

lemald:
Asana ticket: [⚙️ Better Maps | MVE | Logged out vehicles front end changes](https://app.asana.com/0/1200180014510248/1204807139228453/f)

The individual commits should be fairly self-contained if you prefer reviewing them one-at-a-time.

### DIFF
--- a/assets/src/models/vehicle.ts
+++ b/assets/src/models/vehicle.ts
@@ -18,6 +18,9 @@ export const isGhost = (
 export const isLateVehicleIndicator = ({ id }: Ghost): boolean =>
   id.startsWith("ghost-incoming-")
 
+export const isLoggedOut = ({ operatorLogonTime, runId }: Vehicle) =>
+  operatorLogonTime === null && runId === null
+
 export const isRecentlyLoggedOn = (
   vehicleOrGhost: VehicleInScheduledService | Ghost
 ): boolean => {
--- a/assets/src/models/vehicle.ts
+++ b/assets/src/models/vehicle.ts
@@ -36,10 +36,17 @@ export const isRecentlyLoggedOn = (
 }
 
 export const directionName = (
-  { directionId }: Vehicle | Ghost,
+  vehicle: Vehicle | Ghost,
   route: Route | null
-): string =>
-  directionId !== null && route ? route.directionNames[directionId] : ""
+): string => {
+  if (vehicle.directionId !== null && route) {
+    return route.directionNames[vehicle.directionId]
+  } else if (isVehicle(vehicle) && isLoggedOut(vehicle)) {
+    return "N/A"
+  }
+
+  return ""
+}
 
 export const filterVehicles = (
   vehiclesOrGhosts: (Vehicle | Ghost)[] | null
--- a/assets/src/realtime.d.ts
+++ b/assets/src/realtime.d.ts
@@ -123,7 +123,7 @@ export interface Vehicle {
   operatorLogonTime: Date | null
   overloadOffset?: number
   bearing?: number
-  blockId: BlockId
+  blockId: BlockId | null
   previousVehicleId: string | null
   scheduleAdherenceSecs: number | null
   incomingTripDirectionId: DirectionId | null
@@ -145,6 +145,7 @@ export interface Vehicle {
 export interface VehicleInScheduledService extends Vehicle {
   directionId: DirectionId
   routeId: RouteId
+  blockId: BlockId
 }
 
 export type VehicleId = string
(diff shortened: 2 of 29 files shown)


assets/src/models/vehicle.ts
```diff
@@ -18,6 +18,9 @@ export const isGhost = (
 export const isLateVehicleIndicator = ({ id }: Ghost): boolean =>
   id.startsWith("ghost-incoming-")
 
+export const isLoggedOut = ({ operatorLogonTime, runId }: Vehicle) =>
+  operatorLogonTime === null && runId === null
```
KaylaBrady, line 22: question (non-blocking): would it be worthwhile at this point to introduce a LoggedOutVehicle type with this guard? Or does that fit more into https://app.asana.com/0/1152340551558956/1204891002317847/f?

assets/src/realtime.d.ts
```diff
@@ -123,7 +123,7 @@ export interface Vehicle {
   operatorLogonTime: Date | null
   overloadOffset?: number
   bearing?: number
-  blockId: BlockId
+  blockId: BlockId | null
```
KaylaBrady, line 126: question (non-blocking): In the same vein as above, I wonder if it would be helpful to split out subtypes that are LoggedOutVehicle & LoggedInVehicle so that the existing references to the `Vehicle` type still have `block_id` as a required field. Although thinking about it now, I don't think shuttles would have a blockId, either, so maybe not necessary at this point.

assets/src/util/dateTime.ts
```diff
@@ -23,6 +23,15 @@ export const formattedTime = (date: Date): string => {
...
 }
 
+export const formattedDate = (date: Date): string => {
+  const displayMonth = date.getMonth() + 1
+  const day = date.getDate()
+  const displayDay = day < 10 ? `0${day}` : `${day}`
+  const displayYear = date.getFullYear()
+
+  return `${displayMonth}/${displayDay}/${displayYear}`
+}
```
KaylaBrady, line 33: suggestion (non-blocking): I'm a fan of the [date-fns](https://date-fns.org/docs/Getting-Started) library which has some nice date formatting helpers

assets/tests/components/searchResults.test.tsx
```diff
@@ -30,19 +40,39 @@ const state: State = {
...
+      <StateDispatchProvider state={state} dispatch={jest.fn()}>
+        <SearchResults
+          vehicles={[]}
+          onClick={jest.fn()}
+          selectedVehicleId={null}
+        />
+      </StateDispatchProvider>
+    )
 
-    expect(tree).toMatchSnapshot()
```
KaylaBrady, line 45: praise: Thanks for getting rid of this snapshot test!  
suggestion (non-blocking): It may be worth adding a second assertion that the text `NO SEARCH RESULTS` is rendered to match the test name

lib/realtime/server.ex
```diff
@@ -214,7 +235,17 @@ defmodule Realtime.Server do
   def lookup({table, {:vehicle, vehicle_or_ghost_id}}) do
     {table, :logged_in_vehicles}
     |> lookup()
-    |> Enum.filter(&(&1.id == vehicle_or_ghost_id))
+    |> Enum.find(&(&1.id == vehicle_or_ghost_id))
+    |> List.wrap()
```
KaylaBrady, line 239: praise: I like the switch to `find` over `filter`!
question (non-blocking): Is the `List.wrap()` call still necessary?

KaylaBrady (approved):
Looks great!

assets/src/realtime.d.ts
lemald: I think they do have a block ID assigned (otherwise there would have been a type error with shuttle vehicles previously), but I don't think it's meaningful (I think it might be auto-assigned to a placeholder by TM or something). I think either way from Skate's perspective it should be nullable if the values for shuttles don't mean anything, and in an ideal world maybe we would just filter those values out.

assets/src/models/vehicle.ts
lemald: I think given this is already a medium-sized PR it would be better to hold off and wait for that other ticket.

lib/realtime/server.ex
lemald: Yes, since these are still expected to return a list I basically use `List.wrap/1` just to wrap the singleton value in a list, or turn `nil` into an empty list.

KaylaBrady (approved):

PR #2115, opened 2023-07-03
feat(VehiclePropertiesPanel): Update directions button + button styling consistency

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204762161648730/f

This implementation doesn't exactly match what is in the design files. After discussion with Design, we opted to keep this button for the most part consistent with the existing street view button. However, there are also some updates needed for the street view button to add the missing hover/focus states by using the `button-primary` mixin.


assets/css/properties_panel/_vehicle_properties_panel.scss
```diff
@@ -41,9 +41,13 @@
...
 
-.c-vehicle-properties-panel__link {
-  @include font-link;
-  margin-bottom: 0.25rem;
+.c-vehicle-properties-panel__directions {
+  @include button-primary;
+  @include button-icon(1rem);
+  display: inline-flex;
+  align-items: center;
+  gap: 4px;
```
lemald, line 49: **thought:** I know `gap` is a new-ish CSS feature, it might be worth comparing [the browser compatibility info from caniuse](https://caniuse.com/?search=gap) against our supported browser policy, if you haven't already.

assets/src/helpers/icon.tsx
```diff
@@ -149,6 +151,8 @@ export const CommuterRailIcon = svgIcon(commuterRailIconSvg)
 
 export const CrowdingIcon = svgIcon(crowdingIconSvg)
 
+export const DiamondTurnRightSvg = svgIcon(diamondTurnRightSvg)
```
lemald, line 154: **suggestion:** there is actually a test suite for this file that does a basic render of each icon, you could add this one to the list if you want.

lemald (approved):
A couple of minor comments but overall looks good. 👍

assets/css/properties_panel/_vehicle_properties_panel.scss
KaylaBrady: Good call - just checked and our supported browser versions all have support for it!

PR #2117, opened 2023-07-03
feat(ts/components/mapPage): Dedicated street view improvements

firestack:
A few notes:

When implementing "snapping", due to the Street View Control being tied with the Map component, I opted to expose the street view state to children via a context to avoid refactoring the Street View Switch at this time to be more encapsulated.

The overlay SVG is implemented using refs because I made the (untested) assumption that react rendering would be slower than simply setting the CSS property that I want and given that it's connected to mouse movement, I wanted it to feel snappy and update ASAP. happy to revisit this and attempt it using state instead if wanted.

I'm putting this up as it's in a good state to get feedback, but I have a few todo's I'd like to think about some more
todo:
- [ ] Still thinking through ideas for testing
	- [x] Testing the "snapping" functionality?
		- [x] Vehicles
		- [x] Stops
		- [x] Corresponding FS events
	- [ ] Unsure how to test "the overlay"
		- maybe just ensure that the map is clickable/draggable? (Done via "snapping" tests?)
- [x] Gotta check with design on a few things, which will be covered in review. Will note those inline via comments.
	
---

Asana Ticket: [⚙️ Better Maps | Dedicated street view](https://app.asana.com/0/1203014709808707/1204632315151141/f)


assets/css/_leaflet.scss
```diff
@@ -7,9 +7,14 @@ $max-leaflet-z-index: 1000;
...
....57457C12.6648 7.35522 11.5119 6.50222 10.2043 6.49949C9.19197 6.49675 8.63238 6.77562 7.58068 7.18845C6.97329 7.4263 6.47556 7.87741 6.18311 8.45155L5.9947 8.82337C5.77536 9.25533 5.95252 9.78026 6.39401 9.99624C6.83269 10.2122 7.36697 10.0373 7.58912 9.60528L7.77753 9.23346C7.87595 9.04208 8.03904 8.89172 8.24151 8.81243L8.99513 8.51716L8.5677 10.1767C8.42148 10.7453 8.57895 11.3496 8.9867 11.7843L10.6711 13.5723C10.8736 13.7883 11.017 14.048 11.0901 14.3296L11.6047 16.3336C11.7256 16.8011 12.2149 17.0882 12.6958 16.9706C13.1766 16.853 13.4719 16.3773 13.351 15.9098L12.7267 13.4766C12.6536 13.195 12.5102 12.9325 12.3077 12.7193L11.0282 11.3605L11.5119 9.48225L11.6666 9.93336C11.8156 10.3735 12.1362 10.7371 12.558 10.9449L13.2132 11.2675C13.6518 11.4835 14.1861 11.3085 14.4083 10.8766C14.6248 10.4473 14.4476 9.91696 14.0062 9.70097ZM7.56944 13.5477C7.47945 13.7691 7.34447 13.9687 7.17013 14.1355L5.76412 15.5052C5.41261 15.8469 5.41261 16.4019 5.76412 16.7437C6.11562 17.0854 6.68365 17.0854 7.03515 16.7437L8.70549 15.1197C8.87703 14.9529 9.012 14.7534 9.1048 14.5319L9.48443 13.6078C7.92938 11.9592 8.39617 12.465 8.15153 12.1397L7.56944 13.5477Z' fill='white'/%3E%3C/g%3E%3C/svg%3E")
```
firestack, line 10: I'd _really_ like to **not** use a data url for this SVG, but I couldn't figure out how to get our webpack sass loader to comply with relative url's.

assets/css/_cutout_overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+  &,
+  &__rect,
+  &__mask-rect {
+    width: 100%;
+    height: 100%;
+  }
+
+  --circle-radius: 0;
+
+  // Cannot check hover on _this_ component due to `pointer-events: none`
```
firestack, line 15: I feel like `this` is not clear enough.
Maybe somehow explicitly state what `this` is?

```suggestion
  // Cannot check hover on _this_(&) component due to `pointer-events: none`
```
```suggestion
  // Cannot check hover on the `&` component due to `pointer-events: none`
```
```suggestion
  // Cannot check hover on the `c-overlay-cutout` component due to `pointer-events: none`
```

assets/css/_cutout_overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+  :hover > & {
+    --circle-radius: clamp(50px, 10vw, 150px);
```
firestack, line 17: - [ ] TODO: Confirm with design about min and max values here

assets/css/_cutout_overlay.scss
```diff
@@ -0,0 +1,44 @@
...
+  &__mask-circle {
+    fill: black; // Cutout circle in mask
+
+    r: var(--circle-radius);
+
+    transition: none;
+    @media not (prefers-reduced-motion) {
+      // N Frames at 60 hz
+      --transition-frames: 2;
+      transition: r calc((1s / 60) * var(--transition-frames));
```
firestack, line 32: - [ ] TODO: confirm with design on animation, and length

assets/css/_cutout_overlay.scss
```diff
@@ -0,0 +1,44 @@
+.c-overlay-cutout {
+  z-index: 800; // Below controls, above map
```
firestack, line 2: Really would prefer a better way to determine the layer z-index. Is there a sass map that already exists that might help?

assets/css/map/controls/_street_view_switch.scss
```diff
@@ -4,6 +4,7 @@
   // Leaflet uses px for positioning of their controls so we use px here to match
   top: 10px;
   right: 55px;
+  z-index: 1000; // Same control layer as leaflet controls
```
firestack, line 7: [Same as previous comment.]

assets/css/_cutout_overlay.scss
```diff
@@ -0,0 +1,44 @@
+.c-overlay-cutout {
```
firestack, line 1: Not sold on the name of this component tbh

assets/css/map/controls/_street_view_switch.scss
KaylaBrady: I see there is [$max-leaflet-z-index](https://github.com/mbta/skate/blob/e0116fb6190c6a34b12c2b7343fd8805a4e1f56b/assets/css/_leaflet.scss#L2), maybe that could be used here or expanded into a sass-map

assets/css/_cutout_overlay.scss
lemald: There's `$z-page-layout-context` in `app.scss`, that might be what you're looking for.

assets/css/_cutout_overlay.scss
lemald: I like the last of those three suggestions the best, personally.

assets/css/map/controls/_street_view_switch.scss
firestack: PTALA 89d2fca7477460c21728350a6e6a59d651873f0b

assets/css/_leaflet.scss
lemald: Yeah, I ran into this when implementing the original as well and didn't have time to dig into it too deeply.

assets/css/_cutout_overlay.scss
lemald: It looks like the file is called `_cutout_overlay.scss` whereas the class is `c-overlay-cutout`. I like the former better, since it emphasizes that "cutout" is modifying "overlay," which I think is the correct way of thinking about it (it's an overlay, with a cutout). Maybe including the word "map" somewhere in the name would also help indicate what it's for.

assets/src/components/mapMarkers.tsx
```diff
@@ -281,8 +285,31 @@ export const RouteStopMarkers = ({
             key={stop.id}
             stop={stop}
             direction={direction}
-            includeStopCard={includeStopCard}
+            includeStopCard={includeStopCard && !streetViewActive}
```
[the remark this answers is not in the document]
firestack: This was on purpose for two reasons
- It looked really weird to not have any interactivity when in street view mode
- it was less code to write

I do have this question in my design/product review ticket, so this will be confirmed with design before merge.

assets/css/_cutout_overlay.scss
firestack: ah I forgot to update this when I updated the file name, thanks

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -392,11 +394,36 @@ const SelectionDataLayers = ({
...
+              clickedMapAt: {
+                latitude_real: vehicleOrGhost.latitude,
+                longitude_real: vehicleOrGhost.longitude,
+                bearing_real: vehicleOrGhost.bearing,
+              },
+            })
+
+            window.open(url, "_blank")
+          }
+        }
```
lemald, line 426: **thought:** I know it's already on your TODO list (albeit with a question mark), but I want to flag that I think this and the corresponding changes in `mapMarkers.tsx` would be really good to have unit test coverage for given that they're both kind of complicated logic.

assets/src/components/map/controls/StreetViewSwitch.tsx
```diff
@@ -109,5 +109,10 @@ export const StreetViewControl = ({
     </>
   )
 
-  return portalElement ? ReactDOM.createPortal(control, portalElement) : null
+  return (
+    <>
+      {portalElement && ReactDOM.createPortal(control, portalElement)}
+      {streetViewEnabled && <CutoutOverlay.FollowMapMouseMove />}
```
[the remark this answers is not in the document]
firestack: I was fairly confident that this belongs with the `StreetViewSwitch` component so that the `StreetViewSwitch` is encapsulated but I'll have to come back to it after thinking on it.

assets/src/components/mapPage/mapDisplay.tsx
firestack: I was looking for feedback around where the best places to add tests for this would be, and figured I'd seed it with the ideas I already had, so thanks for confirming! Will see what I can add.

assets/css/_vehicle_map.scss
```diff
@@ -1,3 +1,21 @@
...
+  .c-street-view-switch {
+    // Same control layer as leaflet controls
+    z-index: map-get($map-component-z-index, "control-layer");
+  }
+
+  .c-cutout-overlay {
+    // Below controls, above map
+    z-index: map-get($map-component-z-index, "overlay-cutout");
+  }
+}
```
firestack, line 16: @KaylaBrady Continuing from the [comment in `_map_page` on z-index maps](https://github.com/mbta/skate/commit/89d2fca7477460c21728350a6e6a59d651873f0b#r120826285)
I tried something else out in c1c2491bbdd4b49cdd51eb8e088a2222b8bf75cd shown above.

I the commit message has a bit of a description on this, but I think this is a better location since `<Map/>` has `allowStreetView` and therefore the street view control could show up on any map, not just the map page.

assets/src/components/cutoutOverlay.tsx
```diff
@@ -0,0 +1,86 @@
...
+          e.originalEvent
+        const { x: containerScreenX, y: containerScreenY } =
+          hoverTarget.current.getBoundingClientRect()
+
+        const containerOffsetX = cursorScreenX - containerScreenX
+        const containerOffsetY = cursorScreenY - containerScreenY
+
+        const maskPosition = `${containerOffsetX}px ${containerOffsetY}px`
+
+        circleTarget.current.style.translate = maskPosition
```
lemald, line 69: **question:** I think handling this outside of the React component update flow (though if you do get a chance to testing out performance I'd be curious to see results). However, my one concern is if there is any way that React could rerender some components and clobber this `style.translate` change out from under us. That said, if that does happen then as soon as the user moves the pointer the mask will go back to the correct position, so maybe it's not such a big deal.

assets/css/map/controls/_street_view_switch.scss
firestack: Further conversation happening here https://github.com/mbta/skate/pull/2117#discussion_r1254508316, resolving this comment

assets/src/components/mapPage/mapDisplay.tsx
firestack: Tests pushed up! See [`fc4cde4` (#2117)](https://github.com/mbta/skate/pull/2117/commits/fc4cde46c99c95d56bb8ff60025a4b8e0143c118) for the specific commit.

assets/css/_cutout_overlay.scss
lemald: I feel like in a case like this we developers are going to be the primary people interacting with the component name I'm okay with prioritizing making it clearer to devs.

assets/src/components/cutoutOverlay.tsx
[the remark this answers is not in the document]
KaylaBrady: Appreciate the explanation, I think the approach you've taken makes sense!

assets/src/components/cutoutOverlay.tsx
firestack: I just tested this using state instead of refs, it doesn't completely get rid of ref's because you still need to account for the `containerRef` `<rect/>` for calculating offset within the container, and personally it feels slower/laggier. I am using a 165hz screen so it might be unnoticeable on 60hz, but I don't think it's worth it personally.

Additionally, it wasn't as easy to implement using state, and that could be contributing to the issues? Essentially, react does not seem to be updating `style.translate` when I pass it as a prop, and instead I had to use the `transform` prop and provide it the `transform` function in a string containing `transform(x px, y px)`. This is likely irrelevant, but I preferred using `style.translate`, as when I tried to access the `transform` property via javascript my IDE told me it was a readonly property (didn't dig further, might have been possible).

I'm going to leave it as is for now, I'll think on if I can distil any of what I found into useful inline comments.

KaylaBrady (approved):
Resolved my threads, thanks!

assets/src/components/cutoutOverlay.tsx
lemald: Sounds good! I don't know how many users we have with higher refresh rate monitors but if it's noticeably less responsive at 165 hz that's probably good evidence that we should keep the current approach rather than risk something that might become less responsive under certain conditions.

firestack:
Unfortunately one more round of changes for review, a999d5818c02350381d3175c645d7d059a14f9bf.

The circle cutout doesn't disappear when hovering over controls because the `leaflet-container` also contains the `leaflet-control-container`, which means hovering over controls activates the hover state on the `leaflet-container`.
But this also means that when hovering over map elements in the `leaflet-map-pane`, `c-cutout-overlay-hover-target` _loses_ it's hover state, so I needed to "allow-list" that the `circle-expanded` state was enabled when hovering over `leaflet-map-pane` in `_leaflet.scss`). 

I guess alternatively, I could `deny-list` the `leaflet-control-container` with the CSS `:not` pseudo selector.

assets/src/components/mapMarkers.tsx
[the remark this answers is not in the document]
firestack: Kathleen approves, so this is in!

assets/css/_cutout_overlay.scss
firestack: 8538e27d04eb051ee74ef6c249685ddf9a7fd67b

lemald:
> PTALA 🙏🏻 [a999d58](https://github.com/mbta/skate/commit/a999d5818c02350381d3175c645d7d059a14f9bf).
> 
> The circle cutout doesn't disappear when hovering over controls because the `leaflet-container` also contains the `leaflet-control-container`, which means hovering over controls activates the hover state on the `leaflet-container`. But this also means that when hovering over map elements in the `leaflet-map-pane`, `c-cutout-overlay-hover-target` _loses_ it's hover state, so I needed to "allow-list" that the `circle-expanded` state was enabled when hovering over `leaflet-map-pane` in `_leaflet.scss`).
> 
> I guess alternatively, I could `deny-list` the `leaflet-control-container` with the CSS `:not` pseudo selector.
> 
> Let me know if y'all have a preference here for one or the other.

After reading over these changes, I _think_ I understand and agree with the current approach as-is.

KaylaBrady (approved):

PR #2118, opened 2023-07-07
feat: Satellite view for maps

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204547304658756/f

TODO
- [x] Write tests
- [x] Style the control
- [x] Check w/ design about control positioning
- [x] Refactor out generic control parent component (may end up making a separate ticket for this)


assets/tests/components/__snapshots__/mapPage.test.tsx.snap
```diff
@@ -188,7 +188,7 @@ exports[`<MapPage /> Snapshot renders the empty state 1`] = `
                 <img
                   alt=""
                   class="leaflet-tile"
-                  src="/13/2479/3029.png"
+                  src="undefined"
```
KaylaBrady, line 191: May opt to do some global mocking of the tile urls

assets/tests/testHelpers/mockHelpers.ts
```diff
@@ -79,3 +80,14 @@ export const mockFullStoryEvent = (): void => {
...
   })
 }
+
+export const mockTileUrls = (): void => {
+  ;(tilesetUrlForType as jest.Mock).mockImplementation((type: TileType) => {
+    switch (type) {
+      case "base":
+        return "test_base_url.com/{z}/{x}/{y}"
+      case "satellite":
+        return "test_satellite_url.com/{z}/{x}/{y}"
```
firestack, line 90: suggestion: Can we make these url's impossible to resolve?
maybe removing only the tld is enough and should still function?

```suggestion
      case "base":
        return "test_base_url/{z}/{x}/{y}"
      case "satellite":
        return "test_satellite_url/{z}/{x}/{y}"
```

I could see localhost being an option if that doesn't work, but I'd rather that there's a URL that couldn't resolve.

lemald (commented):
blah

assets/src/components/map/controls/LayersControl.tsx
```diff
@@ -0,0 +1,104 @@
...
+  setTileType,
+}: {
+  tileType: TileType
+  setTileType: (tileType: TileType) => void
+}): JSX.Element | null => {
+  const map = useMap()
+  const portalParent = map
+    .getContainer()
+    // TODO: handle two maps?
+    .querySelector(".leaflet-control-container .leaflet-top.leaflet-right")
```
lemald, line 20: **thought:** I played around and logged the output of `map.getContainer()` to console and it's actually the `<div>` that has the classes `c-vehicle-map leaflet-container`, so I think this would be fine if we had multiple maps because they would be under different `c-vehicle-map` components.

assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -71,6 +82,15 @@ describe("Shuttle Map Page", () => {
     expect(result.asFragment()).toMatchSnapshot()
   })
 
+  test("Has the layers control", () => {
+    render(
+      <BrowserRouter>
+        <ShuttleMapPage />
+      </BrowserRouter>
+    )
+    expect(screen.getByRole("button", { name: "Layers" })).toBeInTheDocument()
```
firestack, line 91: suggestion: Since this is replicated across multiple test suites, can you make this into a shared/common selector like the others we have in `assets/tests/testHelpers/selectors`?

assets/src/components/map/controls/LayersControl.tsx
lemald, line 1: **nitpick:** We usually lowercase the tsx filenames (`layersControl.tsx`).

assets/src/components/map/controls/LayersControl.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [portalElement, setPortalElement] = useState<HTMLElement | null>(null)
+
+  const [showLayersList, setShowLayersList] = useState(false)
+
+  useEffect(() => {
+    if (!portalParent || !portalElement) {
```
lemald, line 26: **question:** is the `!portalParent` needed? It looks like this means that `portalParent` could be null but `portalElement` defined and then it would cause us to create a new `portalElement`, only to repeat again whenever `portalParent` becomes defined again.

Aside from that though I _think_ the logic in this hook is sound.

lemald: (Though I do realize that this is how I myself did it in #1834, so maybe I had my reasons?)

assets/src/components/propertiesPanel/miniMap.tsx
```diff
@@ -53,25 +55,26 @@ const MiniMap = ({
...
+
+  return (
     <MapFollowingPrimaryVehicles
       selectedVehicleId={vehicle.id}
       vehicles={[vehicle]}
       shapes={shapes}
       secondaryVehicles={routeVehicles}
       stations={stations}
-      allowFullscreen={false}
+      allowFullscreen={!inMapBetaGroup}
```
lemald, line 67: **question:** this restores the ability of non-beta group users to fullscreen the minimap, is that intentional?

lemald: Oh, wait, I think I just read the code wrong, this is fine.

assets/tests/components/shuttleMapPage.test.tsx
KaylaBrady: Great point, updated in https://github.com/mbta/skate/pull/2118/commits/c68c8fc7dd51195bda25c8e5ede6a12f0703add7

assets/src/state/mapLayersState.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+type MapKey = keyof MapLayersState
+
+export const initialMapLayersState: MapLayersState = {
+  searchMap: {
+    tileType: "base",
+  },
+
```
lemald, line 23: **nitpick:** extraneous blank line

assets/tests/components/map/controls/LayersControl.test.tsx
lemald, line 1: **nitpick:** same note re: capitalization as with `layersControl.tsx`.

assets/tests/components/map/controls/LayersControl.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  test("when the a tile option is clicked, then setTileType is called", async () => {
+    const setTileTypeMock = jest.fn()
+    render(<LayersControl tileType="base" setTileType={setTileTypeMock} />, {
+      wrapper: mapWrapper,
+    })
+    await userEvent.click(layersControlButton.get())
+    await userEvent.click(screen.getByLabelText("Satellite"))
+
+    expect(setTileTypeMock).toHaveBeenCalled()
```
lemald, line 44: **nitpick (non-blocking):** you could use `toHaveBeenCalledWith` to also assert that it's called with the correct value.

assets/tests/testHelpers/mockHelpers.ts
KaylaBrady: Updated in https://github.com/mbta/skate/pull/2118/commits/b33e2d2798df639c1b7907229277d8d65b72544e!

assets/src/components/map/controls/layersControl.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  useEffect(() => {
+    if (!portalParent || !portalElement) {
+      setPortalElement(document.createElement("div"))
+    }
+
+    if (portalParent && portalElement) {
+      portalElement.className = "c-map__layer-control"
+      portalParent.append(portalElement)
+      Leaflet.DomEvent.disableClickPropagation(portalElement)
```
firestack, line 32: issue: based on the [built-in leaflet layers control code](https://github.com/Leaflet/Leaflet/blob/b6b56f6e707142c177fad2f67827a5007e56736a/src/control/Control.Layers.js#L186-L187), I think `disableScrollPropagation` should be added.

```suggestion
      Leaflet.DomEvent.disableClickPropagation(portalElement)
      Leaflet.DomEvent.disableScrollPropagation(portalElement)
```

firestack: Upon checking this out in other controls, it seems like none of the other ones do this, so I'd be fine with leaving it if Design or other Devs are fine with this.

lemald (approved):
I have that one open comment thread but given that it looks like that's already the way the street view control does it I think you can consider that non-blocking.

assets/src/components/map/controls/customControl.tsx
```diff
@@ -0,0 +1,66 @@
...
+        portalParent.append(portalElement)
+      }
+      Leaflet.DomEvent.disableClickPropagation(portalElement)
+    }
+
+    return () => portalElement?.remove()
+  }, [portalElement, portalParent])
+
+  const control = (
+    <div className={`leaflet-control leaflet-bar ${className}`}>{children}</div>
```
firestack, line 62: question: why is `leaflet-bar` added here? 

from what I can tell it's used to group the zoom in and zoom out buttons in the zoom control, but not anywhere else.

KaylaBrady: It can be removed here! It does some styling like the control outline & border radius, but for the layers control at least that should be done on each component of the control.

assets/tests/components/map/controls/customControl.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </CustomControl>
+      </>,
+      {
+        wrapper: mapWrapper,
+      }
+    )
+
+    expect(
+      container.querySelector(".leaflet-control-container")
+    ).toMatchSnapshot()
```
KaylaBrady, line 38: Snapshot testing seemed like the most straight-forward approach to me here in order to verify that the control ended up in exactly the expected place in the dom. Open to suggestions of other approaches though!

assets/src/components/map/controls/customControl.tsx
```diff
@@ -0,0 +1,62 @@
...
+    case "bottomright":
+      return ".leaflet-bottom.leaflet-right"
+  }
+}
+
+export const CustomControl = ({
+  position,
+  children,
+  className,
+  insertAfterSelector,
```
KaylaBrady, line 23: @firestack took this `insertAfterSelector` approach based on one of your suggestions in our conversation last week. I tried using `createControlComponent` to make this generic control but ultimately struggled with passing children to the generic control. I think that in order to insert this control in a particular place in the control bar we would need to continue using portal in some capacity, but please let me know if there is a different approach you see as preferable here!

assets/src/components/map/controls/layersControl.tsx
```diff
@@ -0,0 +1,82 @@
...
+                  Satellite
+                </label>
+              </li>
+            </ul>
+          </div>
+        )}
+      </CustomControl>
+      <TileLayer
+        url={`${tilesetUrlForType(tileType)}`}
+        attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
```
lemald, line 78: **issue:** shouldn't this attribution actually change depending on the layer? The MassGIS aerial imagery doesn't actually require an attribution, but it's not OSM and it seems like it might be good to give information on the source for any users who are curious.

assets/tests/components/map/controls/customControl.test.tsx
lemald: I'm not sure if this is actually better, but would it be possible to do a more traditional test where you select the control and then look up the DOM tree using the `parentElement` property to verify that it's in the appropriate position?

assets/css/map/controls/_layers_control.scss
```diff
@@ -0,0 +1,39 @@
...
+  padding: 5px;
+  span svg {
+    stroke: none;
+  }
+
+  &--selected {
+    background-color: $color-gray-200;
+  }
+  &:focus-visible {
+    outline: revert;
```
firestack, line 20: question(non-blocking): what are we `revert`ing here to, the user-agent style?

KaylaBrady: Yep! The other controls also have the user-agent focus style, so reverting to undo the https://github.com/mbta/skate/blob/400edc0adcb79897b99a2f24c3fbd48b99a22881/assets/css/_base.scss#L14 for buttons.

PR #2124, opened 2023-07-13
feat: New vehicle search channel that accepts a limit

KaylaBrady:

This new channel accepts messages to change the value for limit as well as the other search parameters. Otherwise, if limit was part of the topic, then increasing the limit would result in leaving the old topic, resulting in search results being cleared. I'm not totally a fan of how this means the `vehicles_search` topic is generic... very open to alternative ideas! 

I played around a bit with hooking this up to the frontend in `kb-limited-search-results-hook` to make sure we can communicate with this new channel as expected.

### DIFF
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -96,6 +96,14 @@ defmodule Realtime.Server do
     )
   end
 
+  @spec update_search_subscription(search_params(), GenServer.server()) :: [VehicleOrGhost.t()]
+  def update_search_subscription(search_params, server \\ default_name()) do
+    update_subscription(
+      server,
+      {:search, search_params}
+    )
+  end
+
   @spec subscribe_to_vehicle(String.t(), GenServer.server()) :: [VehicleOrGhost.t()]
   def subscribe_to_vehicle(vehicle_id, server \\ default_name()) do
     subscribe(
@@ -157,6 +165,14 @@ defmodule Realtime.Server do
   @spec subscribe(GenServer.server(), {:alerts, Route.id()}) :: [String.t()]
   defp subscribe(server, subscription_key) do
     {registry_key, ets} = GenServer.call(server, :subscription_info)
+    Registry.register(Realtime.Registry, registry_key, subscription_key)
+    lookup({ets, subscription_key})
+  end
+
+  defp update_subscription(server, {:search, _search_params} = subscription_key) do
+    {registry_key, ets} = GenServer.call(server, :subscription_info)
+    # Replace the old search subscription with the new one
+    Registry.unregister_match(Realtime.Registry, registry_key, {:search, %{}})
     Registry.register(Realtime.Registry, registry_key, subscription_key)
     lookup({ets, subscription_key})
   end
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -91,18 +91,12 @@ defmodule Realtime.Server do
 
   @spec subscribe_to_search(search_params(), GenServer.server()) :: [VehicleOrGhost.t()]
   def subscribe_to_search(search_params, server \\ default_name()) do
-    subscribe(
-      server,
-      {:search, search_params}
-    )
+    subscribe(server, {:search, search_params})
   end
 
   @spec update_search_subscription(search_params(), GenServer.server()) :: [VehicleOrGhost.t()]
   def update_search_subscription(search_params, server \\ default_name()) do
-    update_subscription(
-      server,
-      {:search, search_params}
-    )
+    update_subscription(server, {:search, search_params})
   end
 
   @spec subscribe_to_vehicle(String.t(), GenServer.server()) :: [VehicleOrGhost.t()]
@@ -175,6 +169,7 @@ defmodule Realtime.Server do
     # Replace the old search subscription with the new one
     Registry.unregister_match(Realtime.Registry, registry_key, {:search, %{}})
     Registry.register(Realtime.Registry, registry_key, subscription_key)
+
     lookup({ets, subscription_key})
   end
 
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -44,7 +44,8 @@ defmodule Realtime.Server do
   @type search_params :: %{
           :text => String.t(),
           :property => search_property(),
-          optional(:include_logged_out_vehicles) => boolean()
+          optional(:include_logged_out_vehicles) => boolean(),
+          optional(:limit) => pos_integer()
         }
 
   @type search_property :: :all | :run | :vehicle | :operator
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -35,6 +35,7 @@ defmodule Realtime.Server do
           | :all_shuttles
           | :logged_in_vehicles
           | {:search, search_params()}
+          | {:limited_search, limited_search_params()}
           | {:vehicle, String.t()}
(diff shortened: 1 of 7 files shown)


lib/realtime/server.ex
```diff
@@ -161,6 +164,15 @@ defmodule Realtime.Server do
     lookup({ets, subscription_key})
   end
 
+  defp update_subscription(server, {:search, _search_params} = subscription_key) do
+    {registry_key, ets} = GenServer.call(server, :subscription_info)
+    # Replace the old search subscription with the new one
+    Registry.unregister_match(Realtime.Registry, registry_key, {:search, %{}})
+    Registry.register(Realtime.Registry, registry_key, subscription_key)
```
KaylaBrady, line 171: There is a [Registry.update_value/3](https://hexdocs.pm/elixir/1.6.6/Registry.html#update_value/3) function, but it isn't usable for our use case because we allow duplicate values. 

From the [unregister_match docs](https://hexdocs.pm/elixir/1.6.6/Registry.html#unregister/2):
> Unregisters all entries for the given key associated to the current process in registry.

lib/skate_web/channels/vehicles_search_channel.ex
```diff
@@ -0,0 +1,76 @@
+defmodule SkateWeb.VehiclesSearchChannel do
```
KaylaBrady, line 1: Some duplication here of the existing `VehiclesChannel` search topic - key new piece is `handle_in`

lib/skate_web/channels/vehicles_search_channel.ex
```diff
@@ -0,0 +1,76 @@
...
+  use SkateWeb, :channel
+  use SkateWeb.AuthenticatedChannel
+  require Logger
+
+  alias Realtime.Server
+  alias Util.Duration
+
+  @impl SkateWeb.AuthenticatedChannel
+  def join_authenticated(
+        "vehicles_search:" <> _unique_id,
```
KaylaBrady, line 11: Still thinking through what `_unique_id` might be on the frontend, I'm thinking `property`. It is needed because duplicate subscriptions to the same topic aren't allowed from a single socket, so to support subscribing to multiple searches at once, the topics will have to be different.

lib/realtime/vehicle_or_ghost.ex
```diff
@@ -4,59 +4,87 @@ defmodule Realtime.VehicleOrGhost do
...
-
-  @spec filter_by_prop_matching([t()], atom() | [atom()], String.t()) :: [t()]
+  def find_by(vehicles, %{text: text, property: search_property, limit: limit}) do
+    search_terms = text |> clean_for_matching |> Enum.reject(&(String.length(&1) < 2))
+
+    if search_terms == [] do
+      []
+    else
+      take_by_props_matching(
+        vehicles,
```
KaylaBrady, line 14: this list of vehicles should be sorted by relevance first (logged on time), but this PR is getting meaty. Will create a separate PR for that logic.

lib/skate_web/channels/vehicles_search_channel.ex
KaylaBrady: Maybe as an alternative, the topic could be`vehicles_search: <> search_params` kind of like the existing search, and only accept the limit via message. I wasn't sure before how to access the topic name from within the `handle_in` callback, but now I see you can get the topic as a field from [the socket](https://hexdocs.pm/phoenix/Phoenix.Socket.html#module-socket-fields).

This would mean subscribing to a different topic when the text changes, though I'm not sure that is a problem for regular search results. Maybe it depends on whether or not these channels should also be used for autocomplete.

KaylaBrady:
Moving back into draft mode for a bit - realized that the new channel should also expose a field like `has_more_matches` so that the frontend knows when to display the "Show more" button

test/realtime/server_test.exs
```diff
@@ -397,6 +397,124 @@ defmodule Realtime.ServerTest do
     end
   end
 
+  describe "subscribe_to_limited_search/2" do
```
KaylaBrady, line 400: largely a copy of the above subscribe_to_search tests

lib/skate_web/channels/vehicles_search_channel.ex
KaylaBrady: Resolution from discussion: search params will be put back into the topic name. Autocomplete will be a different topic.

lib/skate_web/channels/vehicles_search_channel.ex
```diff
@@ -0,0 +1,76 @@
...
+    subscribe_args =
+      Map.put(
+        subscribe_args,
+        :include_logged_out_vehicles,
+        Skate.Settings.User.is_in_test_group(user_id, "search-logged-out-vehicles")
+      )
+
+    Logger.info(fn ->
+      "User=#{username} searched for property=#{subscribe_args.property}, text=#{subscribe_args.text} limit=#{subscribe_args.limit}"
+    end)
```
lemald, line 37: **suggestion (non-blocking): it might be good to have a clearer delineation between this logging and the logging in `VehiclesChannel` besides just the presence / absence of the `limit` field.

lemald (approved):
A couple of minor thoughts but overall looks good to me.

lib/skate_web/channels/vehicles_search_channel.ex
KaylaBrady: Good point. I'm thinking of added something like "limited" to the topic name too for further differentiation of these channels.

lib/skate_web/channels/vehicles_search_channel.ex
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  @impl Phoenix.Channel
+  def handle_in(
+        "update_search_query",
+        %{"text" => text, "limit" => limit, "property" => property},
+        socket
+      ) do
+    vehicles =
+      Duration.log_duration(Server, :update_search_subscription, [
```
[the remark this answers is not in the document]
KaylaBrady: Thanks for raising this! I hadn't considered the effects of an update of vehicle data between these two calls. 

>  Since this is calling Server.update_search_description directly and that function along with the private update_subscription/2 does all of its logic without making a GenServer call, this means that all of the calls to the registry functions happen in the calling process, not the Realtime.Server process. 

I _think_ users would still get the latest data for the newest search because the call to `lookup({ets, subscription_key})` from within `update_subscription` would return the latest data available in the `ets` table, even if that `ets` table was updated between the `unregister_match` and `register` calls. But there may still be some nuance about the timing & GenServer that I'm missing.

lib/skate_web/channels/vehicles_search_channel.ex
KaylaBrady: Made both changes in https://github.com/mbta/skate/pull/2124/commits/e320a2364b3e5377b661a6a6d222c0a72fd53b3b

lib/skate_web/channels/vehicles_search_channel.ex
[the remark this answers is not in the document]
lemald: I'm referring to what happens if the Concentrate GenServer pushes updated vehicle information to the `Realtime.Server` process. In that case `Server.broadcast/2` would be called and it would use the subscription information available from Registry at that time. That said, I do still believe it's an extreme edge case so I'm not going to worry too much about it.

lemald (approved):
Latest changes look good!

PR #2127, opened 2023-07-14
Geocoding service backend

lemald:
Asana ticket: [⚙️ Better Maps | Search | Implement geocoding engine for search](https://app.asana.com/0/1200180014510248/1204573236888869/f)

In 5ffd0d44 I also made it so that for local dev use you can use your regular `AWS` credentials in the `~/.aws` directory with ExAws, including for the location search API calls.


lib/skate/location_search/aws_location_request.ex
```diff
@@ -0,0 +1,89 @@
+defmodule Skate.LocationSearch.AwsLocationRequest do
+  alias Skate.LocationSearch.SearchResult
+
+  @base_arguments %{
+    FilterBBox: [-71.55, 42.05, -70.6, 42.65]
+  }
```
KaylaBrady, line 6: question (non-blocking): Would it be worth having this set in config and used from both here and in the `Map.tsx` component? May not be worth it since the values likely won't change much.

lib/skate/location_search/aws_location_request.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  @spec search(String.t()) :: {:ok, map()} | {:error, term()}
+  def search(text) do
```
KaylaBrady, line 9: question(non-blocking): Why not have return the result of `parse_search_response` when the search request is successful?

test/support/factory.ex
```diff
@@ -198,4 +198,34 @@ defmodule Skate.Factory do
       save_changes_to_tab_uuid: nil
     }
   end
+
+  def amazon_location_search_result_factory(attrs) do
```
KaylaBrady, line 202: praise: appreciate this new factory making the tests more readable!

KaylaBrady (approved):
Looks good!

lib/skate/location_search/aws_location_request.ex
firestack: I second this!

lemald: Good point - I hadn't pursued that assuming it would be hard given that one part of the code is Elixir and the other is TypeScript, but I can just include it in one of the `data-*` attributes that we render in the main `<div>` container in Phoenix.

lemald: Addressed in e7508098.

lib/skate/location_search/aws_location_request.ex
lemald: After mulling this over I decided to refactor it as suggested and I'm pretty happy with the end result, let me know what you think. Commit 08c80418.

lemald:
Briefly changed the base on this to the feature branch but then realized that since this doesn't change the user-facing functionality we should just merge it straight into `master`. That does mean I had to rebase to fix a merge conflict though so references to commits I just made in some comments will be out of date.

lib/skate/location_search/aws_location_request.ex
KaylaBrady: Nice, looks good to me!

PR #2128, opened 2023-07-14
feat: Implement new search components

firestack:
This does a few things to create new components based on old components, while avoiding accidentally modifying the old components.
- Copy `SearchForm` to `OldSearchForm`, update references to point to `OldSearchForm`
- Update `SearchForm` to match new designs, use on Map Page.
- Create `FilterAccordion` mostly from `GarageFilter`
	- It's not _exactly_ the same, so I didn't feel comfortable making the `GarageFilter` depend on this component. But I do think we could. Will either implement here or open a ticket based on dev feedback.
- Build a "interactive"(focus state, similar to the bus stop icons) `CircleXIcon`
	- note: I don't like that this component name matches that of our `CircleXIcon` from our `icons.tsx` file, but I was unsure what to name it otherwise.

Additionally, the Filter currently acts like the old ratio buttons. Because the application reducer that remembers these settings needs to change to accommodate this, I've left this to be changed in another PR. If Design pushes back on this, I'll update this PR with those changes.

---

Asana Ticket: https://app.asana.com/0/0/1204650425301364

### DIFF
--- a/assets/css/_search_form.scss
+++ b/assets/css/_search_form.scss
@@ -50,8 +50,9 @@
 .c-search-form__input-controls {
   position: absolute;
   // Stick to right side of box, do not intercept mouse events to the text input
-  inset-block: 0;
-  right: 0;
+  inset: 0;
+  left: unset;
+
   width: 0;
 
   // Layout controls from the right side
--- a/assets/css/_search_form.scss
+++ b/assets/css/_search_form.scss
@@ -72,7 +72,13 @@
 }
 
 .c-search-form__submit {
-  @include button-icon(1rem);
+  span > svg {
+    fill: currentColor;
+
+    display: block;
+    height: 1rem;
+    aspect-ratio: 1/1;
+  }
 
   display: flex;
 
--- a/assets/css/_search_form.scss
+++ b/assets/css/_search_form.scss
@@ -1,95 +1,97 @@
-.c-search-form__row {
+.c-search-form {
   display: flex;
-
-  &:not(:last-child) {
-    margin-bottom: 1rem;
-  }
+  flex-flow: column nowrap;
+  gap: 1rem;
 }
 
-.c-search-form__text {
+.c-search-form__search-control {
+  // Make control area match designs
+  height: 2.25rem;
+}
+
+.c-search-form__search-input-container {
+  // Layout floating `input-controls` over the `search-input-container`
   position: relative;
-  flex: 1 1 auto;
+
+  // Use bootstrap control as base for now
+  @extend .form-control;
+
+  // fill control defined valid area
+  height: 100%;
+  // Make space for `input-controls`
+  padding-right: calc(2.25rem * 2);
+  // Ignore bootstrap padding due to `input-controls`
+  padding-block: 0;
+
+  border-color: $color-gray-400;
+
+  &:focus,
+  &:focus-within,
+  &:active {
+    border: 1px solid $color-eggplant-500;
+    box-shadow: 0px 0px 0px 2px #aa4ef22b;
+    color: $color-gray-800;
+  }
 }
 
 .c-search-form__input {
+  // Align self within parent boundaries
   width: 100%;
-  height: 2.375rem;
-  box-sizing: border-box;
+  height: 100%;
 
-  border-radius: 4px 0 0 4px;
+  // Match font size from the Route Picker Search
+  @include font-small;
+
+  // Remove default displays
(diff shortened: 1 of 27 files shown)


assets/css/_search_form.scss
```diff
@@ -1,95 +1,97 @@
...
 }
 
-.c-search-form__text {
+.c-search-form__search-input-container {
+  // Layout floating `input-controls` over the `search-input-container`
   position: relative;
-  flex: 1 1 auto;
+
+  // Use bootstrap control as base for now
+  @extend .form-control;
```
lemald, line 17: **question:** what are the relative merits of extending the `form-control` style versus customizing it via the available Sass variables? It seems like some of this styling is definitely more specific (mainly leaving that extra space for the controls), but other parts about the colors and focus state look like they could maybe be accommodated via the Sass method.

assets/css/_search_form.scss
```diff
@@ -1,95 +1,97 @@
...
 
-  font-size: var(--font-size-s);
-  font-weight: var(--font-weight-reg);
-  color: var(--font-color-on-light-600);
+// #region Search Input Buttons
+.c-search-form__input-controls {
+  position: absolute;
+  // Stick to right side of box, do not intercept mouse events to the text input
+  inset-block: 0;
+  right: 0;
```
lemald, line 54: **question:** what's the purpose of setting both `inset-block` and `right`? I assume there's some reason but I can't seem to figure it out.

assets/css/_search_form.scss
```diff
@@ -1,95 +1,97 @@
...
+  }
 
-.c-search-form__property-buttons {
-  display: flex;
-  flex: 1 0 auto;
-  list-style-type: none;
-  margin: 0;
-}
+  // flex: 0 0 2.25rem;
+  // flex: 0;
```
lemald, line 81: **nitpick:** looks like this might be some temporary code that got left over as comments.

lemald (approved):
Okay, I left a few questions but I don't think any of them are major blockers - overall this looks good!

assets/css/_search_form.scss
firestack: I needed to set everything _but_ `left`, here, `inset-block` sets both `top` and `bottom` to `0`.

Alternatively I think I could do `inset: 0` and `left: unset`? Unsure about this though (I thought I tried it).
```suggestion
  // Stick to right side of box, do not intercept mouse events to the text input
  inset: 0;
  left: 0;
```

assets/css/_search_form.scss
firestack: > versus customizing it via the available Sass variables
I didn't want to assume we'd be going with bootstrap as a design system, so I didn't want to spend time customizing bootstrap, and instead try and be as close to the way we're currently doing CSS within Skate.

My thought process for extending `form-control` was
- it should give us an easier transistion to bootstrap for this component if we use bootstrap as our design system.
- it should allow us to reference an existing component+style if we're moving to a different design system.

I contemplated inlining bootstraps styles or pruning it down, but I figured the dev time wasn't worth it since this worked.

I'm open to configuring this via SASS/CSS variables if you think it's worth it, I just didn't want to assume that this is what _all_ our `form-controls` should follow in terms of color/focus.

lemald: Makes sense! I wasn't sure exactly how all of your conversations with design had panned out, and I understand not wanting to make premature decisions about what parts of this styling will be universal.

assets/css/_search_form.scss
lemald: Ah, okay, I was slightly misreading the MDN docs on `inset-block` and that's not a property I was previously familiar with. I think it's fine as-is.

lemald: On the other hand maybe it's weird to mix `left` / `right` / `top` / `bottom` with properties that depend on text directionality and writing mode stuff, but I guess the practical chance of that being an issue is extremely low.

firestack: True! I'm gonna rework this

firestack: [`f812089` (#2128)](https://github.com/mbta/skate/pull/2128/commits/f81208973c7f11e338010fdb826601338d164377)

firestack:
@mbta/skate-developers I've gotten final approval from design but there have been a few code changes since last review, I'll plan to merge this and #2132 around 1550-1600 EST if there are no further comments.

PR #2140, opened 2023-07-21
feat(MapPage): Limited search results grouped by property

KaylaBrady:

also dependent on https://github.com/mbta/skate/pull/2124

To limit the scope of this PR & more quickly unblock the dependent work, I've broken the following out into separate subtask tickets that will be addressed by separate PRs
* Getting filter functionality to work
* Showing a single "No results" message when no results for any property

### DIFF
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -141,7 +141,9 @@ $map-page-search-stub-size: 10px;
   color: $color-gray-900;
   font-feature-settings: "clig" off, "liga" off;
   font-size: var(--font-size-xs);
+  font-weight: var(--font-weight-reg);
   line-height: 1;
+  margin-top: 0;
   margin-bottom: 0.5rem;
 }
 
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -133,6 +133,24 @@ $map-page-search-stub-size: 10px;
   padding: 0.5rem;
 }
 
+.c-map-page__search_results_section {
+  margin-bottom: 0.5rem;
+}
+
+.c-map-page__search_results_header {
+  color: $color-gray-900;
+  font-feature-settings: "clig" off, "liga" off;
+  font-size: var(--font-size-xs);
+  line-height: 1;
+  margin-bottom: 0.5rem;
+}
+
+.c-map_page__search_results_actions {
+  display: flex;
+  justify-content: center;
+  padding: 0.5rem;
+}
+
 .c-map-page__new-search-button:only-child {
   justify-content: center;
   flex: 1;
--- a/assets/src/components/mapPage/searchResultsByProperty.tsx
+++ b/assets/src/components/mapPage/searchResultsByProperty.tsx
@@ -33,11 +33,16 @@ const SearchResultSection = ({
   })
 
   return (
-    <div className="c-map-page__search_results_section">
-      <div className="c-map-page__search_results_header">
+    <section
+      className="c-map-page__search_results_section"
+      aria-labelledby={`search-results__${property}`}
+    >
+      <h2
+        className="c-map-page__search_results_header"
+        id={`search-results__${property}`}
+      >
         {searchPropertyDisplayConfig[property].name}
-      </div>
-
+      </h2>
       {limitedSearchResults == null ? (
         <Loading />
       ) : limitedSearchResults.matchingVehicles.length > 0 ? (
@@ -61,7 +66,7 @@ const SearchResultSection = ({
       ) : (
         "No results found"
       )}
-    </div>
+    </section>
   )
 }
 
@@ -80,19 +85,18 @@ const SearchResultsByProperty = ({
             searchPropertyDisplayConfig[second.property].order
         )
         .map(({ property, limit }) => (
-          <div key={property}>
-            <SearchResultSection
-              property={property}
-              text={searchPageState.query.text}
-              limit={limit}
-              selectVehicle={selectSearchResult}
-              showMore={() =>
-                dispatch(
-                  setPropertyMatchLimit(property as SearchProperty, limit + 25)
-                )
-              }
-            />
-          </div>
+          <SearchResultSection
+            key={property}
+            property={property}
+            text={searchPageState.query.text}
+            limit={limit}
+            selectVehicle={selectSearchResult}
+            showMore={() =>
+              dispatch(
+                setPropertyMatchLimit(property as SearchProperty, limit + 25)
+              )
+            }
+          />
         ))}
     </>
   )
--- a/assets/src/components/mapPage/searchResultsByProperty.tsx
+++ b/assets/src/components/mapPage/searchResultsByProperty.tsx
@@ -80,13 +80,19 @@ const SearchResultsByProperty = ({
   selectSearchResult: (result: Vehicle | Ghost | null) => void
 }) => {
   const [{ searchPageState }, dispatch] = useContext(StateDispatchContext)
+
   return (
     <div aria-label="Grouped Search Results">
-      {searchPageState.query.properties
+      {Object.entries(searchPageState.query.properties)
+        .filter(([property, limit]) => limit > 0 && property != "location")
+        .map(([property, limit]) => ({
+          property: property as SearchProperty,
+          limit,
+        }))
         .sort(
-          (first, second) =>
-            searchPropertyDisplayConfig[first.property].order -
-            searchPropertyDisplayConfig[second.property].order
+          ({ property: first_property }, { property: second_property }) =>
+            searchPropertyDisplayConfig[first_property].order -
+            searchPropertyDisplayConfig[second_property].order
         )
         .map(({ property, limit }) => (
           <SearchResultSection
--- a/assets/src/components/mapPage/searchResultsByProperty.tsx
+++ b/assets/src/components/mapPage/searchResultsByProperty.tsx
@@ -1,0 +1,101 @@
+import { useContext } from "react"
+import { SocketContext } from "../../contexts/socketContext"
+import { StateDispatchContext } from "../../contexts/stateDispatchContext"
+import { useLimitedSearchResults } from "../../hooks/useSearchResults"
+import {
+  SearchProperty,
+  searchPropertyDisplayConfig,
+} from "../../models/searchQuery"
+import { Vehicle, Ghost } from "../../realtime"
+import { setPropertyMatchLimit } from "../../state/searchPageState"
+import Loading from "../loading"
+import SearchResults from "../searchResults"
+import React from "react"
+
+const SearchResultSection = ({
+  property,
+  text,
+  limit,
+  selectVehicle,
+  showMore,
+}: {
+  property: SearchProperty
+  text: string
+  limit: number
+  selectVehicle: (vehicle: Vehicle | Ghost) => void
+  showMore: () => void
+}) => {
+  const { socket } = useContext(SocketContext)
+  const limitedSearchResults = useLimitedSearchResults(socket, {
+    property,
+    text,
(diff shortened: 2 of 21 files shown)


assets/css/_map_page.scss
```diff
@@ -133,6 +133,24 @@ $map-page-search-stub-size: 10px;
   padding: 0.5rem;
 }
 
+.c-map-page__search_results_section {
+  margin-bottom: 0.5rem;
+}
+
+.c-map-page__search_results_header {
+  color: $color-gray-900;
+  font-feature-settings: "clig" off, "liga" off;
```
lemald, line 142: **question (non-blocking):** where does this come from? I can't seem to find anything in the designs about it, and it doesn't look like we've used `font-feature-settings` previously. No objections or anything, just curious!

assets/src/components/mapPage/searchResultsByProperty.tsx
```diff
@@ -0,0 +1,105 @@
...
+        .map(({ property, limit }) => (
+          <SearchResultSection
+            key={property}
+            property={property}
+            text={searchPageState.query.text}
+            limit={limit}
+            selectVehicle={selectSearchResult}
+            showMore={() =>
+              dispatch(
+                setPropertyMatchLimit(property as SearchProperty, limit + 25)
```
lemald, line 96: **suggestion:** locally at least I'm able to remove the `as SearchProperty` and it still typechecks (and I prefer to avoid explicit type assertions where possible).

assets/src/hooks/useChannel.ts
```diff
@@ -143,3 +143,82 @@ export const useCheckedChannel = <T, U>({
   ])
   return state
 }
+
+export const useCheckedTwoWayChannel = <T, U>({
```
lemald, line 147: **question:** do you think it would be possible / sensible to make this a special case of `useCheckedChannel`? It seems like a user of the hook could just ignore the `pushUpdate` part of the return value if they don't need to send any data.

assets/src/hooks/useChannel.ts
```diff
@@ -143,3 +143,82 @@ export const useCheckedChannel = <T, U>({
...
+  initialMessage,
+}: {
+  socket: Socket | undefined
+  topic: string | null
+  event: string
+  initialMessage: any
+  dataStruct: Struct<T, any>
+  parser: (data: T) => U
+  loadingState: U
+}): [U, (event: string, payload: any) => void] => {
```
lemald, line 163: **suggestion (non-blocking):** instead of `payload: any`, you could specify `payload: V` where `V` is a new parametrized type on the type signature of `useCheckedTwoWayChannel` - that way if another part of the code using this hook invokes the callback, it will enforce that it has to be used with a specific type.

assets/src/hooks/useChannel.ts
```diff
@@ -143,3 +143,82 @@ export const useCheckedChannel = <T, U>({
...
+            onOk(data)
+          })
+      }
+    },
+    [channel]
+  )
+
+  useEffect(() => {
+    setState(loadingState)
+    let channel: Channel | undefined
```
lemald, line 193: **suggestion:** I think I see why you want to have a separate local definition of `channel` so that you can run through the whole process of joining and attaching handlers before calling `setChannel`, but the variable shadowing with the `channel` associated with the state is a little confusing. Is it possible to come up with some other name? Or somehow rework it so that a different variable isn't required?

assets/src/hooks/useChannel.ts
```diff
@@ -143,3 +143,82 @@ export const useCheckedChannel = <T, U>({
...
+        .receive("timeout", reload)
+    }
+
+    return () => {
+      if (channel !== undefined) {
+        channel.leave()
+        channel = undefined
+      }
+    }
+  }, [socket, topic, event, loadingState, dataStruct, parser])
```
lemald, line 222: *issue:** PSA that I'm getting a React hooks eslint warning about a missing dependency here (`onOk`). Just thought I'd mention it since for now warnings aren't triggering CI failures so it can be easy to miss unless you run `npm check` locally.

assets/src/hooks/useSearchResults.ts
```diff
@@ -33,4 +34,54 @@ const useSearchResults = (
...
+    },
+  })
+
+  useEffect(() => {
+    if (query) {
+      pushUpdate("update_search_query", {
+        limit: query.limit,
+      })
+    }
+  }, [query?.property, query?.text, query?.limit, pushUpdate])
```
lemald, line 82: **suggestion:** note that the React hooks linting rules still don't like this and instead need `query` itself to be a dependency if it's used. To prevent excessive re-renders you can take the approach from some other hooks like `useSwings` where you have a separate state that contains the query, then deep compare the query passed in as an argument to the saved query and only update the state if the deep comparison reveals a difference.

assets/css/_map_page.scss
KaylaBrady: I pulled it from the designs [here](https://www.figma.com/file/sGUlB5cVN61RnXAjUm2JV2/Skate-%7C-Maps-%7C-Geolocation-Search?node-id=25%3A1442&mode=dev) - I *think* this should link directly to the right node, but let me know if that doesn't work!

assets/src/components/mapPage/searchResultsByProperty.tsx
KaylaBrady: Thanks for catching, that is a holdover from an old approach!

assets/src/hooks/useChannel.ts
KaylaBrady: I think it is, I was hesitant to increase the scope of this PR but thinking about it more it should be a pretty minimal change.

assets/src/hooks/useSearchResults.ts
KaylaBrady: Thanks, I'll take a look at `useSwings`! There is a similar hook issue related to `initialMessage` being passed to the channel on join. Will try to do something crafty there so that the value for `initialMessage` is only updated when relevant.

assets/css/_map_page.scss
lemald: Ahh, okay, I was looking at the equivalent node in a separate part of the Figma document.

lib/realtime/server.ex
```diff
@@ -204,6 +239,19 @@ defmodule Realtime.Server do
...
 
+  def lookup({table, {:limited_search, search_params}}) do
+    logged_in_vehicles = lookup({table, :logged_in_vehicles})
+
+    vehicles_to_search =
+      if Map.get(search_params, :include_logged_out_vehicles, false) do
+        logged_in_vehicles ++ lookup({table, :logged_out_vehicles})
+      else
+        logged_in_vehicles
+      end
```
lemald, line 250: **suggestion (non-blocking):** I think we now have this same basic logic repeated a few times in this file, it might be worth splitting out into a helper function.

lib/realtime/vehicle_or_ghost.ex
```diff
@@ -3,60 +3,112 @@ defmodule Realtime.VehicleOrGhost do
 
   @type t :: Vehicle.t() | Ghost.t()
 
+  @spec take_limited_matches([t()], Server.search_params()) :: %{
+          matching_vehicles: [t()],
+          has_more_matches: boolean()
+        }
+  def take_limited_matches(vehicles, %{text: text, property: search_property, limit: limit}) do
+    search_terms = text |> clean_for_matching |> Enum.reject(&(String.length(&1) < 2))
```
lemald, line 11: **question:** this only applies to the new search flows, right? If so, shouldn't it be a minimum of 3 characters, not 2 (though the frontend will also enforce this limit in practice)?

assets/src/hooks/useChannel.ts
```diff
@@ -81,88 +81,39 @@ export const useCheckedChannel = <T, U>({
...
-  initialMessage: any
   dataStruct: Struct<T, any>
   parser: (data: T) => U
   loadingState: U
-}): [U, (event: string, payload: any) => void] => {
+  closeAfterFirstRead?: boolean
+}): [U, (event: string, payload: V) => void] => {
   const [state, setState] = useState<U>(loadingState)
-  const [channel, setChannel] = useState<Channel | undefined>()
+  const [joinedChannel, setJoinedChannel] = useState<Channel | undefined>()
```
lemald, line 116: **praise:** I like the `joinedChannel` name, I think it's nice and clear.

assets/src/hooks/useChannel.ts
```diff
@@ -81,38 +81,86 @@ export const useCheckedChannel = <T, U>({
...
+  const pushUpdate = useCallback(
+    (event: string, payload: any): void => {
+      if (joinedChannel) {
+        joinedChannel
+          .push(event, payload)
+          .receive("ok", (data: { data: unknown }) => {
+            onOk(data)
+          })
+          .receive("error", ({ reason }) => {
+            // eslint-disable-next-line no-console
```
lemald, line 139: **nitpick:** looks like this was left over from a version that logged to the console and this is no longer needed

lib/realtime/vehicle_or_ghost.ex
KaylaBrady: I think we want to keep the limit to 2 based on Mariele's comment [here](https://app.asana.com/0/0/1204650425301364/1205104040335462/f) - autocomplete has a minimum of 3 characters, but search still has a minimum of 2

lemald: Ahh okay I neglected that consideration, but good call.

assets/src/components/searchForm.tsx
```diff
@@ -215,8 +176,8 @@ const SearchFormFromStateDispatchContext = ({
         dispatch(setSearchText(""))
         onClear?.(event)
       }}
-      onFiltersChanged={(name, _) => {
-        dispatch(setSearchProperty(filterNameToSearchProperty(name)))
+      onFiltersChanged={(name, _isActive) => {
+        dispatch(setOldSearchProperty(name))
```
KaylaBrady, line 180: Will update this to dispatch a different action to update the new property list as a different PR

KaylaBrady:
@mbta/skate-developers made some updates so the search property state is more consistent with the representation used in the search form filters. With that change, this is ready for review!

lemald (approved):
One nitpick comment about capitalization of the name of a type, but otherwise this looks good to me!

assets/src/models/searchQuery.ts
```diff
@@ -1,17 +1,37 @@
-export type SearchQueryType = "all" | "run" | "vehicle" | "operator"
+export type OldSearchQueryType = "all" | "run" | "vehicle" | "operator"
+
+export const searchPropertyDisplayConfig = {
```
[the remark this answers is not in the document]
KaylaBrady: Yeah I've run into similar roadblocks in trying different types. I'm running into more of the same type assertions as I start on the separate PR to update the filtering logic - will spend some more time trying to get the types to play nice!

assets/src/components/searchForm.tsx
```diff
@@ -3,66 +3,27 @@ import { StateDispatchContext } from "../contexts/stateDispatchContext"
...
   submitSearch,
 } from "../state/searchPageState"
 
 // #region Search Filters
-/**
- * Unordered enumeration of the UI exposed controls.
- *
- * ---
- *
- * TODO: merge with {@link SearchQueryType}
```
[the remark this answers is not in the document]
firestack: I've just looked over [`fa90a12` (#2140)](https://github.com/mbta/skate/pull/2140/commits/fa90a12c2f58c2642287bd682ea40d8e8106d1db) and I think it looks good!

PR #2151, opened 2023-07-25
feat(SearchForm): Filter toggle functionality

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1205130484527139/f

### DIFF
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -4,14 +4,12 @@ import { SearchIcon } from "../helpers/icon"
 import { CircleXIcon } from "./circleXIcon"
 import { FilterAccordion } from "./filterAccordion"
 import {
-  defaultResultLimit,
   isValidSearchText,
-  PropertyLimits,
   SearchProperty,
   searchPropertyDisplayConfig,
 } from "../models/searchQuery"
 import {
-  setPropertyMatchLimits,
+  setSearchProperties,
   setSearchText,
   submitSearch,
 } from "../state/searchPageState"
@@ -235,17 +233,11 @@ const SearchFormFromStateDispatchContext = ({
         onClear?.(event)
       }}
       onFiltersChanged={(newFilters) => {
-        dispatch(
-          setPropertyMatchLimits(
-            Object.fromEntries(
-              Object.entries(newFilters).map(([property, isActive]) => [
-                property,
-                isActive ? defaultResultLimit : 0,
-              ])
-            ) as PropertyLimits
-          )
-        )
+        const newProperties = Object.entries(newFilters)
+          .filter(([_property, isActive]) => isActive)
+          .map(([property]) => property) as SearchProperty[]
 
+        dispatch(setSearchProperties(newProperties))
         dispatch(submitSearch())
       }}
     />
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -1,15 +1,17 @@
-import React, { useContext, useRef } from "react"
+import React, { useContext, useRef, useState } from "react"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import { SearchIcon } from "../helpers/icon"
 import { CircleXIcon } from "./circleXIcon"
 import { FilterAccordion } from "./filterAccordion"
 import {
+  defaultResultLimit,
   isValidSearchText,
+  PropertyLimits,
   SearchProperty,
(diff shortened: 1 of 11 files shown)


assets/src/state/searchPageState.ts
```diff
@@ -143,17 +153,23 @@ export const reducer = (
...
   switch (action.type) {
-    case "SET_SEARCH_TEXT":
+    case "SET_SEARCH_TEXT": {
       return {
         ...state,
         query: {
           ...state.query,
           text: action.payload.text,
-          properties: defaultAllProperties,
+          properties: Object.fromEntries(
```
KaylaBrady, line 162: This logic and the logic in the `setSearchPropertyLimits` limits has me wondering if it might be worth changing the representation in state to a list `{property: property, limit: limit}[]`, where the only entries in the list are ones where the filters are toggled on, and toggling off a filter would result in removing an item from the list. With the current representation, is limit doing too much by representing both the toggled state (limit === 0) and whether the search result limit has been increased?

assets/src/components/searchForm.tsx
```diff
@@ -54,7 +54,82 @@ type SearchFormProps = SearchFormEventProps & {
...
+      {Object.entries(filters)
+        .map(([property, isActive]) => ({
+          property: property as SearchProperty,
+          isActive,
+        }))
+        .sort(
+          ({ property: first_property }, { property: second_property }) =>
+            searchPropertyDisplayConfig[first_property].order -
+            searchPropertyDisplayConfig[second_property].order
+        )
```
lemald, line 101: **suggestion (non-blocking):** since we sort `SearchProperty`s by order in a few places maybe it would be worth making this a helper function.

assets/src/components/searchForm.tsx
```diff
@@ -54,7 +54,82 @@ type SearchFormProps = SearchFormEventProps & {
...
+  filters: SearchFiltersState
+  onFiltersChanged: (searchFilterState: SearchFiltersState) => void
+}) => {
+  const countActiveFilters = Object.values(filters).filter(
+    (isActive) => isActive
+  ).length
+  const [hasAManuallyActivatedFilter, setHasAManuallyActivatedFilter] =
+    useState(
+      countActiveFilters > 0 && countActiveFilters < Object.keys(filters).length
+    )
```
lemald, line 86: **nitpick (non-blocking):** not sure if this matters, but one thing I'm noticing is that because this is tied to the component state and not the global search page state, if you toggle all of the filters on, go to another mode, and then come back to search it will show the version with all of the toggles inactive again. Maybe a total non-issue but I thought I'd point it out.

assets/src/state/searchPageState.ts
lemald: Hmm, my one concern with a list representation is that it doesn't automatically enforce certain things (like having at most entry with a given `property`).

One random thought: how about instead of `limit = 0` to represent an un-toggled property we use `null` for that purpose? That makes it a bit clearer that we're not dealing with a normal property limit.

assets/tests/state/searchPageState.test.ts
```diff
@@ -148,6 +149,35 @@ describe("reducer", () => {
...
+          properties: { run: 100, vehicle: 0 },
+        }),
+      })
+
+      expect(oldState.query.properties).toEqual({
+        location: defaultResultLimit,
+        operator: defaultResultLimit,
+        vehicle: 0,
+        run: 100,
+      })
```
lemald, line 166: **question:** as I understand it, this assertion is basically testing the factory functions. Do you think it's necessary or is the second assertion the main one that matters for this test?

assets/src/state/searchPageState.ts
KaylaBrady: Good points - I'll switch to `null`!

assets/src/components/searchForm.tsx
KaylaBrady: Ooh interesting! I'll flag that for design review. I'm hoping this is OK as-is, especially knowing there are planned changes to this filter component which may or may not remove the need for this logic.

assets/tests/state/searchPageState.test.ts
KaylaBrady: True, the second assertion is the one that matters! I think alternatively I'll just add location & operator to the defined search query so it is easy to tell that those properties have limits specified and are also affected.

lemald (approved):

PR #2158, opened 2023-07-28
Groundwork for search by location

lemald:

This is getting big so to keep the review manageable and avoid too many merge conflicts down the line I'm going to open a PR for the first part of the work on this feature. It's still all behind the feature flag so we can merge whenever it's ready. Reviewing commit-by-commit is probably going to be easier than looking at it as one big change.

This PR covers the basic functionality up through making a location the selected entity, but not actually showing the selected location on that map (I have something that basically works for that but it lacks polish). Other things that need to be done are design polish of the location cards and integration with autocomplete once that's ready.

### DIFF
--- a/assets/src/components/mapPage.tsx
+++ b/assets/src/components/mapPage.tsx
@@ -69,17 +69,13 @@ const OldSearchResultList = ({
 }
 
 const SearchMode = ({
-  selectSearchResult,
+  selectVehicleResult,
 }: {
-  selectSearchResult: (result: Vehicle | Ghost | null) => void
+  selectVehicleResult: (result: Vehicle | Ghost | null) => void
 }): React.ReactElement => {
   const CurrentSearchForm = inTestGroup(TestGroups.LocationSearch)
     ? SearchFormFromStateDispatchContext
     : OldSearchForm
-
-  const CurrentSearchResults = inTestGroup(TestGroups.LocationSearch)
-    ? SearchResultsByProperty
-    : OldSearchResultList
 
   const [{ searchPageState }] = useContext(StateDispatchContext)
   return (
@@ -98,7 +94,13 @@ const SearchMode = ({
 
       <div className="c-search-display u-hideable">
         {searchPageState.isActive ? (
-          <CurrentSearchResults selectSearchResult={selectSearchResult} />
+          inTestGroup(TestGroups.LocationSearch) ? (
+            <SearchResultsByProperty
+              selectVehicleResult={selectVehicleResult}
+            />
+          ) : (
+            <OldSearchResultList selectSearchResult={selectVehicleResult} />
+          )
         ) : (
           <RecentSearches />
         )}
@@ -221,12 +223,12 @@ const Selection = ({
           vehicleId={selectedEntity.vehicleId}
           setSelection={setSelection}
         />
-      ) : (
+      ) : selectedEntity.type === SelectedEntityType.RoutePattern ? (
         <SelectedRoute
           selectedRoutePattern={selectedEntity}
           selectRoutePattern={selectRoutePattern}
         />
-      )}
+      ) : null}
     </div>
   )
 }
@@ -251,7 +253,7 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
   )
   // #endregion
 
-  const setSelection = useCallback(
+  const setVehicleSelection = useCallback(
     (selectedEntity: SelectedEntity | null) => {
       switch (selectedEntity?.type) {
         case SelectedEntityType.Vehicle:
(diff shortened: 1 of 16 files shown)


assets/src/components/mapPage.tsx
```diff
@@ -269,20 +275,35 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
...
       } else {
-        setSelection(null)
+        setVehicleSelection(null)
       }
     },
-    [setSelection]
+    [setVehicleSelection]
   )
 
+  const selectLocationResult = (
```
KaylaBrady, line 292: question (non-blocking): Is there a benefit to introducing `setLocationResult` over using the `setSelection` function? I wonder if there may be an advantage to keeping it a single function and a single `selectSearchResult` prop for `SearchMode`. That way  if we add something else like Route to search results, we wouldn't have to add another selection prop to `SearchMode`.

assets/src/components/mapPage/searchResultsByProperty.tsx
```diff
@@ -74,17 +92,85 @@ const SearchResultSection = ({
...
+      {locationSearchResults === null ? (
+        <Loading />
+      ) : locationSearchResults.length > 0 ? (
+        <>
+          <ul className="c-search-results__list">
+            {locationSearchResults
+              .slice(0, limit)
+              .map((locationSearchResult) => (
+                <li key={locationSearchResult.id}>
+                  <Card
```
KaylaBrady, line 128: suggestion (non-blocking): This may already be on your radar for polish, but maybe it would be worth breaking this location card out into it's own component

assets/src/components/mapPage/searchResultsByProperty.tsx
```diff
@@ -74,17 +92,85 @@ const SearchResultSection = ({
...
-  selectSearchResult: (result: Vehicle | Ghost | null) => void
+  selectVehicleResult: (result: Vehicle | Ghost | null) => void
+  selectLocationResult: (result: LocationSearchResult | null) => void
 }) => {
   const [{ searchPageState }, dispatch] = useContext(StateDispatchContext)
 
   return (
     <div aria-label="Grouped Search Results">
       {Object.entries(searchPageState.query.properties)
-        .filter(([property, limit]) => limit != null && property != "location")
```
KaylaBrady, line 87: 🎉

assets/src/hooks/useLocationSearchResults.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const useLocationSearchResults = (
+  text: string | null
+): LocationSearchResult[] | null => {
+  const [searchResults, setSearchResults] = useState<
+    LocationSearchResult[] | null
+  >(null)
+
+  useEffect(() => {
+    let shouldUpdate = true
```
KaylaBrady, line 13: praise: Nice work considering the race condition! I often forget this - wonder if it is needed for the vehicle result hooks?

assets/tests/components/mapPage/searchResultsByProperty.test.tsx
```diff
@@ -237,4 +318,40 @@ describe("searchResultsByProperty", () => {
       })
     ).toBeInTheDocument()
   })
+
+  test("When locations section has more matches, includes a 'Show more' button", () => {
```
KaylaBrady, line 322: suggestion (non-blocking): consider adding a test for there not being a 'Show more' button when there are no more matches

KaylaBrady (approved):
Looks good! A few minor comments

assets/src/hooks/useLocationSearchResults.ts
lemald: I think for the vehicle results the current approach is fine - `useChannel` and its derivatives call `channel.leave()` in the cleanup function of their `useEffect` hooks, so presumably once that has been called the callback added to that channel by calling `channel.on()` will no longer be invoked. With HTTP API requests it's different because the various `fetch*` functions return a promise that could resolve after the parameters of the hook have changed.

assets/src/components/mapPage.tsx
lemald: I definitely debated this. My thinking is that the existing function that I renamed `selectVehicleResult` already had vehicle-specific logic in it and having vehicles, locations, and potentially other kinds of results set by the same function would likely end up involving a lot of user-defined type guards to differentiate them, which I'm not a big fan of. That said, I see your point that as we add further search result types, this might actually end up being the more complicated approach. I'll think about it.

assets/src/components/mapPage/searchResultsByProperty.tsx
lemald: Yes, especially as I add things like the icon I suspect that it will make sense to split it out into a separate component, but I think that makes more sense to do in the followup PR once I have a better idea of how those changes are shaking out.

assets/tests/components/mapPage/searchResultsByProperty.test.tsx
lemald: Added in 4a361ecd.

assets/src/components/mapPage.tsx
```diff
@@ -69,18 +70,16 @@ const OldSearchResultList = ({
 }
 
 const SearchMode = ({
-  selectSearchResult,
+  selectVehicleResult,
+  selectLocationResult,
 }: {
-  selectSearchResult: (result: Vehicle | Ghost | null) => void
+  selectVehicleResult: (result: Vehicle | Ghost | null) => void
+  selectLocationResult: (result: LocationSearchResult | null) => void
```
firestack, line 77: suggestion(non-blocking): perhaps prefix these with `on`?

Since working on the autocomplete, I've been trying to align event names with those in html (i.e., "on" + present tense description, like `onSubmit`)

firestack: looks great, thank you!

PR #2159, opened 2023-07-28
feat(searchResultsByProperty): No results found message when no results for any property

KaylaBrady:

### DIFF
--- a/assets/src/components/mapPage/searchResultsByProperty.tsx
+++ b/assets/src/components/mapPage/searchResultsByProperty.tsx
@@ -8,7 +8,7 @@ import {
 } from "../../models/searchQuery"
 import { Vehicle, Ghost } from "../../realtime"
 import { setPropertyMatchLimit } from "../../state/searchPageState"
-import SearchResults from "../searchResults"
+import SearchResults, { NoResults } from "../searchResults"
 import React from "react"
 import { useLocationSearchResults } from "../../hooks/useLocationSearchResults"
 import { Card, CardBody } from "../card"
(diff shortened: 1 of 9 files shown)


assets/src/hooks/useSearchResultsByProperty.ts
```diff
@@ -0,0 +1,82 @@
...
+  operator: VehicleResultType
+  run: VehicleResultType
+  location: LocationResultsType
+}
+
+const useSearchResultsByProperty = (
+  socket: Socket | undefined,
+  text: string,
+  properties: PropertyLimits
+): SearchResultsByProperty => {
```
firestack, line 33: question: do you think it'd be useful to rewrite [`useAutocompleteResults`](https://github.com/mbta/skate/blob/kf/asn/search-autocomplete/assets/src/hooks/useAutocompleteResults.ts) using this function? Seems like we ended up doing the same thing heh!

firestack: No worries!
and a cleanup ticket afterward sounds good to me 👍🏻

assets/src/components/mapPage/searchResultsByProperty.tsx
```diff
@@ -116,83 +82,110 @@ const LocationSearchResultSection = ({
...
+          {results.ok.hasMoreMatches && <ShowMore property={"location"} />}
         </>
-      ) : (
-        "No results found"
       )}
     </section>
   )
 }
 
+const ShowMore = ({ property }: { property: SearchProperty }) => {
```
lemald, line 114: **praise:** thanks for pulling this out into a separate component, I should have done this originally in my PR.

assets/src/hooks/useSearchResultsByProperty.ts
lemald: One other thing to consider is that the functionality for location autocomplete vs. results will actually be different (though that can probably be handled by an extra argument to the hook specifying which one you want).

lemald (approved):

firestack (approved):

assets/src/hooks/useSearchResultsByProperty.ts
```diff
@@ -0,0 +1,82 @@
...
+  | Ok<LimitedSearchResults<LocationSearchResult>>
+  | Loading
+  | null
+
+export type SearchResultsByProperty = {
+  vehicle: VehicleResultType
+  operator: VehicleResultType
+  run: VehicleResultType
+  location: LocationResultsType
+}
```
firestack, line 27: question: Would making this generic make it easier to share this code between autocomplete and search results without making the code too complex?

maybe something along the lines of this?
```suggestion
export type SearchPropertyResults<TVehicleResult, TLocationResult> = {
  vehicle: TVehicleResult
  operator: TVehicleResult
  run: TVehicleResult
  location: TLocationResult
}

export type SearchResultsByProperty = SearchPropertyResults<VehicleResultType, LocationResultsType>
export type AutocompleteResultsByProperty = SearchPropertyResults<VehicleResultType, LocationSuggestionsResultsType>
```

KaylaBrady: Ooh I like that idea! I wonder if it might help to make a generic `SearchPropertyResult` type too so they can share loading / ok / null states? I'm not sure that autocomplete will need to keep the shape of `LimitedSearchResults` though... 

I think I'll hold off on adding generics here for now since I'm not quite seeing the full path to code sharing with autocomplete, but I support adding them as part of that work! 
```
type SearchPropertyResult<T> = Ok<LimitedSearchResults<T>> | Loading | null

export type SearchResultsByProperty<TVehicleResult, TLocationResult> = {
  vehicle: SearchPropertyResult<TVehicleResult>
  operator: SearchPropertyResult<TVehicleResult>
  run: SearchPropertyResult<TVehicleResult>
  location: SearchPropertyResult<TLocationResult>
}
```

PR #2160, opened 2023-07-28
feat(ts/components/GroupedAutocomplete): introduce autocomplete component

firestack:
This creates a new control component for displaying autocomplete results and navigating that control with keyboard & mouse and moving and listening to the document focus state.

The component is then extended to use a hook to retrieve it's results from the back-end.

The the extended component is then embedded into the search form.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1204956501632027/f

### DIFF
--- a/assets/css/_search_form.scss
+++ b/assets/css/_search_form.scss
@@ -129,7 +129,8 @@
     display: block;
   }
 
-  // -- Round out the bottom of the input bar
+  // -- Visually join the bottom of the input bar with autocomplete by squaring
+  //    off the bottom
   .c-search-form__search-input-container {
     border-bottom-right-radius: 0;
     border-bottom-left-radius: 0;
--- a/assets/css/_search_form.scss
+++ b/assets/css/_search_form.scss
@@ -2,6 +2,16 @@
   display: flex;
   flex-flow: column nowrap;
   gap: 1rem;
+
+  // Prevent intersection between accordion and autocomplete popup
+  .c-filter-accordion,
+  .c-search-form__search-control {
+    isolation: isolate;
+  }
+  .c-search-form__search-control {
+    // Stack items within search-control on top of accordion
+    z-index: 1;
+  }
 }
 
 .c-search-form__search-control {
@@ -105,3 +115,40 @@
   }
 }
 // #endregion Search Input Buttons
+
+// #region Autocomplete Control
+// Hide the autocomplete control by default
+.c-search-form__autocomplete-container {
+  display: none;
+}
+
+// When input controls have focus --
+.c-search-form__search-control[data-autocomplete-visible="true"]:focus-within {
+  // -- Show Autocomplete control
+  .c-search-form__autocomplete-container {
+    display: block;
+  }
+
+  // -- Round out the bottom of the input bar
+  .c-search-form__search-input-container {
+    border-bottom-right-radius: 0;
+    border-bottom-left-radius: 0;
+  }
+}
+
+.c-search-form__autocomplete-container {
+  // Isolate to ensure that floating on top of other elements doesn't intersect
+  // in the Z axis.
+  isolation: isolate;
+  // Move behind the input control so focus styles overlap the autocomplete
+  // control
+  position: relative;
+  z-index: -1;
+
+  &:focus-within {
+    // Attempt to put the autocomplete control on top of the input control if
+    // the autocomplete control has focus
+    z-index: 1;
+  }
+}
+// #endregion Autocomplete Control
--- a/assets/src/components/groupedAutocomplete.tsx
+++ b/assets/src/components/groupedAutocomplete.tsx
@@ -703,41 +703,35 @@ export const GroupedAutocompleteFromSearchTextResults = ({
 
   // Build groups and options from search results.
   const groups = [
-    {
-      group: {
-        title: <h2>{searchPropertyDisplayConfig.vehicle.name}</h2>,
-        options: vehicles
-          .slice(0, maxElementsPerGroup)
-          .map((v) =>
-            autocompleteOption(
-              (isVehicle(v) && v.label) || v.id,
-              onSelectVehicleOption(v)
-            )
-          ),
-      },
-    },
-    {
-      group: {
-        title: <h2>{searchPropertyDisplayConfig.operator.name}</h2>,
-        options: operators
-          .filter(isVehicle)
-          .slice(0, maxElementsPerGroup)
-          .map((v) =>
-            autocompleteOption(
-              formatOperatorNameFromVehicle(v),
-              onSelectVehicleOption(v)
-            )
-          ),
-      },
-    },
-    {
-      group: {
-        title: <h2>{searchPropertyDisplayConfig.run.name}</h2>,
-        options: runs
-          .slice(0, maxElementsPerGroup)
-          .map((v) => autocompleteOption(v.runId, onSelectVehicleOption(v))),
-      },
-    },
+    autocompleteGroup(
+      <h2>{searchPropertyDisplayConfig.vehicle.name}</h2>,
+      ...vehicles
+        .slice(0, maxElementsPerGroup)
+        .map((v) =>
+          autocompleteOption(
+            (isVehicle(v) && v.label) || v.id,
+            onSelectVehicleOption(v)
+          )
+        )
+    ),
+    autocompleteGroup(
+      <h2>{searchPropertyDisplayConfig.operator.name}</h2>,
+      ...operators
+        .filter(isVehicle)
+        .slice(0, maxElementsPerGroup)
+        .map((v) =>
+          autocompleteOption(
+            formatOperatorNameFromVehicle(v),
+            onSelectVehicleOption(v)
+          )
+        )
+    ),
+    autocompleteGroup(
+      <h2>{searchPropertyDisplayConfig.run.name}</h2>,
+      ...runs
+        .slice(0, maxElementsPerGroup)
+        .map((v) => autocompleteOption(v.runId, onSelectVehicleOption(v)))
+    ),
   ].filter(({ group: { options } }) => options.length > 0)
 
   return <GroupedAutocomplete {...props} optionGroups={groups} />
--- a/assets/src/components/groupedAutocomplete.tsx
+++ b/assets/src/components/groupedAutocomplete.tsx
@@ -410,13 +410,33 @@ export const autocompleteOption = (
   },
 })
 
-export type AutocompleteDataGroup = {
+/**
+ * Use {@link autocompleteGroup} to construct this type.
+ */
+type AutocompleteDataGroup = {
   group: {
     title: ReactNode
     options: AutocompleteOptionData[]
   }
 }
 
+/**
+ *
+ *
+ * @param title Group Heading
+ * @param options Options belonging to the group
+ */
+export function autocompleteGroup(
+  title: ReactNode,
+  ...options: AutocompleteOptionData[]
+): AutocompleteDataGroup {
+  return {
+    group: {
+      title,
+      options,
+    },
+  }
+}
 /**
  * A keyboard & mouse navigable control containing a list of list of options.
  * Provides callbacks when options are selected.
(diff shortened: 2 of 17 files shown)


assets/css/_search_form.scss
```diff
@@ -105,3 +115,40 @@
...
+}
+
+// When input controls have focus --
+.c-search-form__search-control[data-autocomplete-visible="true"]:focus-within {
+  // -- Show Autocomplete control
+  .c-search-form__autocomplete-container {
+    display: block;
+  }
+
+  // -- Round out the bottom of the input bar
```
lemald, line 132: **most nitpick-y of nitpicks (non-blocking):** isn't this actually squaring off, rather than rounding out, the bottom of the input bar?

firestack: 🤣, thank you for catching this!!!
73bdffb34b3304656a02595cafd307433828c362

firestack: another nit: should we move the border radius styling of the autocomplete into this file and make `.c-autocomplete`'s `border-radius` `4px` all around?

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+                        setCursorLocation({
+                          group: groupIndex,
+                          option: optionIndex,
+                        })
+                      e.stopPropagation()
+                    }}
+                    onClick={onSelectOption}
+                    onKeyDown={(e) => {
+                      // Fire `onSelectOption` if enter is pressed
+                      if (!["Enter"].includes(e.key)) {
```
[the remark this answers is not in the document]
firestack: there's no reason it can't be, I was just leaving it open in case we ended up needing to add more keys. I think at the time I thought it looked cleaner, but we can update it.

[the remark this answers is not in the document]
lemald: Makes sense, I guess it looked slightly weird to me but I have no strong preference in favor of changing it either (although I see you just did so in e5bed05).

assets/src/util/math.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * A helper function to help clarify code when clamping a value between a range.
+ * (Remove when tc39 implements `Math.clamp`)
+ *
+ * @returns {} {@link value} limited to the range defined by {@link min} and
+ * {@link max}
+ */
+export const clamp = (value: number, min: number, max: number): number =>
+  Math.min(Math.max(min, value), max)
```
lemald, line 9: **suggestion:** would be good to also add a test

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+  function findNearestValidLocation() {
+    // Do not snap an undefined cursor.
+    if (cursorState === undefined) {
+      return undefined
+    }
+
+    // Clamp existing cursor into [0, Array.length) in each dimension
+    const group = clamp(cursorState.group, 0, lengthsContext.length - 1)
+    const option = clamp(cursorState.option, 0, lengthsContext[group] - 1)
+    // Check that answer is correct
```
KaylaBrady, line 241: question: When could the answer not be correct?

assets/src/util/operatorFormatting.tsx
```diff
@@ -0,0 +1,21 @@
+import { Vehicle } from "../realtime"
+import { joinTruthy } from "../helpers/dom"
+
+export const formatOperatorName = (
```
lemald, line 4: **suggestion (non-blocking):** not strictly in scope for this PR, but I think this could also be used to get rid of some existing logic that does roughly the same thing in `propertiesList.tsx`.

lemald: Also, it would be good to have tests for this new module as well.

firestack: I wanted to do so, and I avoided it since this was getting big, I'll open a separate PR (or open a ticket otherwise).

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+   * The name of the control. Read to screen readers when control receives focus.
+   */
+  controlName: ReactNode
+  /**
+   * React Component to use as the heading and aria-label for the results list.
+   *
+   * This component must expose the following props
+   * - `aria-hidden`
+   * - `className`
+   * - `id`
```
KaylaBrady, line 365: question (non-blocking): Is this needed as part of the props doc? It seems to me that this is taken care of within the component as long as a string value is passed for "Heading"

assets/src/util/math.ts
firestack: I thought there was a [polyfill](https://www.npmjs.com/package/ecma-proposal-math-extensions) for this I could use but it seems like the repo for the polyfill is gone. 

I had thought I was going to use this function a lot more, hence pulling it out into it's own function, do you feel strongly on keeping this and adding tests vs getting rid of it and inlining?

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+                      e.preventDefault()
+                      e.stopPropagation()
+                      onSelectOption?.(e)
+                    }}
+                    // If this element is selected by the cursor,
+                    // set document focus to this element when mounted.
+                    ref={(selected || null) && ((r) => r?.focus())}
+                    // Allow user to tab into the first option and out of
+                    // the autocomplete control on next `Tab` key.
+                    // > Composite controls should have one tab stop
```
KaylaBrady, line 566: > Composite controls should have one tab stop

question: What is this referencing?

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+            )
+          ),
+      },
+    },
+    {
+      group: {
+        title: <h2>{searchPropertyDisplayConfig.operator.name}</h2>,
+        options: operators
+          .slice(0, maxElementsPerGroup)
+          .filter(isVehicle)
```
KaylaBrady, line 708: question: Should this filtering happen before slicing?

assets/src/components/groupedAutocomplete.tsx
firestack: I had this [article](https://dev.to/s_aitchison/psa-stop-hard-coding-heading-levels-in-your-react-components-2ekp) in mind when writing this, in the case that we did end up with "automatic" heading levels in the app, I was thinking it'd be good to callout what interface was required. Though I agree it's not needed given it's simply trying to predict the future.

assets/src/util/math.ts
lemald: No real preference one way or another.

assets/src/util/operatorFormatting.tsx
firestack: b215d090a4c82ac5219054f6a2c56a898f330e65

assets/src/components/groupedAutocomplete.tsx
firestack: Uhhh, I don't have an answer for that! It's likely that it's always correct, I just didn't trust it at the time of writing.

assets/src/util/operatorFormatting.tsx
```diff
@@ -0,0 +1,37 @@
+import { joinTruthy } from "../helpers/dom"
+import { Vehicle } from "../realtime"
+
+export const DefaultFallbackString = "Not Available"
```
lemald, line 4: **nitpick:** as a `const` I think this should follow our usual capitalization conventions (`defaultFallbackString`).

assets/src/util/operatorFormatting.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type OptionParameters = {
+  fallbackText?: string
+}
+
+export const formatOperatorName = (
+  operatorFirstName: string | null,
+  operatorLastName: string | null,
+  operatorId: string | null,
+  _options?: OptionParameters
```
lemald, line 14: **nitpick:** we usually use underscore-prefixed variables to denote arguments that aren't read (as per Elixir convention), but `_options` is referenced in the function body.

assets/src/components/groupedAutocomplete.tsx
firestack: a5b9495a6bf2bc0910e4a1633fbd70fce141f283

assets/src/util/operatorFormatting.tsx
firestack: 9c174e5575e4c5061d5be97bed41fc5ea26dfbfe

assets/src/util/operatorFormatting.tsx
firestack: I missed this while ensuring tests worked, thank you! 🙇🏻‍♀️

assets/src/util/math.ts
firestack: a55dbf86bd6ba9b874d26258a2354cf67917a592

assets/src/components/groupedAutocomplete.tsx
firestack: I was certain this was in the WCAG or something but I must've unintentionally paraphrased where I read this from because I can't find it. Will revise.

firestack: I thinking of removing this altogether and let arrow keys be the primary method of entering the control, because [WebAIM's documentation on Keyboard Accessibility](https://webaim.org/techniques/keyboard/) says that 

> Tab panel
> - `Tab` - once to navigate into the group of tabs and once to navigate out of the group of tabs
> - `↑`/`↓` or `←`/`→` - choose and activate previous/next tab.

but the autocomplete section states that arrow keys are the way to navigate into and out of the composite control.

> Autocomplete
> - Type to begin filtering
> - `↑`/`↓` - navigate to an option
> - `Enter` - select an option

But I wonder if having it be tab-able is something _our_ users would want? It's probably too early to say so maybe removing this is a better option till usability testing.

firestack: 1242d36129eb26966573a1ae2d2b2906d3bf3519

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -0,0 +1,731 @@
...
+          group: cursorState.group + 1,
+          option: 0,
+        })
+
+      if (validCursorLocation !== undefined) {
+        return validCursorLocation
+      }
+
+      // Otherwise return the current state as we've reached the end
+      onCursor?.onCursorExitEdge?.(CursorExitDirection.ExitEnd)
```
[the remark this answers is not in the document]
firestack: I couldn't think of a way to to this outside of the reducer, but I'm open to suggestions.

This was for making it possible to refocus the input element after the cursor tries to move past the last or first option.
See https://github.com/mbta/skate/blob/da2eb59366c7669c7cdef03f6c189de76d166725/assets/src/components/searchForm.tsx#L308-L309 .

lemald (approved):
Okay, at this point I don't have any more specific comments and I don't have any objections to the overall approach, so I'm going to approve. That said I think it would probably be good to also wait for @KaylaBrady to approve on this one just since this is a big one and she's been catching stuff that I hadn't thought of.

KaylaBrady (approved):
A few minor discussion threads open, but nothing blocking!

assets/src/components/groupedAutocomplete.tsx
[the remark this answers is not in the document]
firestack: Okay so I tried implementing a way I thought would work, but in the end I couldn't quite figure out how to keep the same functionality.
I've pushed up my WIP at b699eb24a8a4c4874b90fe82e4865bb5e7759f89, and I added a condition to the test to fire two keydown events in a row, instead of just one, which shows the problem I'm facing with implementing this.

Essentially, I need the cursor state to exist for one render, and I'm not sure how to go about that at this moment.
If you have any thoughts on how to approch this or where I have gone wrong I'd appreciate the insight, but right now I think I need to take a break and come back to it, and may need to leave this to be addressed in the future.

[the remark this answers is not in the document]
KaylaBrady: Hmmm yeah I see what you mean! A simple change doesn't jump out at me, I think this is fine as-is without trying to move the side-effect.

PR #2167, opened 2023-08-02
fix(VehicleOrGhost): Return matching search vehicles in descending logon time

KaylaBrady:


lemald (approved):
Ah, I could have sworn that I had the ordering correct at first when I was writing the sort function and then something prompted me to change it while writing the tests. Well, thanks for catching this, it would definitely explain the weirdness we were seeing earlier.

PR #2168, opened 2023-08-03
feat(SearchResultsByCategory): Group all vehicle results into a single category

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1205197487452691/f

This PR disconnects the search results from the filters, so all results are always shown. The results will be reconnected to the new filter as part of https://app.asana.com/0/1200180014510248/1205198429752070/f


assets/src/models/searchQuery.ts
```diff
@@ -9,15 +9,23 @@ export const searchPropertyDisplayConfig = {
 
 export type SearchProperty = keyof typeof searchPropertyDisplayConfig
 
+export type VehiclePropertyQuery = "all" | "vehicle" | "run" | "operator"
+export type SearchPropertyQuery = "all" | VehiclePropertyQuery | "location"
```
KaylaBrady, line 13: I struggled with naming here. I think we'll want 3 different types: 
* individual properties excluding all (currently `SearchProperty`)
* all possible vehicle field queries including all (`VehiclePropertyQuery`)
* all possible vehicle + location field queries (`SearchPropertyQuery`)

Maybe renaming `SearchProperty` something like `SearchFields` would convey that those are the individual fields that can be queried, and the other two can become `VehicleProperty` and `SearchProperty`? That doesn't quite feel right to me either - any alternatives come to mind?

assets/tests/state/searchPageState.test.ts
```diff
@@ -142,14 +158,32 @@ describe("reducer", () => {
...
+        query: emptySearchQueryFactory.build({
+          text: "123",
+          categoryResultLimits: { location: 10, vehicle: 20 },
+        }),
+      })
+
+      const newState = reducer(oldState, setCategoryMatchLimit("location", 15))
+      expect(newState.query.categoryResultLimits).toEqual({
+        location: 15,
+
```
lemald, line 181: **nitpick:** blank line (surprised `prettier` didn't catch this>).

lemald (approved):
Okay, one minor syntax nit but overall LGTM.

assets/tests/components/mapPage/searchResultsByCategory.test.tsx
```diff
@@ -241,26 +206,14 @@ describe("searchResultsByProperty", () => {
...
 
-  test("For sections that have more matches, includes a 'Show more' button", () => {
-    ;(useSearchResultsByProperty as jest.Mock).mockReturnValue({
-      run: {
-        ok: {
-          matches: [runMatch],
-          hasMoreMatches: false,
-        },
-      },
+  test("When there are more vehicl matches, includes a 'Show more' button which updates the vehicle result limit on click", async () => {
```
firestack, line 209:
```suggestion
  test("When there are more vehicle matches, includes a 'Show more' button which updates the vehicle result limit on click", async () => {
```

assets/src/models/searchQuery.ts
firestack: I don't have any objections, so far I don't have any ideas for how it should be named.

PR #2171, opened 2023-08-04
feat(ex/SkateWeb): add `ConfigPagesController` to redirect url's in config to external resources

firestack:
I think this is a fairly simple approach to redirecting to external url's, unsure if this is what we _should_ do, and I'm not quite sure how to test it yet (suggestions welcome).

See comments for other considerations. 

## TODO:
- [ ] Add Tests for `ConfigPagesController`

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1205074914378399/f

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -104,6 +104,12 @@ config :skate, SkateWeb.AuthManager,
   issuer: "skate",
   secret_key: nil
 
+config :skate, SkateWeb.ConfigPagesController,
+  pages: %{
+    "/agency-policies/AUP" =>
+      "https://mbta.sharepoint.com/:b:/s/CTD/ER2vUlgzH_xMuNTwKZHsvb0B80yH5XIQFLX7A4e6crycMA?e=GwAHOn"
+  }
+
 config :skate, Skate.Repo,
   database: "skate_dev",
   username: System.get_env("POSTGRES_USERNAME", System.get_env("USER")),
--- a/lib/skate_web/router.ex
+++ b/lib/skate_web/router.ex
@@ -40,6 +40,13 @@ defmodule SkateWeb.Router do
     plug :put_secure_browser_headers
   end
 
+  scope "/docs" do
+    for {key, value} <- Application.compile_env(:skate, SkateWeb.ConfigPagesController)[:pages] do
+      get key, SkateWeb.ConfigPagesController, :index,
+        private: %{SkateWeb.ConfigPagesController => %{external: value}}
+    end
+  end
+
   scope "/auth", SkateWeb do
     pipe_through([:redirect_prod_http, :accepts_html, :browser])
 
--- a/lib/skate_web/controllers/config_pages_controller.ex
+++ b/lib/skate_web/controllers/config_pages_controller.ex
@@ -1,0 +1,14 @@
+defmodule SkateWeb.ConfigPagesController do
+  @moduledoc """
+  Redirects requests to the `:external` parameter in the connection's
+  `SkateWeb.ConfigPagesController` private data.
+  """
+
+  use SkateWeb, :controller
+
+  def index(conn, _params) do
+    url = conn.private[__MODULE__].external
+
+    redirect(conn, external: url)
+  end
+end


lib/skate_web/router.ex
```diff
@@ -40,6 +40,13 @@ defmodule SkateWeb.Router do
     plug :put_secure_browser_headers
   end
 
+  scope "/docs" do
+    for {key, value} <- Application.compile_env(:skate, SkateWeb.ConfigPagesController)[:pages] do
+      get key, SkateWeb.ConfigPagesController, :index,
+        private: %{SkateWeb.ConfigPagesController => %{external: value}}
+    end
+  end
```
firestack, line 48: RFC: Unsure if we would want to write the routes out here, or do this entirely within the `ConfigPagesController`. I like this because it makes the routes statically known to Phoenix instead of needing a wildcard matcher, but it's almost feels like a footgun to have the ability to redirect anywhere based on the `external:` parameter. If there were somehow a way to enforce that the `external` parameter can only be configured at compile time, I think I'd feel better about this.

This also makes testing a little difficult IMO, since we can't swap out the routes per test.

firestack, line 48: RFC: I could try pulling this out into a Plug, so that we don't need to do the `for` loop directly within `router.ex`.

config/config.exs
```diff
@@ -104,6 +104,12 @@ config :skate, SkateWeb.AuthManager,
   issuer: "skate",
   secret_key: nil
 
+config :skate, SkateWeb.ConfigPagesController,
+  pages: %{
+    "/agency-policies/AUP" =>
+      "https://mbta.sharepoint.com/:b:/s/CTD/ER2vUlgzH_xMuNTwKZHsvb0B80yH5XIQFLX7A4e6crycMA?e=GwAHOn"
+  }
+
```
firestack, line 112: question: Is there a good way to document the expected shape of this config?

lib/skate_web/router.ex
KaylaBrady: > If there were somehow a way to enforce that the external parameter can only be configured at compile time, I think I'd feel better about this.
>
I'm not aware of a way to do this. In absence of that, I don't mind the approach of writing out the routes directly in the router since we only have one so far; perhaps as we add more routes we could consider a different approach. 

I wonder if an alternative approach to using `private` might help readability. From the `private` docs: 
> This storage is meant to be used by libraries and frameworks to avoid writing to the user storage
>
The external link doesn't quite seem to fit the need of "user storage" to me; one possible alternative to consider is calling `current_path(conn)` from within the controller and using that to read from the configuration.

firestack: > one possible alternative to consider is calling current_path(conn) from within the controller and using that to read from the configuration.

I did try that at first, but it doesn't show up in the known routes if it's done via a wildcard, which I didn't like.

We could certainly put it in the router, but I don't like that because I don't like assuming that Skate will only ever be used by the MBTA. But, since it's not used by anyone else right now, I'd be fine with that if @lemald agrees.

firestack: > The external link doesn't quite seem to fit the need of "user storage" to me

Another alternative, I could put the `:external` parameter into the `:assigns` instead.

firestack: I just found https://www.viget.com/articles/how-to-redirect-from-the-phoenix-router/ and I feel like I no longer no which way I want to approach this 😭

KaylaBrady: Neat! I tend to like seeing the app path directly in the router like that, with or without wildcards. I do prefer the way you've put the external URLs in config though, so whatever approach you take I'd love to see that aspect stay!

firestack: I just pushed up https://github.com/mbta/skate/tree/kf/asn/aup-/2
I think I like this approach much more, and would be happy to add the external mbta url into the config.

> I do prefer the way you've put the external URLs in config though

Just wanting to clarify, which part of the config approach do you like?
- the map of path -> url
or
- that the external url is in the config, and not inline in the router

KaylaBrady: > that the external url is in the config, and not inline in the router

This one! A non-blocking personal preference though

firestack:
Instead of replacing this PR with new code I'm going to close this for #2174,
Sorry for the "bait and switch"!

PR #2173, opened 2023-08-04
Location search part 2

lemald:

This should address the remaining acceptance criteria aside from the autocomplete functionality, which will follow in another PR.


assets/css/_location_card.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  .c-card__icon {
+    padding-right: 0.5rem;
+
+    svg {
+      width: 1.125rem !important;
+      height: 1.5rem !important;
+
+      path {
+        fill: #ea5382;
```
KaylaBrady, line 43: suggestion (non-blocking): Consider referencing `$color-strawberry-500`

assets/src/components/map.tsx
```diff
@@ -290,6 +292,55 @@ export const autoCenter = (
   }
 }
 
+export const drawerOffsetAutoCenter: UpdateMapFromPointsFn = (map, points) => {
```
KaylaBrady, line 295: praise: I like this rename - `drawerOffsetAutoCenter` is very clear! 
question (non-blocking): does this need to be moved out of `mapDisplay?`

assets/tests/components/mapPage.test.tsx
```diff
@@ -667,6 +675,47 @@ describe("<MapPage />", () => {
...
+        dispatch={jest.fn()}
+      >
+        <MapPage />
+      </StateDispatchProvider>
+    )
+
+    const mapSearchPanel = getMapSearchPanel()
+    expect(mapSearchPanel).toHaveClass("c-map-page__input-and-results--visible")
+
+    const locationCard = screen.getByLabelText(location.name!)
```
KaylaBrady, line 705: thought (non-blocking): maybe it would be worth defining a named region for the selected entity that could be queried `within` to help with asserting that the found card is selected and not a result. That said, I think this is test is OK as-is!

KaylaBrady (approved):
Looks great!

assets/css/map/markers/_location_dot_icon.scss
```diff
@@ -0,0 +1,17 @@
+.c-location-dot-icon {
+  svg {
+    overflow: visible;
+    width: 18px !important;
+    height: 24px !important;
```
firestack, line 5: question(non-blocking): is `width` and `height` necessary if it's specified by the leaflet divIcon?

lemald: Ah, yup, good catch. Leaflet sets the height / width of the element inline so this is redundant.

assets/css/_location_card.scss
lemald: Ah I totally missed that this was a preexisting color, thanks.

assets/src/components/map.tsx
lemald: I chose to move it out of `mapDisplay` to put it in the same place as the other `autoCenter` function, just because it seemed weird to have two functions that do very similar things in different places in the code. I'd be open to other suggestions for how to organize things, though.

assets/src/components/mapMarkers.tsx
```diff
@@ -409,3 +412,26 @@ export const GarageMarkers = ({ zoomLevel }: { zoomLevel: number }) => (
...
+  selected,
+}: {
+  location: LocationSearchResult
+  selected?: boolean
+}) => (
+  <Marker
+    position={[location.latitude, location.longitude]}
+    icon={locationLeafletIcon(selected)}
+  />
+)
```
firestack, line 437: suggestion(non-blocking): consider using `ReactMarker` here

```suggestion

export const LocationMarker = ({
  location,
  selected,
}: {
  location: LocationSearchResult
  selected?: boolean
}) => (
  <ReactMarker
    position={[location.latitude, location.longitude]}
    divIconSettings={{
      iconAnchor: new Leaflet.Point(9, 24),
      iconSize: [24, 18],
      className:
        "c-location-dot-icon" +
        (selected ? " c-location-dot-icon--selected" : ""),
    }}
    icon={<LocationDotIcon />}
  />
)
```

The issue I encountered was that when I have dev tools open and have the `locationDotIcon` svg selected in the DOM. If I pan the map, the selection in the dev tools get's erased, which is kinda annoying, but I also think it means that this `divIcon` is being regenerated every time the surounding state gets updated (such as panning updating the marker posistion in screen space).

When I tried using `ReactMarker` locally for this, I found that this didn't happen anymore, which I believe is because the dom reference for the react portal is preserved across renders.

assets/tests/components/mapPage.test.tsx
lemald: Yeah that's sort of what I was trying to do (indirectly) with the test that the Street View button is visible. I'm actually going to add an assertion that the "New Search" button is there though as a way of making sure that this is a selected result.

assets/css/map/markers/_location_dot_icon.scss
```diff
@@ -0,0 +1,15 @@
+.c-location-dot-icon {
+  svg {
+    overflow: visible;
+
+    path {
+      fill: #ea5382;
```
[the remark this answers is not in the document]
lemald: eb71f1d3

assets/src/components/highlightedMatch.tsx
```diff
@@ -0,0 +1,45 @@
...
+        </span>,
+        <HighlightedMatch
+          content={content.slice(match.index + match[0].length)}
+          highlightText={highlightText}
+          key={`highlighted-extension-${match.index + match[0].length}`}
+        />,
+      ]}
+    </>
+  )
+}
```
firestack, line 39: __question__(non-blocking): Could this be done without recursion? I was finding it a bit hard to follow at first, and I'm partially worried about JS/react having difficulty with this pattern, though less so than the readability.

lemald: It probably is doable, though for the purposes of this PR I just wanted to move the old code to a more sensible place.

firestack: Oh oops, I hadn't noticed that this was moving code and not new code.

firestack (approved):

lemald:
@firestack I actually considered making `<HighlightedMatch/>` non-recursive because I was revisiting it for some design feedback. That said, it ended up looking tough to get everything to work nicely in a `while` loop or similar. That said, in 430a417e I did at least split the actual recursive part out into a helper so that the new, more complicated regex building logic only has to happen once before the recursion is kicked off.

firestack (approved):

PR #2174, opened 2023-08-04
feat(ex/skate_web): add MBTA AUP

firestack:
Supersedes: #2171

I found this online and decided to redo #2171 given a better understanding of plugs and plug_opts.

- (main reference) https://www.viget.com/articles/how-to-redirect-from-the-phoenix-router/
	- (source forum post) https://elixirforum.com/t/phoenix-redirects-in-router-file/26733

I think this is a decent balanced approach to ensuring that the `external` parameter is specified in the router, and that we could move this into the config in the future if we need.
Additionally, the way the routes are specified in the router means that the routes to external redirects are statically known.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1205074914378399/f


test/skate_web/controllers/redirect_controller_test.exs
```diff
@@ -0,0 +1,49 @@
...
+    # Test that a route without a `:external` parameter doesn't pull `external` from query parameter
+    assert_raise Plug.Conn.WrapperError, ~R[Missing required external: option in redirect], fn ->
+      call(Router, :get, "/exceptional?external=https://bad-url")
+    end
+
+    # Test that a route with a `:external` parameter doesn't redirect to a `external` query parameter
+    conn = call(Router, :get, "/1?external=https://bad-url")
+
+    assert Phoenix.ConnTest.redirected_to(conn) == "https://test-site-1"
+  end
```
firestack, line 30: I may be overly cautious here trying to avoid a [CWE-601](https://cwe.mitre.org/data/definitions/601.html). I don't think I know enough about when phoenix calls a plug with `opts` from `init` and when the `opts` contain `params` and what those `params` contain to be able to say if this test is needed or not.

firestack: It doesn't cover all the ways that a attacker could try and sneak a redirect URL into the request (e.g., headers), but I figured this was one of the easy tests to ensure expected functionality.

lib/skate_web/controllers/redirect_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+  Reference: https://www.viget.com/articles/how-to-redirect-from-the-phoenix-router/
+  """
+  import Phoenix.Controller, only: [redirect: 2]
+
+  @spec init(Keyword.t()) :: Keyword.t()
+  def init([external: _] = opts), do: opts
+  def init(_default), do: raise("Missing required external: option in redirect")
+
+  @spec call(Plug.Conn.t(), Keyword.t()) :: Plug.Conn.t()
+  def call(conn, external: url) do
```
lemald, line 16: **suggestion:** related to your other comment about avoiding open redirects, I have a thought on another approach that might be slightly more watertight. Instead of passing a full external URL in, the `Redirect` controller could instead take a single atom argument (say `:aup` in this case). Then, this module could have a map of atoms to URLs for each of the static files that we want to be able to redirect our users to. The controller could then look up the URL from this map and throw an error if the corresponding value is null, which wouldn't allow us to redirect to any URLs not explicitly listed. Thoughts?

lib/skate_web/router.ex
```diff
@@ -40,6 +40,10 @@ defmodule SkateWeb.Router do
     plug :put_secure_browser_headers
   end
 
+  scope "/docs", SkateWeb do
+    get "/agency-policies/AUP", Redirect, external: :aup
```
lemald, line 44: **nitpick:** are Phoenix routers case-sensitive? If so, I think `aup` might be better than `AUP` so that we're not asking our users to type in a mixed-case URL.

lemald (approved):
One usability question but code looks good to me. 👍

lib/skate_web/router.ex
firestack: > are Phoenix routers case-sensitive

From what I can find, they are.


Lowercased in a80efb0d99057c856c6bf4de098a71e0ded741d2
We can revisit this some other time and add a "normalization plug" if we find that we need case insensitivity here https://elixirforum.com/t/how-to-match-uppercase-and-lowercase-route-name-in-phoenix/31037/10

PR #2177, opened 2023-08-09
Provide autocomplete-related data for location search

lemald:
Asana ticket: part of [⚙️ Better Maps | Search | Users can search by location](https://app.asana.com/0/1200180014510248/1204971421569464/f)

Most of this is pretty straightforward data plumbing to get the information we need from Amazon and then provide it to the Skate front-end in the form of hooks.

### DIFF
--- a/lib/skate_web/controllers/location_search_controller.ex
+++ b/lib/skate_web/controllers/location_search_controller.ex
@@ -3,6 +3,15 @@ defmodule SkateWeb.LocationSearchController do
 
   alias Skate.LocationSearch.AwsLocationRequest
 
+  @spec get(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def get(conn, %{"id" => id}) do
+    seach_fn = Application.get_env(:skate, :location_get_fn, &AwsLocationRequest.get/1)
+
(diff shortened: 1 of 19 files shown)


lib/skate_web/controllers/location_search_controller.ex
```diff
@@ -3,6 +3,15 @@ defmodule SkateWeb.LocationSearchController do
 
   alias Skate.LocationSearch.AwsLocationRequest
 
+  @spec get(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def get(conn, %{"id" => id}) do
+    seach_fn = Application.get_env(:skate, :location_get_fn, &AwsLocationRequest.get/1)
```
firestack, line 8: typo: would you update this and the other instances in this file that aren't in the changes?

lemald: ec888e70

I also renamed it to `get_fn` in the `get/2` function to better reflect its meaning.

firestack (commented):
I've done a first pass, and so far it looks good, just gonna mull it over a bit.

firestack (approved):

PR #2182, opened 2023-08-11
Autocomplete for location search

lemald:

This includes at least a first pass at writing some tests, but because there are changes sprinkled across various components with prop drilling between them I'm going to do a further pass of looking through and making sure we have as much coverage as possible. That said the code as-is should be ready for review (and suggestions for specific tests to add are welcome).

### DIFF
--- a/assets/src/components/groupedAutocomplete.tsx
+++ b/assets/src/components/groupedAutocomplete.tsx
@@ -20,9 +20,7 @@ import { Ghost, Vehicle } from "../realtime"
 import { clamp } from "../util/math"
 import { formatOperatorNameFromVehicle } from "../util/operatorFormatting"
 import { SocketContext } from "../contexts/socketContext"
-import { Socket } from "phoenix"
-import { Factory } from "fishery"
-import { LocationSearchSuggestion } from "../models/locationSearchSuggestion"
+import { useLocationSearchSuggestions } from "../hooks/useLocationSearchSuggestions"
 
 // #region Autocomplete Control
 // #region Cursor Reducer
@@ -666,7 +664,7 @@ interface GroupedAutocompleteFromSearchTextResultsProps
   maxElementsPerGroup?: number
 }
 
-export interface GroupedAutocompleteFromSearchTextEventProps{
+export interface GroupedAutocompleteFromSearchTextEventProps {
   /**
    * Callback when a autocomplete vehicle option is selected.
    * @param selectedOption The selected option vehicle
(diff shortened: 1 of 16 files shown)


assets/tests/components/searchForm.test.tsx
```diff
@@ -35,9 +35,13 @@ import {
   option as autocompleteOption,
 } from "../testHelpers/selectors/components/groupedAutocomplete"
 import { useAutocompleteResults } from "../../src/hooks/useAutocompleteResults"
+// import { useLocationSearchResults } from "../../src/hooks/useLocationSearchResults"
```
firestack, line 38: nit: leftover code?
```suggestion
```

assets/src/components/groupedAutocomplete.tsx
```diff
@@ -672,6 +677,9 @@ interface GroupedAutocompleteFromSearchTextResultsProps
    *   type deduction.
    */
   onSelectVehicleOption: (selectedOption: Vehicle | Ghost) => void
+
+  onSelectedLocationId: (selectedPlaceId: string) => void
+  onSelectedLocationText: (selectedLocationText: string) => void
```
firestack, line 682: suggestion: Do you have any thoughts on how to document these? I'm struggling to come up with comments that describe non-obvious parts of these callbacks.

Here's a rough approximation I came up with.
```suggestion
  /**
   * Fired when a autocomplete option with a PlaceId is selected.
   * @param selectedPlaceId Suggested PlaceId.
   */
  onSelectedLocationId: (selectedPlaceId: string) => void
  /**
   * Fired when a autocomplete option without a PlaceId is selected.
   * @param selectedLocationText Suggested Location Search Text.
   */
  onSelectedLocationText: (selectedLocationText: string) => void
```

assets/src/components/searchForm.tsx
```diff
@@ -287,6 +285,11 @@ export const SearchForm = ({
               searchText={inputText}
               fallbackOption={autocompleteOption(inputText, onSubmit)}
               onSelectVehicleOption={onSelectVehicleOption}
+              onSelectedLocationId={onSelectedLocationId}
+              onSelectedLocationText={(text) => {
+                setAutocompleteEnabled(false)
```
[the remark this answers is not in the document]
lemald: The thing is that selecting a text-only suggestion doesn't actually result in a "selection," it results in setting the search query text, so when I was first working on this you would select a text-only suggestion and it would properly set the query and perform the search but not close out the autocomplete. That said I'm not sure this is the best place to trigger this, it just happened to be convenient in terms of where the `setAutocompleteEnabled` callback was, so I'm open to suggestions.

assets/src/components/groupedAutocomplete.tsx
lemald: Yeah I like that!

assets/src/components/searchForm.tsx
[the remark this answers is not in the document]
firestack: Oh right! sorry I misread this as something else, this is good, thanks for explaining!

firestack (approved):

PR #2186, opened 2023-08-15
feat: Fix sentry source maps

KaylaBrady:
ticket: https://app.asana.com/0/1152340551558956/1205212771706510/f

This is largely based on Glides' setup of releases https://github.com/mbta/glides/pull/936. There are some differences in this approach in order to limit the scope of this PR based on Skate's existing setup, but I'll call those out with specific comments.

Note: It seems that Sentry may be moving towards "Artifact Bundles" using debug IDs (see [What are Artifact Bundles](https://docs.sentry.io/platforms/node/sourcemaps/troubleshooting_js/artifact-bundles/), debug ID as mandatory and releases as optional in [Uploading Source Maps](https://docs.sentry.io/platforms/node/sourcemaps/uploading/typescript/)). I attempted to take that approach in [kb-sentry-stack-fix](https://github.com/mbta/skate/tree/kb-sentry-stack-fix) but was not able to get it to work: `debug_meta` was not present on the[ error event JSON](https://mbtace.sentry.io/issues/4396730422/?query=is%3Aunresolved&referrer=issue-stream&stream_index=0) when it is expected based on the [troubleshooting docs](https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/#verify-your-sdk-is-up-to-date) across a few variations of attempts. In favor of fixing the problem at hand, I've opted to stop pursuing the debug ID approach in favor of the working releases approach, but we may want to revisit that approach down the line!

### DIFF
--- a/.github/workflows/deploy-base.yml
+++ b/.github/workflows/deploy-base.yml
@@ -1,0 +1,75 @@
+on:
+  workflow_call:
+    inputs:
+      env:
+        required: true
+        type: string
+    secrets:
+      AWS_ACCESS_KEY_ID:
+        required: true
+      AWS_SECRET_ACCESS_KEY:
+        required: true
+      DOCKER_REPO:
+        required: true
+      SENTRY_AUTH_TOKEN:
+        required: true
+      SENTRY_ORG:
+        required: true
+      SENTRY_PROJECT:
+        required: true
+      SLACK_WEBHOOK:
+        required: true
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.env }}
+    env:
+      ECS_CLUSTER: skate
+      ECS_SERVICE: skate-${{ inputs.env }}
+      AWS_DEFAULT_REGION: us-east-1
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get version ids
+        id: version-ids
+        run: |
+          echo "sentry-release=${{github.ref}}_${{github.sha}}" | tr / - >> "$GITHUB_OUTPUT"
+      - uses: getsentry/action-release@v1
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+        with:
+          environment: ${{ inputs.env }}
+          version: ${{steps.version-ids.outputs.sentry-release}}
+          set_commits: "skip"
+          ignore_missing: true
+      - uses: mbta/actions/build-push-ecr@v1
+        id: build-push
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          docker-repo: ${{ secrets.DOCKER_REPO }}
+          docker-additional-args: --build-arg SENTRY_RELEASE=${{steps.version-ids.outputs.sentry-release}}
+      - name: Upload static assets to S3
+        run: bash upload_assets.sh ${{ steps.build-push.outputs.docker-tag }} ${{steps.version-ids.outputs.sentry-release}}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy-base.yml
```diff
@@ -0,0 +1,75 @@
...
+          version: ${{steps.version-ids.outputs.sentry-release}}
+          set_commits: "skip"
+          ignore_missing: true
+      - uses: mbta/actions/build-push-ecr@v1
+        id: build-push
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          docker-repo: ${{ secrets.DOCKER_REPO }}
+          docker-additional-args: --build-arg SENTRY_RELEASE=${{steps.version-ids.outputs.sentry-release}}
```
KaylaBrady, line 54: Glides also adds the source maps as part of this action - see [here](https://github.com/mbta/glides/blob/ba5773d11e290be55e6bf041de7efdbed529e56a/.github/workflows/deploy.yml#L70). I opted to continue uploading the source maps  as part of the `upload_assets` script for simplicity

upload_assets.sh
```diff
@@ -27,5 +27,5 @@ aws s3 sync "$STATIC_DIR/fonts" "$S3_DIR/fonts" --size-only --exclude "*" --incl
 aws s3 sync $STATIC_DIR $S3_DIR --size-only
 
 # upload source maps to Sentry
-SENTRY_RELEASE=$(npx @sentry/cli@2.3.1 releases propose-version)
-npx @sentry/cli@2.3.1 releases files "$SENTRY_RELEASE" upload-sourcemaps "$STATIC_DIR/js"
+SENTRY_RELEASE=${2-$(npx @sentry/cli@2.3.1 releases propose-version)}
+npx @sentry/cli@2.3.1 releases files "$SENTRY_RELEASE" upload-sourcemaps "$STATIC_DIR/js" --url-prefix "~/${APP}/js"
```
KaylaBrady, line 31: The `--url-prefix` was needed here because the `abs_path` of the error points to the hosted static files, so the path of the uploaded sourcemaps needs to match the path of those hosted static files.

.github/workflows/deploy-base.yml
```diff
@@ -0,0 +1,75 @@
...
+    env:
+      ECS_CLUSTER: skate
+      ECS_SERVICE: skate-${{ inputs.env }}
+      AWS_DEFAULT_REGION: us-east-1
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get version ids
+        id: version-ids
+        run: |
+          echo "sentry-release=${{github.ref}}_${{github.sha}}" | tr / - >> "$GITHUB_OUTPUT"
```
KaylaBrady, line 37: Glides uses the tag as the sentry version when deploying from prod. Since we typically don't do deploys from tags, opting to use the ref + sha for all deploys.

.github/workflows/deploy-base.yml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: getsentry/action-release@v1
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+        with:
+          environment: ${{ inputs.env }}
+          version: ${{steps.version-ids.outputs.sentry-release}}
+          set_commits: "skip"
```
KaylaBrady, line 46: If we enable the sentry => github integration for Skate, then `set_commits` can be set to "auto". This can be done if desired as part of this PR, opted to "skip" for simplicity for now!

.github/workflows/deploy-dev-blue.yml
```diff
@@ -1,44 +1,20 @@
...
-      - uses: mbta/actions/notify-slack-deploy@v1
-        if: ${{ !cancelled() }}
-        with:
-          webhook-url: ${{ secrets.SLACK_WEBHOOK }}
-          job-status: ${{ job.status }}
+    name: Deploy to Dev-Blue
+    concurrency:
+      group: dev-blue
+      cancel-in-progress: true
+    if: ${{ github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'deploy-to-dev-blue') }}
```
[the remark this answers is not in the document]
firestack: I like this refactor, thank you!

PR #2187, opened 2023-08-16
fix(useLimitedSearchResults): Don't subscribe to topic when query is empty

KaylaBrady:
When a user lands on the search map page but hasn't performed a search yet, don't subscribe to an empty search.

### DIFF
--- a/assets/tests/hooks/useSearchResults.test.ts
+++ b/assets/tests/hooks/useSearchResults.test.ts
@@ -315,6 +315,7 @@ describe("useLimitedSearchResults", () => {
     )
     expect(result.current).toEqual(null)
   })
+
   test("initializing the hook subscribes to the search results", () => {
     const mockSocket = makeMockSocket()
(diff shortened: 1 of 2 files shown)


assets/tests/hooks/useSearchResults.test.ts
```diff
@@ -302,6 +302,19 @@ describe("useLimitedSearchResults", () => {
...
+
+    const { result } = renderHook(() =>
+      useLimitedSearchResults(mockSocket, {
+        text: "",
+        property: "run",
+        limit: 5,
+      })
+    )
+    expect(result.current).toEqual(null)
+  })
```
firestack, line 317: nit: I would prefer a newline between tests, though this is another thing that our formatter should just do >.>
```suggestion
  })

```

firestack (approved):
one small thing but otherwise looks good and thanks for fixing this!

PR #2191, opened 2023-08-17
fix(usePersistedStateReducer.test): Fix flaky test of push tab retry

KaylaBrady:
ticket: https://app.asana.com/0/1148853526253437/1204432665252080/f

I was unable to reproduce the flaky test's failure across 1000+ runs, so it has been tricky confirming if these two changes address the root of the problem, but it seems to me that both of these changes would reduce the likeliness of flaky behavior. 

1. awaiting `act`-  this allows us to remove the use of intermediate state variables & `waitFor` in the tests by waiting until all side effects are resolved. Act is still a bit mysterious to me, but I referenced [this explanation](https://github.com/threepointone/react-act-examples/blob/master/sync.md#promises) linked from[ the rtl docs](https://testing-library.com/docs/preact-testing-library/api/#act) for this approach.
2. `useRef` instead of `useState` in the `usePersistedStateReducer` so that updates to `routeTabsPushRetriesLeft` are synchronous. This prevents a possible race conditions between updating `routeTabsPushRetriesLeft` and dispatching the retry action.

### DIFF
--- a/assets/src/hooks/usePersistedStateReducer.ts
+++ b/assets/src/hooks/usePersistedStateReducer.ts
@@ -1,4 +1,4 @@
-import { useEffect, useReducer, useState } from "react"
+import { useEffect, useReducer, useRef } from "react"
 import { putRouteTabs } from "../api"
 import appData from "../appData"
 import { loadState, saveState } from "../localStorage"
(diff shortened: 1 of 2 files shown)


assets/src/hooks/usePersistedStateReducer.ts
```diff
@@ -48,32 +48,31 @@ const usePersistedStateReducer = (): [State, Dispatch] => {
 
   const { routeTabsToPush, routeTabsPushInProgress } = state
 
-  const [routeTabsPushRetriesLeft, setRouteTabsPushRetriesLeft] =
-    useState(routeTabsPushRetries)
+  const routeTabsPushRetriesLeft = useRef(routeTabsPushRetries)
```
lemald, line 51: **question:** I'm wondering if it would even be possible to simply use a local variable in `usePersistedStateReducer` rather than a `ref`, though I haven't thought it all the way through yet and my naive attempt at doing it locally isn't working. This logic is really hard to reason about and I'm hoping there's some way to simplify it a bit further but it might not be possible while fully obeying all of the rules of hooks and everything.

lemald (approved):

firestack (approved):

PR #2193, opened 2023-08-17
Move to Node 20

lemald:
Based on the [Node release schedule](https://github.com/nodejs/release#release-schedule), 20 is the latest version that will be promoted to LTS status, receiving support until 2026. This was mostly a pretty simple upgrade, though ff6bf8c2 involved running `npm install --package-lock-only` to rebuild `package-lock.json` for the new version, which results in a really big diff. In addition, this process caught a conflict in our dependencies that wasn't being flagged previously for some reason, so I upgraded to a new minor version of sass-loader to resolve the issue. I've tested everything locally and also run a successful deploy to dev-green.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 elixir 1.13.3-otp-24
 erlang 24.3.2
-nodejs 18.17.1
+nodejs 20.5.1
 adr-tools 3.0.0
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 elixir 1.13.3-otp-24
 erlang 24.3.2
-nodejs 14.18.2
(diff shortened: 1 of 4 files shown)


firestack:
Do we need to explicitly update anything in our `.github/` CI?

lemald:
> Do we need to explicitly update anything in our `.github/` CI?

Good question - `.tool-versions` should take care of that since our CI uses ASDF to install Elixir / Node. You can see how the first CI job with this new branch installed Node 20 [here](https://github.com/mbta/skate/actions/runs/5895356696/job/15990962795#step:4:63) after getting a cache miss because the hash of `.tool-versions` didn't match the old hash.

firestack (approved):

PR #2195, opened 2023-08-18
feat:fix(ts/components/routePropertiesCard): make direction selector label stretch to size of button

firestack:
Given we have planning Monday, I figured I'd get this done since it was fairly easy and to avoid digging into something deeper at the end of the week.

---

Asana Ticket: https://app.asana.com/0/1200180014510248/1204807646620475/f


lemald (approved):
I like the use of the Bootstrap `stretched-link` here rather than doing it ad-hoc!

PR #2200, opened 2023-08-22
cleanup(ts/hooks/useVehicles): turn Superstruct problems in `useVehicles` into errors

firestack:
Asana Ticket: https://app.asana.com/0/1200180014510248/1204956277851229/f

### DIFF
--- a/assets/tests/hooks/useVehicles.test.ts
+++ b/assets/tests/hooks/useVehicles.test.ts
@@ -75,7 +75,7 @@ describe("useVehicles", () => {
     })
   })
 
-  test("still returns malformed data, but logs to Sentry", async () => {
+  test("does not return malformed data, and logs to Sentry", async () => {
     const mockSocket = makeMockSocket()
     const mockChannel = makeMockChannel()
     mockSocket.channel.mockImplementationOnce(() => mockChannel)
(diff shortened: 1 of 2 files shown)


lemald:
I realize now that the ticket was sort of under-specified, but I think it would be best to handle this situation much like how `useCheckedTwoWayChannel` does, where there is still a `try` / `catch` with a `Sentry.captureException` call in the `catch` block, but the parsing and `dispatch` call to set the results only happens if the exception wasn't hit. (Sorry, should've brought this up when you were asking about testing earlier, I just hadn't put all of the pieces together)

firestack:
>I think it would be best to handle this situation much like how `useCheckedTwoWayChannel` does

Gotcha! Does [`3dce045` (#2200)](https://github.com/mbta/skate/pull/2200/commits/3dce045990b57bb182a0adf56f0fc5f39973222a) look closer to what you're envisioning?

assets/tests/hooks/useVehicles.test.ts
```diff
@@ -88,10 +88,7 @@ describe("useVehicles", () => {
 
     const { result } = renderHook(() => useVehicles(mockSocket, ["1"]))
 
-    expect(result.current).toEqual({
-      "1": [{ ...vehicle, timestamp: null }],
-    })
-
+    expect(result.current).toEqual({})
```
firestack, line 91: question: I dunno if it's worth asserting that it's equal to a empty object, but that is the default value when there is no data.

PR #2201, opened 2023-08-22
feat: Show all nearby stops

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1204906972532767/f

### DIFF
--- a/assets/src/components/mapMarkers.tsx
+++ b/assets/src/components/mapMarkers.tsx
@@ -257,17 +257,24 @@ export const StationMarker = React.memo(
   }
 )
 
-export const RouteStopMarkers = ({
+export const StopMarkers = ({
   stops,
   zoomLevel,
   direction,
   includeStopCard,
+  zoomLevelConfig = {},
 }: {
   stops: Stop[]
   zoomLevel: number
   direction?: DirectionId
   includeStopCard?: boolean
+  zoomLevelConfig?: {
+    minStopZoom?: number
+    minStationZoom?: number
+  }
 }): JSX.Element => {
+  const { minStopZoom = 17, minStationZoom = 15 } = zoomLevelConfig
+
   const seenStopIds = new Set<StopId>()
   // Keep the first occurrence of each stop when there are duplicates
   const uniqueStops: Stop[] = stops.flatMap((stop) => {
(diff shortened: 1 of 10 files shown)


assets/src/components/mapMarkers.tsx
```diff
@@ -257,72 +257,108 @@ export const StationMarker = React.memo(
   }
 )
 
-export const RouteStopMarkers = ({
+const uniqueStopsByLocation = (stops: Stop[]): Stop[] => {
+  let uniqueLocationToStop: Record<string, Stop> = {}
+  // When there are two stops with the same location, prefer the latest station
```
KaylaBrady, line 262: This is to avoid situations where a station & child stop are co-located, but the child stop was rendered on top at close zoom levels.

assets/tests/factories/stop.ts
```diff
@@ -1,15 +1,19 @@
...
-}))
+const stopFactory = Factory.define<Stop>(({ sequence }) => {
+  const coord = localGeoCoordinateFactory.build()
+  return {
+    id: `stop${sequence}`,
+    name: `Some Stop - ${sequence}`,
+    locationType: LocationType.Stop,
+    vehicleType: 3,
+    lat: coord.latitude,
+    lon: coord.longitude,
```
KaylaBrady, line 14: @firestack I really appreciate the `localGeoCoordinateFactory` you added! Using it here at the `stopFactory` seemed like the minimal change to get tests working now that we only render a single stop marker when stop markers are co-located. Interested if you have any concerns about this approach!

assets/src/components/mapMarkers.tsx
```diff
@@ -257,72 +257,113 @@ export const StationMarker = React.memo(
...
 )
 
-export const RouteStopMarkers = ({
+/**
+ * @returns a list of stops at unique locations. Where a platform stop is at the exact location of a station, the list will include the station.
+ */
+const uniqueStopsByLocation = (stops: Stop[]) => {
+  const locationToStop: Record<string, Stop> = {}
+  stops.forEach((stop) => {
+    const key = `${stop.lat}_${stop.lon}`
```
lemald, line 266: **question:** I haven't thought about this in great depth, but do you know if there's any possibility that floating point weirdness might lead to stops that should match not matching?

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -285,14 +293,23 @@ const SelectedVehicleDataLayers = ({
...
-            <RoutePatternLayers
-              routePattern={routePatternForVehicle}
-              isSelected={false}
-            />
+            <>
+              <RoutePatternLayers
+                routePattern={routePatternForVehicle}
+                isSelected={false}
+              />
+            </>
```
lemald, line 301: **nitpick:** since there's only one element here I think you can get rid of the enclosing `<>`...`</>`.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -446,20 +475,66 @@ const SelectionDataLayers = ({
...
+          return (
+            <StopMarkers
+              stops={nearbyStops}
+              zoomLevel={zoomLevel}
+              includeStopCard={true}
+            />
+          )
+        }}
+      </ZoomLevelWrapper>
+    </>
```
lemald, line 534: **nitpick:** I think you can also get rid of the `<>`...`</>` here.

assets/tests/factories/stop.ts
firestack: I looked this over and it looks good to me!
I think we can extend this in the future if we need, so this looks like a great addition 👍🏻

assets/src/components/mapMarkers.tsx
KaylaBrady: Ooh I hadn't considered that! I don't think there should be floating point issues affecting the string equality here, at least nothing jumped out at me in the [toString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference) doc. 

I also suspect if there were an issue that would cause two co-located stops to have slightly different keys, the distance between the center of those stops would be small (<3in based on [this comment](https://github.com/mbta/skate/blob/1f9716979afd3b2abfb8de96e1bec32b291ab5db/assets/tests/factories/geoCoordinate.ts#L11)). If that were to happen, I think the centers would be so close that one marker would effectively be on top of the other and only one would be visible + clickable to the user.

lemald: Sounds good to me!

lemald (approved):

PR #2204, opened 2023-08-30
feat(Skate.Migrate): Support running a subset of migrations async

KaylaBrady:
ticket: https://app.asana.com/0/1200180014510248/1205378257072522/f

References: 
* [Approach for splitting migrations into two directories](https://dashbit.co/blog/automatic-and-manual-ecto-migrations)
* [Approach for running migrations async on application startup (Arrow)
](https://github.com/mbta/arrow/blob/master/lib/arrow/repo/migrator.ex)

Note: Originally I was considering exposing admin UI to run migrations, but in looking at the approach taken in the Arrow repo, I liked the ability to still trigger the migration automatically on deploy but in a way that doesn't block application start. This still seemed fitting with the spirit of the original [infra thread](https://mbta.slack.com/archives/G01E809HUF2/p1682622136324209) that explored alternative approaches to running all queries in a way that blocks application start, and with the suggestion in that thread to consider GenServer migrations.

### DIFF
--- a/lib/skate/migrate.ex
+++ b/lib/skate/migrate.ex
@@ -1,5 +1,53 @@
 defmodule Skate.Migrate do
-  def up() do
-    Ecto.Migrator.run(Skate.Repo, :up, all: true)
+  @moduledoc """
+  GenServer which runs on startup to run Ecto migrations. All migrations stored in the "migrations" directory are run during init. Migrations stored in the "async_migrations" directory will be run after the regular migrations complete and will only log a warning on failure.
+  """
+  use GenServer, restart: :transient
+  require Logger
+
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  @impl GenServer
+  def init(opts) do
+    Logger.info("#{__MODULE__} synchronous migrations starting")
+    Keyword.get(opts, :run_sync_migrations, default_migrator_fn("migrations")).()
+
+    Logger.info("#{__MODULE__} synchronous migrations finished")
+    {:ok, opts, {:continue, :async_migrations}}
+  end
+
+  @impl GenServer
+  def handle_continue(:async_migrations, opts) do
+    Logger.info("#{__MODULE__} async migrations starting")
+
+    try do
+      Keyword.get(
+        opts,
+        :run_async_migrations,
+        default_migrator_fn("async_migrations")
+      ).()
+
+      Logger.info("#{__MODULE__} async migrations finished")
+    rescue
+      e ->
+        Logger.warn("#{__MODULE__} async migrations failed. error=#{inspect(e)}")
+        :ok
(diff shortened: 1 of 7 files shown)


lib/skate/migrate.ex
```diff
@@ -1,5 +1,53 @@
...
+      e ->
+        Logger.warn("#{__MODULE__} async migrations failed. error=#{inspect(e)}")
+        :ok
+    end
+
+    {:stop, :normal, opts}
+  end
+
+  @spec default_migrator_fn(String.t()) :: function()
+  defp default_migrator_fn(migration_directory) do
```
lemald, line 43: **suggestion (non-blocking):** I think it would be slightly clearer to call this `migrator` or `migrate` or something like that, and then call the `opts` keyword `:migrator_fn`.

lib/skate/migrate.ex
```diff
@@ -1,5 +1,53 @@
 defmodule Skate.Migrate do
-  def up() do
-    Ecto.Migrator.run(Skate.Repo, :up, all: true)
+  @moduledoc """
+  GenServer which runs on startup to run Ecto migrations. All migrations stored in the "migrations" directory are run during init. Migrations stored in the "async_migrations" directory will be run after the regular migrations complete and will only log a warning on failure.
```
firestack, line 3: nit(non-blocking): could you fold this doc comment?

lib/skate/migrate.ex
lemald: Oh, I see, this actually _returns_ a function. I think it could be refactored to not do that, though, basically:

```
Keyword.get(opts, :run_sync_migrations, default_migrator_fn("migrations")).()
```

could become

```
Keyword.get(opts, :run_sync_migrations, default_migrator_fn).("migrations")
```

so that there's only one level of function call, and the mock test functions could be updated as needed.

test/skate/migrate_test.exs
```diff
@@ -0,0 +1,119 @@
+defmodule Skate.MigrateTest do
+  use ExUnit.Case
+  import Test.Support.Helpers
+  import ExUnit.CaptureLog
+
+  describe "start_link/1" do
```
lemald, line 6: **suggestion:** for the async migrations the logging is tested in the `handle_continue/1` tests, but there are no tests for the logging for sync migrations - maybe that could be added here, or by testing `init/1` directly?

KaylaBrady: Good idea! I'll add those. For some context on why the async logging wasn't tested directly in the regular tests - there seemed to be some race conditions with `capture_log` not capturing all of the logs from the async step because `start_link` had already returned.

lemald: Ah, yup, makes sense.

lemald (approved):

PR #2205, opened 2023-08-31
Layer to show pull-backs on search map

lemald:
Asana ticket: [⚙️ Better Maps | Show pull-backs](https://app.asana.com/0/1200180014510248/1204946801340572/f)

This PR covers everything except for changes to the VPC for pull-backs to show their destination garage. I'm going to put that into a separate PR and release so that I don't have to start out with the new field being `optional` in Superstruct and then later change it to only `nullable`.

It's a minor thing but I don't think I've been entirely consistent with how I've written variables that involve "pull back" in their names. With snake_case I find that `pull_back` looks natural whereas for whatever reason I find `pullback` without capitalizing the "B" looks better in camelCase, but feel free to push back on that.

### DIFF
--- a/assets/css/map/controls/_layers_control.scss
+++ b/assets/css/map/controls/_layers_control.scss
@@ -34,7 +34,7 @@
   text-align: center;
   color: $white;
   font-size: 11px;
-  line-height: 11px;
+  line-height: 1;
   font-weight: var(--font-weight-semi);
   user-select: none;
 }
--- a/assets/css/map/controls/_layers_control.scss
+++ b/assets/css/map/controls/_layers_control.scss
@@ -71,9 +71,3 @@
     }
   }
 }
-
-// .c-layers-control__tile_layer_control {
-//     display: flex;
-//     flex-direction: column;
-//     gap: 0.5rem;
-// }
--- a/assets/css/map/controls/_layers_control.scss
+++ b/assets/css/map/controls/_layers_control.scss
@@ -21,19 +21,41 @@
   }
 }
 
-.c-layers-control__layers_list {
+.c-layers-control__content {
   position: fixed;
   right: 55px;
   border-radius: 8px;
-  .list-group {
-    margin: 0rem;
-    .list-group-item {
-      display: flex;
-      gap: 0.5rem;
-    }
-  }
+
+  background-color: $white;
+
   font-family: $font-family;
   font-size: var(--font-size-m);
 
-  background-color: $white;
+  .list-group {
+    margin: 0;
+
+    .list-group-item {
+      padding-top: 0;
+      padding-bottom: 0;
+
+      div {
+        * {
+          padding-top: 0.5rem;
+          padding-bottom: 0.5rem;
+        }
+
+        h2 {
+          margin: 0;
+          font-size: var(--font-size-m);
+          font-weight: var(--font-weight-bold);
+        }
+      }
+    }
+  }
 }
+
+// .c-layers-control__tile_layer_control {
+//     display: flex;
+//     flex-direction: column;
+//     gap: 0.5rem;
+// }
--- a/assets/css/map/controls/_layers_control.scss
+++ b/assets/css/map/controls/_layers_control.scss
@@ -21,6 +21,24 @@
   }
 }
 
+.c-layers-control__pill {
+  position: absolute;
+  top: -6px;
+  left: -6px;
+  width: 22px;
+  height: 13px;
+  border-radius: 16px;
+
+  background-color: $color-strawberry-500;
+
+  text-align: center;
+  color: $white;
+  font-size: 11px;
+  line-height: 11px;
+  font-weight: var(--font-weight-semi);
+  user-select: none;
+}
+
 .c-layers-control__content {
   position: fixed;
   right: 55px;
--- a/assets/src/components/map/controls/layersControl.tsx
+++ b/assets/src/components/map/controls/layersControl.tsx
@@ -46,41 +46,51 @@ export const LayersControl = ({
         <MapLayersIcon />
       </button>
       {showLayersList && (
-        <div className="c-layers-control__layers_list">
-          <ul className="list-group">
-            <li className="list-group-item">
-              <input
-                className="form-check-input"
-                type="radio"
-                name="tileType"
-                value=""
-                id="base"
-                checked={tileType === "base"}
-                onChange={() => setTileType("base")}
-              />
-              <label className="form-check-label" htmlFor="base">
-                Map (default)
-              </label>
-            </li>
-            <li className="list-group-item">
-              <input
-                className="form-check-input"
-                type="radio"
-                name="tileType"
-                value=""
-                id="satellite"
-                checked={tileType === "satellite"}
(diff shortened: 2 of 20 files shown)


assets/src/components/map/controls/layersControl.tsx
```diff
@@ -45,42 +52,146 @@ export const LayersControl = ({
       >
         <MapLayersIcon />
       </button>
+
+      {/* For now there's only one vehicle layer, in the future this might need to be an actual count */}
```
KaylaBrady, line 56: 👍

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -397,14 +406,14 @@ const SelectedLocationDataLayer = ({
   )
 }
 
-const DataLayers = ({
+const SelectionLayers = ({
   selectedEntity,
-  setSelection,
+  selectVehicle,
```
KaylaBrady, line 411: praise: I appreciate this refactor to use `setVehicle`!

assets/tests/components/map/controls/layersControl.test.tsx
```diff
@@ -63,4 +74,97 @@ describe("LayersControl", () => {
...
+      }
+    )
+    await userEvent.click(layersControlButton.get())
+    pullbacksSwitch.get().focus()
+    await userEvent.keyboard("{Enter}")
+
+    expect(mockToggle).toHaveBeenCalledTimes(1)
+  })
+
+  test("when pull-back layer is enabled, shows pill", async () => {
```
KaylaBrady, line 153: suggestion (non-blocking): Maybe another test that the pill is hidden when pull-back layer is not enabled?

assets/tests/components/mapPage/mapDisplay.test.tsx
```diff
@@ -611,6 +630,115 @@ describe("<MapDisplay />", () => {
...
+        await userEvent.click(
+          screen.getByRole("switch", { name: "Show pull-backs" })
+        )
+
+        expect(screen.getAllByRole("button", { name: "PULL-B" })).toHaveLength(
+          1
+        )
+      })
+
+      test("if a pulling back vehicle is selected, doesn't render it again in pull-backs layer", async () => {
```
KaylaBrady, line 665: praise: nice edge case consideration! I hadn't thought about this.

test/realtime/vehicle_test.exs
```diff
@@ -579,14 +581,21 @@ defmodule Realtime.VehicleTest do
       assert Vehicle.end_of_trip_type(block, last_trip_of_block, "run2", "middle") == :pull_back
     end
 
+    test "when finished with the last trip of the block but still logged in, returns :pull_back",
```
KaylaBrady, line 584: question: This test name specifies "still logged in", but the change to `end_of_trip_type` doesn't seem to consider whether or not a vehicle is logged in. Should it?

KaylaBrady (commented):
Looks great! Only blocking comment is question about the logic changes to `end_of_trip_type`.

test/realtime/vehicle_test.exs
lemald: It sort of implicitly does because `end_of_trip_type/4` defaults to `:another_trip` when `run_id` is null, so the fact that this test sets a non-null run ID means that the vehicle must be logged in.

assets/css/map/controls/_layers_control.scss
```diff
@@ -21,19 +21,53 @@
...
+  width: 22px;
+  height: 13px;
+  border-radius: 16px;
+
+  background-color: $color-strawberry-500;
+
+  text-align: center;
+  color: $white;
+  font-size: 11px;
+  line-height: 11px;
```
firestack, line 37: suggestion: The line height is the same as the font size, so we should use unitless values

```suggestion
  line-height: 1;
```

> The used value is this unitless `<number>` multiplied by the element's own font size. The computed value is the same as the specified `<number>`. In most cases, this is the ___preferred___ way to set line-height and avoid unexpected results due to inheritance.
>
> -- https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#number

assets/css/map/controls/_layers_control.scss
```diff
@@ -21,19 +21,53 @@
...
+  left: -6px;
+  width: 22px;
+  height: 13px;
+  border-radius: 16px;
+
+  background-color: $color-strawberry-500;
+
+  text-align: center;
+  color: $white;
+  font-size: 11px;
```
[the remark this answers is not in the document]
firestack: I'd be fine with pushing this off for a while, if we can get a ticket for potentially reworking this (and other places?) in Skate. I think this is something our design system should take care of for us honestly, so something to keep in mind for that discussion?

test/realtime/server_test.exs
```diff
@@ -568,17 +605,21 @@ defmodule Realtime.ServerTest do
     end
 
     test "returns empty data when the route is not found", %{ets: ets} do
-      assert Server.lookup({ets, {:route_id, "2"}}) == []
+      assert Server.lookup({ets, {:route_id, "3"}}) == []
```
firestack, line 608: suggestion(non-blocking): can we use -1 or something here?
```suggestion
      assert Server.lookup({ets, {:route_id, "-1"}}) == []
```

assets/css/map/controls/_layers_control.scss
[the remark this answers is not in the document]
lemald: Good idea, here's a ticket: https://app.asana.com/0/1148853526253420/1205411476253505/f

test/realtime/server_test.exs
lemald: Do you just mean that the `"-1"` would better denote "non-existent route ID" to someone reading the code?

firestack: Yes, and I think the change here is diff and review overhead that can be avoided. The only reason this was changed was because something else took 2's spot, so ideally this wouldn't need to be changed unless something pertaining to the specific test and the implementation changed.

firestack: Actually maybe it would be better to be something like this?
```suggestion
      assert Server.lookup({ets, {:route_id, "non-existent-1 "}}) == []
```

assets/tests/hooks/usePullbackVehicles.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // Needs to be kicked to do the effects again after the socket initializes
+    rerender()
+
+    expect(mockSocket.channel).toHaveBeenCalledTimes(1)
+    expect(mockSocket.channel).toHaveBeenCalledWith("vehicles:pull_backs:all")
+    expect(mockChannel.join).toHaveBeenCalledTimes(1)
+  })
+
+  test("does not subscribe to the pull-backs channel when the subscribe argument is flase", () => {
```
firestack, line 36:
```suggestion
  test("does not subscribe to the pull-backs channel when the subscribe argument is false", () => {
```

firestack (approved):

PR #2206, opened 2023-09-01
Revert "feat(Skate.Migrate): Support running a subset of migrations async"

KaylaBrady:
Reverts mbta/skate#2204- in case desired for hotfix


priv/repo/async_migrations/20230830202227_first_async_migration.exs
firestack, line 1: Do you think there'll be any issues from ecto with this being reverted? I'm not sure if the migration state table will complain about this becoming missing.

KaylaBrady: Great catch!!! I'll leave this in

KaylaBrady: okay now that I look at it that is going to be more of a pain to change + un-revert later. I'm going to skip this one since there isn't user-facing changes related to it. 

Merged the revert of nearby stops for confidence that the fix can go out as isolated as possible.

firestack: This doesn't seem to cause an issue, but it doesn't resolve the train vehicles issue, so I don't think we should revert this just yet.

PR #2207, opened 2023-09-01
Revert "feat: Show all nearby stops"

KaylaBrady:
Reverts mbta/skate#2201 in case needed for hotfix


firestack (approved):
This is all frontend components, so I think it's unrelated to the current production issue, if we do end up needing this, I think it'd be fine to revert.

PR #2210, opened 2023-09-07
fix(ex/Api.Stream): handle unknown message types

firestack:
I'm not sold on the testing method, and would like input on if I should consider separating it out into it's own test.

The annoying part about that is it'd basically look the same as the test does now.

---

Asana Ticket: https://app.asana.com/0/1148853526253420/1205367064394200/f


lemald (approved):
I think the test you've got is fine, I would also sort of like something a bit more explicit but I think this is a reasonable approach given the tools at hand.

PR #2211, opened 2023-09-08
feat: handle new tm_trip_id field

lemald:
Asana ticket: background work related to [⚙️ Better Maps | Show pull-backs](https://app.asana.com/0/1200180014510248/1204946801340572/f)

Transit Data is going to no longer using these trip IDs from Busloc in the Concentrate feed, but we still want them. For now we'll check for the new field but fall back to the old one if present, so that we don't need to do deploys simultaneously.

### DIFF
--- a/test/concentrate/parser/gtfs_realtime_enhanced_test.exs
+++ b/test/concentrate/parser/gtfs_realtime_enhanced_test.exs
@@ -90,48 +90,7 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhancedTest do
     end
 
     test "decodes a VehiclePosition JSON map" do
-      input = %{
-        "block_id" => "Q238-135",
-        "capacity" => 18,
-        "congestion_level" => nil,
-        "current_status" => "STOPPED_AT",
-        "current_stop_sequence" => 670,
-        "load" => 12,
-        "location_source" => "samsara",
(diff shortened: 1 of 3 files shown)


test/concentrate/parser/gtfs_realtime_enhanced_test.exs
```diff
@@ -178,6 +178,57 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhancedTest do
...
+          "start_time" => nil,
+          "trip_id" => "37165437-X"
+        },
+        "vehicle" => %{
+          "id" => "G-10098",
+          "label" => "3823-3605",
+          "license_plate" => nil
+        },
+        "revenue" => false
+      }
```
firestack, line 224: Do any factories need to be updated with this? I'm also wondering if this can be condensed via a factory, as it seems like a lot of noise for what fields we're actually testing?

lemald: We have a factory for `%Realtime.Vehicle{}` but the fields on that struct are staying the same, it's just the data input that's changing and we don't have a factory for that. That said I'll take a quick stab at adding such a factory to make this test more concise.

firestack (approved):
Thanks for the factory!

PR #2212, opened 2023-09-11
feat(ex/skate/oban): implement notificaiton table cleanup job

firestack:
This adds Oban, and creates and tests the Oban job. 

Next steps after this is scheduling the job on a time basis and ensuring the job only runs on one node. But that should be able to be independent of the actual implementation proposed here.

---

Asana Ticket: https://app.asana.com/0/1148853526253420/1204432817126652/f


firestack:
I dunno how to test the cron configuration but otherwise I think all that is left is to ensure only the "leader" runs the jobs!

lib/skate/oban/clean_up_notifications.ex
```diff
@@ -0,0 +1,46 @@
...
+    oldest_date = DateTime.utc_now() |> DateTime.add(-cutoff_days * @seconds_per_day)
+
+    Logger.notice("#{__MODULE__} starting cleanup")
+
+    {time, {count, nil}} =
+      :timer.tc(fn ->
+        from(notification_indexed in Notifications.Db.Notification,
+          where:
+            notification_indexed.id in subquery(
+              from(notification_limit in Notifications.Db.Notification,
```
lemald, line 30: **nitpick:** I think calling this `notification_limited` would be a bit clearer (since it's the notifications that have been limited, not the limit of how many notifications to process), and would also match with `notification_indexed`.

lemald (approved):
One minor variable naming nitpick but otherwise LGTM.

lemald:
> Next steps after this is scheduling the job on a time basis and ensuring the job only runs on one node. But that should be able to be independent of the actual implementation proposed here.

Actually, one question - does the Oban cron functionality not take care of this for us? Or do you just mean that we still need to confirm that everything works when running multiple instances?

firestack:
> > Next steps after this is scheduling the job on a time basis and ensuring the job only runs on one node. But that should be able to be independent of the actual implementation proposed here.
> 
> Actually, one question - does the Oban cron functionality not take care of this for us? Or do you just mean that we still need to confirm that everything works when running multiple instances?

As it's implemented right now, I'm pretty sure it would run the job on all nodes.
> [Queue limits are local (per-node), not global (per-cluster). For example, running a queue with a local limit of one on three separate nodes is effectively a global limit of three. If you require a global limit you must restrict the number of nodes running a particular queue.](https://hexdocs.pm/oban/Oban.html#module-configuring-queues)

firestack:
From https://hexdocs.pm/oban/Oban.Peer.html
> Leadership is used by plugins, primarily, to prevent duplicate work across nodes. For example, only the leader's Cron plugin will try inserting new jobs. You can use peer leadership to extend Oban with custom plugins, or even within your own application.
> [...]
> - Without leadership, global plugins (Cron, Lifeline, Stager, etc.), will not run on any node.

So the cron plugin just wouldn't work without the Peer config. But this says nothing about restricting a job to run on a single node, just that the Cron plugin will run on a single node.

lib/skate/oban/clean_up_notifications.ex
firestack: bafaeccf47ea9b54665b0f598d792ea6f928ffb6

firestack:
> But this says nothing about restricting a job to run on a single node, just that the Cron plugin will run on a single node.

Seems like this is one of Oban's "guarantees", that a job will only run on a single node. I've configured the uniqueness check so that it will only add another job if the previous one has completed, so I think this PR is ready to merge if you're alright with that @lemald.

I've tested this locally with two nodes and a job never ran on two nodes at once, the job would be picked up by either node, but it'd never run the same job twice, which I believe is our requirement.

config/config.exs
```diff
@@ -119,6 +119,20 @@ config :skate, Skate.WarmUp,
...
+config :skate, Oban,
+  repo: Skate.Repo,
+  peer: Oban.Peers.Postgres,
+  queues: [default: 10],
+  plugins: [
+    Oban.Plugins.Pruner,
+    {
+      Oban.Plugins.Cron,
+      crontab: [
+        {"*/5 * * * *", Skate.Oban.CleanUpNotifications, args: %{"cutoff_days" => 100}}
```
lemald, line 131: **suggestion:** I think ultimately we do want the cutoff to be something closer to 10 days than 100 in practice given the issues we had had previously with adding the index, but I'm fine with starting at 100 and reducing it later.

lemald (approved):

config/config.exs
firestack: I dunno where I got 100 in my head, so I'll to change this back to 10.

Do you think we should wait to bump the limit up from 100 as well?

lemald: I think it's probably fine either way. The only advantage to having it 100 initially would be putting a more restrictive bound on how much data the job will initially churn through, but if there's some issue we can always just revert the changes rather than wait.

PR #2213, opened 2023-09-11
init: add codeowners file

firestack:
This sets @mbta/skate-developers as the reviewers by default and should request review once a PR is out of draft.


firestack:
@lemald pointed out that a reason _not_ to do this (yet) is that we'll get assigned as reviewers for all dependabot updates. I dunno if we can override this by setting reviewers to `[]` (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers), but likely not.

firestack:
> @lemald pointed out that a reason _not_ to do this (yet) is that we'll get assigned as reviewers for all dependabot updates. I dunno if we can override this by setting reviewers to `[]` ([docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers)), but I expect that will not work.

@lemald it does seem that we could have the codeowners ignore the files dependabot touches? 
- https://stackoverflow.com/questions/75531641/exclude-codeowners-from-dependabot-pr
- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# The Skate team is the default owner for the entire codebase.
+*       @mbta/skate-developers
```
firestack, line 2:
```suggestion
*       @mbta/skate-developers

mix.lock
assets/package-lock.json
```

firestack:
Alternatively, maybe getting bugged for reviews by @dependabot would make us merge more dep updates?

CODEOWNERS
```diff
@@ -0,0 +1,6 @@
...
+
+# Lock-files managed by Dependabot do not automatically add reviewers
+mix.lock
+assets/package-lock.json
```
lemald, line 6: **suggestion:** it's not as common, but dependabot does sometimes change `mix.exs` or `package.json` - see #2154 for one recent example - so we should probably include those as well.

Of course this approach in general does have the downside that if a human opens a PR that only changes / adds / removes a dependency they will have to manually request review, but I think that's acceptable.

lemald:
> Alternatively, maybe getting bugged for reviews by @dependabot would make us merge more dep updates?

I'm sort of torn on this - on the one hand I like keeping deps up-to-date, on the other hand I worry that sometimes there's a bigger volume of them than we can necessarily pay attention to (aside from obvious high-priority things like security updates) and it would crowd out other things while looking at your assigned reviews.

CODEOWNERS
firestack: [`5c8c8eb` (#2213)](https://github.com/mbta/skate/pull/2213/commits/5c8c8eb05d808761dbaaf0d281e312c440acb9cb)

lemald (approved):

PR #2219, opened 2023-09-13
feat: async migration for notifications_users user_id index

lemald:
Asana ticket: [⚙️ Add index to notification_users table](https://app.asana.com/0/1148853526253420/1204432817126661/f)

As of this writing I'm not 100% committed to this approach, but I want to put up a PR to have a space for discussion / comments if necessary.


priv/repo/async_migrations/20230913170707_add_notifications_users_user_id_index.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Skate.Repo.Migrations.AddNotificationsUsersUserIdIndex do
+  use Ecto.Migration
+
+  def change do
+    create index("notifications_users", [:user_id])
+  end
```
firestack, line 6: Since we don't have automated tests for this (and we probably should), does this migration successfully rollback without a down/0?

firestack (approved):
one open question but besides that :shipit:

priv/repo/async_migrations/20230913170707_add_notifications_users_user_id_index.exs
lemald: It does:

```
❯ mix ecto.rollback --migrations-path priv/repo/async_migrations

09:38:52.173 [info]  == Running 20230913170707 Skate.Repo.Migrations.AddNotificationsUsersUserIdIndex.change/0 backward

09:38:52.176 [info]  drop index notifications_users_user_id_index

09:38:52.177 [info]  == Migrated 20230913170707 in 0.0s
```

Usually Ecto is pretty smart with automatically rolling back a simple `create` but I agree it would be nice to have more tests. Of course the way things are set up we don't ever really roll back migrations since it's quicker and easier to just make a new migration to undo the old one rather than customizing the migration runner code, but that's another can of worms.

PR #2220, opened 2023-09-14
init(ts): add storybookjs

firestack:
---

Asana Ticket: https://app.asana.com/0/1148853526253420/1205339217615967/f


assets/package.json
```diff
@@ -20,12 +20,15 @@
...
-    "cypress:open": "cypress open"
+    "cypress:open": "cypress open",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build"
   },
   "dependencies": {
     "@react-hook/media-query": "^1.1.1",
     "@rehooks/component-size": "^1.0.3",
     "@sentry/react": "^7.15.0",
+    "@storybook/preset-scss": "^1.0.3",
```
lemald, line 31: **question:** can this go in `devDependencies` instead?

firestack: Thanks for catching this! 7518ce2c2b906bb25b46869fef1ee0da101adbb3

assets/src/components/streetViewButton.tsx
```diff
@@ -16,16 +16,16 @@ export interface StreetViewButtonProps extends GeographicCoordinateBearing {
...
 
-const StreetViewButton = ({
+export const StreetViewButton = ({
   className,
   text,
   children,
   title,
   ...worldPosition
-}: StreetViewButtonProps): ReactElement<HTMLElement> => {
+}: StreetViewButtonProps) => {
```
lemald, line 28: **question:** is it possible to keep an explicit return type? Not that we've been super consistent about what we use for return types on components.

assets/src/helpers/icon.tsx
```diff
@@ -1,132 +1,132 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 // @ts-ignore
-import alertIconSvg from "../../static/images/icon-alert.svg"
+import alertIconSvg from "../../static/images/icon-alert.svg?raw"
```
lemald, line 3: **question:** is the `?raw` a Storybook thing? Have you tested out how this plays along with the prod assets bundle, not just local dev environment?

assets/src/components/streetViewButton.tsx
firestack: I encountered a storybook github issue where one of the solutions was to remove the return type from the function component, I'm not sure it was actually needed here any more, but I think it's better to leave the return type unspecified in the case of react.

I don't think this is the right return type to use either, so I'd be open to changing it given storybook still works.

lemald: That's fair - I'll confess that I don't really know which of the many `React` and `JSX` types is strictly correct for different use cases.

assets/src/helpers/icon.tsx
firestack: There's a small amount of context in [`0c33423` (#2220)](https://github.com/mbta/skate/pull/2220/commits/0c334236cb4c79589ce2b087286c06c48fa5f6b3)

I'm not sure if this is a storybook thing or a webpack thing, but the `module.rules` that story book uses in it's webpack config has the following defintion
```
  { resourceQuery: /raw/, type: 'asset/source' },
```

This uses https://webpack.js.org/guides/asset-modules/ to import the source instead of `svg-inline-loader`(since it's deprecated and I couldn't get storybook to work with it for some reason).

I've not deployed it yet though, so I'll do so and see if anything is broken.

firestack: I'll spend some time on this and see if I can get this working without this resource url change, and see if it's useful to get storybook to use our webpack config (or if it even works).

firestack: SVG's do appear to be working in dev-green after deployment

lemald (approved):
Looks good to me!

assets/src/helpers/icon.tsx
firestack: Okay I got it working by adding a rule to use asset-modules for svg's to the storybook webpack config. So I'll revert this commit.

What I'm still not sure of is if we should merge our webpack config with storybook.
We _probably_ should, so that our storybook mimics what the application build would look like in prod, but I'm worried about unseen interactions between our config and storybook's "blessed" config.

It would also require a (AFAIK manual) deep merge of configuration which, isn't great.

lemald: > What I'm still not sure of is if we should merge our webpack config with storybook.
We probably should, so that our storybook mimics what the application build would look like in prod, but I'm worried about unseen interactions between our config and storybook's "blessed" config.

My gut instinct is that the closer the configs are to each other the better, but I don't understand the Storybook side well enough to have thoughts on that part.

firestack: If it's compatible, I'd like to convert our webpack.config.js to webpack.config.ts, which would allow us potentially a easier time merging these two configs. (seems like there are some options in storybook that we need to not override for builds to work correctly, like the output file and path)

I'm going to open a ticket to review this in the future instead of gateing this PR on it, since that would involve changes to our webpack config unrelated to storybook.

firestack:
Merges #2225

PR #2222, opened 2023-09-14
cleanup(ex/skate/database): remove users without email

lemald:
Asana ticket: [⚙️ Remove user records without email address](https://app.asana.com/0/1148853526253420/1203392965628668/f)

This is just a `cherry-pick` of the changes from #2025 after resolving some merge conflicts.

PR #2223, opened 2023-09-14
User email non-null constraint

lemald:
Asana ticket: [⚙️ Remove user records without email address](https://app.asana.com/0/1148853526253420/1203392965628668/f)

This should be merged and deployed promptly after #2222.


lib/skate/settings/user.ex
```diff
@@ -37,7 +37,7 @@ defmodule Skate.Settings.User do
   @doc """
   Update the user with the given email if one exists, otherwise insert a new one.
```
firestack, line 38: suggestion(non-blocking): Hmm, I'm wondering if we should update this comment?
```suggestion
  Update or create the user with the given email.
```

firestack (approved):

lib/skate/settings/user.ex
lemald: Good point, that's a more accurate explanation now.

PR #2224, opened 2023-09-14
build: improve means of extracting static assets from Docker image

lemald:
Asana ticket: [⚙️ Fix intermittent deploy failures on "Upload static assets to S3" step](https://app.asana.com/0/1148853526253420/1205398036394618/f)

### DIFF
--- a/upload_assets.sh
+++ b/upload_assets.sh
@@ -12,7 +12,7 @@ STATIC_DIR=$TEMP_DIR/static
 pushd "$TEMP_DIR" > /dev/null
 
 CONTAINER_ID=$(docker create $BUILD_TAG)
-docker cp $CONTAINER_ID:/home/skate/priv/static - | tar -x
+docker cp $CONTAINER_ID:/home/skate/priv/static $TEMP_DIR
 docker rm -v $CONTAINER_ID
 
 popd> /dev/null
--- a/upload_assets.sh
+++ b/upload_assets.sh
@@ -7,10 +7,14 @@ CACHE_CONTROL="public,max-age=31536000"
 S3_DIR=s3://mbta-dotcom/$APP
 BUILD_TAG=${1}
 TEMP_DIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


upload_assets.sh
firestack, line 1: Does this seem to have helped the deploy process flakeyness at all?

upload_assets.sh
```diff
@@ -7,10 +7,14 @@ CACHE_CONTROL="public,max-age=31536000"
...
 BUILD_TAG=${1}
 TEMP_DIR=$(mktemp -d)
-STATIC_DIR=$TEMP_DIR/priv/static
+STATIC_DIR=$TEMP_DIR/static
 
 pushd "$TEMP_DIR" > /dev/null
-sh -c "docker run --rm ${BUILD_TAG} tar -c /home/skate/priv/static" | tar -x --strip-components 2
+
+CONTAINER_ID=$(docker create $BUILD_TAG)
+docker cp $CONTAINER_ID:/home/skate/priv/static - | tar -x
```
firestack, line 15: **question**: is there a benefit to using a pipe here?

would the following work?
```suggestion
docker cp $CONTAINER_ID:/home/skate/priv/static $STATIC_DIR
```

lemald: Oh, good point (though because of the way it handles path stuff I think `$TEMP_DIR` is actually the correct destination since the `/static` part of the path will be preserved).

upload_assets.sh
lemald: I haven't actually tested it enough times that I have a good sense, but I'll try it out a bunch before merging.

upload_assets.sh
lemald: Oh huh I guess I ran into a race condition where I didn't see your edit until I submitted my own comment, but I think your edit is correct.

firestack (approved):

lemald:
I have now performed five deploys to dev-green without incident, so I think it's highly likely that this resolves the problem.

PR #2230, opened 2023-09-22
User location dot

lemald:
Asana ticket: [Implement location dot](https://app.asana.com/0/1148853526253420/1205411312261325/f)

Screenshot 2023-09-22 at 11 22 24

PR #2233, opened 2023-09-25
feat: create user location button and control

firestack:
This adds the new icon and button/control for showing the user location. Integration between the button and the dot( #2230 ) will come in a later PR.

There's one thing missing, which is that I want to display this component on a leaflet map in the storybook to get a better sense of how it'll actually look on the map, but storybook's layout parameter config doesn't do quite what it says (at least until the next major version update, 7->8). I may have a workaround but it's possible it may need to come in a later PR.

---

Asana Ticket: [⚙️ Show user location on Shuttle Map > Implement button](https://app.asana.com/0/1203014709808707/1205336270685222)

---
  - https://app.asana.com/0/0/1205411312261324


assets/css/map/controls/_user_location_control.scss
```diff
@@ -0,0 +1,34 @@
...
+    width: 100%;
+    height: 100%;
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  &__svg {
+    svg {
```
lemald, line 22: **suggestion (non-blocking):** not sure how you feel about this, but instead of nesting `svg` inside of `&__svg` you could also just write one selector (`&__svg svg`) given that the outer rule / selector doesn't contain anything aside from the inner one.

firestack: [`9660272` (#2233)](https://github.com/mbta/skate/pull/2233/commits/96602722816583c57a5510ceb90cd182ec0940e1)

lemald (approved):

PR #2234, opened 2023-09-26
feat(ts/components/ShuttleMapPage): add `UserLocationControl` and `UserLocationMarker`

firestack:
---
  - https://app.asana.com/0/0/1205336270685222

### DIFF
--- a/assets/src/components/shuttleMapPage.tsx
+++ b/assets/src/components/shuttleMapPage.tsx
@@ -142,19 +142,20 @@ export const ShuttleMap = ({
     setFollowActive(controller !== false ? true : false)
   }
 
-  const [location, setLocation] = useState<null | GeolocationCoordinates>(null)
+  const [userLocation, setUserLocation] =
+    useState<null | GeolocationCoordinates>(null)
 
   const followPositions = useMemo(
     () =>
       followerController
         ? {
-            "user-location": location
-              ? [latLng(location.latitude, location.longitude)]
+            "user-location": userLocation
+              ? [latLng(userLocation.latitude, userLocation.longitude)]
               : [],
             "vehicle-location": positions,
           }[followerController]
         : [],
-    [followerController, location, positions]
+    [followerController, userLocation, positions]
   )
 
   useEffect(() => {
@@ -203,11 +204,11 @@ export const ShuttleMap = ({
         />
 
         <>
-          {(followerController === "user-location" || location) && (
-            <UserLocation onLocationUpdate={(l) => setLocation(l)} />
+          {(followerController === "user-location" || userLocation) && (
+            <UserLocation onLocationUpdate={(l) => setUserLocation(l)} />
           )}
         </>
-        <>{location && <UserLocationMarker location={location} />}</>
+        <>{userLocation && <UserLocationMarker location={userLocation} />}</>
 
         <InterruptibleFollower
           {...state}
--- a/assets/src/components/shuttleMapPage.tsx
+++ b/assets/src/components/shuttleMapPage.tsx
@@ -128,7 +128,7 @@ export const ShuttleMap = ({
   tileType,
   setTileType,
 }: ShuttleMapProps) => {
-  const positions = selectedShuttles.map(vehicleToLeafletLatLng)
+  const vehiclePositions = selectedShuttles.map(vehicleToLeafletLatLng)
 
   type FollowerController = false | "user-location" | "vehicle-location"
   const [followerController, setFollowerController] =
@@ -152,10 +152,10 @@ export const ShuttleMap = ({
             "user-location": userLocation
               ? [latLng(userLocation.latitude, userLocation.longitude)]
               : [],
-            "vehicle-location": positions,
+            "vehicle-location": vehiclePositions,
           }[followerController]
         : [],
-    [followerController, userLocation, positions]
+    [followerController, userLocation, vehiclePositions]
   )
 
   useEffect(() => {
--- a/assets/src/components/shuttleMapPage.tsx
+++ b/assets/src/components/shuttleMapPage.tsx
@@ -132,7 +132,7 @@ export const ShuttleMap = ({
 
   type FollowerController = false | "user-location" | "vehicle-location"
   const [followerController, setFollowerController] =
-    useState<FollowerController>(false)
+    useState<FollowerController>("vehicle-location")
 
   const state = useInteractiveFollowerState(),
     { setShouldFollow: setFollowActive } = state
--- a/assets/src/components/shuttleMapPage.tsx
+++ b/assets/src/components/shuttleMapPage.tsx
@@ -2,6 +2,7 @@ import { Socket } from "phoenix"
 import React, {
   Dispatch,
   ReactElement,
+  useCallback,
   useContext,
   useEffect,
   useMemo,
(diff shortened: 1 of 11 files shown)


assets/src/components/shuttleMapPage.tsx
```diff
@@ -57,39 +74,160 @@ const ShuttleMapPage = (): ReactElement<HTMLDivElement> => {
...
-            }
-          />
-        </MapFollowingSelectionKey>
-      </div>
+  const setShouldFollow = (controller: FollowerController) => {
+    setFollowerController(controller)
+    setFollowActive(controller !== false ? true : false)
+  }
+
+  const [location, setLocation] = useState<null | GeolocationCoordinates>(null)
```
lemald, line 140: **suggestion:** given that we're now dealing with both user and vehicle locations, I think it would be best to be explicit about which kind we're talking about when naming variables (like by calling this `userLocation`) for instance). Similarly, we use the term "positions" for vehicles and "location" for user in this component, and I'm sort of torn on whether it makes sense to standardize on one or the other or if there's a meaningful distinction.

assets/src/components/shuttleMapPage.tsx
```diff
@@ -57,39 +74,160 @@ const ShuttleMapPage = (): ReactElement<HTMLDivElement> => {
...
+    // Because we _only_ want to run this when the run id's change, we need to
+    // suppress the exhaustive-deps lint until `useEffectEvent` is stabilized
+    //
+    // See the react docs:
+    // > ## Is it okay to suppress the dependency linter instead?
+    // > After `useEffectEvent` becomes a stable part of React,
+    // > we recommend never suppressing the linter.
+    // > -- https://react.dev/learn/separating-events-from-effects#is-it-okay-to-suppress-the-dependency-linter-instead
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedShuttleRunIds])
```
lemald, line 169: **suggestion:** would it be possible to add `followerController` and `setShouldFollow` as dependencies, but further gate the `setShouldFollow` call behind some logic that verifies that `selectedShuttleRunIds` actually changed? We do sort of similar things in some places where we care about deep equality rather than referential equality in dependencies, like in the `useSwings` hook for instance, where there's an additional check to see if the dependencies actually changed in the specific manner that we care about before performing the logic.

lemald: `useMinischeduleRuns` might actually be a better example.

firestack: I'll look into it. I avoided it because I didn't want to introduce more state to manage, especially state that is essentially just noise to the actual logic of the component, and I didn't want to wrap up that noise in an abstraction too soon. I'm personally a fan of this approch, because it makes it easy to upgrade to using the `useEffectEvent` api and I believe this to be the "correct" way to handle this. But I'll see about reimplementing it.

lemald: I wish it was possible to pass parameters to the linker to have it selectively ignore rules, so for instance in this case we could tell it to ignore `followerController` and `setShouldFollow` but if you introduce any new inputs in the body without adding them to the dependencies array it will catch them.

firestack: Yeah that would certainly be better 😞

assets/src/components/shuttleMapPage.tsx
```diff
@@ -57,39 +74,160 @@ const ShuttleMapPage = (): ReactElement<HTMLDivElement> => {
...
 
+interface UserLocationProps {
+  onLocationUpdate: Dispatch<GeolocationCoordinates | null>
+}
+const UserLocation = ({ onLocationUpdate }: UserLocationProps) => {
+  const location = useGeolocation()
+  useEffect(() => {
+    onLocationUpdate(location)
+  }, [onLocationUpdate, location])
+  return null
```
firestack, line 230: @mbta/skate-developers this is the logic I think is somewhat weird. I needed a way to conditionally call this hook, because we don't want to ask for permission for the user's location until they press the location button. 

I think it's fine, but some extra scrutiny here might be warranted. Though, I do think I prefer this solution instead of say, adding a parameter to `useGeolocation` to conditionally call `watchPosition`.

firestack: There is a test that explicitly tests that `watchPosition` is not called until the button is clicked though.

assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -263,6 +265,41 @@ describe("Map controls", () => {
...
+
+    expect(CurrentLocationMarker.get()).toBeInTheDocument()
+  })
+})
+
+const CurrentLocationMarker = byRole("button", {
+  name: "Your current location",
+})
+const CurrentLocationControl = byRole("button", {
+  name: "Show your current location",
```
firestack, line 301: todo: unsure if I'm keeping these yet

assets/src/components/shuttleMapPage.tsx
```diff
@@ -57,39 +74,161 @@ const ShuttleMapPage = (): ReactElement<HTMLDivElement> => {
...
+          disabled={followerController == "user-location"}
+        />
+        <RecenterControl
+          position="topright"
+          recenter={() => setShouldFollow("vehicle-location")}
+        />
+
+        <>
+          {(followerController === "user-location" || userLocation) && (
+            <UserLocation onLocationUpdate={(l) => setUserLocation(l)} />
```
lemald, line 203:
```suggestion
            <UserLocation onLocationUpdate={setUserLocation} />
```

I think you can just pass `setUserLocation` in rather than wrap it in an anonymous function.

firestack: Oh right, I had a `console.debug` statement here for a bit and hadn't cleaned this up. Thanks!

assets/src/components/shuttleMapPage.tsx
lemald: I think this is fine. Since hooks can't be called conditionally our options are to put the conditional inside of the hook based on an argument, or to conditionally render / not render the component calling the hook, and I think either option is all right depending on the context.

assets/tests/components/shuttleMapPage.test.tsx
```diff
@@ -263,6 +265,40 @@ describe("Map controls", () => {
...
+  test("on initial load, does not show user location", () => {
+    render(<ShuttleMapPage />)
+
+    expect(CurrentLocationMarker.query()).not.toBeInTheDocument()
+  })
+
+  test("after user location button is clicked, show's user location on map", async () => {
+    mockGeolocation()
+    jest
+      .mocked(navigator.geolocation.watchPosition)
```
lemald, line 278: **suggestion:** why not mock the `useGeolocation` hook instead? I think that would be a bit simpler and also more appropriate for testing this functionality.

firestack: I originally did this as a "just get the test working", as it was the easiest entry point. Now, I'm stuck between not wanting to mock implementation details, and wanting to do an integration test between the two. Additionally, mocking `useGeolocation` means the state sematnics don't get tested, i.e., I can't find a way to have the test pass as it is where it returns `null` before returning a `GeolocationCoordinates` object.

I'm re-reviewing the tests though so I'll see about mocking `useGeolocation` instead.

firestack: It does cleanup nicely though 6dbb69ab1cb6d0878db40b72d5f4aa2fcf6d3e75

assets/src/components/shuttleMapPage.tsx
firestack: d9f3abe79a10c26d61a59528064c7c0504ea7e20

lemald (approved):

PR #2235, opened 2023-09-27
Get subway shapes from api

lemald:
Asana ticket: [⚙️ Use API for subway routes on shuttle map](https://app.asana.com/0/1148853526253420/1203629824340404/f)


assets/tests/hooks/useShapes.test.ts
```diff
@@ -34,15 +34,25 @@ describe("useRouteShapes", () => {
     expect(result.current).toEqual([])
   })
 
-  test("loads a subway route shape from hardcoded data", () => {
-    const mockFetchShape: jest.Mock = Api.fetchShapeForRoute as jest.Mock
+  test("loads a subway route shape from API", async () => {
+    const shape = shapeFactory.build({ className: undefined })
+    const mockFetchShape = jest
+      .mocked(Api.fetchShapeForRoute)
+      .mockReturnValueOnce(instantPromise([shape]))
```
firestack, line 41: suggestion(non-blocking): what do you think about using `mockResolvedValueOnce` instead?

```suggestion
      .mockResolvedValueOnce([shape])
```

firestack (commented):
Finished my first pass and I've left a few comments. Gonna wait till the race condition is addressed to do a final review.

assets/src/hooks/useShapes.ts
```diff
@@ -30,18 +37,31 @@ export const useRouteShapes = (selectedRouteIds: RouteId[]): Shape[] => {
...
+              flatten(shapesLists).map((shape) =>
+                enhanceShapeForSubwayRoute(shape, routeId, stations || [])
+              )
+            )
+          })
         }
       }
     })
-  }, [selectedRouteIds, shapesByRouteId])
+  }, [selectedRouteIds, shapesByRouteId, stations])
```
firestack, line 64: question: Hmm, is it weird that `setShapesByRouteId` is not in the dependency list? Given there are no linter errors, It seems like the linter is unsure if `setShapesForRoute`, `setLoadingShapesForRoute` change, dispite  `setShapesByRouteId` being able to change.

Maybe try defineing the funcitons within the `useEffect` and see if the linter then complains about the dependencies list?
https://react.dev/reference/react/useEffect#removing-unnecessary-function-dependencies

lemald: The linter rule actually has some [special logic](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L155) around a few specific kinds of values that don't need to be declared as deps, and state setters are among them.

assets/src/hooks/useShapes.ts
```diff
@@ -30,18 +37,31 @@ export const useRouteShapes = (selectedRouteIds: RouteId[]): Shape[] => {
...
+          setLoadingShapesForRoute(routeId)
 
-        if (isASubwayRoute(routeId)) {
-          setShapesForRoute(routeId, subwayRouteShapes(routeId))
-        } else {
           fetchShapeForRoute(routeId).then((shapes: Shape[]) =>
             setShapesForRoute(routeId, shapes)
           )
+        } else if (stations) {
+          setLoadingShapesForRoute(routeId)
```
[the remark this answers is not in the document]
lemald: It was intentional - basically, if `isASubwayRoute` returns true but the `stations` aren't filled in yet, then the `useEffect` doesn't actually do anything so I figured it would be best to wait on setting the loading state (`null`) until the call to the hook in which we're actually going to fire the API request.

assets/src/hooks/useShapes.ts
firestack: TIL! Thanks!

firestack (approved):

PR #2236, opened 2023-09-29
chore: bump node to version 20.7.0

lemald:
Node 20.8 is out but it's only about a day old at this point and there isn't an image for it on Docker Hub yet, so I'm going with 20.7.


firestack (approved):
🤞🏻

PR #2239, opened 2023-10-02
doc:feat(ex/router): document reason that we're keeping the SSL plug

firestack:
Asana Ticket: https://app.asana.com/0/1148853526253420/1205492553568017/f

### DIFF
--- a/lib/skate_web/router.ex
+++ b/lib/skate_web/router.ex
@@ -5,8 +5,8 @@ defmodule SkateWeb.Router do
 
   # We prefer to redirect at the load balancer level, but for security through
   # redundancy, we're keeping this plug just in case.
-  # A note: this does not affect anything in `endpoint.ex`, meaning static assets
-  # do not get redirected to HTTPS with this config currently.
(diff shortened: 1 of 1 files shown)


lib/skate_web/router.ex
```diff
@@ -3,6 +3,10 @@ defmodule SkateWeb.Router do
   use Plug.ErrorHandler
   use Sentry.Plug
 
+  # We prefer to redirect at the load balancer level, but for security through
+  # redundancy, we're keeping this plug just in case.
+  # A note: this does not affect anything in `endpoint.ex`, meaning static assets
+  # do not get redirected to HTTPS with this config currently.
```
lemald, line 9: **nitpick:** `endpoint.ex` is actually the place that invokes `plug SkateWeb.Router`, so I think it might be more accurate to say that `Plug.Static` specifically isn't affected.

firestack: WWhat do you think about ba72a37e9968f37a6e9d9dbbe2b1f2663f51637a ?

lemald: Looks good!

lemald (approved):

PR #2241, opened 2023-10-03
fix: reload page when channel join fails due to authentication

lemald:
Asana ticket: [⚙️ Gracefully handle sessions where authentication expires](https://app.asana.com/0/1148853526253420/1205451125311872/f)

I've been testing this out locally by adjusting `Skate.Ueberauth.Strategy.Fake` to give a token that's only valid for 10 seconds and finding that it works better than before.

On the search page without an active search, the user won't be making any API requests or subscribed to any channels that would notice that the auth is expired (notably, the data status channel sends an `auth_expired` message from the backend but the frontend doesn't do anything with it - something else to look into). In the old state, this meant that when a user goes and tries to perform a search it would silently fail to join the channel.

### DIFF
--- a/assets/src/hooks/useChannel.ts
+++ b/assets/src/hooks/useChannel.ts
@@ -47,10 +47,15 @@ export const useChannel = <T>({
             channel = undefined
           }
         })
-        .receive("error", ({ reason }) =>
-          // eslint-disable-next-line no-console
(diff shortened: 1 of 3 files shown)


firestack (approved):
looks like a good improvement to me 👍🏻

PR #2243, opened 2023-10-04
AuthenticatedChannel improvements

lemald:
Asana ticket: [⚙️ Gracefully handle sessions where authentication expires](https://app.asana.com/0/1148853526253420/1205451125311872/f)

This extends the `AuthenticatedChannel` behaviour so that `handle_info` and `handle_in` also check the token status before further processing. There are some other callbacks we could implement like `handle_out` but since we don't currently use them I haven't implemented them yet.

I also updated the `useChannel` and related hooks so that any channels using them will auto-reload on `auth_expired` messages.

### DIFF
--- a/assets/src/hooks/useChannel.ts
+++ b/assets/src/hooks/useChannel.ts
@@ -36,6 +36,7 @@ export const useChannel = <T>({
       channel.on(event, ({ data: data }) => {
         setState(parser(data))
       })
+      channel.on("auth_expired", reload)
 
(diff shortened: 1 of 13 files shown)


firestack (commented):
LGTM! Gonna think over the changes in authenticated_channel.ex for a bit but I'll approve this afternoon given that I don't come up with anything

firestack (approved):

PR #2247, opened 2023-10-06
Use FullStory browser API package

lemald:
Asana ticket: first step of [⚙️ Integrate Fullstory and Sentry](https://app.asana.com/0/1148853526253420/1205560495149255/f)

I tested this by deploying to dev-blue and making sure that my session made it into FullStory, including custom events being captured.


assets/tests/components/inputModals/createPresetModal.test.tsx
firestack, line 1: **praise**: I'm really like the use of `jest.mock` and `jest.mocked`, without having worked with it, this seems like a nice cleanup, thanks Eddie!!

lemald: Glad you like it! It was definitely a bit easier to work with.

assets/tests/helpers/fullStory.test.ts
```diff
@@ -1,23 +1,45 @@
...
+jest.mock("@fullstory/browser")
 
-    fullStoryIdentify("username")
+describe("fullStoryInit", () => {
+  test("calls init and identify if organization ID and username are given", () => {
+    const mockedFS = jest.mocked(FullStory)
 
-    expect(window.FS!.identify).toHaveBeenCalledWith("username", {
+    mockedFS.init.mockImplementationOnce(() => {})
+    mockedFS.identify.mockImplementationOnce(() => {})
```
firestack, line 12: question(non-blocking): what does jest/fs do without these being explicitly implemented?

lemald: Oh, that's a good question - because of the `jest.mock("@fullstory/browser")`, I think they're already effectively mocked out with no-op functions. If I remove the `mockImplementationOnce` calls the tests still pass. On the other hand, if I also remove the `jest.mock("@fullstory/browser")` call, the real FullStory init function tries to run and throws an error because of some aspects of the test environment that aren't to its liking.

lemald: Explicit mocks removed in d6c7f385.

firestack (approved):
🎉LGTM!

PR #2251, opened 2023-10-10
fix: only call FullStory.event if FS is initialized

lemald:
Asana ticket: part of [⚙️ Integrate Fullstory and Sentry](https://app.asana.com/0/1148853526253420/1205560495149255/f)


assets/src/helpers/fullStory.ts
```diff
@@ -12,3 +12,12 @@ export const fullStoryInit = (
...
 }
+
+export const fullStoryEvent = (
+  eventName: string,
+  eventProperties?: { [key: string]: any }
+): void => {
+  FullStory.isInitialized() && FullStory.event(eventName, eventProperties || {})
+
+  return
+}
```
firestack, line 23: suggestion(non-blocking): what do you think about doing this?
I get the benifit of defining our own interface is we can control the translation from our events to their API, but I also like presenting the full API of the event function to the caller as suggested below.
I think this would mean that all the places `{}` was deleted from the call, would need to be reverted.
```suggestion
export const fullStoryEvent = (...args: Parameters<typeof FullStory.event>): void => {
  FullStory.isInitialized() && FullStory.event(...args)
}
```

lemald: Note: one way or another we still need a `return`, since if `FullStory.isInitialized()` returns `false`, the whole `&&` statement would evaluate to `false` which is incompatible with `void`.

That said, I'm thinking this suggestion over and overall I think I like it.

firestack: Oh strange, my comment wasn't on the correct lines, sorry about that. It looked right in the editor 😅

firestack (approved):

lemald:
Tested latest changes both locally as well as on dev-blue to make sure that custom events are still being recorded.

PR #2252, opened 2023-10-11
feat(ts/components/RecenterControl): localize `RecenterControl` state and styling

firestack:
## Todo
- [x] Add stories

---
  - https://app.asana.com/0/0/1205709399517280

### DIFF
--- a/assets/css/_vehicle_map.scss
+++ b/assets/css/_vehicle_map.scss
@@ -181,24 +181,6 @@ $map-component-z-index: (
   border-top-color: $color-tooltip-background;
 }
 
-.c-vehicle-map__recenter-button {
-  path {
-    fill: none;
-    stroke: $color-button-secondary;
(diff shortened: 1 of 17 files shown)


lemald:
@firestack okay, I've looked over most things other than the tests and so far I think it looks good, I'll think it over and revisit tomorrow.

lemald:
@firestack okay, having reviewed the whole PR as it stands I think this is looking good! Will wait until it's out of draft status to give it a final look and approve.

firestack:
Okay I'm gonna stop here. I'm still fairly unhappy with the stories and the location triangle svg, but I don't think it's worth spending more time on at this moment, and would instead be better to standardize our icons/buttons for maps (that doesn't rely on leaflet styles to be present, in case we change map providers).

Let me know if there's anything you'd like changed.

lemald (approved):

PR #2254, opened 2023-10-11
[PROTOTYPE] feat(ts/mapPage): respect user's current zoom in follower and initially disable follower on route _selection_

firestack:
Asana Ticket: [⚙️ [Prototype] When clicking on RPC from VPC, maintain zoom level](https://app.asana.com/0/1148853526253420/1205511565935708/f)

### DIFF
--- a/assets/src/components/map/follower.tsx
+++ b/assets/src/components/map/follower.tsx
@@ -180,7 +180,7 @@ export const usePickerContainerFollowerFn = () => {
 }
 
 export const drawerOffsetAutoCenter =
-  (useTargetZoom: boolean): UpdateMapFromPointsFn =>
+  (useCurrentZoom: boolean): UpdateMapFromPointsFn =>
   (map, points) => {
     if (points.length === 0) {
       // If there are no points, blink to default center
@@ -203,11 +203,8 @@ export const drawerOffsetAutoCenter =
     const topLeft = new Point(445, 0)
 
     if (points.length === 1) {
-      const targetZoom = useTargetZoom
-        ? 16
-        : map.getZoom() < 13
-        ? 16
-        : map.getZoom()
+      const currentZoom = map.getZoom()
+      const targetZoom = useCurrentZoom && currentZoom >= 13 ? currentZoom : 16
       const innerBounds = new Bounds(
         topLeft,
         mapContainerBounds.getBottomRight()
@@ -237,6 +234,7 @@ export const drawerOffsetAutoCenter =
     }
   }
 
-export const fixedZoomDrawerOffsetAutoCenterFixed = drawerOffsetAutoCenter(true)
+export const fixedZoomDrawerOffsetAutoCenterFixed =
+  drawerOffsetAutoCenter(false)
 
 // #endregion Follower Update Functions
--- a/assets/src/components/map/follower.tsx
+++ b/assets/src/components/map/follower.tsx
@@ -179,7 +179,7 @@ export const usePickerContainerFollowerFn = () => {
   return onUpdate
 }
 
-export const drawerOffsetAutoCenter2 =
+export const drawerOffsetAutoCenter =
   (useTargetZoom: boolean): UpdateMapFromPointsFn =>
   (map, points) => {
     if (points.length === 0) {
@@ -237,6 +237,6 @@ export const drawerOffsetAutoCenter2 =
     }
   }
 
-export const drawerOffsetAutoCenter = drawerOffsetAutoCenter2(true)
+export const fixedZoomDrawerOffsetAutoCenterFixed = drawerOffsetAutoCenter(true)
 
 // #endregion Follower Update Functions
--- a/assets/src/components/map/follower.tsx
+++ b/assets/src/components/map/follower.tsx
@@ -25,7 +25,7 @@ export type UpdateMapFromPointsFn = (map: LeafletMap, points: LatLng[]) => void
 
 export interface FollowerProps {
   positions: LatLng[]
-  onUpdate: UpdateMapFromPointsFn
+  onUpdate?: UpdateMapFromPointsFn
 }
 
 export const Follower = ({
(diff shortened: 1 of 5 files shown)


assets/src/components/map/follower.tsx
```diff
@@ -177,53 +179,64 @@ export const usePickerContainerFollowerFn = () => {
...
+      const pointsBounds = Leaflet.latLngBounds(points)
+      map.fitBounds(pointsBounds, {
+        paddingBottomRight: [50, 20],
+        paddingTopLeft: topLeft,
+      })
+    }
   }
-}
+
+export const drawerOffsetAutoCenter = drawerOffsetAutoCenter2(true)
```
lemald, line 240: **suggestion:** how about naming this something like `fixedZoomDrawerOffsetAutoCenter`, and keeping `drawerOffsetAutoCenter` as the name for what's now called `drawerOffsetAutoCenter2`.

firestack: 💯 I definitely would reconsider a lot of this if this functionality is wanted by product/design.

assets/src/components/map/follower.tsx
```diff
@@ -177,53 +179,62 @@ export const usePickerContainerFollowerFn = () => {
...
+      const pointsBounds = Leaflet.latLngBounds(points)
+      map.fitBounds(pointsBounds, {
+        paddingBottomRight: [50, 20],
+        paddingTopLeft: topLeft,
+      })
+    }
   }
-}
+
+export const fixedZoomDrawerOffsetAutoCenterFixed =
```
lemald, line 237: **nitpick:** word "fixed" appears twice in the function name

firestack: Whoops, thanks for catching that 369bb03ddc2f9b7ee0107afb1536f658ab1d7f30

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -304,10 +312,22 @@ const SelectedVehicleDataLayers = ({
...
+      <RecenterControl
+        position="topright"
+        active={followerState.shouldFollow}
+        onActivate={() => {
+          followerState.setShouldFollow(true)
+        }}
+      />
+      <InterruptibleFollower
+        // Do nothing if we don't have a vehicle position yet
+        onUpdate={position.length === 0 ? () => {} : onUpdate}
```
lemald, line 324: **suggestion:** how about making `onUpdate` possibly null / undefined and updating `<Follower/>` accordingly to handle that case (i.e. not call it if undefined)? That feels like it might be a bit more natural than the no-op function approach.

firestack: Good idea! e8a1c2332a165a24c5bc46bab8dc23b0e80eeb50

assets/src/components/mapPage.tsx
```diff
@@ -205,6 +205,9 @@ const Selection = ({
 }
 
 const MapPage = (): ReactElement<HTMLDivElement> => {
+  const [followerInitializeState, setFollowerInitializeState] =
+    useState<boolean>(false)
```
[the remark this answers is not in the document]
lemald: One suggestion: drop the "state" from the name.

As for the rest of the name, how about something like `followerDefaultZoomOnInitialize` or something like that? Or the opposite of that if I have the logic backwards.

assets/tests/components/mapPage/mapDisplay.test.tsx
```diff
@@ -860,4 +861,29 @@ describe("<MapDisplay />", () => {
       expect(mockSetSelection).not.toHaveBeenCalled()
     })
   })
+
+  test.each([
+    { enabled: true, expected: "true" },
+    { enabled: false, expected: "false" },
+  ])(
+    "when initializeRouteFollowerEnabled is $enabled and route is selected, follower should be $expected",
```
lemald, line 869: **nitpick:** the "follower should be $expected" part of the name for these tests seems a little weird to me since that will just expand to (say) "follower should be true." Maybe it should actually say "enabled" / "disabled?"

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
firestack: What do you think of `followerHasRun` or `followerHasHadAChanceToRun`?

assets/tests/components/mapPage/mapDisplay.test.tsx
firestack: does this work? 6df71681a397677b9ba2b7b7341ef5b90b6b1e95

lemald: I like it!

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
lemald: Hmm, I see what you're getting at... it sort of approaches this functionality from a different angle (basically it's expressing the condition causes the behavior to change rather than what the behavior is). I'm trying to decide if I feel like that helps or hurts when it comes to clarity. 😅 In any event since "run" is a bit of an overloaded term in Skate I'd prefer `followerHasHadAChanceToRun` out of those two options.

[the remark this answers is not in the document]
firestack: Hah, I didn't even think about how overloaded "run" is at first. I agree the other is better.

I think it reads better `vehicleUseCurrentZoom={followerHasHadAChanceToRun === true}` "vehicle use current zoom if follower has had a chance to run"
but maybe 
`vehicleUseCurrentZoom={followerNeedsToInitializeZoom === false}` would be better?
Because I changed it so that the zoom control state is reset when a item is selected from search (so you don't stay super zoomed out)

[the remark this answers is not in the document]
lemald: Oh I like `followerNeedsToInitializeZoom`!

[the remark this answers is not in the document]
firestack: Okay I changed it once more, let me know if you'd prefer `NeedsTo` instead of `Should`, I was uncertain what would better convey the logic. `Should` seemed better but I'm unsure.

bd56b922151fcddb6e07258e39910167538dc0f5

lemald (approved):

assets/src/components/mapPage.tsx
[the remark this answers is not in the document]
lemald: I'm neutral so if you prefer `Should` that sounds good to me. 👍

PR #2257, opened 2023-10-12
fix: sort through-routed trips numerically

lemald:
Asana ticket: [⚙️ Fix negative duration layovers in minischedules](https://app.asana.com/0/1148853526253420/1205367064394203/f)

It turned out that the issue was that the trips were in reverse order in the GTFS `trips.txt` (which is fine, that file isn't guaranteed to be in any particular order). This adds logic to explicitly sort the trips by their numeric suffix.

I'm sort of of two minds on the `try` / `rescue` part of the code. On the one hand either the pattern matching on the return value of `Regex.scan/3` or the call to `String.to_integer/1` could in theory raise an error, but because of the way `original_id_to_through_routed_trip_ids` is built this case should never happen. That said, if it somehow _did_ happen it could be very bad and would cause the application to fail to fully start up, so it felt better to cover it. One disadvantage of including this logic is that, since it's unreachable, we can't write a test for it.

Note that the trip start and end place still come from HASTUS, and since HASTUS thinks of these through-routed trips as a single trip, that means both of the expanded trips will have the same departure point listed in the schedule. I'm thinking through how we might address that, if at all, but it will be a separate PR.

### DIFF
--- a/lib/schedule/hastus/trip.ex
+++ b/lib/schedule/hastus/trip.ex
@@ -3,6 +3,8 @@ defmodule Schedule.Hastus.Trip do
   alias Schedule.{Block, Route, Trip}
   alias Schedule.Hastus.{Place, Run, Schedule}
 
+  @through_routed_suffix_regex ~r/_(\d+)$/
+
   @type t :: %__MODULE__{
           schedule_id: Schedule.id(),
           run_id: Run.id(),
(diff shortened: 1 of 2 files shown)


firestack (approved):
Seems fine, I agree that the try/rescue feels a bit strange, but I think fixing this is more important than ergonomics right now, and we can fix the ergonomics later.

PR #2258, opened 2023-10-13
fix(ts/components/NearbyStops): refresh stops when inputs to function change

firestack:
Asana Ticket: [⚙️🐞 When selected route direction changes excluded nearby bus stops do not update](https://app.asana.com/0/1148853526253420/1205560497394206/f)

### DIFF
--- a/assets/src/components/mapPage/mapDisplay.tsx
+++ b/assets/src/components/mapPage/mapDisplay.tsx
@@ -455,13 +455,28 @@ const NearbyStops = ({ stops }: { stops: Stop[] }) => {
   )
   const [nearbyStops, setNearbyStops] = useState<Stop[]>([])
   const map = useMap()
-  map.addEventListener("moveend", () => {
+
+  const filterStopsToBounds = useCallback(() => {
     const bounds = map.getBounds()
     // Only show nearby stations or bus stops
     setNearbyStops(
       stationsAndBus.filter((s) => bounds.contains([s.lat, s.lon]))
     )
-  })
+  }, [map, stationsAndBus, setNearbyStops])
+
+  // Filter stops whenever function, and it's inputs, change
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -455,13 +455,28 @@ const NearbyStops = ({ stops }: { stops: Stop[] }) => {
...
-  map.addEventListener("moveend", () => {
+
+  const filterStopsToBounds = useCallback(() => {
     const bounds = map.getBounds()
     // Only show nearby stations or bus stops
     setNearbyStops(
       stationsAndBus.filter((s) => bounds.contains([s.lat, s.lon]))
     )
-  })
+  }, [map, stationsAndBus, setNearbyStops])
```
lemald, line 465: **question:** trying to wrap my head around this, which of these three dependencies actually changes when you toggle the direction in the RPC?

firestack: stationAndBus changes, because it's derived from stops

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -455,13 +455,28 @@ const NearbyStops = ({ stops }: { stops: Stop[] }) => {
...
+
+  // Filter stops whenever function, and it's inputs, change
```
lemald, line 467: **nitpick:** sorry, the most nitpick-y of nitpicks
```suggestion
  // Filter stops whenever function, and its inputs, change
```

lemald:
@firestack general question: do you think it would be possible to write a test for this?

firestack:
> @firestack general question: do you think it would be possible to write a test for this?

I think so, I'll look into it this morning.

lemald (approved):

PR #2265, opened 2023-10-27
Realtime vehicles / alerts PubSub

lemald:
Asana ticket: [⚙️ Simplify how Skate manages subscriptions to realtime data](https://app.asana.com/0/1148853526253420/1205067894630933/f)

### DIFF
--- a/lib/realtime/server.ex
+++ b/lib/realtime/server.ex
@@ -81,6 +81,9 @@ defmodule Realtime.Server do
   @spec default_name() :: GenServer.name()
   def default_name(), do: Realtime.Server
 
+  @spec pubsub_name() :: Phoenix.PubSub.t()
+  def pubsub_name(), do: Realtime.PubSub
+
   @spec start_link(Keyword.t()) :: GenServer.on_start()
   def start_link(start_link_opts) do
     GenServer.start_link(__MODULE__, [], start_link_opts)
@@ -210,17 +213,20 @@ defmodule Realtime.Server do
   @spec subscribe(GenServer.server(), {:run_ids, [Run.id()]}) :: [VehicleOrGhost.t()]
   @spec subscribe(GenServer.server(), {:block_ids, [Block.id()]}) :: [VehicleOrGhost.t()]
   @spec subscribe(GenServer.server(), {:alerts, Route.id()}) :: [String.t()]
+  defp subscribe(server, {:alerts, _route_id} = subscription_key) do
+    {pubsub, ets} = GenServer.call(server, :subscription_info)
+    Phoenix.PubSub.subscribe(pubsub, "realtime_alerts")
+    lookup({ets, subscription_key})
+  end
+
   defp subscribe(server, subscription_key) do
-    {registry_key, ets} = GenServer.call(server, :subscription_info)
-    Registry.register(Realtime.Registry, registry_key, subscription_key)
+    {pubsub, ets} = GenServer.call(server, :subscription_info)
+    Phoenix.PubSub.subscribe(pubsub, "realtime_vehicles")
     lookup({ets, subscription_key})
   end
 
   defp update_subscription(server, {:limited_search, _search_params} = subscription_key) do
-    {registry_key, ets} = GenServer.call(server, :subscription_info)
-    # Replace the old search subscription with the new one
-    Registry.unregister_match(Realtime.Registry, registry_key, {:limited_search, %{}})
-    Registry.register(Realtime.Registry, registry_key, subscription_key)
+    {_pubsub, ets} = GenServer.call(server, :subscription_info)
 
     lookup({ets, subscription_key})
   end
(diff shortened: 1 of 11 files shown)


lib/realtime/server.ex
```diff
@@ -187,17 +213,20 @@ defmodule Realtime.Server do
...
+
   defp subscribe(server, subscription_key) do
-    {registry_key, ets} = GenServer.call(server, :subscription_info)
-    Registry.register(Realtime.Registry, registry_key, subscription_key)
+    {pubsub, ets} = GenServer.call(server, :subscription_info)
+    Phoenix.PubSub.subscribe(pubsub, "realtime_vehicles")
     lookup({ets, subscription_key})
   end
 
   defp update_subscription(server, {:limited_search, _search_params} = subscription_key) do
```
KaylaBrady, line 228: question (non-blocking): Is update_subscription still needed?

lib/skate_web/channels/alerts_channel.ex
```diff
@@ -7,13 +7,18 @@ defmodule SkateWeb.AlertsChannel do
 
   @impl SkateWeb.AuthenticatedChannel
   def join_authenticated("alerts:route:" <> route_id, _message, socket) do
-    alerts = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
-    {:ok, %{data: alerts}, socket}
+    {lookup_key, alerts} = Duration.log_duration(Server, :subscribe_to_alerts, [route_id])
+
+    {:ok, %{data: alerts},
+     Phoenix.Socket.assign(socket, Map.merge(socket.assigns, %{lookup_key: lookup_key}))}
```
firestack, line 13: question: is this merge required? The docs seem to suggest that Socket.assign/2 & /3 already merge the provided keys to the assigns map.
https://hexdocs.pm/phoenix/1.6.16/Phoenix.Socket.html#assign/2

firestack:
```suggestion
     Phoenix.Socket.assign(socket, lookup_key: lookup_key)
```

test/skate_web/channels/vehicle_channel_test.exs
```diff
@@ -78,6 +78,39 @@ defmodule SkateWeb.VehicleChannelTest do
...
+      expected_payload = %{data: nil}
+
+      assert {:ok, ^expected_payload, %Socket{} = _socket} =
+               subscribe_and_join(socket, VehicleChannel, "vehicle:run_ids:123-4567")
+    end
+
+    test "subscribes to the vehicle for given ID when user is not in the test group", %{
+      socket: socket
+    } do
+      Realtime.Server.update_vehicles({%{"1" => [@vehicle]}, [], []})
```
firestack, line 95: suggestion(non-blocking): would you be open to swapping this out for a factory?

lib/skate_web/channels/alerts_channel.ex
firestack: - https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicle_channel.ex#L28
- https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicle_channel.ex#L63
- https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicles_channel.ex#L14-L49
- https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicles_channel.ex#L97
- https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicles_search_channel.ex#L49
- https://github.com/mbta/skate/blob/1478499366e9eb5e65cc0f51da6c67bec8aec83d/lib/skate_web/channels/vehicles_search_channel.ex#L87

lemald: @firestack I actually just force-pushed a change to fix all of those other spots, there should be no more uses of `Map.merge` in any of the calls to `Socket.assign`.

firestack: Thanks!

lib/realtime/server.ex
lemald: The main reason I decided to keep it around is so that `lookup` actually gets called immediately when updating the subscription, rather than having to wait for the next data update (which is usually going to take only around a second, but I figured this way it's more responsive to user input).

test/skate_web/channels/vehicle_channel_test.exs
lemald: 5239851c

lib/realtime/server.ex
KaylaBrady: That makes sense! Maybe a rename would be helpful for this function since it is no longer updating a subscription, though still non-blocking!

lemald: I'm in more-of-less the same boat, I think it needs a new name but I can't determine a good one. `get_latest_results` is probably better than anything I've thought up, though.

PR #2267, opened 2023-10-31
feat: change application supervision strategy

lemald:
Asana ticket: [⚙️ Application can survive minor crashes](https://app.asana.com/0/1148853526253420/1205805990143476/f)

Go from :one_for_all to :rest_for_one to hopefully reduce the impact of processes like Oban crashing

My purpose in bundling the changes to the DB parameters is that I want to avoid the chance that deploying this causes further downtime by causing the Oban process on the existing containers to thrash.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -119,7 +119,8 @@ config :skate, Skate.Repo,
 config :skate, Skate.WarmUp,
   minimum_percent_queries_to_succeed: 0.6,
   max_attempts: 20,
-  seconds_between_attempts: 1
+  seconds_between_attempts: 5,
+  backoff_min: 5_000
 
 config :skate, Oban,
   repo: Skate.Repo,
--- a/config/config.exs
+++ b/config/config.exs
@@ -114,13 +114,13 @@ config :skate, Skate.Repo,
   password: System.get_env("POSTGRES_PASSWORD", ""),
   hostname: System.get_env("POSTGRES_HOSTNAME", "localhost"),
   port: System.get_env("POSTGRES_PORT", "5432") |> String.to_integer(),
-  show_sensitive_data_on_connection_error: true
+  show_sensitive_data_on_connection_error: true,
+  backoff_min: 5_000
 
 config :skate, Skate.WarmUp,
   minimum_percent_queries_to_succeed: 0.6,
   max_attempts: 20,
-  seconds_between_attempts: 5,
-  backoff_min: 5_000
+  seconds_between_attempts: 5
 
 config :skate, Oban,
   repo: Skate.Repo,
--- a/lib/skate/application.ex
+++ b/lib/skate/application.ex
@@ -9,12 +9,21 @@ defmodule Skate.Application do
   def start(_type, _args) do
     load_runtime_config()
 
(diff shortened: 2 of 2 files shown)


config/config.exs
```diff
@@ -119,7 +119,8 @@ config :skate, Skate.Repo,
 config :skate, Skate.WarmUp,
   minimum_percent_queries_to_succeed: 0.6,
   max_attempts: 20,
-  seconds_between_attempts: 1
+  seconds_between_attempts: 5,
+  backoff_min: 5_000
```
firestack, line 123: question: I don't see any reference to this in the `WarmUp` module, what does this do?

lemald: Oops, that's not related to the WarmUp at all, it's supposed to be part of the repo configuration! Will fix. See [here](https://hexdocs.pm/db_connection/2.4.1/DBConnection.html#start_link/2-options) for what it's actually supposed to be.

config/config.exs
```diff
@@ -119,7 +119,8 @@ config :skate, Skate.Repo,
 config :skate, Skate.WarmUp,
   minimum_percent_queries_to_succeed: 0.6,
   max_attempts: 20,
-  seconds_between_attempts: 1
+  seconds_between_attempts: 5,
```
firestack: note(non-blocking): Also just wanting to acknowledge that because we don't have a backoff time that scales with the number of attempts, and that WarmUp prevents the rest of the application from running until it's done (due to init/1), the application won't really start until some multiple of 5.

Maybe we consider starting the `Schedule.Fetcher` before `Repo` and `WarmUp`, so that it's got as much time as possible to complete? I don't see any references to `Skate.Repo` in `Schedule.Fetcher` so I think it'd be okay, but then that breaks up the `start_data_processes` flag.

lemald: Really good point! And yes, you're correct that the schedule process tree shouldn't use the database at all. I think I'll go ahead and make that change.

lib/skate/application.ex
```diff
@@ -9,12 +9,21 @@ defmodule Skate.Application do
...
-      [{Skate.Repo, []}, Skate.WarmUp] ++
-        if Application.get_env(:skate, :start_data_processes) do
+      [{Skate.Repo, []}] ++
+        if start_data_processes? do
+          [
+            Schedule.Supervisor
+          ]
+        else
+          []
+        end ++
```
firestack, line 23: question(non-blocking): is there a reason to start this after `Skate.Repo` instead of before?

firestack (approved):

lib/skate/application.ex
firestack: I'm pretty sure that `Skate.Repo` doesn't block on `init/1`, so I don't think it matters, I am just curious.

lemald: I think it's better to leave it first because that way it has more time to spin up its connection pool in the background while other things happen.

lemald: Which in turn means that hopefully by the time we get to launching `WarmUp` it's more likely that the connection pool is in a good state.

PR #2271, opened 2023-11-06
fix: Fix icon size for location search result marker

joshlarson:
# Before
Screenshot 2023-11-06 at 4 56 41 PM

# After
Screenshot 2023-11-06 at 4 56 23 PM

---
  - https://app.asana.com/0/0/1205658795320766

### DIFF
--- a/assets/css/map/markers/_location_dot_icon.scss
+++ b/assets/css/map/markers/_location_dot_icon.scss
@@ -1,5 +1,7 @@
 .c-location-dot-icon {
   svg {
+    height: 24px;
+
     overflow: visible;
 
     path {


assets/static/images/icon-location-dot.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" height="1em" viewBox="0 0 384 512"><!--! Font Awesome Free 6.4.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc.--><path d="M215.7 499.2C267 435 384 279.4 384 192 384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2 12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512"><!--! Font Awesome Free 6.4.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc.--><path d="M215.7 499.2C267 435 384 279.4 384 192 384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2 12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z"/></svg>
```
firestack, line 1: question(non-blocking): in the case that we regenerate this icon or use an external library, did you find any reason why our style-sheets aren't setting the size correctly?

joshlarson: CSS did the trick!

assets/src/components/mapMarkers.tsx
```diff
@@ -466,7 +466,6 @@ export const LocationMarker = ({
     position={[location.latitude, location.longitude]}
     divIconSettings={{
       iconAnchor: new Leaflet.Point(9, 24),
-      iconSize: [24, 18],
```
firestack, line 469: issue: I don't think we want this removed, as changing this will subtlety offset where leaflet places the marker on the map, because leaflet uses this value for layout and translating the object onto the map. Additionally, this changes the "tap target"/"bounding box" of the marker, so if we did make it clickable, potentially only part of it would be clickable as I understand it.

(If somehow/for some reason we changed the leaflet marker icon class to clip things outside of it's bounding box, the marker would be cut off as well)
image

firestack: > Hmmm. I don't love that the CSS and the divIcon are both there and have to have the same values.

Agreed 😞

joshlarson: But also, the bounding box stuff matters more, and I'll add that back in 😄

firestack (approved):
yay bug fixes!

PR #2273, opened 2023-11-07
feat(ts/app/RightPanel): store right panel state per page

firestack:
This pulls the reducer responsible for handling the right panel state out into it's own reducer and instead of exposing access to the reducer state itself, provides a reduced API via a hook `usePanelState*`.

The new reducer still relies on `StateDispatchContext` and the `state.ts` reducer because there are other systems that do work based on some of the events that `openViewPanelReducer` handles.

Setting the intended tab for the VPP doesn't quite work yet, but this is large enough that I'm going to do that in a separate PR, and it should be more digestible.

Commits should be reviewable in order.

---

Asana Ticket: [⚙️ Panel behavior when navigating across pages](https://app.asana.com/0/1148853526253420/1205762199809559/f)


assets/src/components/app.tsx
```diff
@@ -59,11 +67,7 @@ export const AppRoutes = () => {
       </div>
       <VehiclesByRouteIdProvider vehiclesByRouteId={vehiclesByRouteId}>
         <div className="l-app__main">
-          <Nav
-            allowViews={
-              location.pathname !== mapMode.path || mapMode.supportsRightPanel
-            }
-          >
+          <Nav allowViews>
```
firestack, line 70: note: I'll open a PR after this one to remove this prop altogether.

firestack: https://github.com/mbta/skate/pull/2275

assets/src/state/pagePanelState.ts
```diff
@@ -0,0 +1,218 @@
...
+
+//#region Accessors
+// Keeping this around because this is used by other reducers in `state.ts`.
+export const selectVehicleFromNotification = (
+  vehicle: VehicleType
+): SelectVehicleFromNotificationAction => ({
+  type: "SELECT_VEHICLE_FROM_NOTIFICATION",
+  payload: { vehicle },
+})
+//#endregion Accessors
```
lemald, line 165: **question:** What's your rationale for calling this section "Accessors?" I'm not sure exactly what I'd call it, but something more like "action constructors" would make more sense to me personally.

assets/tests/hooks/usePanelState.test.ts
lemald, line 1: **question:** These tests cover a lot of the actual reducer logic - how would you feel about splitting them up so that the tests in this file focus more on the hooks themselves and tests for the reducer logic go in a new `pagePanelState.test.ts` file?

assets/src/state/pagePanelState.ts
firestack: It was "simply" the first thing that came to mind when I was migrating 😅 
I like "action constructors"! Though I might shorten to just actions if that works.

assets/tests/hooks/usePanelState.test.ts
firestack: Heh, this file was originally named `pagePanelState.test.ts`, but I felt I was actually just testing the public api of `usePanelState`, so I changed it.

Do you have any specifics on what you see is testing the reducer logic or what you would split off and what you would keep?
I actually made a conscious choice to not test the reducer explicitly because I only wanted to test and assert based on the public API, and I am considering `pagePanelState` and it's reducer to be entirely private and an implementation detail, so it could be swapped out for something else and clients of `usePanelState*` would not see a difference.

I do see that I have blurred the lines by adding `initialProps` and `currentState` parameters to the `renderUsePanelState`, and I'd be fine to reconsider/rebuild these tests without those parameters.

assets/src/state/pagePanelState.ts
firestack: ffcab85bea8023ac71cc41cb41bb4b32e53e03ed

assets/tests/hooks/usePanelState.test.ts
lemald: I guess some of the lines are blurry since `pagePanelState` does still have to export things for use by `usePanelState` even if that's the only intended use case... In any event, the way I was thinking about it is that the tests for `usePanelState` would confirm that (say) `usePanelStateFromStateDispatchContext` takes the subset of the state and dispatch function supplied by the `StateDispatchContext`, whereas the tests for `pagePanelState` would test the specific reducer actions. Does that distinction make sense?

firestack: "historical" note: Originally, this hook did live in `pagePanelState`, but when it was co-located, I ended up hitting circular reference issues, so that's why these are not in the same file, where the reducer can be considered more of a implementation detail.

assets/src/components/app.tsx
```diff
@@ -20,28 +19,37 @@ import Modal from "./modal"
...
+    setPath,
+    currentView: { openView, selectedVehicleOrGhost },
+  } = usePanelStateFromStateDispatchContext()
 
-  const location = useLocation()
+  // Keep panel in sync with current path
+  const { pathname: path } = location
+  useEffect(() => {
+    isPagePath(path) && setPath(path)
+  }, [path, setPath])
```
lemald, line 47: **question:** Is there an explicit test for this logic anywhere?

assets/tests/hooks/usePanelState.test.ts
lemald: Gotcha. I guess my feeling is that if the code ends up structured that way then it makes sense to just go with it and structure the tests that way as well, but that's just my $0.02 and I don't think this should be considered a blocking comment.

assets/src/components/app.tsx
firestack: good question, there is not yet, as I wasn't sure how to test stuff with the browser router and such. I'll look into it 👍🏻

assets/src/state/pagePanelState.ts
```diff
@@ -0,0 +1,218 @@
...
+    case "SET_CURRENT_PATH": {
+      return {
+        openView: OpenView.None,
+        previousView: OpenView.None,
+        selectedVehicleOrGhost,
+      }
+    }
+    case "OPEN_NOTIFICATION_DRAWER":
+      return openView === OpenView.NotificationDrawer
+        ? state
```
lemald, line 100: **praise:** I like the refactor in eb4969e, it definitely makes this code a bit more readable.

lemald:
@firestack okay whew, I think I've finally read through all of the commits. It sounds like you're still going to look into that test but otherwise this is looking pretty good overall. 👍

assets/src/components/app.tsx
firestack: I have a test that asserts this works, but I don't have a test that asserts that this doesn't do anything on pages like `/settings`.

I'll go ahead and commit the affirmative test, but I'm not sure if we should keep the one that doesn't work but skip it for now?

## Background
So I've found `app.test.tsx` to be somewhat fragile, I'm honestly still not got my head wrapped around all of this.

My main issue has been that depending on which tests are run, and in which order, some state is being saved across tests within the same module.
I _think_ this is due to JSDOM, because JSDOM is only ["reconfigured"](https://github.com/jestjs/jest/issues/5987) when the test module changes, but not in-between tests.
This means, that if you add a test before the `"render"` test, that "navigates" somewhere using the browser API, that `window` state will not be reset until the next module, and the snapshot changes because the `window.location` is different (and I'm still unclear on how `MemoryRouter` interacts with this). If `window.location` is different, that seems to affect the router, causing snapshot tests to fail because the dom would be in a different state due to being on a different page.

Some of the suggestions in the linked jest issue, suggest using a `reconfigure` method on JSDOM, but I can't seem to find any way to get access to the created `DOM` object that `jest-environment-jsdom` creates https://github.com/jestjs/jest/blob/main/packages/jest-environment-jsdom/src/index.ts#L34 .

I don't think any of the existing tests in `app.test.tsx` actually test navigation within the application, like this new test is. I think this means that since the new test tests navigation, if we were to randomize the order of the tests in this module (something I'd like to turn on in jest), we'd start having random tests fail due to this.

firestack: 10f03586dc03b14cd18fbf69fc58b44ea4010f43

firestack: btw this is what the other test (that fails) looks like
```tsx
  test.skip("doesn't update panel state when page is not an allowed page", async () => {
    const mockedUsePanelState = mockUsePanelState()
    render(
      <StateDispatchProvider state={stateFactory.build()} dispatch={jest.fn()}>
        <MemoryRouter initialEntries={["/"]}>
          <AppRoutes />
        </MemoryRouter>
      </StateDispatchProvider>
    )

    await userEvent.click(screen.getByRole("link", { name: "Settings" }))

    expect(mockedUsePanelState().setPath).toHaveBeenNthCalledWith(1, "/")
    expect(mockedUsePanelState().setPath).toHaveBeenCalledTimes(1)
  })
  ```

File: assets/tests/components/app.test.tsx [hunk 5]
```diff
@@ -90,9 +97,11 @@ describe("App", () => {
       </StateDispatchProvider>
     )
 
-    const routeIds = (useVehicles as jest.Mock).mock.calls[0][1]
-
-    expect(routeIds).toEqual(["1", "15", "22"])
+    expect(useVehicles as jest.Mock).toHaveBeenCalledWith(undefined, [
```
lemald, line 100: **suggestion:** use `jest.mocked(...)` rather than `as jest.Mock`.

firestack: 1457347b0626d434a9abbe05e9bd66d9ed1519b2

assets/tests/hooks/usePanelState.test.ts
firestack: 19d0a4d37add9e7b6ce0351d69dfc33d4b012275

lemald (approved):
Thanks for the changes! Looks good to me.

PR #2275, opened 2023-11-08
 feat(ts/hooks/usePanelState): add `setTabMode` API 

firestack:
When working on #2273, I ended up breaking the MapPage feature which allowed opening the VPP to a specific page. I considered adding the first part of this PR in #2273, but I figured it was getting large enough it'd be easier to digest as a separate PR.

This "stores" the state of the active VPP Tab in the `usePanelState*` hooks because the `initialTab` prop was insufficient because switching pages wouldn't reset the state of `tabMode`, therefore `initialState` was only respected once. By pulling the state out into the panel state, the tab can now be remembered across pages, and each page has it's own tab mode.

---

Depends On: #2273 
Asana Ticket: https://app.asana.com/0/1148853526253420/1205762199809559/f

---
  - https://app.asana.com/0/0/1205877298615016

### DIFF
--- a/assets/src/components/propertiesPanel/header.tsx
+++ b/assets/src/components/propertiesPanel/header.tsx
@@ -23,9 +23,9 @@ import TabList from "./tabList"
 import { currentRouteTab } from "../../models/routeTab"
 import ViewHeader from "../viewHeader"
 import { usePanelStateFromStateDispatchContext } from "../../hooks/usePanelState"
-import { IndividualPropertiesPanelProps } from "../propertiesPanel"
+import { ClosePanelProps, TabModeProps } from "../propertiesPanel"
 
-interface Props extends IndividualPropertiesPanelProps {
+interface Props extends TabModeProps, ClosePanelProps {
   vehicle: Vehicle | Ghost
 }
 
--- a/assets/src/components/propertiesPanel/header.tsx
+++ b/assets/src/components/propertiesPanel/header.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useContext } from "react"
+import React, { useContext } from "react"
 import { useRoute } from "../../contexts/routesContext"
 import { StateDispatchContext } from "../../contexts/stateDispatchContext"
 import { joinClasses } from "../../helpers/dom"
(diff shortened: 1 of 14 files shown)


assets/src/components/rightPanel.tsx
```diff
@@ -1,35 +1,30 @@
...
-    )
-  } else if (openView === OpenView.Swings) {
+const RightPanel = ({
+  openView,
+  propertiesPanel,
+}: RightPanelProps): ReactElement | null => {
+  if (propertiesPanel !== undefined) {
+    return <>{propertiesPanel}</>
+  }
+  if (openView === OpenView.Swings) {
```
lemald, line 19: **suggestion (non-blocking):** I think I sort of get why the first `if` statement is separated out rather than being part of the main series of `if` / `else` statements, but I personally feel like it would be more readable if it was all one series of `if` / `else`s.

assets/src/hooks/usePanelState.ts
```diff
@@ -53,13 +53,16 @@ export const usePanelStateForViewState = (
         })
       }
     },
-    openVehiclePropertiesPanel: (
-      vehicle: VehicleType,
-      _initialView?: TabMode
-    ) =>
+    setTabMode(tabMode: TabMode) {
```
lemald, line 56: **nitpick:** is there a reason to use the different notation without the `=>` for this function?

firestack: Oh y'know, I ment for all of them to follow this convention, instead of the arrow function convention, but given the rest are using arrow functions, I'll update this to match.

assets/src/components/rightPanel.tsx
firestack: Okay I meant to not merge them yet, but I did, so I've fixed this in the latest force push.
13d46768878283a58025d4054f9e40d0f4197520

assets/src/hooks/usePanelState.ts
firestack: Squashed into commit 13d46768878283a58025d4054f9e40d0f4197520

lemald: I actually don't have a strong preference between the two as long as it's consistent, though I think the arrow notation is more common in our codebase overall.

lemald (approved):

assets/src/components/propertiesPanel/header.tsx
```diff
@@ -20,16 +20,13 @@ import { Ghost, Vehicle, VehicleInScheduledService } from "../../realtime"
...
 import VehicleIcon, { Size, vehicleOrientation } from "../vehicleIcon"
 import TabList from "./tabList"
-import { TabMode } from "./tabPanels"
 import { currentRouteTab } from "../../models/routeTab"
 import ViewHeader from "../viewHeader"
 import { usePanelStateFromStateDispatchContext } from "../../hooks/usePanelState"
+import { IndividualPropertiesPanelProps } from "../propertiesPanel"
 
-interface Props {
+interface Props extends IndividualPropertiesPanelProps {
```
lemald, line 28: **suggestion (non-blocking):** Extending `IndividualPropertiesPanelProps` feels like the wrong "level" to be working at because the properties panel as a whole could have extra stuff not related to the header in the future. Instead, we could consider having some third type (unsure of the name) that captures just the shared properties that would be relevant and `IndividualPropertiesPanelProps` could (for now at least) just be equal to that type.

firestack: what do you think of 6abf46e3?

lemald (approved):

PR #2276, opened 2023-11-08
 cleanup(ts/components/nav): remove unused `allowViews` prop 

firestack:
Depends On: #2273
Context: https://github.com/mbta/skate/pull/2273#discussion_r1386682374
Asana Ticket: https://app.asana.com/0/1148853526253420/1205762199809559/f

---
  - https://app.asana.com/0/0/1205855740316544

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -71,7 +71,7 @@ export const AppRoutes = () => {
       </div>
       <VehiclesByRouteIdProvider vehiclesByRouteId={vehiclesByRouteId}>
         <div className="l-app__main">
-          <Nav allowViews>
+          <Nav>
             <Routes>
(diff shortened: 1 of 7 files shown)


lemald (approved):

firestack:
I'm still unsure on if we should remove this, @lemald, do you have a preference for keeping this and closing this PR (in the case we want to merge it in the future), or removing it now and adding it back later if we need it?

lemald:
> I'm still unsure on if we should remove this, @lemald, do you have a preference for keeping this and closing this PR (in the case we want to merge it in the future), or removing it now and adding it back later if we need it?

Hmm I'd err on the side of removing it if we don't need it knowing that we can always add it back later, but that's just my personal feeling.

PR #2277, opened 2023-11-08
Replace street view with directions in VPC

lemald:
Asana ticket: [⚙️ Switch Street View button in VPC to Get directions to bus](https://app.asana.com/0/1148853526253420/1205762196567449/f)

Note that per the ticket this doesn't need to be gated behind the test group.

### DIFF
--- a/assets/src/components/mapPage/vehiclePropertiesCard.tsx
+++ b/assets/src/components/mapPage/vehiclePropertiesCard.tsx
@@ -199,7 +199,7 @@ const VehicleNearestIntersection = ({
   )
 }
 
-const VehicleLocationStreetViewButton = ({ vehicle }: { vehicle: Vehicle }) => (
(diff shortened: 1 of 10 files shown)


assets/src/components/mapPage/vehiclePropertiesCard.tsx
```diff
@@ -200,12 +200,7 @@ const VehicleNearestIntersection = ({
 }
 
 const VehicleLocationStreetViewButton = ({ vehicle }: { vehicle: Vehicle }) => (
```
firestack, line 202: suggestion: maybe `VehicleLocationStreetViewButton` should be renamed or collapsed?

lemald: Oh heh, thanks for catching this, I definitely meant to rename it.

lemald: a8d12a49

PR #2278, opened 2023-11-13
feat(storybook): Add story for location marker icon

joshlarson:
This is a follow-up to [this PR](https://github.com/mbta/skate/pull/2271) to add a storybook story for the affected icon.

Question: I wasn't able to get the icon to load without being in the map context. It _feels_ like we should have a non-map version of the icon, but I sort of timeboxed the effort to add it. What do you all think?

### DIFF
--- a/assets/stories/skate-components/map/markers/locationMarker.stories.tsx
+++ b/assets/stories/skate-components/map/markers/locationMarker.stories.tsx
@@ -1,0 +1,67 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import React from "react"
+import { LocationMarker } from "../../../../src/components/mapMarkers"
+import { LocationDotIcon } from "../../../../src/helpers/icon"
+import { inMapDecorator } from "../../../../.storybook/inMapDecorator"
+import locationSearchResultFactory from "../../../../tests/factories/locationSearchResult"
+
+const location = locationSearchResultFactory.build({
+  latitude: 42.360082,
+  longitude: -71.05888,
+})
+
+const meta = {
+  args: { location },
+  argTypes: {
+    location: { table: { disable: true } },
+  },
+  render: ({ selected }) => {
+    return (
+      <LocationDotIcon
+        className={
+          "c-location-dot-icon" +
+          (selected ? " c-location-dot-icon--selected" : "")
+        }
+      />
+    )
+  },
+  component: LocationMarker,
+  parameters: {
+    layout: "centered",
+    stretch: false,
+  },
+} satisfies Meta<typeof LocationMarker>
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Unselected: Story = {
+  args: {
+    selected: false,
+  },
+}
+
+export const Selected: Story = {
+  args: {
+    selected: true,
+  },
+}
+
+const InMap: Story = {
+  decorators: [inMapDecorator],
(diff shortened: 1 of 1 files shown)


lemald:
> Question: I wasn't able to get the icon to load without being in the map context. It feels like we should have a non-map version of the icon, but I sort of timeboxed the effort to add it. What do you all think?

I'd say this component only really makes sense in a Leaflet map setting - it uses `<ReactMarker/>` which in turn uses the react-leaflet `<Marker/>` that does the under-the-hood effect synchronization between React and Leaflet. The icon itself is the `<LocationDotIcon/>` component. Because of the CSS involved in Leaflet itself I think it's most meaningful to look at how the icon renders in that environment, unless we decide we're also using this icon in other places, in which case it could make sense to have `<LocationDotIcon/>` surfaced directly in StoryBook somehow.

assets/stories/skate-components/map/markers/locationDotMarker.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const meta = {
+  component: ({ selected }) => {
+    return <LocationMarker location={location} selected={selected} />
+  },
+  parameters: {
+    layout: "fullscreen",
+    stretch: true,
+  },
+  decorators: [(Story) => <Story />, inMapDecorator],
```
lemald, line 22: **suggestion:** the anonymous function here appears to be a no-op.
```suggestion
  decorators: [inMapDecorator],
```

assets/stories/skate-components/map/markers/locationDotMarker.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import React from "react"
+import { LocationMarker } from "../../../../src/components/mapMarkers"
+import { inMapDecorator } from "../../../../.storybook/inMapDecorator"
+
+const location: LocationSearchResult = {
```
lemald, line 6: **suggestion:** not sure how much luck you've had working through the CI errors you're seeing, but adding `import { LocationSearchResult } from "../../../../src/models/locationSearchResult"` will help (and then you just have to add location to the args for the two specific stories).

assets/stories/skate-components/map/markers/locationDotMarker.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  name: "MBTA Offices",
+  address: "10 Park Plaza",
+  latitude: 42.360082,
+  longitude: -71.05888,
+}
+
+const meta = {
+  component: ({ selected }) => {
+    return <LocationMarker location={location} selected={selected} />
+  },
```
firestack, line 17: question: [is there benefit/what happens] if you pass the `LocationMarker` directly?
```suggestion
  component: LocationMarker,
  args: {
    location
  },
```

assets/stories/skate-components/map/markers/locationDotMarker.stories.tsx
joshlarson: I originally had that, but that led to this kind of meaningless input:
Screenshot 2023-11-14 at 10 49 22 AM

I guess it could be helpful for showing usage, but on the storybook page, that input doesn't have any effect and looks kind of ugly.

firestack (approved):

PR #2281, opened 2023-11-14
refactor: Create LocationMarkerIcon component and use it in Storybook

joshlarson:


firestack (approved):

assets/src/components/map/markers/locationMarker.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <LocationDotIcon
+      className={
+        "c-location-dot-icon" +
+        (selected ? " c-location-dot-icon--selected" : "")
+      }
+    />
+  )
+}
+
+export const LocationMarker = ({
```
lemald, line 18: **suggestion (non-blocking):** now that we also have the `<UserLocationMarker/>` component I've sort of wanted to come up with a more descriptive name for this to differentiate it (maybe something indicating that it's a geocoded location from a search result?), but just separating it out into its own module is already an improvement on its own.

joshlarson: `<LocationSearchMarker />`?

lemald: @joshlarson I like `<LocationSearchMarker/>`, were you going to make that change?

joshlarson: Thanks for reminding me though - it had completely slipped my mind.

joshlarson: @lemald Done!

lemald (approved):

PR #2282, opened 2023-11-14
feat: hide open map button when already on map page

lemald:
Asana ticket: [⚙️ Don't show open maps button in VPP on Search Maps](https://app.asana.com/0/1148853526253420/1205718173592000/f)

### DIFF
--- a/assets/src/components/app.tsx
+++ b/assets/src/components/app.tsx
@@ -22,13 +22,14 @@ import { allOpenRouteIds } from "../models/routeTab"
 import Nav from "./nav"
 import RightPanel from "./rightPanel"
 import { mapModeForUser } from "../util/mapMode"
-import { Ghost, VehicleInScheduledService } from "../realtime"
+import { Ghost, Vehicle, VehicleInScheduledService } from "../realtime"
 import MapPage from "./mapPage"
 import SearchPage from "./searchPage"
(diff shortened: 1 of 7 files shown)


assets/src/components/app.tsx
```diff
@@ -100,6 +91,19 @@ export const AppRoutes = () => {
...
+                  <RouteElement
+                    selectedVehicleOrGhost={selectedVehicleOrGhost}
+                    openView={openView}
+                    openMapEnabled={false}
+                    vppTabMode={vppTabMode}
+                    setTabMode={setTabMode}
+                    closeView={closeView}
+                  />
+                }
+              >
```
firestack, line 106: question: what's the rationale of splitting this(`RouteElement`) out into it's own component and putting `path={mapMode.path}` into it's own `Route`?

lemald: They're split out into separate `<Route/>`s to allow different values of `openMapEnabled` to be set. Splitting out `<RouteElement/>` just seemed natural because otherwise I'd be duplicating the logic in there about the placement of the `<Outlet/>`, conditional rendering of `<RightPanel/>`, etc.

firestack: oooooh that's super clean, I like it!

firestack (approved):

PR #2283, opened 2023-11-16
feat(frontend/MapPage):  fit drawer on screen for all viewport sizes

firestack:
Asana Ticket: https://app.asana.com/0/1148853526253420/1205703270277847/f

### DIFF
--- a/assets/css/_map_page.scss
+++ b/assets/css/_map_page.scss
@@ -45,7 +45,12 @@ $z-map-page-context: (
   flex: 0 0 auto;
   flex-direction: column;
   gap: 1rem;
-  width: $map-page-search-panel-width;
+  width: Min(
+    $map-page-search-panel-width,
+    /* Panel + Drawer(2rem) should have 100% viewport width - 10px to match leaflet margin */
+    calc((100vw - 2rem) - 10px)
+  );
+
   height: 100%;
   box-sizing: border-box;
   position: absolute;
@@ -77,6 +82,7 @@ $z-map-page-context: (
   }
 
   &--hidden {
+    width: $map-page-search-panel-width;
     left: calc(-1 * #{$map-page-search-panel-width} + 1rem);
 
     .c-search-display {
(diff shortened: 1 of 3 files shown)


assets/src/components/mapPage.tsx
```diff
@@ -324,6 +324,10 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
             />
           )}
         </div>
+        <button
+          onClick={toggleSearchDrawer}
+          className="c-map-page__panel-backdrop-button"
+        ></button>
```
firestack, line 330: I'm not a fan of this, I would've preferred to add a pseudo element to the `<DrawerTab/>` that stretched it's clickable area when on certain screen sizes and states, but this was easiest since I couldn't get that to work.

assets/css/_map_page.scss
```diff
@@ -85,6 +97,26 @@ $z-map-page-context: (
...
+      display: block;
+      // Force Flex + Posistion to fill area
+      width: 100%;
+
+      // Ignore flex layout and fill screen
+      position: absolute;
+      inset: 0;
+
+      // Inbetween panel and map
+      z-index: 150;
```
lemald, line 115: **suggestion:** could you define this in terms of the values from `$z-map-page-context`?

assets/src/components/mapPage.tsx
lemald: **suggestion:** I think this could be a self-closing tag.

assets/css/_map_page.scss
firestack: Thanks for the reminder!
12a3a0a2a54997d2e8a2ccba36eb0d39de5b7923

assets/src/components/mapPage.tsx
firestack: > **suggestion:** I think this could be a self-closing tag.

ebe5948b2a594acfefb5a8fee3fd70637df523a1

lemald (approved):

joshlarson (approved):

PR #2285, opened 2023-11-20
feat: Don't open the drawer on mobile when a search result is visible

joshlarson:
---
  - https://app.asana.com/0/0/1205703270277843

### DIFF
--- a/assets/src/components/mapPage.tsx
+++ b/assets/src/components/mapPage.tsx
@@ -223,6 +223,7 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
 
   // #region Search Drawer Logic
   const [searchOpen, setSearchOpen] = useState<boolean>(defaultSearchOpen)
+
   const toggleSearchDrawer = useCallback(
     () => setSearchOpen((open) => !open),
     [setSearchOpen]
@@ -249,13 +250,10 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
         case SelectedEntityType.RoutePattern:
           fullStoryEvent("RPC Opened", {})
       }
-      if (selectedEntity) {
-        setSearchOpen(!isMobile)
-      }
 
       dispatch(setSelectedEntity(selectedEntity))
     },
-    [dispatch, isMobile]
+    [dispatch]
   )
 
   const selectVehicleResult = useCallback(
@@ -265,11 +263,12 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
           type: SelectedEntityType.Vehicle,
           vehicleId: vehicleOrGhost.id,
         })
+        setSearchOpen(!isMobile)
       } else {
         setVehicleSelection(null)
       }
     },
-    [setVehicleSelection]
+    [setVehicleSelection, isMobile]
   )
 
   const selectLocationResult = (
@@ -282,6 +281,7 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
           location,
         })
       )
+      setSearchOpen(!isMobile)
     } else {
       dispatch(setSelectedEntity(null))
     }
@@ -347,6 +347,7 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
               setSelection={(...args) => {
                 setFollowerShouldSetZoomLevel(false)
                 setVehicleSelection(...args)
+                setSearchOpen(true)
               }}
               fetchedSelectedLocation={fetchedSelectedLocation}
               initializeRouteFollowerEnabled={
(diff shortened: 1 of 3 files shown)


lemald:
**question:** do you think it's possible to add an automated test for this logic? The existing logic to open the panel has a test in `mapPage.test.tsx`, you could probably do something similar and mock out `useScreenSize`.

joshlarson:
@lemald I added a test. I'm not sure I _love_ it, since it has a lot of setup, but the use case is also kind of a fiddly one. I would love to get a broader overview of how we test, and how we want tests to look in general.

And possibly whether I gave this feature the coverage it deserves with just one additional test (I did validate that the test fails when I get rid of the open/close logic).

assets/tests/testHelpers/mockHelpers.ts
```diff
@@ -92,3 +94,7 @@ export const mockTileUrls = (): void => {
     }
   })
 }
+
+export const mockScreenSize = (screenSize: DeviceType): void => {
+  ;(useScreenSize as jest.Mock).mockImplementation(() => screenSize)
```
firestack, line 99:
```suggestion
  jest.mocked(useScreenSize).mockReturnValue(screenSize)
```

firestack: We have a lot of this `;(<thing> as jest.Mock)` syntax in our codebase, and it's a little unclear why that was chosen over say, `jest.mocked`, so as of recently, we've been trying to convert to using `jest.mocked`.

But this isn't documented anywhere, hmm.

assets/tests/components/mapPage.test.tsx
```diff
@@ -696,7 +697,43 @@ describe("<MapPage />", () => {
...
+            query: { text: vehicle.runId! },
+          }),
+        })}
+      >
+        <MapPage />
+      </RealDispatchWrapper>
+    )
+
+    const mapSearchPanel = getMapSearchPanel()
+    expect(mapSearchPanel).toHaveClass("c-map-page__input-and-results--visible")
```
lemald, line 724: **suggestion (non-blocking):** I'd argue that we could remove this assertion since it's not actually related to the property being tested (that the drawer gets closed after selecting a result) - it's more testing the setup than the actual result.

lemald (approved):
One suggestion but overall LGTM!

assets/tests/components/mapPage.test.tsx
firestack: Since we cannot control the visible state of the panel via props or state, I'd argue that this is necessary to test that it's closing, because if this assumes that it starts at `--visible`, but it actually starts with `--hidden` then this test would silently succeed when the functionality may be failing.

This test file is fairly large though, so if another test is asserting that this same setup and state opens as `--visible`, I'd agree that it could be removed.

joshlarson: The reason I added it was what Kayla said - basically to highlight the contrast. I don't feel that strongly about it, but I also don't think that it's asserted elsewhere.

That highlighted contrast could be in another test right next to this one though. Let's see how that looks.

joshlarson: Hmmm - okay actually... two thoughts:

1. I copied the structure of this test from the one above it, and I worry about confusion if we have a bunch of tests right near each other that are structured differently. 
2. Splitting this test up will increase the length of the test file. Not necessarily the worst thing in the world, but I do want to be considerate about that kind of thing.

lemald: Yeah, those are both good points. I'm fine with keeping it as-is, but I wanted to throw it out there as an idea. Maybe something to consider as we write tests in the future.

assets/src/components/mapPage.tsx
```diff
@@ -243,9 +250,6 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
         case SelectedEntityType.RoutePattern:
           fullStoryEvent("RPC Opened", {})
       }
-      if (selectedEntity) {
```
joshlarson, line 246: @lemald @firestack I want to highlight this change, since it's different from what you approved. 

I moved the call to `setSearchOpen(true)` outside of `setVehicleSelection`, since on mobile some calls to `setVehicleSelection` (clicking on a vehicle from the map) should open the drawer, and others (clicking on a vehicle search result) shouldn't.

joshlarson: Another note: The collection of use cases has gotten complicated enough that I will feel a lot more comfortable merging this once I've added better test coverage.

PR #2286, opened 2023-11-20
fix: Center selected entity correctly based on whether the drawer is open

joshlarson:
---
  - https://app.asana.com/0/0/1205730670134122
  - https://app.asana.com/0/0/1204330988641900

### DIFF
--- a/assets/src/components/map/follower.tsx
+++ b/assets/src/components/map/follower.tsx
@@ -21,11 +21,16 @@ import { equalByElements } from "../../helpers/array"
 import { RecenterControl } from "./controls/recenterControl"
 import { defaultCenter } from "../map"
 
-export type UpdateMapFromPointsFn = (map: LeafletMap, points: LatLng[]) => void
+export type UpdateMapFromPointsFn = (
+  map: LeafletMap,
+  points: LatLng[],
+  shouldOffset: boolean
+) => void
 
 export interface FollowerProps {
   positions: LatLng[]
   onUpdate?: UpdateMapFromPointsFn
+  shouldOffset?: boolean
 }
 
 export const Follower = ({
@@ -33,6 +38,7 @@ export const Follower = ({
   onUpdate,
   isAnimatingFollowUpdate,
   shouldFollow = true,
+  shouldOffset = true,
 }: FollowerProps & InteractiveFollowState) => {
   const map = useMap()
   const [currentLatLngs, setCurrentLatLngs] = useState<LatLng[]>(positions)
@@ -48,9 +54,16 @@ export const Follower = ({
       if (isAnimatingFollowUpdate !== undefined) {
         isAnimatingFollowUpdate.current = true
       }
-      onUpdate?.(map, currentLatLngs)
+      onUpdate?.(map, currentLatLngs, shouldOffset)
     }
-  }, [map, shouldFollow, isAnimatingFollowUpdate, currentLatLngs, onUpdate])
+  }, [
+    map,
+    shouldFollow,
+    shouldOffset,
+    isAnimatingFollowUpdate,
+    currentLatLngs,
+    onUpdate,
+  ])
 
   return null
 }
@@ -61,7 +74,7 @@ export interface InteractiveFollowState {
   shouldFollow: boolean
   setShouldFollow: Dispatch<SetStateAction<boolean>>
 }
-// Gathers all state needed for the Follower to be able to display it's state
+// Gathers all state needed for the Follower to be able to display its state
 // as well as support turning off when interrupted
 
 export const useInteractiveFollowerState = (
@@ -181,7 +194,7 @@ export const usePickerContainerFollowerFn = () => {
 
 export const drawerOffsetAutoCenter =
   (useCurrentZoom: boolean): UpdateMapFromPointsFn =>
-  (map, points) => {
+  (map, points, shouldOffset) => {
     if (points.length === 0) {
       // If there are no points, blink to default center
       map.setView(defaultCenter, 13, { animate: false })
@@ -200,7 +213,7 @@ export const drawerOffsetAutoCenter =
     // In this case, we get the top left of the inner bounds by padding the left
     // with the distance from the right side of the VPC to the left side of the
     // map container
-    const topLeft = new Point(445, 0)
+    const topLeft = new Point(shouldOffset ? 445 : 0, 0)
 
     if (points.length === 1) {
       const currentZoom = map.getZoom()
--- a/assets/src/components/map/follower.tsx
+++ b/assets/src/components/map/follower.tsx
@@ -3,7 +3,7 @@ import Leaflet, {
   LatLng,
   LatLngExpression,
   Map as LeafletMap,
-  Point,
+  PointExpression,
 } from "leaflet"
 import React, {
   Dispatch,
@@ -180,7 +180,7 @@ export const usePickerContainerFollowerFn = () => {
 }
 
 export const drawerOffsetAutoCenter =
-  (useCurrentZoom: boolean, shouldOffset: boolean): UpdateMapFromPointsFn =>
+  (useCurrentZoom: boolean, topLeft: PointExpression): UpdateMapFromPointsFn =>
   (map, points) => {
     if (points.length === 0) {
       // If there are no points, blink to default center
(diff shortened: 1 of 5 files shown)


assets/src/components/map/follower.tsx
```diff
@@ -136,16 +149,18 @@ export const StatefulInteractiveFollower = (props: FollowerProps) => {
...
+    <>
+      <InterruptibleFollower {...props} />
+      <RecenterControl
+        position="topright"
+        active={props.shouldFollow}
+        onActivate={() => props.setShouldFollow(true)}
+      />
+    </>
+  )
+}
```
firestack, line 163: question: Why were braces added here?

joshlarson: d6aaf720

assets/src/components/map/follower.tsx
```diff
@@ -21,18 +21,24 @@ import { equalByElements } from "../../helpers/array"
...
   onUpdate?: UpdateMapFromPointsFn
+  shouldOffset?: boolean
 }
 
 export const Follower = ({
   positions,
   onUpdate,
   isAnimatingFollowUpdate,
   shouldFollow = true,
+  shouldOffset = true,
```
firestack, line 41: suggestion: what do you think about doing this via currying `drawerOffsetAutoCenter` instead of adding another prop and function argument to the base signature?

assets/src/components/map/follower.tsx
```diff
@@ -181,7 +195,7 @@ export const usePickerContainerFollowerFn = () => {
 
 export const drawerOffsetAutoCenter =
   (useCurrentZoom: boolean): UpdateMapFromPointsFn =>
-  (map, points) => {
+  (map, points, shouldOffset) => {
```
firestack, line 198: > https://github.com/mbta/skate/pull/2286#discussion_r1399475866

Maybe something like this? 

TBH, I'm not even sold that points should be an argument to this function or the base <Follower/> but I think that's a slightly larger refactor.
```suggestion
  (useCurrentZoom: boolean, topLeft: Point): UpdateMapFromPointsFn =>
  (map, points) => {
```

And then remove the [`topLeft` definition](https://github.com/mbta/skate/pull/2286/files#diff-fec870c97fb7c88e5929bb4949758d5dbbc8137119df5b14a1efcfe744971ab0R217)

assets/src/components/map/follower.tsx
firestack: Yep! Happy to pair whenever works! I'll be back from lunch after 1300

joshlarson: I did a thing: https://github.com/mbta/skate/pull/2286/commits/fe97cff54bd5c3e4ceaa8b125d372c886867847c

firestack: Cool!
I notice that even on mobile, the stop card still is following the safe area set here 

https://github.com/mbta/skate/blob/master/assets/src/components/mapPage/mapDisplay.tsx#L675-L676

What do you think of using the `MapSafeAreaContext` to give `drawerOffsetAutoCenter` a safe bounds and then updating the `MapSafeAreaContext.Provider` value from `mapPage` based on if the drawer is open?

assets/src/components/map/follower.tsx
```diff
@@ -234,6 +223,9 @@ export const drawerOffsetAutoCenter =
     }
   }
 
-export const fixedZoomDrawerOffsetAutoCenter = drawerOffsetAutoCenter(false)
+export const fixedZoomDrawerOffsetAutoCenter = drawerOffsetAutoCenter(
+  false,
+  [0, 0]
+)
```
firestack, line 229: related to snapshot changes
```suggestion
    // ```
    // vpcElement.getBoundingClientRect().right - mapElement.getBoundingClientRect().left
    //  -> 445
    // ```
    // Create a new inner bounds from the map bounds + "padding" to shrink the
    // inner bounds
    // In this case, we get the top left of the inner bounds by padding the left
    // with the distance from the right side of the VPC to the left side of the
    // map container
    new Point(445, 0)
)
```

joshlarson: Oooh I did change the defaults, didn't I

joshlarson: Related question - is this actually a problem? Having a non-zero default creates cognitive overhead, and maybe we don't need it?

firestack: Yes & no. If we don't need this default, we don't actually need `fixedZoomDrawerOffsetAutoCenter` at all, but then we need (and probably should) update all usages of `fixedZoomDrawerOffsetAutoCenter` to actually use `drawerOffsetAutoCenter`

or `fixedZoomDrawerOffsetAutoCenter` should forward the `topLeft` parameter, since it is setting `useCurrentZoom` for a reason.

joshlarson: Yeah, I just tested out location search, and that's still broken, so I think you're right that we need to rethink `fixedZoomDrawerOffsetAutoCenter`.

joshlarson: (Mentioned in Slack, and mentioning here for posterity as well...) the change suggested here doesn't affect the snapshot diff.

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -361,6 +366,7 @@ const SelectedRouteDataLayers = ({
     (selectedRoutePattern?.shape?.stops || []).map((s) => s.id)
   )
 
+  const { paddingTopLeft } = useContext(MapSafeAreaContext)
```
joshlarson, line 369: This isn't my favorite, because we have to repeat this like four times, but I don't ... thoroughly hate it? Since the thing that's repeated is just one line.

assets/src/components/stopCard.tsx
```diff
@@ -73,6 +73,13 @@ const StopCard = ({
       autoPanPadding={autoPanPadding?.padding || [20, 20]}
       autoPanPaddingTopLeft={autoPanPadding?.paddingTopLeft}
       autoPanPaddingBottomRight={autoPanPadding?.paddingBottomRight}
+      // Leaflet implements Popup as a singleton, which means that the
+      // autoPanPadding settings won't be updated while the map is
+      // in-memory. Setting a key that has a stringified version of
```
firestack, line 78:
```suggestion
      // autoPanPadding setting won't be updated because it is only
      // applied when the `<Popup/>` is created.
      // Setting a key that has a stringified version of
```

assets/src/components/stopCard.tsx
```diff
@@ -73,6 +73,13 @@ const StopCard = ({
...
+      // the autoPanPadding settings will cause the Popup to get
+      // re-rendered if that key changes, which means that changing
```
firestack, line 80:
```suggestion
      // re-created if that key changes, which means that changing
```

assets/src/components/mapPage/mapDisplay.tsx
firestack: What if you made a hook that wraps the context and the function?
```
const useFixedZoomDrawerOffsetFollowerFn() => {
	const { paddingTopLeft } = useContext(MapSafeAreaContext)
	return fixedZoomDrawerOffsetAutoCenter(paddingTopLeft)
}
```

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -672,7 +681,7 @@ const MapDisplay = ({
     >
       <MapSafeAreaContext.Provider
         value={{
-          paddingTopLeft: [445, 54],
+          paddingTopLeft: [shouldOffset ? 445 : 54, 54],
```
firestack, line 684: **suggestion**: If `MapPage` ever changes the size of it's panels, we'll have to also remember to update these values, in the case that `MapPage` needs to control the safe area, what do you think about making the safe area provided via a prop?

I think it'd also be simpler for testing (not that we've got many tests for this) because instead of asserting it's either 54 or 445, we can assert that the value updates when the prop updates.

joshlarson: Are you suggesting that the value of `value` should be passed in as a prop instead of `shouldOffset`? Or that in some way, we could pass the entire provider in?

firestack: That was what I was suggesting, but your comment made me realize that we could pull `MapSafeAreaContext.Provider` up into `MapPage` containing `MapDisplay` instead of needing to pass args into `MapDisplay` or manage any composition within `MapDisplay`.

Happy to walk through this together in a huddle or provide a code sample if you'd like 👍🏻.

firestack (approved):
I'd like to further discuss my [`MapSafeAreaContext` PR comment](https://github.com/mbta/skate/pull/2286#discussion_r1402481705) but otherwise LGTM 👍🏻

assets/src/components/mapPage/mapDisplay.tsx
firestack: Addressed in https://github.com/mbta/skate/pull/2286/commits/dfa22163d6a2ea3dcafd04fc49c92549c5bd977c

PR #2288, opened 2023-11-27
fix: Improve error handling on the test-groups page

joshlarson:
---
  - https://app.asana.com/0/0/1203373689404088

### DIFF
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -15,7 +15,7 @@ defmodule Skate.Settings.TestGroup do
   @spec create(String.t()) :: {:ok, t()} | {:error, Ecto.Changeset.t()}
   def create(name) do
     %DbTestGroup{}
-    |> DbTestGroup.changeset(%{name: name})
+    |> DbTestGroup.changeset(%{name: String.trim(name)})
     |> Skate.Repo.insert()
     |> case do
       {:ok, new_test_group} ->
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -81,4 +81,12 @@ defmodule Skate.Settings.TestGroup do
       users: db_test_group.users
     }
   end
+
+  @doc """
+  Deletes a test group with the given ID
+  """
+  @spec delete(integer()) :: nil
+  def delete(id) do
+    Skate.Repo.delete(%DbTestGroup{id: id})
+  end
 end
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -12,12 +12,21 @@ defmodule Skate.Settings.TestGroup do
 
   defstruct [:id, :name, users: []]
 
-  @spec create(String.t()) :: __MODULE__.t()
+  @spec create(String.t()) :: {:ok, t()} | {:error, Ecto.Changeset.t()}
   def create(name) do
-    %DbTestGroup{name: name}
-    |> Skate.Repo.insert!()
-    |> Skate.Repo.preload(:users)
-    |> convert_from_db_test_group()
+    %DbTestGroup{}
+    |> DbTestGroup.changeset(%{name: name})
+    |> Skate.Repo.insert()
+    |> case do
+      {:ok, new_test_group} ->
+        {:ok,
+         new_test_group
+         |> Skate.Repo.preload(:users)
+         |> convert_from_db_test_group()}
+
+      {:error, errored_changeset} ->
+        {:error, errored_changeset}
+    end
   end
 
   @spec get(integer()) :: t() | nil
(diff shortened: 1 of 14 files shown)


lib/skate_web/views/error_helpers.ex
```diff
@@ -4,4 +4,17 @@ defmodule SkateWeb.ErrorHelpers do
...
 
   use Phoenix.HTML
+
+  @doc """
+  Generates a tag for inlined form input errors.
+  """
+  def error_tag(form, field) do
+    case form.errors[field] do
+      {error, _} ->
+        content_tag(:span, error)
```
joshlarson, line 14: This has basically no styling right now, so it doesn't look particularly nice. I'm on the fence about whether to change that, because:
1. It's an internal tool that's used basically only by us.
2. The exact way to give it nice styles will change with HEEx.
3. If we do declarative test groups (which I think is a good idea), then this error state will go away.

lib/skate/settings/test_group.ex
```diff
@@ -12,12 +12,20 @@ defmodule Skate.Settings.TestGroup do
...
+         |> DbTestGroup.changeset(%{name: name})
+         |> Skate.Repo.insert() do
+      {:ok, new_test_group} ->
+        {:ok,
+         new_test_group
+         |> Skate.Repo.preload(:users)
+         |> convert_from_db_test_group()}
+
+      {:error, errored_changeset} ->
+        {:error, errored_changeset}
```
lemald, line 27: **suggestion:** should also write tests for this error case in `Skate.Settings.TestGroupTest`.

lib/skate/settings/test_group.ex
```diff
@@ -12,12 +12,20 @@ defmodule Skate.Settings.TestGroup do
...
-  @spec create(String.t()) :: __MODULE__.t()
+  @spec create(String.t()) :: {:ok, t()} | {:error, Ecto.Changeset.t()}
   def create(name) do
-    %DbTestGroup{name: name}
-    |> Skate.Repo.insert!()
-    |> Skate.Repo.preload(:users)
-    |> convert_from_db_test_group()
+    case %DbTestGroup{}
+         |> DbTestGroup.changeset(%{name: name})
+         |> Skate.Repo.insert() do
```
firestack, line 19: suggestion(non-blocking): AFAIK, you can add the case as another pipeline entry, and I think it's easier to read this way.
```suggestion
    %DbTestGroup{}
    |> DbTestGroup.changeset(%{name: name})
    |> Skate.Repo.insert() 
    |> case do
```

joshlarson: Oh cool - I didn't know you could do that

lemald (approved):

lib/skate_web/templates/layout/barebones.html.eex
```diff
@@ -17,6 +17,13 @@
         <link rel="apple-touch-icon" href="<%= static_href.(@conn, "/images/mbta-logo-t-180.png") %>" type="image/png">
         <link rel="icon" href="<%= static_href.(@conn, "/images/mbta-logo-t-favicon.png") %>" sizes="32x32" type="image/png">
         <link rel="icon" href="<%= static_href.(@conn, "/favicon.ico") %>" sizes="16x16" type="image/vnd.microsoft.icon">
+
+        <style>
+            .error-tag {
```
joshlarson, line 22: Added this here because it feels like the style attached to `error_tag` ought to be available to anything in admin-land.

lib/skate_web/controllers/test_group_controller.ex
```diff
@@ -15,15 +15,27 @@ defmodule SkateWeb.TestGroupController do
...
   def post(conn, params) do
-    TestGroup.create(params["name"])
-
-    redirect(conn, to: SkateWeb.Router.Helpers.test_group_path(conn, :index))
+    case TestGroup.create(params["test_group"]["name"]) do
+      {:ok, _} ->
+        redirect(conn, to: SkateWeb.Router.Helpers.test_group_path(conn, :index))
+
+      {:error, changeset} ->
+        test_groups = TestGroup.get_all() |> Enum.sort_by(& &1.name)
```
lemald, line 30: **nitpick:** see previous note on single pipes

test/skate/settings/test_group_test.exs
```diff
@@ -67,4 +90,14 @@ defmodule Skate.Settings.TestGroupTest do
...
     end
   end
+
+  describe "delete/1" do
+    test "deletes a test group" do
+      {:ok, test_group} = TestGroup.create("group to delete")
+
+      TestGroup.delete(test_group.id)
+
+      assert TestGroup.get_all() |> Enum.empty?()
```
lemald, line 100: **nitpick:** single pipe

lemald (approved):
Additional changes look good to go! 👍

PR #2289, opened 2023-11-28
Full-screen panels on small screens

lemald:
Asana ticket: [⚙️ Views or panels should be full screen for < 800px](https://app.asana.com/0/1148853526253420/1205785606005457/f)

There were a couple of places that referenced a breakpoint of 480px from the old breakpoint variables, but I updated them to point to the new one because the new breakpoint was actually the relevant one for that context (the thing that determines which breakpoint the CSS should apply at is based on the `useScreenSize` breakpoints).

I followed the same pattern as with the mobile portrait nav where the nav components are explicitly hidden when a view is opened over them - for one thing the order of the z-indexes isn't right to have the panel go on top of the nav and switching them around causes issues (the drop shadow of some panels bleeds out over the top nav on desktop).

### DIFF
--- a/assets/css/_late_view.scss
+++ b/assets/css/_late_view.scss
@@ -10,8 +10,8 @@ $late-view-background-color: $color-bg-light;
   top: 0;
   z-index: map-get($z-page-layout-context, "view");
 
-  @media screen and (max-width: 700px) {
-    width: 100%;
+  @media screen and (max-width: map-get($breakpoints, "max-mobile-landscape-tablet-portrait-width")) {
+    width: 100vw;
   }
 }
 
--- a/assets/src/components/nav.tsx
+++ b/assets/src/components/nav.tsx
@@ -6,6 +6,8 @@ import LeftNav from "./nav/leftNav"
 import TopNav from "./nav/topNav"
 import MobilePortraitNav from "./nav/mobilePortraitNav"
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
+import { usePanelStateFromStateDispatchContext } from "../hooks/usePanelState"
+import { OpenView } from "../state/pagePanelState"
 
 interface Props {
   children?: React.ReactNode
@@ -15,19 +17,30 @@ const Nav: React.FC<Props> = ({ children }) => {
   const [, dispatch] = useContext(StateDispatchContext)
   const deviceType = useScreenSize()
 
+  const {
+    currentView: { openView, selectedVehicleOrGhost },
+  } = usePanelStateFromStateDispatchContext()
+  const isViewOpen =
+    openView !== OpenView.None || (selectedVehicleOrGhost && true) || false
+
+  const navVisibilityStyle = isViewOpen ? "hidden" : "visible"
+
   switch (deviceType) {
     case "mobile":
       return (
         <div className="l-nav--narrow">
           <div className="l-nav__app-content">{children}</div>
-          <MobilePortraitNav />
+          <MobilePortraitNav isViewOpen={isViewOpen} />
         </div>
       )
     case "mobile_landscape_tablet_portrait":
       return (
         <div className="l-nav--medium">
           <div className="l-nav__app-content">{children}</div>
-          <div className="l-nav__nav-bar l-nav__nav-bar--left">
+          <div
+            className="l-nav__nav-bar l-nav__nav-bar--left"
+            style={{ visibility: navVisibilityStyle }}
+          >
             <LeftNav
               toggleMobileMenu={() => dispatch(toggleMobileMenu())}
               defaultToCollapsed={true}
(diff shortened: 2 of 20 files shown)


assets/src/components/nav.tsx
```diff
@@ -15,19 +17,30 @@ const Nav: React.FC<Props> = ({ children }) => {
   const [, dispatch] = useContext(StateDispatchContext)
   const deviceType = useScreenSize()
 
+  const {
+    currentView: { openView, selectedVehicleOrGhost },
+  } = usePanelStateFromStateDispatchContext()
+  const isViewOpen =
+    openView !== OpenView.None || (selectedVehicleOrGhost && true) || false
+
```
firestack, line 25: suggestion(non-blocking): what do you think about computing this in `usePanelStateForViewState`?

lemald: Oh I like that idea! Will make the change.

lemald: 26a627da

firestack: I like it!
I do have a question: Do you think that it is confusing that `isViewOpen` is not under `currentView`?
I prefer that `currentView` contains only the (reducer) state of the page, but that makes me question if we should be exposing `curentView` as it's own object vs it's properties being added to the "base object".

I am fine either way, just wondering your opinion.

lemald: Hmm, good question, though I think for the reason you stated I prefer not adding `isViewOpen` to `currentView`. Maybe hoisting the properties of `currentView` to the main object would make sense at some point but I don't think that it's worth doing right now.

assets/css/_late_view.scss
```diff
@@ -10,8 +10,8 @@ $late-view-background-color: $color-bg-light;
   top: 0;
   z-index: map-get($z-page-layout-context, "view");
 
-  @media screen and (max-width: 700px) {
-    width: 100%;
+  @media screen and (max-width: map-get($breakpoints, "max-mobile-landscape-tablet-portrait-width")) {
+    width: 100vw;
```
firestack, line 14: suggestion: what do you think about using `inset: 0` instead?
```suggestion
    inset: 0;
```

assets/src/components/nav.tsx
```diff
@@ -15,19 +16,26 @@ const Nav: React.FC<Props> = ({ children }) => {
...
         </div>
       )
     case "mobile_landscape_tablet_portrait":
       return (
         <div className="l-nav--medium">
           <div className="l-nav__app-content">{children}</div>
-          <div className="l-nav__nav-bar l-nav__nav-bar--left">
+          <div
+            className="l-nav__nav-bar l-nav__nav-bar--left"
+            style={{ visibility: navVisibilityStyle }}
```
firestack, line 37: question: just wondering, why `visibility` here instead of something like the `hidden` attribute?

firestack: Hmm, I feel like I'd need to know why that was chosen in `<MobilePortraitNav/>` originally. Using `hidden` instead of visibility would change the way it works(`visibility: visible` keeps the element taking up layout space, where `hidden` does `display: none`), and I'm not sure if there are assumptions else where that require that. But I would prefer `hidden` in both cases if there are no issues.

assets/tests/hooks/usePanelState.test.tsx
```diff
@@ -140,6 +140,53 @@ describe("usePanelStateForViewState", () => {
...
+  describe("isViewOpen", () => {
+    test("returns false when no view open", () => {
+      const path = PagePath.Ladders
+      const pageViewState = pageViewFactory.build({
+        openView: OpenView.None,
+        selectedVehicleOrGhost: null,
+      })
+
+      const { isViewOpen } = usePanelStateForViewState(
+        viewFactory.build({ state: { [path]: pageViewState } }),
```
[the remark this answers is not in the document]
lemald: Ah, I was sort of struggling with finding the right combination of factories to use here. That is simpler and I like it better, I'll go ahead and use it.

assets/css/_late_view.scss
lemald: So unfortunately in part because of how the Late View page is implemented (read: not very responsive) this doesn't actually result in it cleanly taking up the full page (I don't know how to best describe it but you can try it out quickly and see). There's also the matter that we want it to cover the nav, but `inset: 0` would just cause it to take up the full size of its container which is the app container element.

assets/src/components/nav.tsx
firestack: It _seems_ like using `hidden`` would work? https://github.com/mbta/skate/pull/1719#discussion_r967366692

assets/css/_late_view.scss
firestack: ah right, we'd need to also add `position: fixed;` to make it use the viewport as it's target.

Part of me worries about using `100vw` if the panel could accidentally get moved horizontally somehow, so I would _prefer_ that we use `fixed` and `inset`, but I'm considering this non-blocking and up to you.

assets/src/components/nav.tsx
lemald: I'll play around with it a bit and see if there are any drawbacks to `hidden`.

PR #2290, opened 2023-11-29
feat(sb/colors): configure bootstrap colors and export to storybook

firestack:
# Summary
This does a few things
1. Replaces our usage of `node-sass` with dart-sass (`sass`) | (commit: "[cleanup: replace node-sass with sass](https://github.com/mbta/skate/commit/c4fff648933747971a70ba3d0fc4b6c51740f609)")
    - This was done so that we can use the `@use` scss rule to import colors selectively
    - And because the official docs for `sass-loader` say to prefer using `dart-sass`
    - We also only implicitly depended on `node-sass`, so this makes our dependency on our sass compiler explicit.
2. Pulls out color variables into their own file
3. Displays those color values in Storybook | (commit: "[feat(ts/stories/colors): display colors in Colors.mdx](https://github.com/mbta/skate/commit/3a5321994ce20326bcb17a8691fa3fb3f26c1d64)")

## Details about "2. Pull out color variables into their own file"
Pulling out color variables into their own file that could be transparently used by our existing scss without any deep changes, and making it available to typescript and bootstrap ended up with me making a few choices.
I essentially took two approaches. One tries to have all color variables and "patch panels" in a single file, but mixes concerns between css variables and scss variables. The other accepts the "tiers" of colors that we have and reorganizes them into their own importable files which depend on each other in a single direction.

The thinking was that ideally, when writing bespoke components or customizing _after_ bootstrap, we would use css variables for clarity in the output styles instead of "raw" hex values. The benefit here would be that when you look at, say a `background-color` property, in dev tools, the color would be easily identifiable via it's name in css variables (e.g., `var(--color-info)` or `var(--color-strawberry-900)` instead of `#269c95` and `#4b0828` respectively)

So any input on the second approach I tried would be appreciated, if you'd like to review that, you can review the state of the repo before and after the commit "[refactor(css): remove circular reference in color definitions](https://github.com/mbta/skate/commit/cb2920ea12764815d2662f251323c4c48b5874ef)".

---

Asana Ticket: [⚙️ Dev gets bootstrap colors set up and in storybook](https://app.asana.com/0/1148853526253420/1205836854144669/f)


lemald:
@firestack I'll need to chew over this some more tomorrow, but I've given it a first pass and I'm thinking through the approach. 👍

lemald (approved):
Okay, after thinking over the organization and trying to wrap my head around all of it I think this is a good approach for now. Maybe as we move forward with building out our design system and CSS code there will be opportunities to simplify, but in the meantime this opens up possibilities to use Bootstrap and improves the usefulness of Storybook, which is nice. 👍

PR #2294, opened 2023-12-01
feat(elixir/1.15): update to elixir 1.15 and OTP 26

firestack:
Asana Ticket: [⚙️ Upgrade to Elixir 1.15 and Erlang/OTP 26](https://app.asana.com/0/1148853526253420/1206028632842971/f)


lemald (approved):
Looks good if CI continues to pass after that test fix! 🤞

joshlarson (approved):

lemald:
@firestack I see you're still having failures - `Bypass.pass` might also be worth a look?

firestack:
> @firestack I see you're still having failures - `Bypass.pass` might also be worth a look?

I was looking at using that but I was unclear on what it actually would do based on it's documentation. I'm trying it in the latest push.

test/realtime/vehicle_test.exs
```diff
@@ -682,7 +682,7 @@ defmodule Realtime.VehicleTest do
       assert encoded_string =~ "\"route_id\":\"28\""
 
       assert encoded_string =~
-               "\"data_discrepancies\":[{\"attribute\":\"trip_id\",\"sources\":[{\"id\":\"swiftly\",\"value\":\"swiftly-trip-id\"},{\"id\":\"busloc\",\"value\":\"busloc-trip-id\"}]},{\"attribute\":\"route_id\",\"sources\":[{\"id\":\"swiftly\",\"value\":null},{\"id\":\"busloc\",\"value\":\"busloc-route-id\"}]}]"
+               "\"data_discrepancies\":[{\"sources\":[{\"id\":\"swiftly\",\"value\":\"swiftly-trip-id\"},{\"id\":\"busloc\",\"value\":\"busloc-trip-id\"}],\"attribute\":\"trip_id\"},{\"sources\":[{\"id\":\"swiftly\",\"value\":null},{\"id\":\"busloc\",\"value\":\"busloc-route-id\"}],\"attribute\":\"route_id\"}]"
```
firestack, line 685: Maybe this should be removed and be replaced with something like this?
```
assert vehicle == (vehicle |> Jason.encode!() |> Jason.decode!())
```

firestack: We could even reduce the scope down to just the `data_discrepancies` field to be closer to this assertion.

firestack:
^ Running 6 pipelines in CI to look for any more flakeyness

lemald:
> I think the latest failures are related to map key order not being specified?

Oh huh I must have been looking at an old test run, then, so it seems like the use of `stub` fixed that particular error?

test/realtime/vehicle_test.exs
lemald: Yeah you might also need to deal with string vs. atom keys but overall I like both of those ideas, maybe you could also integrate the other couple of assertions in this test somehow? Like make a `decoded_vehicle` by decoding `encoded_string` and then asserting things about specific fields?

firestack:
> > @firestack I see you're still having failures - `Bypass.pass` might also be worth a look?
> 
> I was looking at using that but I was unclear on what it actually would do based on it's documentation. I'm trying it in the latest push.

> > I think the latest failures are related to map key order not being specified?
> 
> Oh huh I must have been looking at an old test run, then, so it seems like the use of `stub` fixed that particular error?

@lemald Seems that didn't solve everything :c 

<figure>

```
  1) test start_link handles api events (Api.StreamTest)
     test/api/stream_test.exs:61
     ** (exit) shutdown
```

<figcaption>

-- https://github.com/mbta/skate/actions/runs/7089344655/job/19293889162?pr=2294

</figcaption>
</figure>

Do you know what the `exit` signifies in this context? I don't really understand how the GenServer & Process lifecycle fits into testing.

lemald:
> > > @firestack I see you're still having failures - `Bypass.pass` might also be worth a look?
> > 
> > 
> > I was looking at using that but I was unclear on what it actually would do based on it's documentation. I'm trying it in the latest push.
> 
> > > I think the latest failures are related to map key order not being specified?
> > 
> > 
> > Oh huh I must have been looking at an old test run, then, so it seems like the use of `stub` fixed that particular error?
> 
> @lemald Seems that didn't solve everything :c
> 
> ```
>   1) test start_link handles api events (Api.StreamTest)
>      test/api/stream_test.exs:61
>      ** (exit) shutdown
> ```
> 
> -- https://github.com/mbta/skate/actions/runs/7089344655/job/19293889162?pr=2294
> 
> Do you know what the `exit` signifies in this context? I don't really understand how the GenServer & Process lifecycle fits into testing.

Ah right that's the failure I'm thinking of, but it's not the same test you were just working on (I had assumed it was). That is kind of a weird failure, and it doesn't give us much to work with. :-/

test/realtime/vehicle_test.exs
firestack: PTALA 35b21a404b8e78099b3f4226ce3a27b7c0d8a878 🙏🏻

test/skate_web/controllers/page_controller_test.exs
```diff
@@ -173,7 +173,7 @@ defmodule SkateWeb.PageControllerTest do
       conn = get(conn, "/")
 
       assert html_response(conn, 200) =~
-               "data-map-limits=\"{&quot;east&quot;:3,&quot;north&quot;:1,&quot;south&quot;:2,&quot;west&quot;:4}\""
+               "data-map-limits=\"{&quot;north&quot;:1,&quot;south&quot;:2,&quot;east&quot;:3,&quot;west&quot;:4}\""
```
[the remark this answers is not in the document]
joshlarson: A possible option could be asserting on the substrings that are guaranteed to be there:
```
assert html_response(conn, 200) =~ "data-map-limits=\"{&quot;"
assert html_response(conn, 200) =~ "&quot;east&quot;:3"
assert html_response(conn, 200) =~ "&quot;north&quot;:1"
assert html_response(conn, 200) =~ "&quot;south&quot;:2"
assert html_response(conn, 200) =~ "&quot;west&quot;:4"
```

This has the downside that it's _technically_ not as strong an assertion as what we currently have, but also what we currently have is a little too strong, since I don't think there are any guarantees that the keys will stay in that order.

But also happy to wait on changing this until this test actually starts flaking (if it ever does).

test/realtime/vehicle_test.exs
```diff
@@ -675,14 +675,47 @@ defmodule Realtime.VehicleTest do
...
-      assert encoded_string =~ "\"id\":\"y1261\""
-
-      assert encoded_string =~ "\"route_id\":\"28\""
-
-      assert encoded_string =~
-               "\"data_discrepancies\":[{\"attribute\":\"trip_id\",\"sources\":[{\"id\":\"swiftly\",\"value\":\"swiftly-trip-id\"},{\"id\":\"busloc\",\"value\":\"busloc-trip-id\"}]},{\"attribute\":\"route_id\",\"sources\":[{\"id\":\"swiftly\",\"value\":null},{\"id\":\"busloc\",\"value\":\"busloc-route-id\"}]}]"
+      decoded_vehicle =
+        vehicle
+        |> Jason.encode!()
+        |> Jason.decode!(keys: :atoms!)
```
lemald, line 681: Oooh I didn't know about `keys: :atoms!`, that's a good thing to be aware of.

lemald (approved):
Latest changes all look good. 👍

PR #2296, opened 2023-12-01
fix(ex/credo): Fix single-pipe issues

joshlarson:
No ticket

Inspired by [this comment](https://github.com/mbta/skate/pull/2288#discussion_r1411198007) on an earlier PR, I decided to enable the single pipe rule and see what happened. 

...this is what happened!


lib/schedule/health/checkers/timepoints_checker.ex
```diff
@@ -19,7 +19,7 @@ defmodule Schedule.Health.Checkers.TimepointsChecker do
     timepoints_on_route_fn =
       Application.get_env(:skate_web, :timepoints_on_route_fn, &Schedule.timepoints_on_route/1)
 
-    length = timepoints_on_route_fn.(route_id) |> length()
+    length = route_id |> timepoints_on_route_fn.() |> length()
```
joshlarson, line 22: In most cases, I replaced the single pipe with no pipe at all. There were a few, like this one, where it seemed better to make it a multiple-pipe deal.

test/skate_web/controllers/report_controller_test.exs
```diff
@@ -8,25 +8,23 @@ defmodule SkateWeb.ReportControllerTest do
...
 
       assert html_response(conn, 200) =~ "User settings"
     end
   end
 
   describe "run/2" do
     @tag :authenticated
-    test "when not an admin, redirects to unauthorized page", %{conn: conn} do
-      conn = conn |> get(SkateWeb.Router.Helpers.report_path(conn, :run, "user_settings"))
+    test "when not an admin, redirects to unauthorized page", %{onn} do
```
joshlarson, line 26: A macro gone rogue!

test/skate_web/controllers/notification_read_states_controller_test.exs
```diff
@@ -104,24 +102,23 @@ defmodule SkateWeb.NotificationReadStatesControllerTest do
...
-            select: {nu.notification_id, nu.user_id},
-            where: nu.state == ^read_state
+        Enum.sort(
+          Skate.Repo.all(
+            from(nu in DbNotificationUser,
+              select: {nu.notification_id, nu.user_id},
+              where: nu.state == ^read_state
+            )
           )
         )
```
firestack, line 112: suggestion(non-blocking): what do you think about turning this into a multi-pipe again?
```suggestion
        from(nu in DbNotificationUser,
            select: {nu.notification_id, nu.user_id},
            where: nu.state == ^read_state
        )
        |> Skate.Repo.all()
        |> Enum.sort()
```

firestack: Though, I would expect `Credo.Check.Readability.NestedFunctionCalls` to call this out...

joshlarson: Done.

Perhaps I'll do a follow-up PR to enable `NestedFunctionCalls` as well.

joshlarson: 👍 Trying to think of how to do that while avoiding conflicts

firestack (approved):

PR #2301, opened 2023-12-06
Add Typography page to Storybook

lemald:
Asana ticket: [⚙️ Dev gets bootstrap typography set up and in storybook](https://app.asana.com/0/1148853526253420/1205836854144672/f)


assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,100 @@
...
+import "../../css/storybook_reboot.scss"
+
+<Meta title="Typography" />
+
+# Global Settings / Style
+
+Examples of typography styles found in Skate, based on defaults from
+Reboot with some modifications.
+
+<div class="sb-unstyled">
```
firestack, line 11: suggestion: Can we use `<Unstyled/>` here instead? 
https://storybook.js.org/docs/api/doc-block-unstyled

firestack: Same for the rest of the document which uses this class

lemald: Ah I missed that when skimming through the available doc blocks, I'll give it a try!

assets/stories/skate-design/Typography.mdx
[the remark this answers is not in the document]
firestack: suggestion(non-blocking): Depending on if design wants this as well, I think we should have a few more things from the [bootstrap typography doc page](https://getbootstrap.com/docs/5.3/content/typography/#global-settings) in here.

For instance, I think in addition to the unstyled list, we should show off the inline list, and maybe the definition list examples that are in the docs.

assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+You can use the <ins>inserted tag</ins> for text that has been added to the document.
+
+You can use the <u>underline tag</u> for underlined text.
+
+You can use the <small>small tag</small> for text that is to be treated as fine print.
+
+You can use the <strong>strong tag</strong> for bold text.
+
+You can use the <em>emphasis tag</em> for italicized text.
```
firestack, line 67: suggestion: I like this! What do you think about adding a `<Source/>` block which shows this code so it's more explicit where the tags are in the text, and how it's being used?

assets/stories/skate-design/Typography.mdx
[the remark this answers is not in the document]
lemald: I think in any event we should definitely link to the part of the Bootstrap docs that this page roughly corresponds to.

firestack, line 1: suggestion(non-blocking): Depending on if design wants this as well, I think we should have a few more things from the [bootstrap typography doc page](https://getbootstrap.com/docs/5.3/content/typography/#global-settings) in here.

For instance, I think in addition to the unstyled list, we should show off the inline list, and maybe the definition list examples that are in the docs.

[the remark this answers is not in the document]
lemald: 42b00513

assets/stories/skate-design/Typography.mdx
lemald: It looks like the `<Source/>` block is actually a little idiosyncratic (you can't just wrap it around code and get what you'd expect, to insert literal source code you need to pass it in as a prop): https://storybook.js.org/docs/api/doc-block-source

How do you feel about just putting it in a `<code>` block instead?

assets/stories/skate-design/Typography.mdx
lemald: My feedback from design so far has been that directly mimicking what's in the Figma document is good, but I can ask if there's any interest in those specific additions. 👍

assets/stories/skate-design/Typography.mdx
firestack: I'd be interested to see a solution using `<code>`, but I'm not sure you're going to get what you expect given the markdown parser is also present.

A few things I like about the `<Source/>` element is
1. It can skip out on the markdown rendering by passing in the code as a prop string
1. it comes with some decent styling 
image
1. It has syntax highlighting

It's not quite perfect, and if you'd prefer something else I'd be interested to see.

Unfortunately, you can't just take the existing markdown and put it in the source block, because it won't have the `<p>` tags, but it's also fine without them IMO.

```mdx
<Source language="html" code={`
You can use the <mark>mark tag</mark> to highlight text.

You can use the <del>deleted tag</del> for deleted text.

You can use the <s>strikethrough tag</s> for text that is no longer relevant or accurate.

You can use the <ins>inserted tag</ins> for text that has been added to the document.

You can use the <u>underline tag</u> for underlined text.

You can use the <small>small tag</small> for text that is to be treated as fine print.

You can use the <strong>strong tag</strong> for bold text.

You can use the <em>emphasis tag</em> for italicized text.
`}/>
```

assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,126 @@
...
+The sections that follow below are all rendered with Skate, as opposed
+to Storybook, styling.
+
+<Unstyled>
+  <p>
+    {
+      "The default font is Inter. Body text is 1rem, 400 weight, $gray-900(#141524)."
+    }
+    <br />
+    {"In paragraphs, we use 150% line height."}
```
firestack, line 21: suggestion: It's a little confusing to go from reading about the font to talking about the line height. Could we move the font text into it's own paragraph and copy the text from figma?
```suggestion
  The default font is Inter. Body text is 1rem, 400 weight, $gray-900(#141524).

  <p>
    {"Paragraph text. When body text is"}
    <br />
    {"in paragraphs, we use 150% line height."}
```

I'd also be happy with using the text from [bootstrap's line-height docs](https://getbootstrap.com/docs/5.3/utilities/text/#line-height), instead of trying for force a break within text. It might be good to have the different line-height utilities shown anyway?

assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,126 @@
...
+Reboot with some modifications. This roughly follows the outline of
+the <a href="https://getbootstrap.com/docs/5.3/content/typography/">
+Bootstrap typography documentation</a> adapted specifically for Skate.
+The sections that follow below are all rendered with Skate, as opposed
+to Storybook, styling.
+
+<Unstyled>
+  <p>
+    {
+      "The default font is Inter. Body text is 1rem, 400 weight, $gray-900(#141524)."
```
firestack, line 18: suggestion(non-blocking): since we're "in the land of links", if you have time to figure it out, do you think you could try finding a way to link to this color on the colors page from this text?

These storybook docs _may_ help with figuring out linking? https://storybook.js.org/docs/writing-docs/mdx#linking-to-other-stories-and-pages

The biggest challenge I see is that the `gray` swatches don't have a targetable `id`, and neither does the `gray-900` varient. So somehow you'd need to add an id/anchor to the colors page that could be linked to? (which, actually may be easier than I'm expecting)

assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,126 @@
...
+      </span>
+      <ul>
+        <li>Like this.</li>
+        <li>For example.</li>
+      </ul>
+    </li>
+    <li>
+      <span>
+        Note that for now <code>_base.scss</code> interferes with the above
+        example.
```
firestack, line 122: suggestion(non-blocking): what do you think about using markdown for the inline code instead of tags? The styling is slightly different since we're getting storybook styles applied when it uses markdown, but I think that's potentially good because it's more uniform?

```suggestion
        This is a list with the `list-unstyled` style applied.
      </span>
    </li>
    <li>It appears completely unstyled but is still structurally a list.</li>
    <li>However, it only applies to its immediate children.</li>
    <li>
      <span>
        {
          "A nested list will get normal formatting unless the class is added to it."
        }
      </span>
      <ul>
        <li>Like this.</li>
        <li>For example.</li>
      </ul>
    </li>
    <li>
      <span>
        Note that for now `_base.scss` interferes with the above
        example.
```

assets/stories/skate-design/Typography.mdx
```diff
@@ -0,0 +1,126 @@
...
+    <li>
+      <span>
+        {
+          "A nested list will get normal formatting unless the class is added to it."
+        }
+      </span>
+      <ul>
+        <li>Like this.</li>
+        <li>For example.</li>
+      </ul>
```
firestack, line 117: question(non-blocking): do you know why this is rendering without a `::marker`, unlike the [bootstrap typography docs on unstyled lists](https://getbootstrap.com/docs/5.3/content/typography/#unstyled)?

I'm a little confused, especially since the copy says that the markers should be there. I tried looking into the CSS last night but I didn't have any luck figuring this out so I figured I'd ask.
image.png

firestack: Is this related to this comment?
<figure>

> issue with the MDX markdown towards the end of the page

<figcaption>
&mdash; https://github.com/mbta/skate/pull/2301#issuecomment-1843715069
</figcaption>

lemald: `_base.scss` sets `list-style-type: none` on all `ul` elements, which is the interference I was alluding to there. I guess this raises a philosophical point about how much of the current styling this page should reflect or if it should just be a sort of "clean slate" view on what using only Bootstrap-based styles would do, which I have mixed feelings on.

firestack: suggestion(non-blocking): (not committed to this suggestion, just a thought) What do you think about moving this note about the interference into this part? 

> A nested list will get normal formatting unless the class is added to it

lemald: 0a4e3515

assets/stories/skate-design/Typography.mdx
lemald: Hmm, I'm looking at the color page with the inspector and not finding much in the way of useful IDs to link to particular parts on the page. Unless you have a specific idea in mind I haven't stumbled upon yet I don't think this is feasible, for now at least.

firestack: This may be the wrong tag(but it also strangely feels like the right one?), but I think you could add an ID to use by putting an anchor above the gray `<ColorItem/> entry

```html
<a id="color-gray" />
```

lemald: ba49be33

firestack: Thanks!

firestack (approved):
LGTM, thanks @lemald!

PR #2322, opened 2023-12-11
feat(ex/phoenix): Migrate EEx templates to HEEx

joshlarson:
---
  - https://app.asana.com/0/0/1206113388125709


lib/skate_web/templates/layout/app.html.heex
```diff
@@ -51,7 +47,7 @@
     <main role="main" class="container">
       <%= @inner_content %>
     </main>
-    <script type="text/javascript" src="<%= static_href.(@conn, "/js/app.js") %>"></script>
-    <%= _drift(assigns) %>
+    <script type="text/javascript" src={~p"/js/app.js"}></script>
```
lemald, line 50: **question:** on this and other places where static paths are replaced - does this still get turned into a URL pointing to cdn.mbta.com rather than skate.mbta.com?

lemald:
Are you sure about renaming the files under `rel/`? According to [the docs on releases in the Elixir version we're using](https://hexdocs.pm/mix/1.15/Mix.Tasks.Release.html#module-vm-args-and-env-sh-env-bat), those are still supposed to be EEx templates, with no mention of HEEx.

lemald:
Also, you should add HEEx formatter to our configuration: https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.HTMLFormatter.html It currently reports a few inconsistencies.

lib/skate_web/templates/layout/app.html.heex
joshlarson: I bet it does not. I'll probably have to return these to their previous forms. Will check in dev-blue.

lib/skate_web/components/core_components.ex
```diff
@@ -0,0 +1,13 @@
...
+  @moduledoc """
+  Provides core UI components.
+  """
+
+  def cdn_route(conn, path) do
+    {static_href_module, static_href_fn} = Application.get_env(:skate, :static_href)
+    static_href = fn conn, path -> apply(static_href_module, static_href_fn, [conn, path]) end
+
+    static_href.(conn, path)
+    "https://localhost:4000#{static_href.(conn, path)}"
```
lemald, line 11: Looks like some leftover local testing code?

lib/skate_web/components/core_components.ex~
lemald, line 1: **issue:** based on the suffix it looks like this is an Emacs backup file that ended up getting committed.

joshlarson: Aaaarrrrgh

lib/skate_web/components/core_components.ex
joshlarson: Yep! It works on my machine, and now dev-blue doesn't load :)

lib/skate_web/components/core_components.ex
```diff
@@ -0,0 +1,12 @@
+defmodule SkateWeb.CoreComponents do
+  @moduledoc """
+  Provides core UI components.
+  """
+
+  def cdn_route(conn, path) do
```
lemald, line 6: **suggestion (non-blocking):** I might call this something like `static_content_route` since I feel like that better reflects what it does from the perspective of somebody writing a HEEx page (and whether that links to a CDN or not is transparent to them).

lemald (approved):
Looks good to go!

PR #2325, opened 2023-12-14
refactor(ex/skate_web): misc improvments using Phoenix 1.7

firestack:
While reviewing #2323 there were a few updates I wanted made, but it was a little outside the scope of #2323. 

This converts us to using more components in phoenix as well as follows conventions for `render`ing HTML.


lemald (approved):
👍

PR #2326, opened 2023-12-14
fix(dependabot): ignore JS path based dependencies

firestack:
# Summary

Dependabot stopped updating our JS dependencies without notification. Upon looking into dependabot jobs, the error message tells us that our path based dependencies are the issue. This commit ignores dependencies that are managed by `mix` like `phoenix` and `phoenix_html`.


firestack:
@dependabot help

lemald (approved):
🤞

PR #2327, opened 2023-12-18
feat(ts/components/DiversionPage): create diversion page stub

firestack:
This is only available & visible via Storybook at this moment.
I've styled this using bootstrap classes which some of we currently do not import. Given this is "lightly styled" I focused on CSS layout functionality and getting a few bootstrap classes right.

---

Asana Ticket: [⚙️🚧 Create lightly styled stub screen to display directions/missed stops](https://app.asana.com/0/1148853526253420/1206038957525484/f)

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -1,0 +1,56 @@
+.l-diversion-page {
+  display: grid;
+  // Default layout is linear
+  grid-template:
+    "heading" min-content
+    "panel" auto
+    "map" 80vh
+    / 1fr;
+
+  &__header {
+    grid-area: heading;
+  }
+
+  &__panel {
+    grid-area: panel;
+  }
+
+  &__map {
+    grid-area: map;
+  }
+
+  @include media-breakpoint-up(md) {
+    // For larger breakpoints, layout into grid
+    grid-template:
+      "heading heading" min-content
+      "panel map" auto
+      / fit-content(28ch) auto;
+
+    // Overlay the panel and make the panel header match
+    // the size of the page header if subgrid is supported
+    @supports (grid-template-rows: subgrid) {
+      grid-template:
+        [panel-start] "heading heading" min-content
+        "panel map" auto
+        / fit-content(28ch) auto;
+
+      .c-diversion-panel,
+      .l-diversion-page__panel {
+        display: grid;
+
+        grid-template-rows: subgrid;
+        grid-template-columns: 1fr;
+
+        grid-area: panel;
+
+        .c-diversion-panel__header {
+          grid-row: heading;
+        }
+
+        .c-diversion-panel__body {
+          grid-row: 2/-1;
+        }
+      }
+    }
+  }
+}
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -1,0 +1,35 @@
+import React from "react"
+import { DiversionPanel, DiversionPanelProps } from "./diversionPanel"
+import MapDisplay from "../mapPage/mapDisplay"
+
+export const DiversionPage = ({
+  directions,
+  missedStops,
+  routeName,
+  routeDescription,
+  routeDirection,
+  routeOrigin,
+}: DiversionPanelProps) => (
+  <article className="l-diversion-page h-100 border-box">
+    <header className="l-diversion-page__header text-bg-light border-bottom">
+      <h1 className="h3 text-center">Create Detour</h1>
+    </header>
+    <div className="l-diversion-page__panel bg-light">
+      <DiversionPanel
+        directions={directions}
+        missedStops={missedStops}
+        routeName={routeName}
+        routeDescription={routeDescription}
+        routeOrigin={routeOrigin}
+        routeDirection={routeDirection}
+      />
+    </div>
+    <div className="l-diversion-page__map">
+      <MapDisplay
+        selectedEntity={null}
+        setSelection={() => {}}
+        fetchedSelectedLocation={null}
+      />
+    </div>
+  </article>
+)
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -27,10 +27,13 @@ export const DiversionPanel = ({
       <section className="py-3 border-bottom">
         <h3 className="h4">Affected route</h3>
 
-        <div className="d-flex gap-2">
-          <RoutePill routeName={routeName} />
+        <div>
+          <RoutePill
+            className="d-inline-block me-2 align-top"
+            routeName={routeName}
+          />
 
-          <div>
+          <div className="d-inline-block">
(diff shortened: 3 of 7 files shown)


assets/css/_diversion_page.scss
```diff
@@ -0,0 +1,56 @@
...
+    // Overlay the panel and make the panel header match
+    // the size of the page header if subgrid is supported
+    @supports (grid-template-rows: subgrid) {
+      grid-template:
+        [panel-start] "heading heading" min-content
+        "panel map" auto
+        / fit-content(28ch) auto;
+
+      .c-diversion-panel,
+      .l-diversion-page__panel {
```
firestack, line 38: I'm a little uncertain on the prefixes in this document, and would like input on picking a direction.

Because these are all layout classes pertaining to the Diversion _Page_, I opted to use the `l-`(layout) prefix for things directly within the Pages's control. Subcomponents, like the `c-diversion-panel`, use the `c-` prefix because it's targeting specific regions of a component. I could go either way on if `c-diversion-panel` should be a `c-`(component) prefix or a `l-`(layout) prefix.

lemald:
Overall this looks good to me! I just want to think over the CSS a bit and revisit it tomorrow before giving it final approval.

assets/css/_diversion_page.scss
lemald: I think I like this current breakdown of the use of `c-` versus `l-`.

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+export interface DiversionPanelProps {
+  directions?: ReactNode
+  missedStops?: ReactNode
+  routeName: string
+  routeDescription: string
+  routeOrigin: string
+  routeDirection: string
+}
+
+export const DiversionPanel = ({
```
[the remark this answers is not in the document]
lemald: Yeah I'd double-check with design but agree that we should go with whatever terminology they're using.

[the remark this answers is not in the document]
firestack: (As you may have heard) Design is good with calling this a panel, _for now_.

lemald (approved):

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+  directions,
+  missedStops,
+  routeName,
+  routeDescription,
+  routeOrigin,
+  routeDirection,
+}: DiversionPanelProps) => (
+  <article className="c-diversion-panel h-100 bg-light border-end">
+    <header className="c-diversion-panel__header align-items-center border-bottom d-flex justify-content-around px-3">
```
joshlarson, line 22: General question: How do we decide what of these styles should live in `_*.scss` and what should go directly here?

joshlarson (approved):

assets/src/components/detours/diversionPage.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react"
+import { DiversionPanel, DiversionPanelProps } from "./diversionPanel"
+import MapDisplay from "../mapPage/mapDisplay"
+
+export const DiversionPage = ({
+  directions,
+  missedStops,
```
joshlarson, line 7: I think I see this collection of attributes being its own object at some point, like a `Detour` object. Otherwise, we could wind up passing each of these around separate a whole lot, which could get kind of messy.

assets/src/components/detours/diversionPanel.tsx
firestack: Good question!
I don't think we have an established practice yet. Part of me wants to reduce the amount of classes and use scss `@extend` but I think it may be better to do it within the DOM as it is?

So far I was adding classes "ad-hoc" as needed to get close to the styles in the Figma, and anything I couldn't do with bootstrap classes (bespoke grid) I did in the `_*.scss` file.

If you have any input on how to approach this I'd be happy to, at the very least, consider it!

assets/src/components/detours/diversionPage.tsx
joshlarson: I could justify this change as a refactor in the future, once we see more clearly how these attributes are being used.

firestack: Defintely agree! So far, I built this to give us as much flexibility over the content and not include any sort of object grouping yet. I'd imagine us creating a `DiversionPageFromDetour` or something of the like which would be the translation from say, a `Detour` object, into the actual properties the base component takes. I think this also gives us more flexibility in Storybook to play around with the content, instead of an object.

assets/src/components/detours/diversionPanel.tsx
joshlarson: How come it would be better to do it within the DOM? 

Right now we sort of have a hybrid between SCSS and utility classes, which is a bit confusing to me. I dunno that I have a strong opinion about which direction we should go, but I do think that we'd be better off picking one approach and going all in on that.

I'll spend some time thinking about which approach I personally would prefer.

Aaaaand since this seems like a "problem" (at least a feature) throughout our codebase, I don't think we need to solve it now - another refactor opportunity!

joshlarson: I mean... moving to Tailwind is an option :)

(I personally DO like Tailwind and its utility class concept - it makes things very flexible, but also React components [or Phoenix components if we ever go that route] give us plenty of reusability)

Curious to see that diff!

assets/src/components/detours/diversionPage.tsx
joshlarson: At some point, I'd at least like to try and see.

<details>
<summary>
Rant
</summary>
In a certain sense, if we have patterns that exist in order to accommodate Storybook, then those patterns are Storybook-only patterns, even if they live outside of the stories/ directory.

In those cases, I'd rather make the rationale explicit by making Storybook do the extra work and putting the extra patterns in Storybook.

But..... if the "extra" pattern also makes the production code easier to work with, then it's not Storybook-only (the location marker thing we paired on a while ago is a good example of that - we made the Storybook integration easier AND made the production code better at the same time)
</details>

assets/src/components/detours/diversionPanel.tsx
firestack: I did immediately run into an issue that wouldn't happen if all the classes were in the HTML.
Right now we don't have typography imported, which means styles like `.h3` are not available, and SCSS fails to build when doing `@extend .h3` because the style is missing.

firestack: it does seem like I can use `!optional`` to get around this

firestack: I'm not sure if this is what you had imagined, but here it is rewritten using scss @extend and bootstrap. I opted to keep using bootstrap and `@extend` so that any design tokens edited in the bootstrap config would automatically apply here.
dda154f6d5c78d8aded644836837b93fc7474267

PR #2328, opened 2023-12-18
chore: Add sobelow for static code security analysis

joshlarson:
Screenshot 2023-12-18 at 3 36 10 PM

Screenshot 2023-12-18 at 3 36 46 PM

---
  - https://app.asana.com/0/0/1204362820164138


lib/skate_web/router.ex
```diff
@@ -29,6 +29,7 @@ defmodule SkateWeb.Router do
 
   pipeline :accepts_html do
     plug :accepts, ["html"]
+    plug :put_secure_browser_headers, %{"content-security-policy" => "default-src 'self'"}
```
firestack, line 32: question: have you checked on a dev env if all our vendored code and external services (drift, appcues, fullstory, etc) if everything still works with this security policy?

_If_ I recall correctly, I think we may have some inline code executing from some of our libraries and this CSP may block that.

mix.exs
```diff
@@ -104,7 +105,8 @@ defmodule Skate.MixProject do
       "assets.setup": ["cmd npm --prefix=assets install"],
       "assets.reset": ["cmd npm --prefix=assets ci"],
       test: ["ecto.create --quiet", "ecto.migrate_all", "test"],
-      fmt: ["format", "cmd npm --prefix=assets run format"]
+      fmt: ["format", "cmd npm --prefix=assets run format"],
+      sobelow: "sobelow --skip --exit=low --ignore Config.HTTPS,Config.CSWH"
```
firestack, line 109: question: what is the rationale for ignoring CSWH / can we address the issues?

lib/skate_web/router.ex
firestack: btw, I think fullstory is only present on dev-blue

lib/skate_web/router.ex
```diff
@@ -39,7 +40,6 @@ defmodule SkateWeb.Router do
     plug :fetch_session
     plug :fetch_flash
     plug :protect_from_forgery
-    plug :put_secure_browser_headers
```
lemald, line 42: **question:** what impact does this have on the `/api` endpoints? Those have `:browser` but not `:accepts_html`, so they will no longer get `:put_secure_browser_headers` after this change.

mix.exs
joshlarson: The reason I picked this approach for ignoring these errors is because `# sobelow_skip` only works on functions, and the `HTTPS` and `CSWH` issues were flagged in `router.ex`, which doesn't have much in the way of functions, and I couldn't figure out how to actually _resolve_ them.

There is another option, which is to use the `--mark_skip_all` flag to generate a `.sobelow_skips` file. The problem with that is that `.sobelow_skips` is kind of incomprehensible.

joshlarson: And for `CSWH` in particular, I'm reasonably certain that it's already not an issue, because we do have `check_origin` [set for websockets](https://github.com/mbta/skate/blob/7c0347fab359c43b849a63033e7cb8cb37499651/lib/skate_web/endpoint.ex#L5-L7), and that environment config is [always set for prod environments](https://github.com/mbta/skate/blob/7c0347fab359c43b849a63033e7cb8cb37499651/config/prod.exs#L36-L39), but I don't know how to convince Sobelow that that's true. Hence the ignore here.

lib/skate_web/router.ex
joshlarson: Augh - I thought that was a pure refactor, but I didn't notice that this effectively removed `:put_secure_browser_headers` from `:accepts_json`.

lib/skate_web/router.ex
joshlarson: I have not. I'm kind of tempted to roll at least part of this change back in order to keep the addition of Sobelow as lite as possible, and then add some follow-up tickets to investigate the specific issues it found, including the secure headers that I tried to fix here.

joshlarson: Yeah... it don't work :(

mix.exs
```diff
@@ -105,7 +106,9 @@ defmodule Skate.MixProject do
       "assets.setup": ["cmd npm --prefix=assets install"],
       "assets.reset": ["cmd npm --prefix=assets ci"],
       test: ["ecto.create --quiet", "ecto.migrate_all", "test"],
-      fmt: ["format", "cmd npm --prefix=assets run format"]
+      fmt: ["format", "cmd npm --prefix=assets run format"],
+      sobelow:
+        "sobelow --skip --exit=low --ignore Config.HTTPS,Config.CSWH,Config.Headers,Config.CSP"
```
joshlarson, line 111: Added `Config.Headers` and `Config.CSP` here rather than trying to fix it (and risking "fixing" it wrong) in this PR

lemald (approved):
LGTM! We just discussed and agreed that this is a net improvement and we'll create followup tickets to address things like the security headers and CSP.

PR #2329, opened 2023-12-18
chore: Add excellent_migrations

joshlarson:
Screenshot 2023-12-18 at 3 58 55 PM

Screenshot 2023-12-18 at 4 00 05 PM

---
  - https://app.asana.com/0/0/1205085395491448


config/config.exs
```diff
@@ -155,6 +155,10 @@ config :ueberauth, Ueberauth,
     cognito: nil
   ]
 
+# Don't report issues with migrations from before we introduced excellent_migrations
+# config :excellent_migrations, start_after: "20230908133314"
+config :excellent_migrations, start_after: "20220927170631"
```
joshlarson, line 160: This is intentionally wrong in order to make this fail the build.

mix.exs
```diff
@@ -60,6 +60,7 @@ defmodule Skate.MixProject do
       {:ex_aws, "~> 2.1"},
       {:ex_doc, "~> 0.27", only: :dev, runtime: false},
       {:ex_machina, "~> 2.7.0", only: :test},
+      {:excellent_migrations, "~> 0.1", only: [:dev, :test], runtime: false},
```
firestack, line 63: question(non-blocking): why do we need this in `:test`?

firestack (approved):
🎉

lemald (approved):

mix.exs
joshlarson: Huh... I just grabbed that line from the `excellent_migrations` docs. Let's see if CI likes it if we move it to only be `:dev`.

joshlarson: CI liked it!

PR #2332, opened 2023-12-19
chore: update AWS RDS cert bundle hash

lemald:
This is based on the value I got when downloading it on my personal connection.


firestack (approved):
Conditional approval given CI succeeds.

lemald:
> Conditional approval given CI succeeds.

I'm also deploying to dev-blue just to make sure it actually works.

PR #2334, opened 2023-12-20
Detour pilot: basic detour route drawing

lemald:
Asana ticket: [⚙️🚧 Click on the map to draw, click on waypoints, and click on route to end](https://app.asana.com/0/1148853526253420/1206038957525487/f)


joshlarson (changes requested):
How would you feel about adding the dummy detour page back in gated behind a test group? That would make this work and anything building on it easier (possible?) to validate easily, and to validate in dev and/or prod.

assets/stories/skate-components/detours/diversionPage.stories.tsx
```diff
@@ -14,6 +15,7 @@ const meta = {
   args: {
     // Provide default route settings
     ...DiversionPanelMeta.args,
+    shape: route39shape,
```
joshlarson, line 18: Nit: It's throwing me off a bit that the route shape is the 39, but the pill in the story says "66".
```suggestion
    routeName: "39",
    shape: route39shape,
```

assets/src/components/mapPage/mapDisplay.tsx
```diff
@@ -200,35 +200,33 @@ const RoutePatternLayers = ({
...
-              {routePattern.shape && (
-                <>
-                  <RouteShape
-                    shape={routePattern.shape}
-                    isSelected={isSelected}
+    <ZoomLevelWrapper>
+      {(zoomLevel) => {
+        return (
+          <>
+            {routePattern.shape && (
```
joshlarson, line 207: Is this necessary? Don't we have the guarantee that `routePattern.shape` will be truthy because of the ternary on line 202?

assets/src/components/detours/detourMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+        onSetEndPoint={setEndPoint}
+        detourPositions={detourPositions}
+        onAddDetourPosition={onAddDetourPosition}
+      />
+    </Map>
+  )
+}
+
+const RouteShapeWithDetour = ({
+  shape,
```
joshlarson, line 52: If `shape` here means "the original un-diverted route shape", can we call it `original` or `originalShape` or something like that?

lemald:
> How would you feel about adding the dummy detour page back in gated behind a test group? That would make this work and anything building on it easier (possible?) to validate easily, and to validate in dev and/or prod.

Just for the sake of posterity, we decided that we are going to do this as an intermediate step but not as part of this particular PR.

assets/src/components/mapPage/mapDisplay.tsx
lemald: That's true, but unfortunately (I think because this use of `routePattern.shape` being wrapped inside of a closure), TypeScript doesn't realize that and throws an error without the check in place.

assets/src/components/detours/detourMap.tsx
lemald: Yeah, I like `originalShape`, I think I'll go with that.

lemald: 55ba0b74

assets/src/components/detours/detourMap.tsx
```diff
@@ -0,0 +1,203 @@
...
+import Leaflet, { Map as LeafletMap } from "leaflet"
+import Map from "../map"
+import { CustomControl } from "../map/controls/customControl"
+import { Button } from "react-bootstrap"
+import { ReactMarker } from "../map/utilities/reactMarker"
+
+export const DetourMap = ({ shape }: { shape: Shape }) => {
+  const [startPoint, setStartPoint] = useState<LatLngExpression | null>(null)
+  const [endPoint, setEndPoint] = useState<LatLngExpression | null>(null)
+  const [detourPositions, setDetourPositions] = useState<LatLngExpression[]>([])
```
firestack, line 14: random-thought(non-blocking): I think it's too early to refactor this now, but I think we should consider doing the following
1. Pull out this state into it's own hook, e.g., `useDetourCreationObject` or something.
2. Pass this state into `DetourMap` via props.
3. remove the existance of `startPoint` and `endPoint` and instead contain the implementation within a hook which provides an API.
    - I'm imagining something like a `state` variable, which would be a state machine. The states may be something like `awaitingStart` -> `awaitingPoints` -> `finished`, and then the `start` and `end` points are just the first and last elements of the `detourPositions`.
    - And a few functions, which can interact with the state machine, like `setStartPoint`, `addWayPoint`, `setEndPoint`.

lemald: Yes, I really like this idea! Let's keep it in mind as we move forward and the interfaces between the different components and stuff start to settle down.

assets/src/components/detours/detourMap.tsx
```diff
@@ -0,0 +1,203 @@
...
+              onSetStartPoint(position)
+              position && onAddDetourPosition(position)
+            } else if (endPoint === null) {
+              const position = closestPosition(
+                routeShapePositions,
+                e.latlng,
+                map
+              )
+              onSetEndPoint(position)
+              position && onAddDetourPosition(position)
```
joshlarson, line 110: This causes `onAddDetourPosition` to get called twice for the `endPoint`, since the `useMapEvent` handler also gets called for the same point. 
```suggestion
```

lemald: It's not quite the same - in the `useMapEvent` handler, it directly adds the corresponding position the user clicked on the map. In this code, we specifically take the closest point on the shape first. So actually you're right that there's a problem here with it firing twice, but it's actually the other instance of it firing that we want to remove. I don't immediately see an obvious answer and just doing `e.originalEvent.stopPropagation()` didn't seem to do the trick - I'll have to keep digging.

joshlarson: It might also not be a huge deal, except that it'll make some silly turn-by-turn directions.

joshlarson (approved):

assets/src/components/detours/detourMap.tsx
lemald: 7bf105e2 (thanks @firestack!)

PR #2335, opened 2023-12-21
feat: Add override for test groups

joshlarson:
---
  - https://app.asana.com/0/0/1206195348204443

### DIFF
--- a/lib/skate/settings/test_group.ex
+++ b/lib/skate/settings/test_group.ex
@@ -4,15 +4,17 @@ defmodule Skate.Settings.TestGroup do
   alias Skate.Settings.Db.TestGroup, as: DbTestGroup
   alias Skate.Settings.Db.User, as: DbUser
 
+  import Ecto.Query, only: [from: 2]
+
   @type t() :: %__MODULE__{
           id: integer(),
           name: String.t(),
           users: [DbUser.t()]
         }
 
-  @enforce_keys [:id, :name, :users]
+  @enforce_keys [:id, :name, :users, :override]
 
-  defstruct [:id, :name, users: []]
+  defstruct [:id, :name, users: [], override: :none]
 
   @spec create(String.t()) :: {:ok, t()} | {:error, Ecto.Changeset.t()}
   def create(name) do
@@ -50,6 +52,13 @@ defmodule Skate.Settings.TestGroup do
     |> Enum.map(&convert_from_db_test_group(&1))
   end
 
+  @spec get_override_enabled() :: [t()]
+  def get_override_enabled() do
+    from(tg in DbTestGroup, where: tg.override == :enabled)
+    |> Skate.Repo.all()
+    |> Enum.map(&convert_from_db_test_group(&1))
+  end
+
   @spec update(t()) :: t()
   def update(test_group) do
     existing_test_group =
@@ -80,7 +89,8 @@ defmodule Skate.Settings.TestGroup do
     %__MODULE__{
       id: db_test_group.id,
       name: db_test_group.name,
-      users: db_test_group.users
+      users: db_test_group.users,
+      override: db_test_group.override
     }
   end
 
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -2,6 +2,7 @@ defmodule Skate.Settings.User do
   @moduledoc false
 
   alias Skate.Settings.Db.User, as: DbUser
+  alias Skate.Settings.TestGroup
 
   import Ecto.Query
   require Logger
@@ -75,12 +76,20 @@ defmodule Skate.Settings.User do
   end
 
   def is_in_test_group(user_id, test_group_name) do
-    %{test_groups: user_test_groups} =
-      DbUser
-      |> Skate.Repo.get(user_id)
-      |> Skate.Repo.preload(:test_groups)
+    user_id
+    |> get_by_id!()
+    |> all_test_group_names()
+    |> Enum.member?(test_group_name)
+  end
 
-    Enum.any?(user_test_groups, &(&1.name == test_group_name))
+  def all_test_group_names(user) do
+    user = Skate.Repo.preload(user, :test_groups)
+
+    enabled_test_groups = TestGroup.get_override_enabled()
+
+    (enabled_test_groups ++ user.test_groups)
+    |> Enum.map(& &1.name)
+    |> Enum.dedup()
   end
 
   defp users_for_route_ids_query(route_ids) do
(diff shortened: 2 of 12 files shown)


lib/skate/settings/user.ex
```diff
@@ -75,12 +76,20 @@ defmodule Skate.Settings.User do
...
-      |> Skate.Repo.get(user_id)
-      |> Skate.Repo.preload(:test_groups)
+    user_id
+    |> get_by_id!()
+    |> all_test_group_names()
+    |> Enum.member?(test_group_name)
+  end
+
+  def all_test_group_names(user) do
+    user = Skate.Repo.preload(user, :test_groups)
```
joshlarson, line 86: I would eventually like for this to be just one query, but I wasn't convinced that that was necessary right now. Unless there's something about Ecto that I'm missing, this will run 2 queries, not like.. 20.

priv/repo/migrations/20231220204054_add_override_for_test_groups.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Skate.Repo.Migrations.AddOverrideForTestGroups do
+  use Ecto.Migration
+
+  def change do
+    # excellent_migrations:safety-assured-for-this-file raw_sql_executed
```
joshlarson, line 5: Just our luck - the first migration we write with `excellent_migrations` in place, and we basically have to ignore it 🤦

test/skate/settings/test_group_test.exs
```diff
@@ -97,6 +119,37 @@ defmodule Skate.Settings.TestGroupTest do
...
+
+    test "override shows as :none if it is nil in the database" do
+      {:ok, test_group} = TestGroup.create("group name")
+
+      new_test_group = TestGroup.update(%{test_group | override: nil})
+
+      assert new_test_group.override == :none
+    end
+
+    @tag :skip
```
joshlarson, line 145: I could be persuaded to remove this test for now, since there isn't really a good way of denoting that an update failed. I was planning on eventually switching `TestGroup.update/1` to also return `{:ok, tg} | {:error, errors}`, like I did for `TestGroup.create/1` a while ago.

lib/skate/settings/test_group.ex
```diff
@@ -50,6 +52,13 @@ defmodule Skate.Settings.TestGroup do
     |> Enum.map(&convert_from_db_test_group(&1))
   end
 
+  @spec get_enabled() :: [t()]
+  def get_enabled() do
```
lemald, line 56: **suggestion:** The name `get_enabled` is a little confusing to me because it only includes test groups that are enabled via override, not test groups that are enabled on a per-user basis for a given user (basically, I feel like "enabled" is still a function of both a test group and a user, not just a test group). I think something like `get_override_enabled` would make more sense personally.

priv/repo/migrations/20231220204054_add_override_for_test_groups.exs
```diff
@@ -0,0 +1,15 @@
...
+    execute(
+      "CREATE TYPE test_group_override AS ENUM ('none', 'enabled', 'disabled')",
+      "DROP TYPE test_group_override"
+    )
+
+    alter table(:test_groups) do
+      add(:override, :test_group_override)
```
lemald, line 12: **suggestion:** can we add a default value of `none` and thereby avoid handling null values? Both in terms of the application code and in terms of marking the column itself as non-nullable.

joshlarson: [`excellent_migrations` doesn't like adding columns with default values](https://hexdocs.pm/excellent_migrations/readme.html#adding-a-column-with-a-default-value)

I would probably be okay with doing it anyway and making `excellent_migrations` ignore the "error" though, especially since this table is pretty small.

lib/skate/settings/user.ex
```diff
@@ -75,12 +76,20 @@ defmodule Skate.Settings.User do
...
+
+    enabled_test_groups = TestGroup.get_enabled()
 
-    Enum.any?(user_test_groups, &(&1.name == test_group_name))
+    (enabled_test_groups ++ user.test_groups)
```
lemald, line 90: **suggestion:** I'm sort of torn because this logic (combine the user-specific test groups with the globally0enabled test groups) feels more like test group logic than user logic, but I can't quite think of another way to split it up.

priv/repo/migrations/20231220204054_add_override_for_test_groups.exs
lemald: That page also says this becomes safe in Postgres 11+, and we're on 15. It's understandable since `excellent_migrations` has no way of knowing what Postgres version we're on, but we should probably just disable that warning somehow.

lib/skate/settings/user.ex
lemald: Count this piece of feedback as optional, though, as I'm not really sure what it should look like instead.

lemald (changes requested):
Okay done with a first pass review. I think the three things that need addressing are the default value of the column, the name of `get_enabled`, and the skipped test.

lib/skate/settings/user.ex
joshlarson: Honestly, this felt like the sort of dumb hacky code that we'll refactor in the future. I don't like it either, but this is easy to fix after-the-fact.

priv/repo/migrations/20231220204054_add_override_for_test_groups.exs
joshlarson: Touché - I had been meaning to ask about that. 

Updated, and added config to disable that check in particular.

test/skate/settings/test_group_test.exs
joshlarson: I think so, but I didn't bother investigating it too thoroughly, since there isn't _really_ a way of triggering that behavior.

lib/skate/settings/test_group.ex
joshlarson: Done

lib/skate/settings/user.ex
joshlarson: After a few minutes of thought, it feels like there are three separate things going on here:
 - Admin view / User management (seeing all users, for a given user, seeing what test groups they're in, etc)
 - Admin view / Test group management (a lot of the behavior in test_group.ex)
 - User view / Test group membership (when a user loads the regular skate app, what test groups are they a member of for purposes of `data-user-test-groups` and any other backend `is_in_test_group` calls)

I'm thinking that in the medium-term, each of these different questions belongs in its own module, with a new `TestGroupMembership` module (probably with a better name than that) that takes care of all of the increasingly complex logic for figuring out what test groups a user is a part of.

(Maybe that could also be an opportunity to move all of this test group stuff out of the `Settings` context, where it sort of feels out of place)

lemald (approved):
Looks good to me and I was playing around with it locally and everything seems to work as expected. 👍

PR #2337, opened 2023-12-22
refactor: Improve error handling if TestGroup.update/1 is given a bad override value

joshlarson:
Just leaving this here for now to flag that the contract of `TestGroup.update/1` could use some improvement.

It's not a big deal. I don't think there's any way from the UI to get "bad" behavior out of `TestGroup.update/1` right now, but it sort of offends my sensibilities, and it's a (very minor) source of risk in the future to have a context that exposes methods with no graceful error handling.


joshlarson:
Closed and Asana-ified: https://app.asana.com/0/1148853526253420/1206320701968038/f

PR #2340, opened 2023-12-28
refactor(assets/nav): Separate nav data from nav rendering

joshlarson:


assets/src/components/nav/navLinkData.ts
```diff
@@ -0,0 +1,38 @@
...
+type HTMLElementProps = React.PropsWithoutRef<React.HTMLAttributes<HTMLElement>>
+
+export interface LinkData {
+  title: string
+  path: string
+  navIcon: React.JSXElementConstructor<HTMLElementProps>
+  onClick?: () => void
+}
+
+export const useNavLinkData: () => LinkData[] = () => {
```
lemald, line 15: **question:** does this have to be a hook?

assets/src/components/nav/leftNav.tsx
```diff
@@ -20,11 +18,33 @@ import {
...
+}: LeftNavLinkProps): JSX.Element => (
+  <NavLink
+    className={({ isActive }) =>
+      "c-left-nav__link" + (isActive ? " c-left-nav__link--active" : "")
+    }
+    title={linkData.title}
+    to={linkData.path}
+    onClick={linkData.onClick}
+  >
+    <linkData.navIcon className="c-left-nav__icon" />{" "}
```
lemald, line 44: **nitpick:** unnecessary space character (I think this is what caused the snapshot diffs):
```suggestion
    <linkData.navIcon className="c-left-nav__icon" />
```

assets/src/components/nav/navLinkData.ts
joshlarson: This might be my lack of react fundamentals speaking, but the nav link data is dependent on what test groups you're in, so I don't _think_ it can just be a constant. 

Is there a different pattern that would make more sense in this context?

firestack: I'm not sure if it's the right decision, but you _could_ implement this as a component I think.

lemald: You can also make it a plain, non-hook function. Hooks have specific logic in React around triggering re-renders of components that I don't think is really relevant here.

assets/src/components/nav/leftNav.tsx
joshlarson: Ooooh I think prettier must have done that without me noticing

lemald: Yeah, it does that sometimes.

assets/src/components/nav/navLinkData.ts
joshlarson: What makes this a hook as opposed to a plain old function?

I just tried writing `getNavLinkData()` as a plain old function, and the only difference was that it was named `get...` instead of `use....`. Is that all it takes?

lemald: Yes, the `use` prefix is special: https://react.dev/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use I'm not exactly sure what under-the-hood JavaScript magic React uses to accomplish this.

firestack: For what it's worth, I'm _fairly_ sure the `use` prefix is just convention, and I don't think React does any compilation related to **hooks**, rather custom hooks often call react hooks to get reactivity (`useEffect`, `useState`, etc).

But because this function doesn't use any react hooks, it's probably good to follow convention and not use the `use` prefix.
> This convention guarantees that you can always look at a component and know where its state, Effects, and other React features might “hide”. For example, if you see a getColor() function call inside your component, you can be sure that it can’t possibly contain React state inside because its name doesn’t start with use. However, a function call like useOnlineStatus() will most likely contain calls to other Hooks inside!

lemald: I think it does a little bit of stuff to provide runtime warnings when you break the Rules of Hooks: https://react.dev/warnings/invalid-hook-call-warning

lemald: Also the ESLint plugin for React hooks will similarly complain about such things in CI. But yes, I think in practice it's mostly convention. In any event since this function doesn't need to do any "hook-y" things I'd advocate for renaming it.

joshlarson: WELP!

And here I was thinking that I was being clever, but I guess I wasn't.

Just renamed it to `getNavLinkData`

assets/src/components/nav/navLinkData.ts
lemald, line 1: **nitpick:** okay sorry, everything else is looking great but one last thing that I just noticed is that this is a non-component module that's under the `components` directory. I'm not sure if it's sufficiently "component-adjacent" that we just want to leave it here, but I wanted to point it out. cc @firestack in case you have thoughts.

joshlarson: I left it in the `nav/` folder because it's only used by other things in that folder, but would it make sense to move it from `src/components/nav` into `src/nav` or something like that?

lemald: I think just under the top level `src` directory would be fine since I don't see there being much need for other files in a `src/nav` directory.

joshlarson: Moved it to `src/`

lemald (approved):

PR #2345, opened 2023-12-30
feat: Make API call when waypoints are added

joshlarson:
---
  - https://app.asana.com/0/0/1206038961362403


.envrc.template
```diff
@@ -87,3 +87,7 @@ export ERL_FLAGS="+MIscs 2048"
 ## Used for locally building (and eventually testing) Storybook and Snapshot tests
 ## The project token can be found in Chromatic, (Manage > Configure > Project > Project Token)
 # export CHROMATIC_PROJECT_TOKEN=
+
+## Open Route Service
+export OPEN_ROUTE_SERVICE_API_URL=https://api.openrouteservice.org/
```
lemald, line 92: **nitpick:** with the other environment variables in this file we've just been documenting them here and then for information on actual values to use putting them in Notion. I don't think there's any inherent issue with putting this URL in the repo (it's not secret or anything), but I think it would probably be best to be consistent with what we've done so far.

assets/src/api.ts
```diff
@@ -147,6 +150,24 @@ export const fetchShapeForRoute = (routeId: RouteId): Promise<Shape[]> =>
...
+    url: "/api/detours/directions",
+    parser: detourShapeFromData,
+    dataStruct: DetourShapeData,
+    defaultResult: null,
+    fetchArgs: {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "x-csrf-token": getCsrfToken(),
+      },
```
lemald, line 166: **idea (non-blocking):** now that we've got a few `POST` requests in this file that need these headers, it might be good to factor them out to a helper function or something.

assets/src/components/detours/detourMap.tsx
```diff
@@ -1,33 +1,56 @@
...
+      if (detourShape) {
+        setDetourShapePositions(
+          detourShape.coordinates.map((position: ShapePoint) => {
+            const { lat, lon } = position
+            return { lat, lng: lon }
+          })
+        )
+      }
+    })
+  }, [waypoints])
```
lemald, line 42: **issue:** don't forget to handle the race condition where two `fetchDetourDirections` calls happen in quick succession and then resolve out-of-order. The `useNearestIntersection` hook contains a good example of how to guard against this.

assets/src/models/detourShapeData.ts
```diff
@@ -0,0 +1,20 @@
...
+    })
+  ),
+})
+export type DetourShapeData = Infer<typeof DetourShapeData>
+
+export const detourShapeFromData = (
+  shapeData: DetourShapeData
+): DetourShape => ({
+  coordinates: shapeData.coordinates.map((pointData) => {
+    return { lat: pointData.lat, lon: pointData.lon }
```
lemald, line 18: **suggestion:** TypeScript syntax can actually make this even more concise:
```suggestion
    return { ...pointData }
```

config/runtime.exs
```diff
@@ -11,6 +11,10 @@ config :skate,
   satellite_tileset_url: System.get_env("SATELLITE_TILESET_URL"),
   aws_place_index: System.get_env("AWS_PLACE_INDEX")
 
+config :skate, Skate.OpenRouteServiceAPI,
+  api_base_url: System.get_env("OPEN_ROUTE_SERVICES_API_URL"),
+  api_key: System.get_env("OPEN_ROUTE_SERVICES_API_KEY")
```
firestack, line 16: __issue__: I don't think these should be plural?
```suggestion
  api_base_url: System.get_env("OPEN_ROUTE_SERVICE_API_URL"),
  api_key: System.get_env("OPEN_ROUTE_SERVICE_API_KEY")
```

joshlarson: They should not be - you're right!

lib/skate/open_route_service_api/directions_request.ex
lemald, line 1: **issue:** this should also have a type definition

lib/skate/open_route_service_api/directions_response.ex
lemald, line 1: **issue:** see other comment, should have type definition

test/skate_web/controllers/detour_route_controller_test.exs
```diff
@@ -0,0 +1,42 @@
...
+      %{bypass: bypass}
+    end
+
+    @tag :authenticated
+    test "returns shape data as geojson", %{conn: conn, bypass: bypass} do
+      Bypass.expect(bypass, fn conn ->
+        assert %Plug.Conn{
+                 request_path: "/v2/directions/driving-hgv/geojson",
+                 method: "POST"
+                 #  req_headers: [{"authorization", "nil"}, {"content-type", "application/json"}]
```
lemald, line 26: **nitpick:** leftover commented out code

test/skate_web/controllers/detour_route_controller_test.exs
```diff
@@ -0,0 +1,42 @@
...
+  use SkateWeb.ConnCase
+  import Test.Support.Helpers
+
+  setup %{conn: conn} do
+    {:ok, conn: put_req_header(conn, "accept", "application/json")}
+  end
+
+  describe "directions" do
+    setup do
+      bypass = Bypass.open()
```
lemald, line 11: **issue (non-blocking):** if this test works I think it's fine, but just a note that Bypass seems pretty much unmaintained at this point, so we should probably try to use alternatives going forward.

lemald:
@joshlarson okay I've completed a first pass, I've left a few comments but at a base level I think this is looking good. 👍

test/skate_web/controllers/detour_route_controller_test.exs
joshlarson: Yup! I'm working on replacing it with Mox, per _all the recommendations_

assets/src/components/detours/detourMap.tsx
joshlarson: Ah yeah I was wondering how I should handle that. 

Just to clarify: Your suggestion here is to make the `waypoints` array part of a bigger struct that includes `loading` state as well, and to only call `fetchDetourDirections` when `loading` is `false`? NOT to create a hook.

assets/src/models/detourShapeData.ts
lemald: Oh, if TypeScript likes that then that's good too. If that works you could also get rid of the `{ ... }` curly braces and `return` statement. But I'm not sure if it's quite smart enough for that.

assets/src/components/detours/detourMap.tsx
lemald: I'm agnostic on how to solve it, what you described is one approach but `useNearestIntersection` just uses a local variable within the `useEffect` function. As for making this its own hook I could go either way, but if the logic gets more complex it might be worth moving it out.

assets/src/api.ts
joshlarson: I might do that as a follow-up refactor PR, but I like the change.

assets/src/models/detourShapeData.ts
joshlarson: Wait, I think at that point, we could just get rid of the `.map` call altogether. 🤯

assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -10,6 +10,10 @@ beforeEach(() => {
   jest.spyOn(global, "scrollTo").mockImplementationOnce(jest.fn())
 })
 
+jest.mock("../../../src/api", () => ({
+  fetchDetourDirections: jest.fn(() => Promise.resolve(null)),
+}))
+
```
firestack, line 16: suggestion(non-blocking): just wondering, did you try this to see if it worked?

```suggestion
jest.mock("../../../src/api")
```

We're intending to [⚙️ Ensure frontend test isolation by resetting mocks](https://app.asana.com/0/1148853526253420/1203482629607208/f) and this default implementation would be lost when resetting the mock.

Instead, you'd want to mock the function in a `beforeEach` call so that it's mocked for every test individually. And then we can use the `mockResolvedValue` helper for this instance.

```suggestion
jest.mock("../../../src/api")

beforeEach(() => {
	jest.mocked(fetchDetourDirections).mockResolvedValue(null)
})
```

test/skate/open_route_service_api/client_est.exs
lemald, line 1: **nitpick:** typo in file name (`client_est` rather than `client_test`)

joshlarson: wups!

assets/src/components/detours/detourMap.tsx
```diff
@@ -1,34 +1,57 @@
...
+    setWaypoints((positions) => [...positions, p])
+  }
+
+  useEffect(() => {
+    const shapePoints: ShapePoint[] = waypoints.map(
+      (waypoint: LatLngLiteral) => {
+        const { lat, lng } = waypoint
+        return { lat, lon: lng }
+      }
+    )
```
lemald, line 31: **suggestion:** since this operation and its inverse happen in a few places it might be worth splitting it out into utility functions, even if it's small / simple logic.

lib/skate/open_route_service_api/client.ex
```diff
@@ -0,0 +1,90 @@
...
+
+  alias Skate.OpenRouteServiceAPI.DirectionsRequest
+
+  @callback get_directions(DirectionsRequest.t()) :: {:ok, map()} | {:error, any()}
+
+  @doc """
+  Sends `request` to the OpenRouteService API and then sends the response through
+  `parse_response/1` before returning it
+  """
+  @spec get_directions(DirectionsRequest.t()) :: {:ok, map()} | {:error, any()}
```
lemald, line 14: **question:** I think you can just do `@impl true` rather than repeating the typespec from the `@callback`.

assets/src/api.ts
joshlarson: https://app.asana.com/0/1148853526253420/1206332571064068/f

lib/skate/open_route_service_api/client.ex
lemald: Ahh, okay. Yeah if changing it would make the docs worse then I'm fine leaving it as-is.

assets/src/util/pointLiterals.ts
```diff
@@ -0,0 +1,16 @@
...
+  const { lat, lng } = latLngLiteral
+  return { lat, lon: lng }
+}
+
+export const shapePointToLatLngLiteral = (
+  shapePoint: ShapePoint
+): LatLngLiteral => {
+  const { lat, lon } = shapePoint
+  return { lat, lng: lon }
+}
```
firestack, line 16: suggestion(non-blocking):
```suggestion
export const latLngLiteralToShapePoint = (
  {lat, lng}: LatLngLiteral
): ShapePoint => ({ lat, lon: lng })

export const shapePointToLatLngLiteral = (
  {lat, lon}: ShapePoint
): LatLngLiteral => ({ lat, lng: lon })
```

firestack (commented):
I believe you've addressed all of @lemald's comments, and this LGTM :+1:

firestack (approved):

PR #2346, opened 2024-01-02
feat: Add dummy detour page behind a test group

joshlarson:
Follow-up to https://github.com/mbta/skate/pull/2334 to add a page where we can interact with the line-drawing, and eventually interact with the API calls that that line-drawing initiates.

This also includes the changes from https://github.com/mbta/skate/pull/2340, so 👈 PR should be merged first


assets/src/navLinkData.ts
```diff
@@ -10,11 +11,23 @@ export interface LinkData {
   path: string
   navIcon: React.JSXElementConstructor<HTMLElementProps>
   onClick?: () => void
+  hideOnMobile?: boolean
 }
 
 export const getNavLinkData: () => LinkData[] = () => {
   const mapMode = mapModeForUser()
 
+  const maybeDetours = inTestGroup(TestGroups.DetourPanel)
```
joshlarson, line 20: This is a silly name, and I don't _hate_ it, but maybe it would be better to call it like... `optionalDetoursLink` or something like that?

assets/tests/components/nav/bottomNavMobile.test.tsx
```diff
@@ -59,4 +59,19 @@ describe("BottomNavMobile", () => {
     expect(screen.queryByTitle("Search")).toBeNull()
     expect(screen.queryByTitle("Search Map")).toBeInTheDocument()
   })
+
+  test("does not render nav item with title 'Detours' even if in test group", () => {
+    ;(getTestGroups as jest.Mock).mockReturnValue([TestGroups.DetourPanel])
```
lemald, line 64: **suggestion:** you can use `jest.mocked` instead:
```suggestion
    jest.mocked(getTestGroups).mockReturnValue([TestGroups.DetourPanel])
```

lemald: See also similar cases in `leftNav.test.tsx`.

assets/src/userInTestGroup.ts
```diff
@@ -2,6 +2,7 @@ import getTestGroups from "./userTestGroups"
 
 export enum TestGroups {
   DemoMode = "demo-mode",
+  DetourPanel = "detour-panel",
```
lemald, line 5: **question:** I forget, have we decided on an overall test group for this? Should this be that test group or should the dummy page have its own test group? Either way I'm not sure `detour-panel` is really the best name, maybe just `detours` if it's for the overall feature or `dummy-detour-page` if it's just for this page.

assets/src/navLinkData.ts
lemald: I actually think it's fine, to be honest.

assets/src/userInTestGroup.ts
joshlarson: We did not decide on a name. I just came up with this spur-of-the-moment.

Let's chat tomorrow about what a good name and scope would be for this. I think I've sort of talked myself into liking `dummy-detour-panel` so that it becomes pretty obvious that we should get rid of it once we don't need the dummy panel anymore.

But maybe we want like... `detour-alpha` or something like that so that we can just add ourselves to that group and then forget?

lemald: > Let's chat tomorrow about what a good name and scope would be for this. I think I've sort of talked myself into liking `dummy-detour-panel` so that it becomes pretty obvious that we should get rid of it once we don't need the dummy panel anymore.

Sounds good!

assets/tests/components/nav/bottomNavMobile.test.tsx
joshlarson: I thought about that, but I didn't do it in this PR because I was faced with the choice of either:
 - Make this test different from all the surrounding ones; OR
 - Explode this diff by fixing all of them.

Maybe this would be worth fixing as a follow-up? Or maybe as a pre-factor?

lemald: I think having these tests be slightly different from the surrounding ones is fine.

assets/src/components/dummyDetourPage.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  useEffect(() => {
+    fetchShapeForRoute("39").then((shapes) => {
+      setRouteShape(shapes[0])
+    })
+  }, [])
+
+  return (
+    <div className="l-page">
+      {routeShape && <DetourMap shape={routeShape} />}
```
firestack, line 17: question: is there a reason for specifically using `DetourMap` here and not `DiversionPage`?

lemald: Oh this is probably just left over from the version of the code that I had in my branch when I was solely working on my own `<DetourMap/>` component.

firestack: Oh okay, I'm fine with it as is if it's changing incrementally in another branch.

lemald (approved):

PR #2353, opened 2024-01-08
fix(ts/components/minischedule): rename components to avoid collision with imported type names

firestack:
Found when adding new storybook pages, due to babel complaining about the conflict.

Basically the same as #2352.


lemald (approved):
Honestly this is helpful for readability regardless of whether or not there are any issues with Storybook.

PR #2354, opened 2024-01-08
feat(sb/components): add main Skate app and Pages

firestack:
---
  - https://app.asana.com/0/0/1206275026123290


assets/stories/skate-components/pages/app.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+    (StoryFn) => (
+      <MemoryRouter initialEntries={["/"]}>
+        <StoryFn />
+      </MemoryRouter>
+    ),
+    (StoryFn) => (
+      <AppStateProvider>
+        <StoryFn />
+      </AppStateProvider>
+    ),
```
lemald, line 24: **question (non-blocking, just curious):** do you see a specific advantage to chaining decorators rather than having a single decorator that wraps the story in both the memory router and app state providers?

assets/stories/skate-components/pages/ladderPage.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { RoutesProvider } from "../../../src/contexts/routesContext"
+import { LadderDirection } from "../../../src/models/ladderDirection"
+
+import stateFactory from "../../../tests/factories/applicationState"
+import routeFactory from "../../../tests/factories/route"
+import routeTabFactory from "../../../tests/factories/routeTab"
+import { RealDispatchWrapper } from "../../../tests/testHelpers/wrappers"
+
+routeFactory.rewindSequence()
+const routes = routeFactory.buildList(6, { garages: ["Garage A", "Garage B"] })
```
lemald, line 14: **nitpick:** the `routeFactory` by default creates routes with names like `Route 1`, which has been fine for our unit tests and serves to distinguish the name from the ID. However, for visual purposes these look very unlike our actual route designations. What do you think about updating it so that the route names look more "realistic?"

lemald (approved):
A couple of non-blocking questions / notes, but overall looks good!

assets/stories/skate-components/pages/app.stories.tsx
firestack: I was having trouble deciding on this.
On one hand, I think doing them "chained" better fits the "single responsibility principle", but that's more a coincidence than a intention. I did it like this to make diff's nicer, but couldn't decide if I wanted them merged or not.

assets/stories/skate-components/pages/ladderPage.stories.tsx
firestack: I was half expecting the change to end up affecting a bunch of tests, but it worked out of the box!
I like how descriptive it is, but I'm good either way.

Should we just use the sequence number? or randomly add extra tokens, like `16L` or `443/444` etc.

firestack: f2eeaf87fb153c3e90888de836231f37c0a28c7d

lemald: Oh, sorry, I meant to respond to your above comment when I first saw it! Yeah, I think anything that results in unique strings of somewhere between 1-3 numerals is pretty realistic for our use case here.

assets/stories/skate-components/pages/app.stories.tsx
firestack: I'm leaning toward leaving it, just because it keeps indentation lower, do you have any preferences on this before I merge?
cc: @joshlarson

lemald: I'm fine with how it is! Was mainly just wondering what your thoughts were.

PR #2355, opened 2024-01-09
fix:test(ts/factories/route): fix route name generation

firestack:
Seems that when addressing this [feedback on storybook page addition PR](https://github.com/mbta/skate/pull/2354#discussion_r1445104632), I didn't actually check that the change did what I was expecting. So I've fixed it this time and double checked that the names are rendering correctly.


lemald (approved):
Gah, I should have caught this one in code review! Thanks for fixing.

PR #2357, opened 2024-01-10
chore: Remove User.in_test_group?/2, which isn't called anywhere

joshlarson:


lemald (approved):

lemald:
I have slightly mixed feelings on this because we're not using it at the moment but we'll probably use it again plenty of times, but thanks to git history and everything it will be very easy to resurrect this function whenever we need it.

joshlarson:
@lemald I actually wonder if we might be able to use something like
```elixir
user
|> User.all_test_group_names()
|> Enum.member?("whatever")
```
instead?

Which would then have an a priori guarantee that listing a user's test groups and checking whether a user is in a test group will never get out of sync with one another (I almost goofed on that when I was adding [the override feature](https://github.com/mbta/skate/pull/2335).

joshlarson:
But if we think we're going to want to resurrect this, then I almost would rather _not_ delete it.

lemald:
> @lemald I actually wonder if we might be able to use something like
> 
> ```elixir
> user
> |> User.all_test_group_names()
> |> Enum.member?("whatever")
> ```
> 
> instead?
> 
> Which would then have an a priori guarantee that listing a user's test groups and checking whether a user is in a test group will never get out of sync with one another (I almost goofed on that when I was adding [the override feature](https://github.com/mbta/skate/pull/2335).

Yes, that's a good point, and it's simple enough logic for a user of the test group API to implement. I say go ahead and delete!

PR #2358, opened 2024-01-10
Keycloak SSO

lemald:
Asana ticket: [⚙️ migrate skate from Cognito to Keycloak](https://app.asana.com/0/1148853526253420/1206141963656816/f)

This completely removes the application code related to Cognito and replaces it with Keycloak. We might also want to take a different approach where the application code uses Keycloak or Cognito depending on which is configured, but that would involve more work and for now we haven't decided on a deployment strategy for this. If it does become necessary we can make those changes.


lib/skate/ueberauth/strategy/fake.ex
```diff
@@ -6,7 +6,7 @@ defmodule Skate.Ueberauth.Strategy.Fake do
   @impl Ueberauth.Strategy
   def handle_request!(conn) do
     conn
-    |> redirect!("/auth/cognito/callback")
+    |> redirect!("/auth/keycloak/callback")
```
firestack, line 9: suggestion(non-blocking): can we use verified routes here?

lemald: We can! c960b43d

firestack: Cool! there's a few other places I noticed this url being used/updated without verified routes, I can go call them out if you'd like, but I consider this to be non blocking. So if you felt like updating them, I wouldn't say no 👀

lemald: So we should definitely use verified routes for things like calls to `get`, what's your opinion on using it in assertions? Like should `assert redirected_to(conn) == "/foo"` become `assert redirected_to(conn) == ~p"/foo"`?

firestack: (Sorry didn't see your comment till just now) If it's possible to use verified routes for assertions, I think it's worthwhile, as it attempts to convey that we intend the route to exist and therefore get compilation errors?

joshlarson:
> We might also want to take a different approach where the application code uses Keycloak or Cognito depending on which is configured

Oof - this is annoying... I just typed out a whole argument for why we should use test groups/feature flags for this, but you can't do login-based feature flagging on the login page itself 🤦‍♂️ 

I guess the advantage of having it be configurable would be that then like... Lena or Ashli could flip the switch, instead of it being a deploy? And it would be easier to roll back if something goes wrong (like inspectors write in en-masse saying that they can't login)? And we could merge this PR now without having to remember that the next prod deploy will make this abruptly go  live.

So ... I guess a lot of advantages. Not sure how much work though, or whether the advantages are worth it.

config/dev.exs
```diff
@@ -44,7 +44,16 @@ config :ex_aws,
 
 config :ueberauth, Ueberauth,
   providers: [
-    cognito: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher", "skate-admin"]]}
+    keycloak: {Skate.Ueberauth.Strategy.Fake, [groups: ["skate-dispatcher", "skate-admin"]]}
+  ]
+
+config :ueberauth_oidcc,
```
joshlarson, line 50: Are we using Keycloak locally?

lemald: We're using the fake strategy that always logs the user in, but the key part is that it's called "keycloak" for the purposes of URL paths and such so that no other application code needs to change.

lemald:
> > We might also want to take a different approach where the application code uses Keycloak or Cognito depending on which is configured
> 
> Oof - this is annoying... I just typed out a whole argument for why we should use test groups/feature flags for this, but you can't do login-based feature flagging on the login page itself 🤦‍♂️
> 
> I guess the advantage of having it be configurable would be that then like... Lena or Ashli could flip the switch, instead of it being a deploy? And it would be easier to roll back if something goes wrong (like inspectors write in en-masse saying that they can't login)? And we could merge this PR now without having to remember that the next prod deploy will make this abruptly go live.
> 
> So ... I guess a lot of advantages. Not sure how much work though, or whether the advantages are worth it.

The question is - configurable how? Where would this switch live? If it lives in the Skate application, how does someone access it to switch it back to the old value if there are problems, given they would need to access Skate to do so? Does this configuration bit live in a new global application configuration table in our database, or somewhere else? We could make it configurable out of band from the application itself somehow via environment variables, but that's actually harder to roll forward and back given that it requires someone outside of this team to review a Terraform PR and merge / apply it. I think sort of counterintuitively the current approach is the most flexible one available without major additional work.

joshlarson:
> The question is - configurable how? Where would this switch live? If it lives in the Skate application, how does someone access it to switch it back to the old value if there are problems, given they would need to access Skate to do so?

I was thinking in the Skate database. You're right that that wouldn't work for problems that are like "Keycloak is completely broken and we need to switch back to Cognito", but I can think of other types of problems too. Problems like "some inspectors can't login because of some weird Cognito/Keycloak inconsistency", or "we rolled this out too fast and people are confused and annoyed". 

The "everything is broken everybody panic" problem is the only one where ☝️ doesn't work.

> Does this configuration bit live in a new global application configuration table in our database, or somewhere else? We could make it configurable out of band from the application itself somehow via environment variables, but that's actually harder to roll forward and back given that it requires someone outside of this team to review a Terraform PR and merge / apply it. I think sort of counterintuitively the current approach is the most flexible one available without major additional work.

Based on ☝️, I would vote for "live in a new global application configuration table in our database".

---

Honestly, I don't think it's a huge deal either way. The only thing is that if we merge this as-is, then the next time we deploy is when we force everyone to switch to the new login flow. I think it's a product question whether or not that's acceptable, or whether we need to give them time to communicate to inspectors and dispatchers about the changes.

If we're not up for doing that, then this PR has to sit open until we're ready, which might be a little while.

joshlarson (approved):

lib/skate_web/auth_manager/error_handler.ex
```diff
@@ -7,16 +7,6 @@ defmodule SkateWeb.AuthManager.ErrorHandler do
...
-    if auth_retries > 0 do
-      conn
-      |> put_session(:auth_retries, auth_retries - 1)
-      |> Phoenix.Controller.redirect(to: ~p"/auth/cognito")
-    else
-      conn
-      |> delete_session(:auth_retries)
-      |> send_resp(:unauthorized, "unauthorized")
-    end
+    Phoenix.Controller.redirect(conn, to: ~p"/auth/keycloak")
```
firestack, line 10: question(non-blocking): just wondering, if we changed the redirect based on a _global_ feature flag (i.e., a feature flag that doesn't have any users, is just on or off), could we temporaroly support keycloak AND cognito so we could roll this out and release it without requiring a deploy?

Because this has been sitting a while and we're getting a bit ansty about changes building up, maybe we could merge this and use a feature flag to release both the new keycloak login AND the new logout and other UI changes all at once (and test internally in prod without affecting users).

> I guess the advantage of having it be configurable would be that then like... Lena or Ashli could flip the switch, instead of it being a deploy? And it would be easier to roll back if something goes wrong (like inspectors write in en-masse saying that they can't login)? And we could merge this PR now without having to remember that the next prod deploy will make this abruptly go live.
> -- https://github.com/mbta/skate/pull/2358#issuecomment-1907095808

I also like the idea of being able to release this without requiring a deploy.

> The question is - configurable how? Where would this switch live? If it lives in the Skate application, how does someone access it to switch it back to the old value if there are problems, given they would need to access Skate to do so?
> -- https://github.com/mbta/skate/pull/2358#issuecomment-1908193335

Since we've confirmed that we can log-in in dev blue I'm a bit more comfortable with the idea of deploying this behind a feature flag as noted at the beginning of this.
I'm thinking that if we have cognito "enabled" (globally) then the Skate team can test that keycloak is working in prod before release by explicitly going to `<skate>/auth/keycloak`.
once we cutover to keycloak, if things are failing, we can login by explicitly going to `<skate>/auth/cognito` which would allow us access to the test group panel, where we could disable the keycloak flag, so we wouldn't need out of band ways to disable it.

I'm not sure that I've thought of every edge case, but I'd like to put this up for discussion since we're running into more difficulties with how dev would like to merge things and test iteratively and design+product would like to release it all at once.


cc: @mbta/skate-developers

lemald: I have also been thinking about this as we've discussed the plans for logout. I suspect it may be possible to do something like this with our existing test group functionality and the override feature, though I still think it's a bit of a hack. I'm going to spend some time this morning investigating and trying it out. One thing I want to make sure we're all clear on is that, while this may help coordinate rollout, this is the sort of thing where if it breaks it will probably break for everyone, so for a rollback we still need to be prepared to do a deploy.

joshlarson: I agree with pretty much all of this.

The one thing I would want to keep in mind is that we have an ambitious timeline goal that we're trying to meet, so if we do anything that might cause this to go past the end of the sprint, we should let everyone else know.

lemald: Okay with 51832764 / b2fc5410 I've got something that seems to work up on dev-blue.

firestack (approved):
@lemald What do you think about rebasing the Cognito removal commits out of this and into it's own branch so we've got a starting point when we do remove Cognito?
I figured since you've done the work of removing it, it'd not be worth losing, but it would still exist in the history of this PR, so I don't mind if you don't.

lemald:
> @lemald What do you think about rebasing the Cognito removal commits out of this and into it's own branch so we've got a starting point when we do remove Cognito? I figured since you've done the work of removing it, it'd be worth keeping. But the commits would still exist in the history of this PR, so I don't mind if you don't.

Having done the removal originally, I personally think just glancing at the commits in this PR and redoing it wouldn't really be harder than the git work.

firestack (approved):

joshlarson (approved):
Re-approving because I like the on/off switch

PR #2362, opened 2024-01-11
fix(ts/math): fix and test `closestPosition`

firestack:
The old implementation was not tested and returned the input `position` instead of the closest point within `positions`.

### DIFF
--- a/assets/src/util/math.ts
+++ b/assets/src/util/math.ts
@@ -1,3 +1,5 @@
+import { LatLng, LatLngExpression } from "leaflet"
+
 /**
  * A helper function to help clarify code when clamping a value between a range.
  * (Remove when tc39 implements `Math.clamp`)
@@ -7,3 +9,26 @@
  */
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(min, value), max)
+
+/**
+ * Finds the element in {@link positions} that is closest to {@link point}
+ * @param positions List of coordinates to search
+ * @param point Point of reference to check distance to
+ * @returns The closest point in {@link positions} to {@link point}
+ */
+export const closestPosition = (
+  positions: LatLngExpression[],
+  point: LatLng
+) => {
+  const positionsByDistance = positions
+    .map((currentPosition, index) => ({
+      distance: point.distanceTo(currentPosition),
+      position: currentPosition,
+      index,
+    }))
+    .sort((lhs, rhs) => lhs.distance - rhs.distance)
+
+  return positionsByDistance[0] ?? undefined
+
+  // Interpolate here
+}
--- a/assets/src/util/math.ts
+++ b/assets/src/util/math.ts
@@ -10,16 +10,29 @@ import { LatLng, LatLngExpression } from "leaflet"
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(min, value), max)
 
+
+interface ClosestPosition {
+  position: LatLngExpression,
+  index: number
+  distance: number
+}
+
 /**
  * Finds the element in {@link positions} that is closest to {@link point}
  * @param positions List of coordinates to search
  * @param point Point of reference to check distance to
- * @returns The closest point in {@link positions} to {@link point}
+ * @returns
+ * if {@link positions} is empty, returns `undefined`
+ *
+ * Otherwise, returns a {@link ClosestPosition} object containing
+ * - the closest element in {@link positions} to {@link point}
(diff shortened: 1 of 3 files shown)


assets/src/util/math.ts
```diff
@@ -7,3 +9,26 @@
...
+    .map((currentPosition, index) => ({
+      distance: point.distanceTo(currentPosition),
+      position: currentPosition,
+      index,
+    }))
+    .sort((lhs, rhs) => lhs.distance - rhs.distance)
+
+  return positionsByDistance[0] ?? undefined
+
+  // Interpolate here
```
firestack, line 33: I didn't quite mean to leave this in, but also, I definitely think we'll need to do this at some point, because it'll otherwise feel weird if someone clicks on the line, and there isn't a point nearby, and it snaps to that point, instead of the point on the line.

lemald: Agreed it's fine to leave a comment, but maybe expand on it a bit to explain that if we do want to interpolate in the future, this is where we'd do it (as it stands I can imagine someone being confused by seeing this comment and wondering where the interpolation logic is).

assets/src/util/math.ts
```diff
@@ -7,3 +9,26 @@
...
+ * @param point Point of reference to check distance to
+ * @returns The closest point in {@link positions} to {@link point}
+ */
+export const closestPosition = (
+  positions: LatLngExpression[],
+  point: LatLng
+) => {
+  const positionsByDistance = positions
+    .map((currentPosition, index) => ({
+      distance: point.distanceTo(currentPosition),
```
lemald, line 25: **praise:** nice, I didn't realize there was a handy method you could use for distance without needing to reference the map object.

assets/src/util/math.ts
```diff
@@ -7,3 +9,26 @@
  */
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(min, value), max)
+
+/**
+ * Finds the element in {@link positions} that is closest to {@link point}
+ * @param positions List of coordinates to search
+ * @param point Point of reference to check distance to
+ * @returns The closest point in {@link positions} to {@link point}
```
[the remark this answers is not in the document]
firestack: What do you think of 16fdac1faadca6f1f1112b8ae0665416a32db45a (and f8dc7a151776b680836eab46d35407a105e000b1) ?

assets/src/util/math.ts
firestack: I'm open to/would appreciate feedback on the wording if anything stands out, PTALA bbe5cb2839e837f93f93365779714be33f1179c0

lemald: Looks good to me!

lemald (approved):
Thanks for catching and fixing this.

assets/src/util/math.ts
```diff
@@ -7,3 +9,46 @@
...
+  positions: LatLngExpression[],
+  point: LatLng
+): ClosestPosition | undefined => {
+  const positionsByDistance = positions
+    .map((currentPosition, index) => ({
+      distance: point.distanceTo(currentPosition),
+      position: currentPosition,
+      index,
+    }))
+    .sort((lhs, rhs) => lhs.distance - rhs.distance)
```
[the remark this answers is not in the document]
firestack: I don't believe we have a function like that right now, aside from implementing it as a `reduce` (https://developer.mozilla.org/en-US/search?q=maxby)

I _personally_ found the `reduce` operation harder to read, and this sets us up to do interpolation afterward, and I imagine the `sort` is negligible compared to a single pass `reduce` at the number of points we're working with.

I'd be happy to look into it further if you have any suggestions!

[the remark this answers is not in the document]
joshlarson: Nahh - what you said makes sense. If there are a lot of points, then I could see `sort` being slower, but like you said, marginally - maybe by a factor of 2 or something.

If we start seeing that this is a problem, then we can revisit it - my idea would be to get the two nearest points (which I think could be done in two passes, rather than potentially `log_base_2(n)`).

assets/src/util/math.ts
firestack: > Do we have a ticket in the backlog to do this interpolation?

Not yet, because I figured we'd wait till it's a problem, but I can write one!

assets/src/util/math.ts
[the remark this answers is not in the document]
firestack: Another benefit, I'm returning `index` because it's possible that a non index adjacent point could be closer than other points nearby

svg

PR #2364, opened 2024-01-17
refactor(ts/components/detourMap): move state up in tree

firestack:
Refactors the state from `<DetourMap/>` into `<DiversionPage/>`


assets/src/components/detours/detourMap.tsx
```diff
@@ -1,152 +1,120 @@
...
-        positions={detourShapePositions}
-        className="c-detour_map--detour-route-shape"
-      />
-      {uniqueWaypoints.map((position) => (
-        <DetourPointMarker key={position.toString()} position={position} />
-      ))}
-    </>
-  )
+
+  onClickOriginalShape: onAddConnectionPoint,
```
joshlarson, line 65: Nit - should we stick with either `onClickOriginalShape` or `clickOriginalShape` for this, rather than `addConnectionPoint`? Given this refactor, it feels like the fact that this operation adds a connection point is a diversion-page fact, not a detour-map fact.

assets/src/components/detours/detourMap.tsx
```diff
@@ -1,152 +1,120 @@
...
+  onClickMap: onAddWaypoint,
+
+  undoDisabled,
+  onUndoLastWaypoint: undoLastWaypoint,
```
joshlarson, line 69: I'm honestly torn about this. On the one hand, it's a little confusing to me to have the variable have a different name than the parameter. On the other hand, it does make each side of this boundary read more nicely:

```typescript
/* ... */
onUndoLastWaypoint={foobar}

/* ... */
/* Someone clicks on the undo button */
undoLastWaypoint() // calls foobar
// versus
onUndoLastWaypoint() // is more obviously linked to foobar, but doesn't read as nicely
```

assets/src/components/detours/diversionPage.tsx
```diff
@@ -1,7 +1,107 @@
...
+    fetchDetourDirections(shapePoints).then((detourShape) => {
+      if (detourShape && shouldUpdate) {
+        setDetourShape(detourShape.coordinates.map(shapePointToLatLngLiteral))
+      }
+    })
+
+    return () => {
+      shouldUpdate = false
+    }
+  }, [shapePoints])
```
joshlarson, line 27: I don't think this is memoizing properly. When I run this, `fetchDetourDirections` is being called potentially infinitely many times (well, 40 per minute before the API 500's)

assets/src/components/detours/diversionPage.tsx
```diff
@@ -1,7 +1,107 @@
...
+const useDetour = () => {
+  const [startPoint, setStartPoint] = useState<LatLngLiteral | null>(null)
+  const [endPoint, setEndPoint] = useState<LatLngLiteral | null>(null)
+  const [waypoints, setWaypoints] = useState<LatLngLiteral[]>([])
+
+  const detourShape = useDetourDirections(
+    [startPoint, ...waypoints, endPoint]
+      .filter((v): v is LatLngLiteral => !!v)
+      .map(latLngLiteralToShapePoint)
+  )
```
joshlarson, line 41:
```suggestion
  const allPoints = useMemo(
    () =>
      [startPoint, ...waypoints, endPoint]
        .filter((v): v is LatLngLiteral => !!v)
        .map(latLngLiteralToShapePoint),
    [startPoint, endPoint, waypoints]
  )

  const detourShape = useDetourDirections(allPoints)
```

I think this fixes the infinite API call issue above. (You've unleashed a monster! A memoized monster! So I guess only one monster at a time unless parameters change.)

assets/src/components/detours/diversionPage.tsx
joshlarson: I think [this](https://github.com/mbta/skate/pull/2364/files#r1456500283) fixes it

assets/src/components/detours/diversionPage.tsx
joshlarson: I have no attachment at all to the name `allPoints`, fwiw

assets/src/components/detours/detourMap.tsx
firestack: Whoops! that was unintentional, sorry about that.
See 382b93547fa632bf40e5d5862965ba62eb99acd8

assets/src/components/detours/detourMap.tsx
firestack: https://github.com/mbta/skate/pull/2364#discussion_r1457496176
382b93547fa632bf40e5d5862965ba62eb99acd8

assets/src/components/detours/diversionPage.tsx
```diff
@@ -1,7 +1,107 @@
...
+
+    return () => {
+      shouldUpdate = false
+    }
+  }, [shapePoints])
+
+  return detourShape
+}
+
+const useDetour = () => {
```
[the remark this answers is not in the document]
firestack: ce25b300
fba38eb7

assets/src/components/detours/diversionPage.tsx
firestack: Thanks! 84afffa8

assets/src/components/detours/detourMap.tsx
firestack: This was a side-effect of using VSCode to rename the symbols and forgetting to clean up after it 🫠

assets/tests/hooks/useDetour.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      start,
+      end,
+    ])
+
+    await waitFor(() =>
+      expect(result.current.detourShape).toStrictEqual([start, end])
+    )
+  })
+
+  test("when `undoLastWaypoint` is called, removes the last `waypoint`", async () => {
```
joshlarson, line 90: Do we also want a test asserting that when `undoLastWaypoint` is called, the API is called again?

assets/tests/hooks/useDetour.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  test("when `addWaypoint` is called, `detourShape` is updated", async () => {
+    const start = { lat: 0, lon: 0 }
+    const end = { lat: 1, lon: 1 }
+
+    jest
+      .mocked(fetchDetourDirections)
+      .mockImplementation((coordinates) =>
+        Promise.resolve({ coordinates: coordinates })
+      )
```
joshlarson, line 71: Nit: This test would pass if the [`useEffect` in `useDetourDirections`](https://github.com/mbta/skate/pull/2364/files#diff-ea299f4c918f1648d9f7ab5eb825737dd5e598ed5f3d3a9db036ae03d447d224R8-R24) just ignored the API and returned the input. 

I'd recommend changing it to something like 👇, along with the associated 

```suggestion
    jest
      .mocked(fetchDetourDirections)
      .mockImplementation((coordinates) =>
        Promise.resolve({ coordinates: [
          { lat: 0, lon: 0 },
          { lat: 1, lon: 0},
          { lat: 1, lon: 1},
        ] })
      )
```

firestack: baef5e3196b0a44cd57d1a515c7c2c765f5214bb

assets/tests/hooks/useDetour.test.ts
firestack: 9c76ca2af8f7810282a09c3577d9f1b8ced44dcb

lemald (approved):
I personally think this looks good!

PR #2366, opened 2024-01-17
feat: Add tooltip that says "Click to start detour"

joshlarson:


assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1,7 +1,6 @@
 import { jest, describe, test, expect, beforeEach } from "@jest/globals"
-import { render, screen } from "@testing-library/react"
+import { fireEvent, render, screen } from "@testing-library/react"
```
joshlarson, line 2: Without this, almost all tests in this file fail with this error: 
```
 Invalid LatLng object: (NaN, NaN)
```

Maybe this change should be moved to a PR that comes before this one?

joshlarson: Why these tests suddenly start failing with that error as the result of a tooltip... I have absolutely no idea.

assets/css/detours/_detour_map.scss
```diff
@@ -15,3 +15,18 @@
...
+.c-detour_map--tooltip {
+  background-color: black;
+  border-color: black;
+  color: white;
+  opacity: 100% !important;
+  font-size: 14px;
+  padding: 4px 8px 4px 8px;
+  font-family: var(--bs-body-font-family);
+  border-radius: var(--bs-border-radius);
+}
```
lemald, line 32: **suggestion:** you could get at least some of this styling from just importing the Bootstrap Tooltip styling (uncommenting like 52 in `css/bootstrap.scss`). You could probably also get some more of the styling by updating Bootstrap SCSS variables rather than writing your own CSS rules (though some may still be necessary).

assets/src/components/detours/detourMap.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react"
 import { LatLngLiteral } from "leaflet"
-import { Polyline, useMapEvents } from "react-leaflet"
+import { Polyline, Tooltip, useMapEvents } from "react-leaflet"
```
lemald, line 3: **issue:** I see - this is actually a Leaflet tooltip, not a Bootstrap tooltip. I think the idea is that we wanted to use Bootstrap, not just have a style that visually matches Bootstrap. I could very well have missed some conversation last week, though.

lemald: Granted, using Bootstrap / React-Bootstrap does have potential hurdles when it comes to the interactions with Leaflet on both the JavaScript and CSS fronts, but I'd like to at least think it through as an option.

joshlarson: So @firestack and I paired for a little bit last week to try to get this to be a Bootstrap tooltip, and we couldn't really figure it out. From looking through the bootstrap docs, I couldn't even figure out how to attach a boostrap tooltip to a Polyline and have it actually show up, and even trying to style the Leaflet tooltip using bootstrap styles didn't work (the width was getting compressed for reasons that neither Kayla nor I was able to figure out). 

Do you have any idea how to go about that? Maybe I missed something simple in the bootstrap or react-bootstrap docs?

And also, isn't it an implementation detail whether we use Leaflet tooltips versus Bootstrap tooltips under the hood?

lemald: Okay, if you all spent that much time on it and couldn't figure something out then I believe you. I just ask because the original ticket specifically said "Bootstrap tooltip."

> And also, isn't it an implementation detail whether we use Leaflet tooltips versus Bootstrap tooltips under the hood?

Sort of - from a purely user-facing functionality perspective, yes. However from a tooling perspective, both dev and design get certain benefits if it's a real Bootstrap component in terms of obeying certain application-wide styles and updating if they update. That said as we've established the "real Bootsrap" approach probably isn't technically feasible here, so it's fine to take another route.

assets/tests/components/detours/diversionPage.test.tsx
lemald: Yeah, this is also stumping me and I can't find anything relevant in the Testing Library docs. I don't think it's strictly a problem, but still weird.

lemald (approved):

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: Yeah, I only "figured it out" by asking Kayla about it last week, with the hope (which turned out to be correct!) that this was an error that the team had seen before.

assets/src/components/detours/detourMap.tsx
joshlarson: Do we get some of those benefits by pulling in the bootstrap variables the way I did in the CSS?

And/Or is there any way to get some or all of those benefits anyway?

assets/stories/skate-components/map/tooltip.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const meta = {
+  component: MapTooltip,
+  args: {
+    title: "Click to start detour",
+    permanent: true,
+  },
+  argTypes: {
+    permanent: { table: { disable: true } },
+  },
```
firestack, line 15: question(non-blocking): why disable this arg?

assets/src/components/map/tooltip.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+interface MapTooltipProps {
+  title: string
+  permanent?: boolean
+}
+
+export const MapTooltip = ({ title, permanent }: MapTooltipProps) => {
+  return (
+    <Tooltip
+      key={`tooltip-permanent-${permanent}`}
```
firestack, line 12: question: is this `key` prop needed?

assets/src/components/map/tooltip.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import { Tooltip } from "react-leaflet"
+
+interface MapTooltipProps {
+  title: string
+  permanent?: boolean
+}
+
+export const MapTooltip = ({ title, permanent }: MapTooltipProps) => {
+  return (
```
firestack, line 10: nit(non-blocking): the braces and `return` keyword could be removed

assets/src/components/map/tooltip.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <Tooltip
+      key={`tooltip-permanent-${permanent}`}
+      className="c-detour_map--tooltip"
```
firestack, line 13: If this is becoming a reusable component, maybe we should change this class to be it's own component, `c-map-tooltip`?

assets/src/components/map/tooltip.tsx
joshlarson: Sooooort of not really?

`<Tooltip />` doesn't re-render properly because Leaflet and React aren't the best of friends, so giving this a key to force a re-render is necessary for that reason.

BUT.... that's not really a use case I see us using - a tooltip will either be permanent or not.

assets/stories/skate-components/map/tooltip.stories.tsx
joshlarson: Without the `key` prop, this toggle doesn't actually do anything (because leaflet `Tooltip`'s don't re-render properly).

(With the `key` prop, this toggle makes the tooltip appear and disappear with a pretty glitchy animation 😂)

The reason I added `permanent` at all was so that in Storybook, when you navigate to the page for this component, there's actually a tooltip visible. Otherwise, you'd have to hover in order to see it, which is kind of annoying.

assets/src/components/map/tooltip.tsx
firestack: Oooh I see, that makes sense, I wanna think about it some more but I like this solution so far.

firestack: at the very least, we probably wanna document this

assets/src/components/map/tooltip.tsx
```diff
@@ -6,17 +6,15 @@ interface MapTooltipProps {
...
-      permanent={permanent}
-    >
-      {title}
-    </Tooltip>
-  )
-}
+export const MapTooltip = ({ title, permanent }: MapTooltipProps) => (
+  <Tooltip
+    key={`tooltip-permanent-${permanent}`}
+    className="c-map_tooltip"
```
firestack, line 12: nit: I would name this using dashes instead of underlines
```suggestion
    className="c-map-tooltip"
```

assets/src/components/map/tooltip.tsx
joshlarson: https://github.com/mbta/skate/pull/2366/commits/22f35ce287a76ed7db9662696bf1e56cda339011

assets/src/components/map/tooltip.tsx
joshlarson: https://github.com/mbta/skate/pull/2366/commits/2b518692c199a95ffb667592e3b3d4342943429b

assets/src/components/map/tooltip.tsx
joshlarson: https://github.com/mbta/skate/pull/2366/commits/0510b487eced731c65607671ae33452a40f15120

firestack: Thanks!

assets/src/components/map/tooltip.tsx
```diff
@@ -0,0 +1,20 @@
...
+    key={`tooltip-permanent-${permanent}`}
+    className="c-map-tooltip"
+    direction="top"
+    offset={[0, -15]}
+    sticky
+    permanent={permanent}
+  >
+    {title}
+  </Tooltip>
+)
```
firestack, line 20: question: just wondering, do you think the API of this component is better off using a `title` prop instead of `children`? 
Right now, this component ignores that `<Tooltip>` can take a `ReactNode` and not just a `string`, as well as essentially moves `<Tooltip>`'s `children` into the `title` prop.

joshlarson: Ooooh that's a good point.

This was a hack because I didn't know the right types to make `children` behave correctly. `PropsWithChildren`.... good to know!

firestack: and `PropsWithChildren` is just a short cut to get it right "Always™️" (like, in the case they change the type signature or something), but if you look, it's just defined as 

```ts
type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };
```

So you _could_, define `title` like `{ title?: ReactNode | undefined }` if you really preferred to keep `title`

joshlarson: https://github.com/mbta/skate/pull/2366/commits/7d8713434b6f83f4aeff7231045929b6de98ca51
https://github.com/mbta/skate/pull/2366/commits/02fecf76217896a4a6483b54ab9bd51f8976bffb

firestack (approved):

PR #2367, opened 2024-01-18
feat(skate/detours/turn-by-turn): add turn-by-turn directions to detours page

firestack:
---
  - https://app.asana.com/0/0/1206038961362406


lib/skate/open_route_service_api.ex
```diff
@@ -21,6 +21,18 @@ defmodule Skate.OpenRouteServiceAPI do
...
             %{"lat" => 1, "lon" => 0}
+          ],
+          directions: [
+            %{
+              name: "1st Avenue",
+              type: :right,
+              instruction: "Turn right onto 1st Avenue"
+            },
+            %{
+              name: "2nd Place",
```
joshlarson, line 32: 🥈

joshlarson:
*praise:* It makes me very happy that `detourMap.tsx` didn't need to change at all!

assets/src/components/detours/diversionPage.tsx
```diff
@@ -3,9 +3,9 @@ import { DiversionPanel, DiversionPanelProps } from "./diversionPanel"
 import { DetourMap } from "./detourMap"
 import { Shape } from "../../schedule"
 import { useDetour } from "../../hooks/useDetour"
+import { ListGroup } from "react-bootstrap"
 
 export const DiversionPage = ({
-  directions,
   missedStops,
```
joshlarson, line 9: Not  for this PR, but will `missedStops` also get dropped from the public contract for this component? That also seems like something that might get computed _within_ `DiversionPage` rather than outside of it.

firestack: I'm still on the fence about where stuff should live ([to use a MVC analogy] right now, I see `useDetour` as like, the model, `DiversionPage` is acting as a controller in a way, but `DiversionPage` is also the view)

But I think that `missedStops` will _likely_ be removed, still mulling it over though.

joshlarson: I imagine that sometime soonish, it'll make sense to give `DiversionPage` a treatment similar to the one you gave to `DetourMap`, so that it's pretty much only dealing with the rendering side of the world (and noticing that there's input, I guess?). 

I almost would rather see that as a follow-up PR, since as it stands, this doesn't look too painful, and I like the idea of keeping feature work and refactor work separate. I don't feel too strongly about it though, if doing that refactor here feels _right_ to you, then I have no objection.

assets/src/detour.ts
```diff
@@ -0,0 +1,26 @@
...
+  | "goal"
+  | "depart"
+  | "keep_left"
+  | "keep_right"
+
+export interface DetourShape {
+  coordinates: ShapePoint[]
+  directions: {
+    instruction: string
+    name: string
```
joshlarson, line 23: Nit: `name` and `type` aren't being used at all, which I think has implications for the backend as well.

assets/tests/hooks/useDetour.test.ts
```diff
@@ -60,15 +60,20 @@ describe("useDetour", () => {
...
-        Promise.resolve({ coordinates: coordinates })
-      )
+    jest.mocked(fetchDetourDirections).mockImplementation((coordinates) =>
+      Promise.resolve({
+        coordinates: coordinates,
+        directions: [...Array(coordinates.length)].map((_, idx) => ({
+          instruction: idx.toString(),
+          name: idx.toString(),
+          type: "left",
+        })),
```
joshlarson, line 74:
```suggestion
        directions: [
          { instruction: "Turn left on Main St", name: "Main St", type: "left" },
          { instruction: "Slight right on Centre Ave", name: "Centre Ave", type: "slight_right" },
        ],
```

(Along with the corresponding change to the assertion)

I find tests easier to read when they have more explicit test data, and when two things are supposed to be exactly the same, like the `directions` return value from `fetchDetourDirections` and `directions` in the `result.current` assertion below, I'd prefer that exactly-the-sameness to be more explicit. Having these obviously-canned values in the setup and assertion does that.

(Using variables also works for that, like `start` and `end` above, but I tend to prefer the super explicit do-everything-as-literals approach)

Basically, I'd rather not have to spend time thinking about how `[...Array(coordinates.length)].map(...)` works unless that's a core part of what the test is asserting, which I don't think it is in this case.

test/skate_web/controllers/detour_route_controller_test.exs
```diff
@@ -12,7 +12,33 @@ defmodule SkateWeb.DetourRouteControllerTest do
...
+                            "type" => 1
+                          }
+                        ]
+                      }
+                    end)
+                end
+            }
+          }
+        ]
+      }
```
joshlarson, line 41: Sort of the same comment that I made [earlier](https://github.com/mbta/skate/pull/2367/files#r1458081312)... I'd rather have the test data be more explicit. For instance, 

```suggestion
      %{
        "features" => [
          %{
            "geometry" => %{"coordinates" => coordinates},
            "properties" => %{
              "segments" => turn_by_turn_segments,
            }
          }
        ]
      }
```

Where `turn_by_turn_segments` is an argument passed in like `coordinates` (Github won't let me suggest a change to a line you didn't edit 😢).

Then `turn_by_turn_segments` can be passed in as part of the setup of each test (or given a sane default for any test that doesn't care about the turn-by-turns ((I think in order to do sane defaults, we'll need to torture Elixir a bit - keyword args with default values is not one of Elixir's strong core competencies))).

As an added bonus, we could make the test `"returns directions as a flat list"` below a lot clearer, since right now, it's dealing with a coordinate array, rather than the nested `segment` structure.

assets/tests/hooks/useDetour.test.ts
firestack: (posting for discussion)
We could use factories here to abstract away the actual generation of the data so we're not also having to handle that mental overhead _in_ the tests.

A factory for this could look like the following code block, and it could be improved by using the https://fakerjs.dev library.

```ts
import { Factory } from "fishery";
import { DetourShape, DirectionType } from "../../src/detour";
import { localGeoCoordinateFactory } from "./geoCoordinate";

export const directionTypeFactory = Factory.define<DirectionType>(({sequence}) => {
  const directions: DirectionType[] = [
    "depart",

    "sharp_right",
    "left",
    "keep_right",
    "sharp_left",
    "right",
    "keep_left",

    "straight",

    "enter_roundabout",
    "exit_roundabout",

    "u_turn",

    "goal",
  ]
  return directions[sequence % directions.length]
})

export const directionNameFactory = Factory.define<string>(({sequence}) => {
  const things = [
    "Main St",
    "High St",
    "Park Plaza",
    "Boylston St",
    "Massachusetts Ave",
  ]
  return things[sequence % things.length]
})

const makeInstruction = (type: DirectionType, name: string) => {
  switch (type) {
    case "left":
      return `Turn Left onto ${name}`
    case "right":
      return `Turn Right onto ${name}`
    case "keep_left":
      return `Keep Left on ${name}`
    case "keep_right":
      return `Keep Right on ${name}`
    case "sharp_left":
      return `Make a Sharp Left on ${name}`
    case "sharp_right":
      return `Make a Sharp Right on ${name}`
    case "slight_left":
      return `Make a Slight Left on ${name}`
    case "slight_right":
      return `Make a Slight Right on ${name}`
    case "enter_roundabout":
      return `Enter the roundabout and take the 1st exit onto ${name}`
    case "exit_roundabout":
      return `Exit the roundabout at the the 1st exit onto ${name}`
    case "straight":
      return `Continue Straight on ${name}`
    case "depart":
      return `Depart from ${name}`

    default:
      `${type} onto ${name}`;
  }
}

export const directionsFactory = Factory.define<DetourShape["directions"][0]>(({sequence}) => {
  const name = directionNameFactory.build();
  const type = directionTypeFactory.build();
  return ({
    instruction: makeInstruction(type, name),
    name,
    type,
  });
})

export const detourShapeFactory = Factory.define<DetourShape>(({}) => ({
  coordinates: localGeoCoordinateFactory.buildList(3).map(v => ({ lat: v.latitude, lon: v.longitude})),
  directions: directionsFactory.buildList(3)
}))
```

Usage of this factory would look like
```ts
const detourShape = detourShapeFactory.build()
```

I'm also not even certain that we should be mocking out instruction data in `makeInstruction`, as I think the code should treat it as an opaque blob.

test/skate_web/controllers/detour_route_controller_test.exs
firestack: (posting for discussion)

I'd also like to try using factories for this, and have an example implementation in elixir to review
```elixir

  def ors_directions_step_json_factory do
    %{
      "instruction" => sequence("ors_instruction_step"),
      "name" => sequence("ors_instruction_name"),
      "type" => sequence("ors_instruction_type", [0, 1, 2, 3, 4, 5, 6, 7, 12, 13])
    }
  end

  def ors_directions_segment_json_factory do
    %{
      "steps" =>
        build_list(
          sequence("ors_segment_json_num_steps", [4, 1, 3, 2]),
          :ors_directions_step_json
        )
    }
  end

  def ors_directions_json_factory(attrs) do
    coordinates = Map.get(attrs, :coordinates, [[0, 0], [1, 1], [2, 2]])

    segments =
      Map.get(attrs, :segments, build_list(3, :ors_segment_json))

    %{
      "features" => [
        %{
          "geometry" => %{"coordinates" => coordinates},
          "properties" => %{
            "segments" => segments
          }
        }
      ]
    }
    |> evaluate_lazy_attributes()
  end
```

Usage of this is would/could be along the lines of
```elixir
Skate.Factory.build(:ors_directions_json, coordinates: [...])
Skate.Factory.build(:ors_directions_json, %{coordinates: [...]})
Skate.Factory.build(:ors_directions_json, segments: [...], coordinates: [...])
Skate.Factory.build(:ors_directions_json, segments: [...])

Skate.Factory.build(:ors_directions_json, 
	segments: [
		Skate.Factory.build(:ors_directions_segment_json, 
			%{"steps" => Skate.Factory.build_list(2, :ors_directions_step_json, 
				[{"type", 1}])}),
		Skate.Factory.build(:ors_directions_segment_json, 
			%{"steps" => Skate.Factory.build_list(2, :ors_directions_step_json, 
				%{"type" => 0, "name" => "test-name"})})
	]
)

File: assets/tests/hooks/useDetour.test.ts (hunk 4 above)
Comment (joshlarson) [reply assets/tests/hooks/useDetour.test.ts:74]: TL;DR...

- I definitely want to see us either explicitly use factories, or at least use objects that sort of behave like them. I think that'll really help to reduce boilerplate. 
- I strongly prefer to have any data that's asserted on in a test be because of something that was set up explicitly in that test, rather than relying on factory defaults.
- I weakly prefer factories to have their defaults be the simplest possible values that make them valid.
- I agree that we probably don't need `makeInstruction`, and in general I prefer to avoid having too much nontrivial logic in tests.

---

> We could use factories here to abstract away the actual generation of the data so we're not also having to handle that mental overhead in the tests.

If the test cares about a particular chunk of data, then IMO, it creates more overhead if the data generation happens somewhere else.

For instance, a test like:

```javascript
test("handles directions properly", () => {
  const apiData = detourShapeFactory.build()
  jest.mocked(fetchTheThings).mockReturnValue(apiData)

  const directionsContainer = however.we.find.that.on.the.page()

  expect(directionsContainer).toInclude("Depart from Main St")
})
```

In the context of that test, the `"Depart from Main St"` comes out of nowhere, and I have to figure out whether that data came from the factory, some default in the code, or somewhere else. If I saw a test like that, I would assume that it was saying that the _production_ code should default to `"Depart from Main St"` if it wasn't given explicit directions. 

Even something like this:

```javascript
test("handles directions properly", () => {
  const apiData = detourShapeFactory.build()
  jest.mocked(fetchTheThings).mockReturnValue(apiData)

  const directionsContainer = however.we.find.that.on.the.page()

  expect(directionsContainer).toInclude(extractAUsefulStringFrom(apiData.directions))
})
```

...still doesn't give us much insight into what `directions` actually _is_. I don't know if there are any edge cases being validated here, like maybe the factory defaults directions to `null`, or `[]`. (Not to mention that `extractAUsefulStringFrom()` is its own can of worms, and could wind up being almost a re-implementation of the production code)

This is what I want to see:

```javascript
test("handles directions properly", () => {
  const apiData = detourShapeFactory.build({
    directions: [{ instruction: "Turn Right on Centre Blvd" }, /* Etc */]
  })
  jest.mocked(fetchTheThings).mockReturnValue(apiData)

  const directionsContainer = however.we.find.that.on.the.page()

  expect(directionsContainer).toInclude("Turn Right on Centre Blvd")
})
```

So the data in the assertion is set up explicitly when we invoke the factory.

That said, I think factories have their place here very solidly. `apiData` needs to have both `directions` and `coordinates` in order to avoid causing errors, but I like that ☝️ test doesn't say anything about `coordinates`. It just lets the factory go ahead and put _something_ in `coordinates` so that the test can run, but then lets us focus on `directions`-y stuff. 

Basically, they save us from something like 👇 

```javascript
test("handles directions properly", () => {
  const apiData = createAnApiData({
    coordinates: [], // has to be here so that the test doesn't console.error, but we don't actually care
    directions: [{ instruction: "Turn Right on Centre Blvd" }, /* Etc */]
  })
  jest.mocked(fetchTheThings).mockReturnValue(apiData)

  const directionsContainer = however.we.find.that.on.the.page()

  expect(directionsContainer).toInclude("Turn Right on Centre Blvd")
})
```

Where the parameters to `createAnApiData()` get more and more annoying as the API gets more sophisticated.

---

Related to ☝️, I have a weak preference for factories to have the simplest defaults possible, so in this case, `coordinates` would default to `[]` unless we explicitly set it to something else. It's a weak preference, mostly related to the fact that if data in the test has anything interesting going on, then I'd rather have that be more obvious in the test setup itself.

---

> I'm also not even certain that we should be mocking out instruction data in `makeInstruction`, as I think the code should treat it as an opaque blob.

Yeah I agree that we don't need this. Having it would set us up for a "Wait, where did `"Make a Sharp Right on Main St"` come from??" situation, because I'm not sure that it would be obvious that it came from the factory rather than the production code.

In general, I prefer to avoid having nontrivial logic in tests. Enough to get us away from annoyingly repetitive boilerplate, sure, but I'd mostly rather see the data exactly as it's going into the system-under-test, and then assert on the data exactly as it comes back out.

assets/src/detour.ts
firestack: I'm considering dropping these because we're not using them and they were added in speculation.
I'm hesitent because that would mean the `Skate.OpenRouteServiceAPI.map_type/1` function would no longer be used.

I'm fairly certain we will not be using ORS's generated `instruction`(speculation), but that was the easiest to do at first. Given that, do you still think it should be dropped?

assets/tests/hooks/useDetour.test.ts
lemald: I'll caveat this by saying that I haven't read the whole PR or fully processed every bit of this discussion, but my $0.02 is that I think I agree with @joshlarson that if the specific values in a factory-generated piece of data matter for the test, the test should set those explicitly rather than relying on factory defaults..

assets/src/detour.ts
firestack: @lemald do you have any thoughts on this?

joshlarson: A few options:
 - Just remove it, and we'll recreate it if needed (hopefully that wouldn't take too much effort)
 - Remove it, and create a branch reverting the removal so that it'll just be there if/when we need it
 - Mayyyyyybe... generate our own "instruction" based on `type` and `name`, rather than using ORS's? Even if our result would be the same as ORS's, this would still have the advantage of making pretty easy to make any adjustments in the future

I'm fine with any of these... I think I have a slight preference for the third option.

firestack: Just noting this, I found the localization file which actually generates the instructions `https://github.com/GIScience/openrouteservice/blob/ab2be549acc21455a2ad7ca925c74ffb9db39cd9/ors-engine/src/main/resources/resources/locales/ors_en-US.resources#L64`

Something good to keep in mind, this shows that for roundabout instructions, there's an extra `exit_number` value to parse out of the object.

test/skate_web/controllers/detour_route_controller_test.exs
joshlarson: Definitely interested in pursuing this. I'm interested in learning more about the advantages that this has over a slightly expanded `directions_json` (expanded to add `segments` as a parameter).

I assume that we wouldn't want to support both `%{"type" => 0, "name" => "test-name"}` format and `[{"type", 1}]`? Or is there some zaniness in ORS's API that I'm not thinking of?

assets/tests/hooks/useDetour.test.ts
```diff
@@ -80,14 +77,19 @@ describe("useDetour", () => {
...
     expect(jest.mocked(fetchDetourDirections)).toHaveBeenNthCalledWith(1, [
       start,
       end,
     ])
 
-    await waitFor(() =>
-      expect(result.current.detourShape).toStrictEqual(apiResult)
-    )
+    await waitFor(() => {
+      expect(result.current.detourShape).toStrictEqual([start, end])
```
joshlarson, line 87:
```suggestion
      expect(result.current.detourShape).toStrictEqual(apiResult)
```

test/skate_web/controllers/detour_route_controller_test.exs
firestack: The library [ex_machina](https://hexdocs.pm/ex_machina/readme.html) which provides the `Skate.Factory.build` family of functions takes either a keyword list, or a map as input. I think behind the scenes it collapses them both into a map, but the syntax here is just the different ways that the `build` function _can_ be used, all are valid expressions (y'know, if the `[...]` were valid data).

The `[{"type", 1}]` is just required because it's different from `[type: 1]` and `["type": 1]` is not a valid expression, so instead, you make a tuple of the key name as a string, and the value.

assets/tests/hooks/useDetour.test.ts
firestack: PTALA [`78f43f6` (#2367)](https://github.com/mbta/skate/pull/2367/commits/78f43f6b4f09e0b3e327acd95aa61e0587079b6b) 🙏🏻

test/skate_web/controllers/detour_route_controller_test.exs
firestack: PTALA [`1abe9cc` (#2367)](https://github.com/mbta/skate/pull/2367/commits/1abe9cc486b7ee4df3bd457f91c7b5b4171f6440) 🙏🏻

assets/src/detour.ts
firestack: I opted to remove it, because we don't know the design requirements here yet.
My thinking is
- We _could_ format it on the **frontend**
	- Pro: gives us a lot of flexibility in the rendering part
		- like wrapping the instruction (`"left"`, `"right"`, `"sharp right"`) and the name in `<strong>` or `<em>`
			- ex: 
			  > **Turn right** onto _Boylston St_
	- Con: Potentially need to calculate the instruction twice, once on frontend, and once on backend, to ensure data integrity when writing to the DB
- We _could_ format it on the **backend**
	- Pro: all the data is there, less stuff to send to frontend
	- Con: but then the frontend can't fancy rendering stuff as easily,

joshlarson (approved):

PR #2368, opened 2024-01-18
feat(gh/actions/chromatic): require label for automatic chromatic runs

firestack:
[Failing Dependabot PR](https://github.com/mbta/skate/pull/2324)'s are due to the `chromatic` job having issues with secrets permissions. This forces chromatic to only run if the `chromatic` label is present on a PR, essentially opting in to running chromatic.

Additionally, to support building arbitrary commits without a PR, this enables `workflow_dispatch` so that chromatic can be triggered manually.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,6 +1,9 @@
 name: "Chromatic"
 
-on: push
+on:
+  workflow_dispatch:
+  push:
+
 
 jobs:
   chromatic:
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -2,12 +2,14 @@ name: "Chromatic"
 
 on:
   workflow_dispatch:
-  push:
+  pull_request:
 
 
 jobs:
   chromatic:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'chromatic') }}
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4


.github/workflows/chromatic.yml
```diff
@@ -1,10 +1,15 @@
 name: "Chromatic"
 
-on: push
+on:
+  workflow_dispatch:
```
joshlarson, line 4: Do we also want this to run on merges to `main`?

firestack: Oh yes, thank you!

firestack: 90f08eea5d073462592ee65daba2ec606d875daf

joshlarson: 🥳

.github/workflows/chromatic.yml
```diff
@@ -1,10 +1,18 @@
...
+  pull_request:
+  push:
+    branches:
+      - main
+
 
 jobs:
   chromatic:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'chromatic') }}
```
joshlarson, line 14:
```suggestion
    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'chromatic') }}
```

I'm guessing otherwise it still won't actually run on merge to `main`.

firestack: Oh, yep, thank you!

I'm also wondering why it didn't trigger the chromatic run when I added the tag, because it seems like when the deploy tags are added, it runs them immediately, but chromatic didn't run till the next push. 

Maybe it has something to do with it being a deploy to an environment?

.github/workflows/chromatic.yml
```diff
@@ -1,10 +1,18 @@
 name: "Chromatic"
 
-on: push
+on:
+  workflow_dispatch:
+  pull_request:
```
joshlarson, line 5:
```suggestion
  pull_request:
    types:
    - synchronize
    - labeled
```

I think this will fix [this issue](https://github.com/mbta/skate/pull/2368/files#r1457935969)

.github/workflows/chromatic.yml
joshlarson: I was wondering the same thing!

I think [this](https://github.com/mbta/skate/pull/2368/files#r1457945443) is the fix. Otherwise the workflow doesn't interpret labeling as an event worthy of rerunning (or considering whether to rerun).

firestack: That seems to have fixed it, Thanks!

joshlarson (approved):

PR #2370, opened 2024-01-22
Initial fixes to tab order

lemald:
Asana ticket: [⚙️🚧 Investigate/implement tabbing order](https://app.asana.com/0/1148853526253420/1206260856814692/f)

This provides some very basic fixes to the high-level navigation of Skate by tabbing, in particular on the route ladders page. For the most part this is accomplished through improvement to the layout of the DOM rather than explicitly setting `tabindex`. This uncovers a lot of additional issues like missing focus state, but it's at least a start.

### DIFF
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -73,7 +73,7 @@ export const DetourMap = ({
   undoDisabled,
   onUndoLastWaypoint,
 }: DetourMapProps) => (
-  <Map vehicles={[]}>
+  <Map vehicles={[]} allowStreetView>
     <CustomControl position="topleft" className="leaflet-bar">
       <Button
         variant="primary"
--- a/assets/src/components/ladderPage.tsx
+++ b/assets/src/components/ladderPage.tsx
@@ -79,7 +79,8 @@ const LadderTab = ({
         }
       }}
       role="tab"
-      tabIndex={tab.isCurrentTab ? 0 : -1}
+      tabIndex={tab.isCurrentTab ? -1 : 0}
+      aria-selected={tab.isCurrentTab}
     >
       <div className="c-ladder-page__tab-contents">
         <div
--- a/assets/src/components/map/controls/StreetViewSwitch.tsx
+++ b/assets/src/components/map/controls/StreetViewSwitch.tsx
@@ -56,6 +56,7 @@ export const StreetViewControl = ({
           "c-street-view-switch",
           "position-absolute",
         ])}
+        insertFirst
       >
         <StreetViewSwitch
           streetViewEnabled={streetViewEnabled}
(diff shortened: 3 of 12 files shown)


joshlarson (approved):
It's sort of hard for me to wrap my mind around what the scope of this is. As an incremental improvement, this seems like a good change (and good to merge anyway to avoid annoying conflicts in case anyone goes in and fusses with `nav.tsx` or something), but the nav order stuff still has a long way to go.

It's kind of wild to me that like half of our nav links just "randomly" get skipped in tab order (I think there's something funky going on with `<NavLink />`, since all of the links that are skipped are `<NavLink />`'s).

firestack (approved):

lemald:
> It's kind of wild to me that like half of our nav links just "randomly" get skipped in tab order (I think there's something funky going on with `<NavLink />`, since all of the links that are skipped are `<NavLink />`'s).

What browser / OS combo are you on? On Macs, Firefox at least has some weird default behavior with tabbing and links, see [this Stackoverflow post](https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox).

joshlarson:
😭 I'm on MacOS / Firefox 😭

firestack:
> 😭 I'm on MacOS / Firefox 😭

@joshlarson do you have the setting enabled?
Under `Settings.app > Keyboard > Keyboard navigation (boolean)`

joshlarson:
@firestack Just enabled it and that does fix the problem.

joshlarson:
I wonder if we ought to document anywhere that that's a thing you might need to do, or if enough of our users are Windows/!Firefox users that we can assume it's not a problem.

lemald:
> I wonder if we ought to document anywhere that that's a thing you might need to do, or if enough of our users are Windows/!Firefox users that we can assume it's not a problem.

Maybe not this specific issue because I doubt we have any desktop Mac users outside of our team, but it does make me wonder if (say) Edge on Windows or other browser / OS combos we don't tend to use might have similar quirks we aren't aware of. 🤔

lemald:
@firestack @joshlarson I'd be curious to hear what you all think about the last few commits to move the street view control around in the DOM to put it first in tab order among the controls, as design asked. The use of portals makes it really tricky and the order that it shows up in the JSX doesn't automatically translate to where it ends up in the DOM. I could further pull the prop out and make it an option for `<StreetViewControl/>` itself, but that felt like overkill for the time being.

lemald:
Oh also I could have probably put more work into making the `insertAfterSelector` and `insertFirst` props mutually exclusive, but my one attempt at it got kind of annoying. Open to suggestions on how I might approach that.

lemald:
@firestack I'd be very curious to hear more about your ideas about how to redo `<CustomControl>`! Can you create Asana tickets for us to discuss later?

firestack:
> @firestack I'd be very curious to hear more about your ideas about how to redo `<CustomControl>`! Can you create Asana tickets for us to discuss later?

👍🏻 https://app.asana.com/0/1148853526253420/1206436289420248/f
Let me know if there's anything I can clarify or expand on.

PR #2373, opened 2024-01-29
feat: Add dropdown when right-clicking on a vehicle marker

joshlarson:
---
  - https://app.asana.com/0/0/1206038961362412

### DIFF
--- a/assets/src/components/mapMarkers.tsx
+++ b/assets/src/components/mapMarkers.tsx
@@ -1,7 +1,7 @@
-import Leaflet, { LatLngExpression, PointTuple } from "leaflet"
+import Leaflet, { LatLngExpression } from "leaflet"
 import "leaflet-defaulticon-compatibility" // see https://github.com/Leaflet/Leaflet/issues/4968#issuecomment-483402699
 import "leaflet.fullscreen"
-import React, { useContext, useRef } from "react"
+import React, { PropsWithChildren, useContext, useRef } from "react"
 import { Marker, Polyline, Popup, Tooltip } from "react-leaflet"
 
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
@@ -28,9 +28,6 @@ import { streetViewUrl } from "../util/streetViewUrl"
 import { TileTypeContext } from "../contexts/tileTypeContext"
 import { ReactMarker } from "./map/utilities/reactMarker"
 import { fullStoryEvent } from "../helpers/fullStory"
-import { DropdownItem, DropdownMenu } from "./map/dropdown"
-import inTestGroup, { TestGroups } from "../userInTestGroup"
-import useScreenSize from "../hooks/useScreenSize"
 
 /*  eslint-enable @typescript-eslint/ban-ts-comment */
 
@@ -99,20 +96,22 @@ const makeLabelIcon = (
   })
 }
 
-export const VehicleMarker = ({
-  vehicle,
-  isPrimary,
-  onSelect,
-  isSelected = false,
-}: {
+interface VehicleMarkerProps extends PropsWithChildren {
   vehicle: Vehicle
   isPrimary: boolean
   isSelected?: boolean
   onSelect?: (vehicle: Vehicle) => void
-}) => {
+}
+
+export const VehicleMarker = ({
+  children,
+  vehicle,
+  isPrimary,
+  onSelect,
+  isSelected = false,
+}: VehicleMarkerProps) => {
   const [{ userSettings }] = useContext(StateDispatchContext)
   const markerRef = useRef<Leaflet.Marker<any>>(null)
-  const screenSize = useScreenSize()
 
   const suppressPopup = () => {
     markerRef.current?.closePopup()
@@ -147,17 +146,6 @@ export const VehicleMarker = ({
   // > [specify] a high value like 1000 [...]
   const zIndexOffset = isSelected ? 1000 : 0
 
-  // This offset is here because, due to a limitation of Leaflet
-  // popups, we weren't able to render the popup at the bottom-right
-  // corner of the marker, where it's supposed to go. This effectively
-  // renders it centered and above the marker, and then uses the
-  // offset to reposition it to the bottom-right corner.
-  const dropdownOffset: PointTuple = [140, 97]
-
-  const dropdownEnabled =
-    inTestGroup(TestGroups.DetoursPilot) &&
-    ["desktop", "tablet"].includes(screenSize)
-
   return (
     <>
       <Marker
@@ -167,15 +155,7 @@ export const VehicleMarker = ({
         zIndexOffset={zIndexOffset}
         ref={markerRef}
       >
-        {dropdownEnabled && (
-          <Popup className="c-dropdown-popup-wrapper" offset={dropdownOffset}>
-            <DropdownMenu>
-              <DropdownItem>
-                Start a detour on route {vehicle.routeId}
-              </DropdownItem>
-            </DropdownMenu>
-          </Popup>
-        )}
+        {children}
       </Marker>
 
       <Marker
--- a/assets/src/components/mapMarkers.tsx
+++ b/assets/src/components/mapMarkers.tsx
@@ -1,7 +1,7 @@
-import Leaflet, { LatLngExpression } from "leaflet"
+import Leaflet, { LatLngExpression, PointTuple } from "leaflet"
 import "leaflet-defaulticon-compatibility" // see https://github.com/Leaflet/Leaflet/issues/4968#issuecomment-483402699
 import "leaflet.fullscreen"
-import React, { useContext } from "react"
+import React, { useContext, useRef } from "react"
 import { Marker, Polyline, Popup, Tooltip } from "react-leaflet"
 
 import { StateDispatchContext } from "../contexts/stateDispatchContext"
@@ -28,6 +28,7 @@ import { streetViewUrl } from "../util/streetViewUrl"
 import { TileTypeContext } from "../contexts/tileTypeContext"
 import { ReactMarker } from "./map/utilities/reactMarker"
 import { fullStoryEvent } from "../helpers/fullStory"
+import { Dropdown } from "react-bootstrap"
 
 /*  eslint-enable @typescript-eslint/ban-ts-comment */
 
@@ -108,7 +109,21 @@ export const VehicleMarker = ({
   onSelect?: (vehicle: Vehicle) => void
 }) => {
   const [{ userSettings }] = useContext(StateDispatchContext)
-  const eventHandlers = onSelect ? { click: () => onSelect(vehicle) } : {}
+  const markerRef = useRef<Leaflet.Marker<any>>(null)
+
+  const suppressPopup = () => {
+    markerRef.current?.closePopup()
+  }
+
+  const eventHandlers = {
+    click: () => {
+      onSelect && onSelect(vehicle)
+      suppressPopup()
+    },
+    contextmenu: () => {
+      markerRef.current?.openPopup()
+    },
+  }
   const position: LatLngExpression = [vehicle.latitude, vehicle.longitude]
   const vehicleIcon: Leaflet.DivIcon = makeVehicleIcon(
     vehicle,
@@ -128,6 +143,14 @@ export const VehicleMarker = ({
   // > [...] if you want to put the marker on top of all others,
   // > [specify] a high value like 1000 [...]
   const zIndexOffset = isSelected ? 1000 : 0
+
+  // This offset is here because, due to a limitation of Leaflet
+  // popups, we weren't able to render the popup at the bottom-right
+  // corner of the marker, where it's supposed to go. This effectively
+  // renders it centered and above the marker, and then uses the
+  // offset to reposition it to the bottom-right corner.
+  const dropdownOffset: PointTuple = [140, 97]
+
   return (
     <>
       <Marker
@@ -135,7 +158,17 @@ export const VehicleMarker = ({
         icon={vehicleIcon}
         eventHandlers={eventHandlers}
         zIndexOffset={zIndexOffset}
-      />
+        ref={markerRef}
+      >
+        <Popup className="c-dropdown-popup-wrapper" offset={dropdownOffset}>
+          <Dropdown.Menu className="c-dropdown-popup-menu" show>
+            <Dropdown.Item>
+              Start a detour on route {vehicle.routeId}
+            </Dropdown.Item>
+          </Dropdown.Menu>
+        </Popup>
+      </Marker>
+
       <Marker
         position={position}
         icon={labelIcon}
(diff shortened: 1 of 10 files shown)


assets/css/dropdown.scss
lemald, line 1: **nitpick:** this file should be called `_dropdown.scss` to indicate that it's a [partial](https://sass-lang.com/documentation/at-rules/import/#partials).

assets/src/components/dropdown.tsx
lemald, line 1: **suggestion:** since a lot of what's special about this `Dropdown` component is related to making it work in the Leaflet context, would it make sense to put this somewhere under the `map/` directory? Or do you think that enough about this is applicable to other contexts?

assets/src/components/mapMarkers.tsx
```diff
@@ -108,7 +110,27 @@ export const VehicleMarker = ({
...
 }) => {
   const [{ userSettings }] = useContext(StateDispatchContext)
-  const eventHandlers = onSelect ? { click: () => onSelect(vehicle) } : {}
+  const markerRef = useRef<Leaflet.Marker<any>>(null)
+
+  const suppressPopup = () => {
+    const marker = markerRef.current
+    if (marker) {
+      marker.closePopup()
+    }
```
lemald, line 119: **suggestion (non-blocking):** up to you if you feel that this is more or less readable, but it looks like TypeScript is smart enough to realize that this will work and it's somewhat more concise.

```suggestion
    markerRef.current?.closePopup()
```

assets/src/components/mapMarkers.tsx
```diff
@@ -108,7 +110,27 @@ export const VehicleMarker = ({
...
+  const eventHandlers = {
+    click: () => {
+      onSelect && onSelect(vehicle)
+      suppressPopup()
+    },
+    contextmenu: () => {
+      const marker = markerRef.current
+      if (marker) {
+        marker.openPopup()
+      }
```
lemald, line 131: **suggestion:** similar to the above, you could do:
```suggestion
      markerRef.current?.openPopup()
```

assets/src/components/mapMarkers.tsx
```diff
@@ -135,7 +157,19 @@ export const VehicleMarker = ({
         icon={vehicleIcon}
         eventHandlers={eventHandlers}
         zIndexOffset={zIndexOffset}
-      />
+        ref={markerRef}
+      >
+        {inTestGroup(TestGroups.DetoursPilot) && (
+          <Popup className="c-dropdown-popup-wrapper" offset={[140, 97]}>
```
lemald, line 163: **suggestion:** I think per some of our discussion having the hardcoded offset is somewhat inevitable, but if it's possible I think a quick comment explaining how this particular value was derived would be helpful, especially if we go mess with the styling in the future and the current offset no longer looks good.

assets/src/components/mapMarkers.tsx
```diff
@@ -135,7 +157,19 @@ export const VehicleMarker = ({
...
+            <DropdownMenu>
+              <DropdownItem>
+                Start a detour on route {vehicle.routeId}
+              </DropdownItem>
+            </DropdownMenu>
+          </Popup>
+        )}
```
firestack, line 170: question(non-blocking): do you think we could do this via composistion so it's opt in?
```
<VehicleMarker>
  <DropdownMenu.Popup>
    <DropdownItem>Item</DropdownItem>
  </DropdownMenu.Popup>
</VehicleMarker>
```

I was trying to find a way to do it without having to run a bunch of "wires" but didn't want to spend much more time on it.
(Example of what "wires" I was hoping to avoid. Essentially, it'd be great if `ref` didn't need to be interacted with at all on the top level)

```
<VehicleMarker ref={markerRef} eventHandlers={{click: () => markerRef.current?.closePopup(), contextmenu: () => markerRef.current?.openPopup()}}>
  <DropdownMenu.Popup>
    <DropdownItem>Item</DropdownItem>
  </DropdownMenu.Popup>
</VehicleMarker>
```

assets/css/dropdown.scss
joshlarson: https://github.com/mbta/skate/pull/2373/commits/e9c52d4d212c152354640e4bcd0e33bdd0314847

assets/src/components/dropdown.tsx
joshlarson: I had sort of intended this component to be usable inside or outside of a map - it renders properly (finally O_o) in Storybook, for instance.

I could go either way, TBH.

assets/src/components/mapMarkers.tsx
joshlarson: https://github.com/mbta/skate/pull/2373/commits/3c35d54b355f58d41f2797009ce6d1cdf2cfc685

assets/src/components/mapMarkers.tsx
joshlarson: https://github.com/mbta/skate/pull/2373/commits/3c35d54b355f58d41f2797009ce6d1cdf2cfc685

assets/src/components/dropdown.tsx
firestack: > I had sort of intended this component to be usable inside or outside of a map

I don't see anything special about this component that makes it useful outside the map though? The styling on it seems to only be needed to fight leaflet, and otherwise, I think we'd want to prefer using `Dropdown.Menu` when working outside the map, because more options are exposed than with `<DropdownMenu>`?

`<DropdownMenu/>` could be reconfigured to use `props: ComponentProps<typeof Dropdown.Menu>` which would then allow defaults to be reconfigured, but otherwise, I think we'd prefer the react bootstrap component, unless I'm missing something.

assets/src/components/mapMarkers.tsx
joshlarson: https://github.com/mbta/skate/pull/2373/commits/6e682ea8fa4b97d1b319f6b60e6378cd2f8b6628

Happy to keep iterating on the wording in ☝️ comment - this is just a first draft.

assets/src/components/mapMarkers.tsx
joshlarson: Hmmmm - I'll play around with this and see if I can come up with anything. I agree that running a bunch of `ref` wires through the interface of `VehicleMarker` will be annoying.

assets/src/components/mapMarkers.tsx
lemald: I think that's fine, but it would also be nice if we could say something more specific like "this was derived from taking the rendered size of the `.c-foo` element and adding `X` pixels of padding to it" or whatever you did. Or maybe I'm imagining something more systematic when it's more a result of experimenting until it looked good.

assets/src/components/mapMarkers.tsx
```diff
@@ -128,14 +144,34 @@ export const VehicleMarker = ({
...
         zIndexOffset={zIndexOffset}
-      />
+        ref={markerRef}
+      >
+        {inTestGroup(TestGroups.DetoursPilot) && (
+          <Popup className="c-dropdown-popup-wrapper" offset={dropdownOffset}>
+            <DropdownMenu>
+              <DropdownItem>
+                Start a detour on route {vehicle.routeId}
+              </DropdownItem>
```
firestack, line 169: question: have you thought about how we're going to get the callback that this option is selected? 
Also, _I think_, that once an option is selected, we'd need to close the popup, do you see that happening in another ticket?

joshlarson: I envisioned having it close itself as part of the modal ticket, but .... actually maybe it would make more sense to put as part of this ticket, since it really is a behavior of the menu, not of the modal.

assets/src/components/mapMarkers.tsx
joshlarson: I think we can avoid the `ref` wires by keeping `suppressPopup()` where it is and just having the `Popup` itself be passed into `<VehicleMarker />` as a `children`, like you suggested. 

But....

If every vehicle marker should get the popup, then I think I'd rather keep it like this. There are five instances of `<VehicleMarker />` throughout the code, and I'm worried about duplicating the `<DropdownMenu />` call in five places. 

Basically, is the popup an inherent feature of a `<VehicleMarker />`, or is it a feature only of specific `<VehicleMarker />`'s? I _think_ it's the former, but that's more of a design/product question.

assets/src/components/dropdown.tsx
joshlarson: I thought we wanted to use bootstrap components in large part because we wanted to use the same component in-map and outside-of-map whenever possible? Like... if we update something about the way this is styled, then it 

And... when/if we ever do decide to use `<DropdownMenu />` outside of a map, wouldn't we just do the thing that you suggested, and configure `<DropdownMenu />` to use `props: ComponentProps<typeof Dropdown.Menu>`?

I feel like I'm missing something here, but also I stand by what I said earlier that I'm fine either way, so I'll move it.

assets/src/components/mapMarkers.tsx
joshlarson: I experimented until it looked good 😅

assets/src/components/dropdown.tsx
joshlarson: https://github.com/mbta/skate/pull/2373/commits/e3432bfda0e943750ff18e9ebb25876d4a56ed3c
https://github.com/mbta/skate/pull/2373/commits/ad3470a0657add48a43f01b95024ef9f764e358d
https://github.com/mbta/skate/pull/2373/commits/9e2924b6e62d08e611a37f47591bc668c3b3dd8f

assets/src/components/mapMarkers.tsx
firestack: > If every vehicle marker should get the popup, then I think I'd rather keep it like this. There are five instances of `<VehicleMarker />` throughout the code, and I'm worried about duplicating the `<DropdownMenu />` call in five places.

I don't think every `<VehicleMarker />` is supposed to have it though, AFAIK from the designs, we're only supposed to have it on the Map Page. So we should double check with Mariele on this.

> Basically, is the popup an inherent feature of a `<VehicleMarker />`, or is it a feature only of specific `<VehicleMarker />`'s? I _think_ it's the former, but that's more of a design/product question.

I still think it's better practice to compose these rather than build them together, because if we need to pull them apart in the future, it may get harder if we add more logic in here, but if we build with composition from the start we're offered a lot more flexibility as well as less to test per component and less each component is individually responsible for. If we end up wanting to use them together in more than two locations, building a component which composes them would be my choice. 
I'm less stubborn on this if we do want this _everywhere_ we use vehicle markers, but that is essentially every map right now, including the minimap, which I'm relatively certain the menu is not supposed to be there.

But this is non-blocking, as we _can_ pull these apart later if needed.

assets/src/components/mapMarkers.tsx
lemald: Heh okay totally valid! In that case I think this comment is fine.

assets/src/components/dropdown.tsx
lemald: LGTM - @firestack since you also chimed in in this thread I'll give you a chance to take a look but if you agree it's good then go ahead and resolve this conversation.

firestack: TLDR: no blocking comments, good to resolve this thread

---

> I thought we wanted to use bootstrap components in large part because we wanted to use the same component in-map and outside-of-map whenever possible? Like... if we update something about the way this is styled, then it

Agreed, and I don't want to have to make a special component to make it work on the map, but it seems like that's the best option for now.

> [!NOTE]
> it'd be ideal if using a dropdown in leaflet was "simply".
> ```
> <Popup>
> 	<Dropdown.Menu>
> 		<Dropdown.Item>Item</Dropdown.Item>
>	</Dropdown.Menu>
> </Popup>
> ```
> I've written an example of a way we could write the style to allow this below.

If we were to use the following style, then we wouldn't need a special class or component to make `dropdown-menu` work on the leaflet map.
```
.leaflet-popup:has(.dropdown-menu) {
  visibility: hidden;

  .dropdown-menu {
    position: static;
    visibility: visible;
  }
}
```

But I'm not sure that we can use the [`:has` selector](https://caniuse.com/css-has)

assets/tests/components/map.test.tsx
```diff
@@ -293,6 +298,90 @@ describe("<MapFollowingPrimaryVehicles />", () => {
     expect(onClick).not.toHaveBeenCalled()
   })
 
+  test("starts out with no dropdown menu", async () => {
+    jest.mocked(getTestGroups).mockReturnValue([TestGroups.DetoursPilot])
```
joshlarson, line 302: These tests assert on the wrong use case. Gotta move them

assets/src/components/dropdown.tsx
joshlarson: I mean, I could just inline the `Dropdown.Menu` and `Dropdown.Item`. I extracted them as sort of a future-proofing exercise, and so that it would be a _guarantee_ that what's in storybook is the same as what's actually rendered. 

But I don't feel strongly about that. I'll think about which approach makes more sense to me - maybe I'll make a PR off of this PR just to see what we all think

assets/src/components/mapMarkers.tsx
joshlarson: @firestack These two commits are the approach I landed on - what do you think?
https://github.com/mbta/skate/pull/2373/commits/164175e24b59701c0ef0de19b8ee6853c3fa0714 <-- refactor
https://github.com/mbta/skate/pull/2373/commits/01b162362108627fcb5c2b9033da82533edd0be2 <-- ~1-liner to actually alert the on-click behavior

firestack: I'm good with this!

assets/tests/components/map.test.tsx
```diff
@@ -293,6 +298,90 @@ describe("<MapFollowingPrimaryVehicles />", () => {
...
+    )
+
+    await userEvent.pointer({
+      keys: "[MouseRight>]",
+      target: screen.getByText(runIdToLabel(vehicle.runId!)),
+    })
+
+    expect(
+      container.querySelector(".c-dropdown-popup-wrapper")
+    ).not.toBeInTheDocument()
```
firestack, line 359: suggestion/nit(non-blocking): would using `getByRole` and `queryByRole` instead of `container.querySelector` work for these tests?

When I was looking at the accessibility tree for this, the dropdown shows up as a button, so maybe this would work?
```
expect(screen.getByRole("button", { name: "Start a detour on route"})).toBeVisible()
```
and
```
expect(screen.queryByRole("button", { name: "Start a detour on route"})).not.toBeInTheDocument()
```

joshlarson: Yeah good call - I like that a lot better!

https://github.com/mbta/skate/pull/2373/commits/f64e7ebda4a250a47913dcb9f9fca76e82f13450

assets/tests/components/map.test.tsx
joshlarson: So moved: https://github.com/mbta/skate/pull/2373/commits/f3f5131d808a44465915323514641d675aa4809a

lemald (approved):
Looks good on my end!

firestack (approved):

PR #2376, opened 2024-02-05
feat: add missed stops algorithim

firestack:
Asana Ticket: https://app.asana.com/0/0/1206536428201924


lib/skate/detours/missed_stops.ex
```diff
@@ -0,0 +1,128 @@
...
+
+      []
+    else
+      # We don't want the stop associated with the segment that the detour rejoins.
+      # So subtract one from the end index to ignore that segment
+      Enum.slice(segmented_shape, start_index..(end_index - 1))
+    end
+  end
+
+  defp segment_shape_by_stops([] = _shape, [] = _stops),
```
lemald, line 56: **suggestion:** it would probably be more for documentation than for actual type-safety, but I think a typespec for this function might help make it slightly more readable.

lemald: I think the same probably goes for a couple of the other private functions in this module.

test/skate/detours/missed_stops_test.exs
```diff
@@ -0,0 +1,99 @@
+defmodule Skate.Detours.MissedStopsTest do
+  use ExUnit.Case
+
+  alias Util.Location
+
+  describe "missed_stops" do
+    test "given a straight line, should return missed stops" do
+      param = %Skate.Detours.MissedStops{
+        connection_start: Location.new(-0.001, 1),
```
lemald, line 9: **suggestion (non-blocking):** I wonder if little ASCII art diagrams for these different test cases would make them easier to read - obviously they wouldn't be to scale, but I wonder if you could at least capture the relative positions of points and stops to give a sense of what's going on.

lib/skate/detours/shape_segment.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Skate.Detours.ShapeSegment do
+  @moduledoc """
+  Represents the shape points which form a segment associated
+  with stops on the shape.
+  """
+
+  @type t :: %__MODULE__{
+          points: [Util.Location.From.t()],
+          stops: [Util.Location.From.t()]
```
joshlarson, line 9: If I remember correctly, `ShapeSegment` will never have more than one stop, right? If so, could we make that more explicit in the type here? Something like:

```suggestion
          stops: {:some, Util.Location.From.t()} | {:none}
```

Or even

```suggestion
          stop: Util.Location.From.t() | nil
```

joshlarson: That would necessarily mean changing the `stops` invocations to use singular stops instead of arrays, but I think that would be an improvement, since there's never more than one stop.

firestack: > If I remember correctly, `ShapeSegment` will never have more than one stop, right? If so, could we make that more explicit in the type here?

_currently_ `ShapeSegment` will not have more than one stop, but, this is something that I 100% believe could happen as we improve the algorithm. 

I personally would keep it as is, but I'm maybe trying to over-fit for the (unknown) future.

test/skate/detours/missed_stops_test.exs
firestack: Thoughts?
```ex
      ##                                                 (Missed Stop)
      ##        Stops: v(0.001, 0)                        v(0.001, 5)   v(0.001, 5)
      ##               o                                  o             o
      ##
      ##               o------o------o------o------o------o------o------o
      ## Shape Points: ^(0,0) ^(0,1) ^(0,2) ^(0,3) ^(0,4) ^(0,5) ^(0,6) ^(0,7)
      ##                                                        
      ## Connection Points :  o                                  o
      ##                      ^(-0.001, 1)                       ^(-0.00, 6)
      ##                   (connection_start)                  (connection_end)
```

joshlarson: I don't care what anyone else says - this is art.

(Nit: I think `connection_end` in the diagram should be `-0.001` rather than `-0.00`)

test/skate/detours/missed_stops_test.exs
```diff
@@ -0,0 +1,99 @@
...
+          Location.new(0, 0),
+          Location.new(1, 0),
+          Location.new(1, 1),
+          Location.new(0, 1),
+          Location.new(0, 0),
+          Location.new(0, -1)
+        ]
+      }
+
+      assert Enum.slice(param.stops, 1..3) ==
```
joshlarson, line 81: Nit: Could we update this assertion to make it clearer what the assertion actually says? At a glance, I can't tell if this is saying that we miss `duplicate_stop` both times, or if we miss one instance of `duplicate` and two other stops.

test/skate/detours/missed_stops_test.exs
firestack: I can continue making diagrams like this, but it was pretty tricky and annoying to make this (let alone maintain it??) and took me longer than I'd have liked.

And it's only gonna get more difficult with the "cross" test and others.

Thoughts on maybe using Excalidraw? (or miro)
Example: https://excalidraw.com/#json=OrMM928mw4CR3Qy8sc6oL,sXiFCU1s-K1ugEQvgSvh3g

I'm not happy with the idea of having this diagram hosted by an external service, so while a link would probably be the easiest option if we used Excalidraw, we could instead save the diagrams as SVG, PNG, or Export the `.excalidraw` (json) document and version it into Skate.

firestack: Updated visual https://excalidraw.com/#json=91bSifAkQwmz3uy4x8k02,Eu4K8LXTkYOtPcBchBjkUg

lemald: Yeah, if you think it's too tricky don't worry about it. I figured it would be a neat idea to try but I don't think it's worth sinking too much time into.

firestack: I do like the idea of adding a diagram though, do y'all have issues with using Excalidraw/preferences on how(link, svg, png, `.excalidraw`) it's used?

firestack: I'm personally okay with using links for now, and I think I'd otherwise prefer SVG to make it easy to view, or `.excalidraw` for easy to edit.

lib/skate/detours/missed_stops.ex
```diff
@@ -0,0 +1,128 @@
...
+      points = sort_by_distance(points, reference)
+      # Pick closest points within fuzzy_dist=(0)
+      # Discard adjacent points
+      # Pick highest rank or return :indeterminate
+      %{segment: segment, sorted_points: points, index: idx}
+    end)
+    |> Enum.min_by(fn %{sorted_points: [%{dist: dist} | _]} -> dist end)
+  end
+
+  defp sort_by_distance(coordinates, reference) do
```
joshlarson, line 112: A bunch of these calls to `sort_by` (maybe all of them?) are only being used to then retrieve the first item - that is, the minimum, per some criterion. I think it would be simpler to understand and computationally faster to do `min_by` instead, and replace even more of these arrays with single items.

joshlarson: I have an approach for doing this kind of refactor that I generally quite like, and I would be happy to pair on it.

firestack: Something to keep in mind, I've been working on a few "bugs" in the algorithim, and have quickly needed to use the rest of the values in these arrays, so I think we're going to run right back into this if we switch to `min_by`

firestack: See uhh, this mess 😅

```
      [closest | next] = points

      points =
        ([closest] ++
           Enum.filter(next, fn %{dist: dist_from_reference} ->
             dist_from_reference - closest.dist < min_diff
           end))
        |> Enum.sort_by(& &1.index, :asc)
        |> Enum.chunk_while(
          [],
          fn
            element, [] ->
              {:cont, [element]}

            %{index: e_idx} = element, [%{index: acc_idx} | _] = acc ->
              case e_idx - acc_idx do
                1 -> {:cont, [element | acc]}
                0 -> throw(:duplicate_index)
                _ -> {:cont, acc, [element]}
              end
          end,
          fn acc -> {:cont, acc, []} end
        )
        |> Enum.map(&Enum.min_by(&1, fn %{dist: dist} -> dist end))

```

joshlarson: Oh jeez!

Yeah okay maybe let's revisit once the 🐛 🐞 are squared away!

lib/skate/detours/missed_stops.ex
firestack: PTALA 572fa5784ae87befbfc81532f905843a93e7ca25 🙏🏻

lemald: Awesome, thanks!

test/skate/detours/missed_stops_test.exs
lemald: My concern with having an external link or something like an SVG in the repo is that it would also be hard to maintain.

test/skate/detours/missed_stops_test.exs
firestack: Does 52ba7cf836e15989ace924df5bca97802980d714 work?

lemald (approved):
Per some discussion with @firestack earlier I want to hereby register that I'm good with this PR in its current state, while also acknowledging that there are some outstanding comments from @joshlarson but I don't want to block my own review on those.

lib/skate/detours/missed_stops.ex
firestack: a78eec61114f040b1e27dc69522db7933c0dc219

lib/skate/detours/shape_segment.ex
firestack: @joshlarson thoughts? 2b5f726494f0dbc21a1cccd105059734daba6cf5

lib/skate/detours/missed_stops.ex
```diff
@@ -0,0 +1,177 @@
...
+          []
+
+        # If there's more than one stop in this segment,
+        # Then return all the stops
+        %Skate.Detours.ShapeSegment{
+          stop: stop
+        } ->
+          [stop]
+      end
+
```
joshlarson, line 67:
```suggestion
```

If I'm reading this correctly, the results of this `case` statement aren't being used, since it's followed immediately by just returning `[]` (which makes sense - if the connection end-point is before the collection of segments we're currently looking at, then there shouldn't be any missed stops!)

firestack: 61b816a880a0b694b6db68acd79f4771aabb91ee

lib/skate/detours/missed_stops.ex
```diff
@@ -0,0 +1,177 @@
...
+    shape_segments
+    |> Enum.with_index(fn %Skate.Detours.ShapeSegment{points: points} = segment, idx ->
+      points = closest_point(points, reference)
+      # Pick closest points within fuzzy_dist=(0)
+      # Discard adjacent points
+      # Pick highest rank or return :indeterminate
+      %{segment: segment, sorted_points: points, index: idx}
+    end)
+    |> Enum.min_by(fn %{sorted_points: %{dist: dist}} -> dist end)
+  end
```
joshlarson, line 130:
```suggestion
  @spec get_index_by_min_dist(
          shape_segments :: [Skate.Detours.ShapeSegment.t()],
          reference :: Util.Location.From.t()
        ) ::
          %{
            segment: Skate.Detours.ShapeSegment.t(),
            closest_point: point_dist_index(),
            index: non_neg_integer()
          }
  defp get_index_by_min_dist(shape_segments, reference) do
    shape_segments
    |> Enum.with_index(fn %Skate.Detours.ShapeSegment{points: points} = segment, idx ->
      closest_point = closest_point(points, reference)
      # Pick closest points within fuzzy_dist=(0)
      # Discard adjacent points
      # Pick highest rank or return :indeterminate
      %{segment: segment, closest_point: closest_point, index: idx}
    end)
    |> Enum.min_by(fn %{sorted_points: %{dist: dist}} -> dist end)
  end
```

Nit: Both the names `points` and `sorted_points` are a bit confusing now that each refers to a singular point.

(Also I know that Elixir lets you re-bind variable names to new data, but I prefer to avoid it when possible, since then it gets hard to tell which definition is the most recent)

I'm also wondering why Dialyxir didn't complain about the type of `sorted_points` being wrong. (I think it was wrong?)

lib/skate/detours/missed_stops.ex
joshlarson: Oh nice - that's even simpler

lib/skate/detours/missed_stops.ex
firestack: 0146770708ef7a8b020c0b62544d8bd38dc25fad

joshlarson (approved):

PR #2377, opened 2024-02-06
feat: Add a detour-editing modal when a user selects that option on the dropdown menu

joshlarson:
---
  - https://app.asana.com/0/0/1206260856814696

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -57,11 +57,11 @@
 
 .c-diversion-panel__origin,
 .c-diversion-panel__direction {
-  font-size: 14px;
+  font-size: var(--font-size-s);
 }
 
 .c-diversion-panel__description {
-  font-weight: 600;
+  font-weight: var(--font-weight-semi);
 }
 
 .l-diversion-page__header {
@@ -71,15 +71,15 @@
 }
 
 .c-diversion-panel__h2 {
-  font-size: 24px;
-  font-weight: 600;
+  font-size: var(--font-size-xl);
+  font-weight: var(--font-weight-semi);
 }
 
 .c-diversion-panel__h3 {
-  font-size: 20px;
-  font-weight: 600;
+  font-size: var(--font-size-l);
+  font-weight: var(--font-weight-semi);
 }
 
 .c-diversion-panel__help-text {
-  line-height: 24px;
+  line-height: 1.5;
 }
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -68,7 +68,6 @@
   display: flex;
   justify-content: flex-end;
   align-items: center;
-  padding: 16px;
 }
 
 .c-diversion-panel__h2 {
@@ -79,4 +78,8 @@
 .c-diversion-panel__h3 {
   font-size: 20px;
   font-weight: 600;
+}
+
+.c-diversion-panel__help-text {
+  line-height: 24px;
 }
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -63,3 +63,10 @@
 .c-diversion-panel__description {
   font-weight: 600;
 }
+
+.l-diversion-page__header {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  padding: 16px;
+}
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -54,3 +54,12 @@
     }
   }
 }
+
+.c-diversion-panel__origin,
+.c-diversion-panel__direction {
+  font-size: 14px;
+}
+
+.c-diversion-panel__description {
+  font-weight: 600;
+}
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -70,3 +70,13 @@
   align-items: center;
   padding: 16px;
 }
+
+.c-diversion-panel__h2 {
+  font-size: 24px;
+  font-weight: 600;
+}
+
+.c-diversion-panel__h3 {
+  font-size: 20px;
+  font-weight: 600;
+}
--- a/assets/css/_modal.scss
+++ b/assets/css/_modal.scss
@@ -14,11 +14,3 @@
   @include modal-backdrop;
   z-index: 2000;
 }
-
-.c-modal-fullscreen {
-  z-index: 2001;
-}
-
-.c-modal-fullscreen-body {
-  padding: 0px;
-}
--- a/assets/css/_modal.scss
+++ b/assets/css/_modal.scss
@@ -14,3 +14,7 @@
   @include modal-backdrop;
   z-index: 2000;
 }
+
+.c-modal-fullscreen {
+  z-index: 2001;
+}
--- a/assets/css/_modal.scss
+++ b/assets/css/_modal.scss
@@ -18,3 +18,7 @@
 .c-modal-fullscreen {
   z-index: 2001;
 }
+
+.c-modal-fullscreen-body {
+  padding: 0px;
+}
--- a/assets/src/components/detours/detourDropdown.tsx
+++ b/assets/src/components/detours/detourDropdown.tsx
@@ -1,7 +1,8 @@
 import React from "react"
-import { DropdownItem } from "../map/dropdown"
-import { Route, RouteId, RoutePattern, Shape } from "../../schedule"
-import { Vehicle } from "../../realtime"
+import { DropdownItem, DropdownMenu } from "../map/dropdown"
+import { Route, RoutePattern, Shape } from "../../schedule"
+import { Popup } from "react-leaflet"
+import { PointTuple } from "leaflet"
 
 export interface StartDetourProps {
   routeName: string
@@ -11,48 +12,62 @@ export interface StartDetourProps {
   shape: Shape
 }
 
-export interface DetourDropdownItemProps {
+export interface DetourDropdownProps {
   setShouldShowPopup: (newValue: boolean) => void
   routePatternForVehicle: RoutePattern | null
-  routeId: RouteId | undefined
   route: Route | null
   onStartDetour?: (props: StartDetourProps) => void
-  selectedVehicleOrGhost: Vehicle
+  onClick: (props: StartDetourProps) => void
 }
 
-export const DetourDropdownItem = ({
-  setShouldShowPopup,
+export const DetourDropdown = ({
   routePatternForVehicle,
-  routeId,
   route,
-  onStartDetour,
-  selectedVehicleOrGhost,
-}: DetourDropdownItemProps) => {
+  onClick,
+}: DetourDropdownProps) => {
+  // This offset is here because, due to a limitation of Leaflet
+  // popups, we weren't able to render the popup at the bottom-right
+  // corner of the marker, where it's supposed to go. This effectively
+  // renders it centered and above the marker, and then uses the
+  // offset to reposition it to the bottom-right corner.
+  const dropdownOffset: PointTuple = [140, 97]
+
+  if (!routePatternForVehicle || !route) {
+    return <></>
+  }
+
+  const routeId = routePatternForVehicle.routeId
+
+  const routeDescription = routePatternForVehicle.headsign
+
+  const routeOrigin = routePatternForVehicle.name
+
+  const routeDirection =
+    route.directionNames[routePatternForVehicle.directionId]
+
+  const shape = routePatternForVehicle.shape
+
+  if (!routeDescription || !shape) {
+    return <></>
+  }
+
   return (
-    <DropdownItem
-      onClick={() => {
-        setShouldShowPopup(false)
-
-        const directionName =
-          routePatternForVehicle?.directionId != undefined &&
-          route?.directionNames[routePatternForVehicle?.directionId]
-
-        onStartDetour &&
-          routeId &&
-          routePatternForVehicle?.shape &&
-          routePatternForVehicle.headsign &&
-          routePatternForVehicle.name &&
-          directionName &&
-          onStartDetour({
-            routeName: routeId,
-            routeDescription: routePatternForVehicle.headsign,
-            routeOrigin: routePatternForVehicle.name,
-            routeDirection: directionName,
-            shape: routePatternForVehicle.shape,
-          })
-      }}
-    >
-      Start a detour on route {selectedVehicleOrGhost.routeId}
-    </DropdownItem>
+    <Popup className="c-dropdown-popup-wrapper" offset={dropdownOffset}>
+      <DropdownMenu>
+        <DropdownItem
+          onClick={() => {
+            onClick({
+              routeName: routeId,
+              routeDescription,
+              routeOrigin,
+              routeDirection,
+              shape,
+            })
+          }}
+        >
+          Start a detour on route {routeId}
+        </DropdownItem>
+      </DropdownMenu>
+    </Popup>
   )
 }
--- a/assets/src/components/detours/detourDropdown.tsx
+++ b/assets/src/components/detours/detourDropdown.tsx
@@ -1,0 +1,58 @@
+import React from "react"
+import { DropdownItem } from "../map/dropdown"
+import { Route, RouteId, RoutePattern, Shape } from "../../schedule"
+import { Vehicle } from "../../realtime"
+
+export interface StartDetourProps {
+  routeName: string
+  routeDescription: string
+  routeOrigin: string
+  routeDirection: string
+  shape: Shape
+}
+
+export interface DetourDropdownItemProps {
+  setShouldShowPopup: (newValue: boolean) => void
+  routePatternForVehicle: RoutePattern | null
+  routeId: RouteId | undefined
+  route: Route | null
+  onStartDetour?: (props: StartDetourProps) => void
+  selectedVehicleOrGhost: Vehicle
+}
+
+export const DetourDropdownItem = ({
+  setShouldShowPopup,
+  routePatternForVehicle,
+  routeId,
+  route,
+  onStartDetour,
+  selectedVehicleOrGhost,
+}: DetourDropdownItemProps) => {
+  return (
+    <DropdownItem
+      onClick={() => {
+        setShouldShowPopup(false)
+
+        const directionName =
+          routePatternForVehicle?.directionId != undefined &&
+          route?.directionNames[routePatternForVehicle?.directionId]
+
+        onStartDetour &&
+          routeId &&
+          routePatternForVehicle?.shape &&
+          routePatternForVehicle.headsign &&
+          routePatternForVehicle.name &&
+          directionName &&
+          onStartDetour({
+            routeName: routeId,
+            routeDescription: routePatternForVehicle.headsign,
+            routeOrigin: routePatternForVehicle.name,
+            routeDirection: directionName,
+            shape: routePatternForVehicle.shape,
+          })
+      }}
+    >
+      Start a detour on route {selectedVehicleOrGhost.routeId}
+    </DropdownItem>
+  )
+}
--- a/assets/src/components/detours/detourDropdown.tsx
+++ b/assets/src/components/detours/detourDropdown.tsx
@@ -40,7 +40,7 @@ export const DetourDropdown = ({
   const dropdownOffset: PointTuple = [140, 97]
 
   if (!routePatternForVehicle || !route) {
-    return <></>
+    return null
   }
 
   const routeName = route.name
@@ -55,7 +55,7 @@ export const DetourDropdown = ({
   const shape = routePatternForVehicle.shape
 
   if (!routeDescription || !shape) {
-    return <></>
+    return null
   }
 
   return (
--- a/assets/src/components/detours/detourDropdown.tsx
+++ b/assets/src/components/detours/detourDropdown.tsx
@@ -3,6 +3,7 @@ import { DropdownItem, DropdownMenu } from "../map/dropdown"
 import { Route, RoutePattern, Shape } from "../../schedule"
 import { Popup } from "react-leaflet"
 import { PointTuple } from "leaflet"
+import { LatLngLiteral } from "leaflet"
 
 export interface StartDetourProps {
   routeName: string
@@ -10,6 +11,8 @@ export interface StartDetourProps {
   routeOrigin: string
   routeDirection: string
   shape: Shape
+  center: LatLngLiteral
+  zoom: number
 }
 
 export interface DetourDropdownProps {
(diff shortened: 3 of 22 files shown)


assets/css/_modal.scss
```diff
@@ -14,3 +14,11 @@
   @include modal-backdrop;
   z-index: 2000;
 }
+
+.c-modal-fullscreen {
+  z-index: 2001;
```
lemald, line 19: **suggestion:** if you want, a quick and easy refactor could be to move these `z-index` values in this file into the `$z-page-layout-context` map in `app.scss`.

assets/src/components/detours/detourDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return <></>
+  }
+
+  return (
+    <Popup className="c-dropdown-popup-wrapper" offset={dropdownOffset}>
+      <DropdownMenu>
+        <DropdownItem
+          onClick={() => {
+            onClick({
+              routeName: routeId,
```
lemald, line 67: **issue:** be careful not to mix up route name and ID - they're mostly the same, but there are examples where they aren't (for instance the SL1 has route ID 741).

lemald: Thankfully there is an application-wide context providing all of the route objects, so you can use that if you need to do the mapping.

joshlarson: Ahhhhh I was wondering about that.

assets/src/components/detours/detourDropdown.tsx
```diff
@@ -0,0 +1,82 @@
+import React from "react"
+import { DropdownItem, DropdownMenu } from "../map/dropdown"
+import { Route, RoutePattern, Shape } from "../../schedule"
+import { Popup } from "react-leaflet"
+import { PointTuple } from "leaflet"
+import { LatLngLiteral } from "leaflet"
+
+export interface StartDetourProps {
```
joshlarson, line 8: @mbta/skate-developers I feel like these types have become kind of a mess. I feel like `StartDetourProps` (or something like it) should exist (because otherwise we'll end up prop-drilling like a million separate params), but I'm not sure it should live here.

assets/src/components/detours/diversionPage.tsx
```diff
@@ -12,7 +13,15 @@ export const DiversionPage = ({
   routeDirection,
   routeOrigin,
   shape,
-}: DiversionPanelProps & { shape: Shape }) => {
+  onClose,
+  center,
+  zoom,
+}: DiversionPanelProps & {
```
joshlarson, line 19: I think this needs a separate type. There's some overlap between `DiversionPanelProps` and ... effectively `DiversionPageProps`, but less and less - for instance, `directions` are part of `DiversionPanelProps`, but `DiversionPage` (correctly) ignores them in favor of the directions it gets from `useDetour()`.

assets/src/components/detours/detourModal.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react"
+import { DiversionPage } from "./diversionPage"
+import { StartDetourProps } from "./detourDropdown"
+
+export const DetourModal = ({
```
joshlarson, line 5: This is cool and all, but I wonder if it can be combined with `DiversionPage`...

But maybe not, because this does take care of the modal logic itself, which... actually might be worth keeping separate.

Thoughts?

assets/src/components/detours/detourDropdown.tsx
firestack: I think you should store the route pattern id, and rederive the route name, description, origin, direction, and shape in the `detourModal`.

assets/src/components/detours/detourMap.tsx
```diff
@@ -59,6 +59,12 @@ interface DetourMapProps {
    * Callback fired when the undo button is clicked.
    */
   onUndoLastWaypoint: () => void
+
+  /*
+   * Center and zoom to position the map correctly when it renders
+   */
+  center: LatLngLiteral
```
joshlarson, line 66: I was thinking about this, and it actually does feel right to put this here, since it does govern behavior of the detour map when it renders. But I wanted to raise it anyway, since I put this here more out of convenience than anything else.

assets/src/components/detours/detourDropdown.tsx
joshlarson: 👍 Is there a hook already that will let me exchange a route pattern ID for all of that info?

If not, feels like a good idea to make that hook, huh

assets/src/components/detours/detourDropdown.tsx
joshlarson: How do I actually get the route name out of the routePattern?

When I console.logged `routePatternForVehicle` for an SL2 bus, it didn't have any fields whose value was `"SL2"`

lemald: It's the route, not the route pattern, that has the name. You'll need to key off of `routePatternForVehicle.routeId` to look up the appropriate route.

assets/src/components/detours/detourDropdown.tsx
joshlarson: Oooorrrr... I could pass the route pattern itself as a prop?

assets/src/components/detours/detourDropdown.tsx
joshlarson: Ooh actually this was a pretty easy fix, since I had access to the `route` in this context as well: https://github.com/mbta/skate/pull/2377/commits/386996b161b5d6f55b2211cb2af70050a2e66786

assets/src/components/detours/detourModal.tsx
```diff
@@ -0,0 +1,28 @@
...
+  detourInfo,
+  onClose,
+  show,
+}: {
+  detourInfo: StartDetourProps
+  onClose: () => void
+  show: boolean
+}) => {
+  return (
```
joshlarson, line 14: There is a bug that I introduced here, and I'm trying to think about the best way to resolve it.

---

### Steps to Reproduce
 - Open the search maps page.
 - Find a bus, right-click, open the detour modal, and drop a start point.
 - Close the detour modal.
 - Without refreshing, change the zoom level and/or pan the map.
 - Re-open the detour modal from the bus.

### Expected Behavior
 - The start point is gone
 - The map opens to zoom level and center that the search map screen currently shows

### Actual Behavior
 - The start point is still there
 - The map opens to the same zoom level and center that the previous detour modal was previously showing

---

The offending commit is this one, https://github.com/mbta/skate/pull/2377/commits/2dacf205c273e4cd15ef72a30cbdfbb3abe01a6e, which adds a CSS transition to make the modal animate in and out, rather than just _appearing_ (which is pretty jarring, and sort of makes it look like clicking on the dropdown menu didn't do anything).

I'm pretty sure that the root cause is that before ☝️ commit, the modal was being added or removed from the DOM, but afterwards, it's just being given an opacity of 0 and hidden behind everything. That is... in the before-times, React saw the detour modal leave the DOM, and then when it was time to create a new one, it was just a completely brand new one, but in the after-times, as far as React is concerned, it's the same object, so it keeps all of the same state.

I'm thinking that the answer is a `key=` on one of these objects - probably either `DetourModal` or `DiversionPage`. Maybe `key`ing on the timestamp that the user clicked on the dropdown? Or even a random nonce generated when the user clicks the dropdown? I think either of these would work, since they would persist across a user's journey through a single detour modal, and then be different if they right-click again.

Interested to hear what @mbta/skate-developers think about this.

assets/css/_diversion_page.scss
```diff
@@ -54,3 +54,32 @@
...
+  font-weight: 600;
+}
+
+.c-diversion-panel__h3 {
+  font-size: 20px;
+  font-weight: 600;
+}
+
+.c-diversion-panel__help-text {
+  line-height: 24px;
```
firestack, line 84: question: can unitless values be used instead here?
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values

assets/src/components/detours/detourDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const routeOrigin = routePatternForVehicle.name
+
+  const routeDirection =
+    route.directionNames[routePatternForVehicle.directionId]
+
+  const shape = routePatternForVehicle.shape
+
+  if (!routeDescription || !shape) {
+    return <></>
+  }
```
firestack, line 59: suggestion: I believe you can return null here instead of constructing a document fragment
```suggestion
  if (!routePatternForVehicle || !route) {
    return null
  }

  const routeName = route.name

  const routeDescription = routePatternForVehicle.headsign

  const routeOrigin = routePatternForVehicle.name

  const routeDirection =
    route.directionNames[routePatternForVehicle.directionId]

  const shape = routePatternForVehicle.shape

  if (!routeDescription || !shape) {
    return null
  }
```

assets/src/components/detours/detourModal.tsx
firestack: I'd prefer to compose them instead of collapsing them.

assets/src/components/detours/detourModal.tsx
firestack: `key` is probably the easiest method, and I don't think you'd need to make it very complex, 
```
const [modalKey, setModalKey] = useState(0)

"onModalOpen": () => setModalKey(current => current + 1)
```

assets/css/_diversion_page.scss
joshlarson: Thanks for catching that! Did that and a few other CSS things here - https://github.com/mbta/skate/pull/2377/commits/b4f29bbd9c77cf9b409366e0cf73870128eeaff9

assets/src/components/detours/detourModal.tsx
joshlarson: That makes a lot of sense - giving it a try now...

joshlarson: [It worked!](https://github.com/mbta/skate/pull/2377/commits/4b3d067a7b5eaabca2ca6e9d716cf62da23c4d36)

assets/src/components/detours/detourModal.tsx
joshlarson: @firestack are you thinking something like 👇?

```typescript
<DetourModal show={showDetourModal} onClose={() => {doTheRightThing()}}>
  <DiversionPage {...allTheProps} />
</DetourModal>
```

firestack: I think what I was trying to convey is that, similarly to how the Mobile Nav Menu is somewhat annoying to use (and display in storybook) because it handles both displaying itself on the screen, positioning, AND being a nav, I like that `DetourModal` and `(Detour/Diversion)Page` are separate and that the page doesn't concern itself with creating it's parent container and positioning it (though I'll admit it's breaking some encapsulation boundaries by having the close button be in the Page and not the modal, but going that direction is maybe too "pure")

The example you show is certainly one way to achieve that, but you could also let the modal do the work of figuring out `allTheProps` and "hardcode" the `DiversionPage` as the `DetourModal`'s only child for now. so the example would be more like
```
<DetourModal
	show={showDetourModal}
	onClose={() => {doTheRightThing()}}
	routePatternId={...}
	{any other id's needed to derive information for the `DiversionPage`}
/>
```

I'm assuming that "hard-coding" it would be simpler/easier for now.

But if the number of props that `DetourModal` needs to prop-drill, without modification, 
becomes large enough that we're thinking of putting them in a struct to move it, then I think it'd be better to do via composition, like your [example above](https://github.com/mbta/skate/pull/2377#discussion_r1490110281).

assets/src/components/detours/detourDropdown.tsx
firestack: there is `usePatternsByIdForRoute` but there isn't anything that directly goes from `routePatternId` to `RoutePattern` it seems. I agree that having that would be a great addition.

firestack: I think it's fine if the `routeId` is needed in addition to the `routePatternId` to satisfy `usePatternsByIdForRoute`

assets/src/components/detours/detourModal.tsx
joshlarson: I think I also prefer the idea of composition, and using `grid` and `subgrid` (like `DiversionPage` and `DiversionPanel` do), I think it might be possible to get the close button to be fully a part of the `DetourModal` without the `DiversionPage` knowing about it at all.

I would be interested in trying that out, although I might choose to do that as a follow-up instead.

firestack: Heh, using `grid` and `subgrid` to get the button to belong to the modal but appear on the diversion page was exactly what I was thinking. If it's easy, than yay! but otherwise it's _probably_ not worth our effort at this time?

firestack (commented):
Still discussing 
- https://github.com/mbta/skate/pull/2377#discussion_r1491153625
- https://github.com/mbta/skate/pull/2377#discussion_r1484542013

but otherwise LGTM

assets/src/components/detours/detourModal.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const DetourModal = ({
+  detourInfo,
+  onClose,
+  show,
+}: {
+  detourInfo: StartDetourProps
+  onClose: () => void
+  show: boolean
+}) => {
```
lemald, line 13: **nitpick (non-blocking):** since this function body is of the form

```
{
  return (
    < ...JSX Code... />
  )
}
```

you can rewrite it as a simple value without a `return` statement or curly braces.

assets/src/components/detours/detourModal.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div className={`c-detour-modal${show ? "" : " c-detour-modal__hidden"}`}>
```
lemald, line 15: **suggestion:** this is a great use case for our `joinClasses` helper function (you'll also need to import it)

```suggestion
    <div className={joinClasses(["c-detour-modal", show ? "c-detour-modal__hidden" : null])}>
```

assets/src/components/detours/detourModal.tsx
joshlarson: I'll prioritize this last so that the PR can get merged without nice composition, and then follow up with nice composition later on if I can figure it out.

(Or I'll sneak it into this PR if I _can_ figure it out reasonably quickly)

assets/src/components/detours/detourDropdown.tsx
joshlarson: Hmmmmm I'm becoming a little less sold on this for a few reasons:

1. I don't love the idea of re-deriving information that we already know.
2. `StartDetourProps` has all of its fields as not-null - if any of them are `null`, then the `DetourDropdown` simply doesn't appear. I'd rather not do a bunch of null-checking on fields that we already know in-context are not null.

I'm leaning a bit more towards having `StartDetourProps` move into `detour.ts` and be like... `OriginalRoutePattern` or something like that, while keeping its existing structure, and then instead of passing around each of the fields separately, passing them together.

firestack: > I don't love the idea of re-deriving information that we already know.

The problem I have is assuming that we're storing information that'll never update, whereas pulling this information from hooks via ID's allows us to centralize the way we access this data (allowing us to do more refactors and caching that we've had queued up)

Maybe this data shouldn't change, but it's already an established pattern within `state.ts` that we store Id's, not objects. There are a few exceptions to this, but I'd prefer to get rid of those as well.

---

All in all though, I'm fine with this being non-blocking and you continuing with what you think is best, as we can refactor later if it's a problem.

joshlarson: Kind of a big diff, but not a huge change as it turns out.

[This](https://github.com/mbta/skate/pull/2377/commits/9d600e4108d7ac49cdd64c8147da8939529d14e7) is what I landed on.

- Renamed `StartDetourProps` to `OriginalRoute`.
- Moved `OriginalRoute` into `detour.ts` because it's shared around and is a pretty core part of the detour-editing world.
- Wherever _all_ of the fields of `OriginalRoute` were being passed around, replaced the collection of fields with just one field of type `OriginalRoute` (the fields are still passed "by hand" into `DetourMap` and `DiversionPanel`, but that's a problem for another day)

joshlarson: (And then [this](https://github.com/mbta/skate/pull/2377/commits/4e701387fb08ffd236589e02f124a3e9ac5cdaae) to fix the tests and other compilation issues)

assets/src/detour.ts
```diff
@@ -1,8 +1,19 @@
...
   }[]
 }
+
+export interface OriginalRoute {
+  routeName: string
+  routeDescription: string
+  routeOrigin: string
+  routeDirection: string
+  shape: Shape
+  center: LatLngLiteral
```
joshlarson, line 17: This feels like a case of making wrong code look wrong - in this case the fact that there are leaflet coordinates and schedule `Shape` objects mixed together here.

This was an issue for `StartDetourProps` (where this came from) as well, but it feels more glaring here, since this is lifted up a layer and feels a lot further away from the map itself.

I don't actually want to fix this now, because I think we should decide as a team what the right approach is for dealing with latitude/longitude pairs, and I don't want that discussion to block merging this PR.

assets/src/components/detours/detourModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  show: boolean
+}) => {
+  useEffect(() => {
+    const handleEscape = (event: KeyboardEvent) => {
+      if (event.key === "Escape") {
+        onClose()
+      }
+    }
+
+    window.addEventListener("keydown", handleEscape)
```
lemald, line 22: **question:** do we necessarily want this event listener on `window`? Does it work all right if we just attach it to the top-level `<div>` of the modal?

assets/src/components/mapPage.tsx
```diff
@@ -356,9 +369,20 @@ const MapPage = (): ReactElement<HTMLDivElement> => {
                   setFollowerShouldSetZoomLevel(false)
                 })
               }
+              onStartDetour={onStartDetour}
             />
           </MapSafeAreaContext.Provider>
         </div>
+        {originalRouteForDetour && (
+          <DetourModal
+            key={detourModalKey}
```
lemald, line 378: **question (non-blocking):** can you help me understand what necessitates the `detourModalKey`? If `originalRouteForDetour` gets reset to `null` in the `onClose` callback I would think you could avoid it.

assets/tests/components/mapPage.test.tsx
```diff
@@ -148,10 +150,13 @@ jest.mock("../../src/hooks/useSearchResultsByCategory", () => ({
...
 jest.mock("../../src/helpers/fullStory")
 
-jest.mock("../../src/userInTestGroup")
-
 jest.mock("../../src/hooks/usePanelState")
 
+jest.mock("../../src/userTestGroups", () => ({
+  __esModule: true,
+  default: jest.fn(() => []),
+}))
```
firestack, line 158: question/suggestion(non-blocking): does this work instead?
```suggestion
jest.mock("../../src/userTestGroups")
```

firestack (approved):
I agree with https://github.com/mbta/skate/pull/2377#discussion_r1492957001 but I'm fine with this being merged without.

assets/src/components/mapPage.tsx
joshlarson: It fixes this bug 👉 https://github.com/mbta/skate/pull/2377#discussion_r1484798109

TL;DR: Leaflet holds onto state even when a React mindset says that it shouldn't. When the `detourModalKey` changes, that convinces React that it's really a different modal, not the same one reappearing, which then lets React convince Leaflet of the same thing.

assets/src/components/detours/detourModal.tsx
firestack: I know that with having this event listener on `window` we don't need to manage focus because _something_ will hear it, but I agree that it'd be nicer to have this event in the actual tree of the component, and manage focus instead.

Fine with this being non-blocking though.

firestack: It is kinda weird that when I open the modal and press tab nothing happens because the focus is still on the map page (god I hate modals >.>)

assets/src/components/mapPage.tsx
lemald: Got it, thanks.

lemald (approved):

assets/tests/components/mapPage.test.tsx
joshlarson: It doesn't :(

firestack: Can you show or explain what happens when it doesn't work?
My assumption is that this would break in the same way if we did https://app.asana.com/0/1148853526253420/1203482629607208/f

So my hypothesis is that we could do [the above](https://github.com/mbta/skate/pull/2377#discussion_r1492977637), and then  add
```
beforeEach(() => {
	jest.mocked(inTestGroup).mockReturnValue([])
})
```

firestack: Alternatively, you can find if it breaks in the same way if you try
```
beforeEach(() => {
	jest.resetAllMocks()
})
```

assets/src/components/detours/detourModal.tsx
joshlarson: I created two follow-up tickets:

 - [⚙️ Have the detour modal grab focus when it renders](https://app.asana.com/0/1148853526253420/1206648035312400/f)
 - [⚙️ Move the escape-key handler onto the detour modal instead of the whole window](https://app.asana.com/0/1152340551558956/1206648035312404/f)

assets/tests/components/mapPage.test.tsx
joshlarson: Oof! Sorry for my brevity on Friday! I meant to fill in more details, and I guess I wound up focused on other things 🤦 

The first suggestion - getting rid of the `__esModule: true, default: ...` leads to any test that interacts with test groups failing with 👇 this error 👇 

```
 TypeError: Cannot read properties of undefined (reading 'includes')

   10 |
   11 | const inTestGroup = (key: TestGroups): boolean => {
 > 12 |   return getTestGroups().includes(key)
      |                         ^
   13 | }
   14 |
   15 | export default inTestGroup
```

I think that's because without the default behavior given in the `__esModule: true, default: ...` block, `getTestGroups()` is returning `undefined` instead of `[]`, and then `userInTestGroup()` is blowing up.

The second suggestion resolves that by, as a `beforeEach`, restoring that default behavior, so the second suggestion does work properly. (I won't pretend to know exactly how `jest.mock(...)` and `jest.mocked(...)` play with each other, so I'm honestly not sure, but I'm guessing that putting this in a `beforeEach` rather than an initializer that just runs once when the test file loads is a better choice.)

The third suggestion, `jest.resetAllMocks()`, causes all tests in this file to fail with 👇 this error 👇 
```
TypeError: Cannot read properties of undefined (reading 'openVehiclePropertiesPanel')

  221 |   const defaultSearchOpen = !selectedEntity || !isMobile
  222 |
> 223 |   const { openVehiclePropertiesPanel } = usePanelStateFromStateDispatchContext()
      |                                                                                 ^
  224 |
  225 |   // #region Search Drawer Logic
  226 |   const [searchOpen, setSearchOpen] = useState<boolean>(defaultSearchOpen)

```

At a quick glance, I'm guessing that these tests just rely on a few mocks that aren't set up to accommodate those tests. I'd rather punt that setup to the Asana ticket you linked.

PR #2379, opened 2024-02-08
Breakpoint simplification

lemald:
Asana ticket: [⚙️ Finish up layout/breakpoint changes](https://app.asana.com/0/1148853526253420/1206526459943482/f)

### DIFF
--- a/assets/src/components/nav.tsx
+++ b/assets/src/components/nav.tsx
@@ -1,11 +1,9 @@
-import React, { useContext } from "react"
-import { toggleMobileMenu } from "../state"
+import React from "react"
 import appData from "../appData"
 import useScreenSize from "../hooks/useScreenSize"
 import LeftNav from "./nav/leftNav"
 import TopNav from "./nav/topNav"
 import MobilePortraitNav from "./nav/mobilePortraitNav"
-import { StateDispatchContext } from "../contexts/stateDispatchContext"
 import { usePanelStateFromStateDispatchContext } from "../hooks/usePanelState"
 
 interface Props {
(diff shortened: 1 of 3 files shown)


assets/src/components/nav.tsx
```diff
@@ -26,46 +23,30 @@ const Nav: React.FC<Props> = ({ children }) => {
...
     default:
       return (
         <div className="l-nav--wide">
           <>
-            <div className="l-nav__nav-bar l-nav__nav-bar--top">
+            <div
+              className="l-nav__nav-bar l-nav__nav-bar--top"
+              hidden={
+                isViewOpen && deviceType === "mobile_landscape_tablet_portrait"
+              }
```
firestack, line 34: question(non-blocking): I pulled this down locally and it worked fine, but it didn't work for me in dev green, any ideas why?

firestack (approved):

assets/src/components/nav.tsx
lemald: Yeah I just ran into the same thing with design review, I'm pulling my hair out trying to figure it out.

lemald: Ah, the deploy action was marked as successful but the circuit breaker actually rolled it back. Gonna retry.

joshlarson: Wuttt... why would it do that?

lemald: I think it hit the ECS deploy circuit breaker that eventually gives up on failing deploys (it was having issues connecting to the DB). That's great and all but it isn't well-integrated into our deploy action so it basically sees that the deploy finished without detecting the particular error state it hit.

PR #2380, opened 2024-02-08
feat: add logout for keycloak

firestack:
Depends on: #2358 

References:
- https://github.com/mbta/arrow/pull/950/files#diff-94eca8259513c68e11b1b5d47d9c571dd7c0d8233a813ca95626507bffd65399R7-R75
- https://github.com/mbta/my_charlie/blob/main/lib/my_charlie_web/router.ex#L159-L163
- https://github.com/mbta/my_charlie/blob/main/lib/my_charlie_web/controllers/auth_controller.ex#L112-L154

---

Asana Ticket: https://app.asana.com/0/1148853526253420/1206509056963769/f

### DIFF
--- a/lib/skate_web/controllers/auth_controller.ex
+++ b/lib/skate_web/controllers/auth_controller.ex
@@ -82,4 +82,29 @@ defmodule SkateWeb.AuthController do
         nil
     end
   end
+
+  @spec logout(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def logout(conn, %{"provider" => "keycloak"}) do
+    case Guardian.Plug.current_claims(conn) do
+      %{"sign_out_url" => sign_out_url} when not is_nil(sign_out_url) ->
+        conn
+        |> session_cleanup()
+        |> redirect(external: sign_out_url)
+
+      # The router makes sure we can't call `/auth/:provider/keycloak`
+      # unless we have a session.
+      # So the potential `nil` from `current_claims` and the potential map with
+      # `sign_out_url=nil` can be handled the same
+      _ ->
+        conn
+        |> session_cleanup()
+        |> redirect(to: "/")
+    end
+  end
+
+  defp session_cleanup(conn) do
+    conn
+    |> SkateWeb.AuthManager.Plug.sign_out()
+    |> configure_session(drop: true)
+  end
 end
--- a/lib/skate_web/controllers/auth_controller.ex
+++ b/lib/skate_web/controllers/auth_controller.ex
@@ -74,7 +74,7 @@ defmodule SkateWeb.AuthController do
 
   # https://github.com/mbta/arrow/blob/372c279e04866509f1e287e07844d61cc243850b/lib/arrow_web/controllers/auth_controller.ex#L57-L66
   defp sign_out_url(auth) do
-    case UeberauthOidcc.initiate_logout_url(auth) do
+    case initiate_logout_url(auth) do
       {:ok, url} ->
         url
 
@@ -82,6 +82,9 @@ defmodule SkateWeb.AuthController do
         nil
     end
   end
+
+  defp initiate_logout_url(auth),
+    do: Application.get_env(:skate, :logout_url_fn, &UeberauthOidcc.initiate_logout_url/1).(auth)
 
   @spec logout(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def logout(conn, %{"provider" => "keycloak"}) do
--- a/lib/skate_web/controllers/auth_controller.ex
+++ b/lib/skate_web/controllers/auth_controller.ex
@@ -27,7 +27,7 @@ defmodule SkateWeb.AuthController do
       |> Guardian.Plug.sign_in(
         AuthManager,
         %{id: user_id},
-        %{groups: groups},
+        %{groups: groups, sign_out_url: sign_out_url(auth)},
         ttl: {expiration - current_time, :seconds}
       )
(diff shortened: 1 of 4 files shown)


lib/skate_web/controllers/auth_controller.ex
```diff
@@ -71,4 +71,45 @@ defmodule SkateWeb.AuthController do
   def callback(%{assigns: %{ueberauth_failure: %{provider: :keycloak}}} = conn, _params) do
     send_resp(conn, :unauthorized, "unauthenticated")
   end
+
+  # https://github.com/mbta/arrow/blob/372c279e04866509f1e287e07844d61cc243850b/lib/arrow_web/controllers/auth_controller.ex#L57-L66
+  defp sign_out_url(auth) do
+    case UeberauthOidcc.initiate_logout_url(
+           auth,
+           %{
+             #  post_logout_redirect_uri: "https://www.mbta.com/"
```
firestack, line 80: question: I didn't realize I left this in here commented out, I pulled this from https://github.com/mbta/arrow/pull/950/files#diff-94eca8259513c68e11b1b5d47d9c571dd7c0d8233a813ca95626507bffd65399R58-R60 but, any ideas on where we should redirect to after?

firestack: I don't really know what the user flow looks like with this, so I'm testing in [`5b2d866` (#2380)](https://github.com/mbta/skate/pull/2380/commits/5b2d86647fc0d12af13a8f3d9d9d443f170f58a1)

lib/skate_web/controllers/auth_controller.ex
```diff
@@ -71,4 +71,45 @@ defmodule SkateWeb.AuthController do
...
+
+  @spec logout(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def logout(conn, %{"provider" => "keycloak"}) do
+    case Guardian.Plug.current_claims(conn) do
+      %{"sign_out_url" => sign_out_url} when not is_nil(sign_out_url) ->
+        conn
+        |> session_cleanup()
+        |> redirect(external: sign_out_url)
+
+      # The router makes sure we can't call `/auth/:provider/keycloak`
```
lemald, line 99: **nitpick:** I think there's a typo in the path, is that supposed to be `/auth/:provider/callback`?

firestack: Yep! thanks!
```suggestion
      # The router makes sure we can't call `/auth/:provider/callback`
```

test/support/factory.ex
```diff
@@ -335,4 +335,25 @@ defmodule Skate.Factory do
...
+        raw_info: %UeberauthOidcc.RawInfo{
+          userinfo: %{
+            "resource_access" => %{
+              "test-client" => %{"roles" => ["test1", "skate-readonly"]}
+            }
+          }
+        }
+      }
+    }
+  end
```
firestack, line 358: I took the lazy route and just pulled one of the working auth structs from `auth_controller_test` but I'm happy to make this factory "better". I figured we could punt on it till later though

lib/skate_web/router.ex
```diff
@@ -53,6 +53,12 @@ defmodule SkateWeb.Router do
     get("/:provider/callback", AuthController, :callback)
   end
 
+  scope "/auth", SkateWeb do
+    pipe_through([:redirect_prod_http, :accepts_html, :browser, :auth])
```
firestack, line 57: I'm not actually sure if `:ensure_auth` is needed or not?

firestack (commented):
posting just a few questions/notes

test/skate_web/controllers/auth_controller_test.exs
```diff
@@ -103,6 +106,49 @@ defmodule SkateWeb.AuthControllerTest do
...
+        conn
+        |> init_test_session(%{})
+        |> assign(:ueberauth_auth, Skate.Factory.build(:ueberauth_auth))
+        |> get(~p"/auth/keycloak/callback")
+
+      assert Guardian.Plug.authenticated?(conn)
+
+      conn = get(conn, ~p"/auth/keycloak/logout")
+
+      refute Guardian.Plug.authenticated?(conn)
```
lemald, line 127: **nitpick (non-blocking):** I guess we don't _strictly_ need these Guardian-related assertions in this test since other tests cover the Guardian functionality, but I suppose they're nice guardrails to make sure that the sign in / out flow works, so up to you on keeping them or not.

test/support/factory.ex
lemald: Yeah I'm fine with the "get it working and then improve it when we need to" approach on this one.

lib/skate_web/router.ex
lemald: I think we actually do want it because the other plugs used in `SkateWeb.AuthManager.Pipeline` don't do anything if there's no Guardian session information at all, whereas `:ensure_auth` uses `Guardian.Plug.EnsureAuthenticated` which does.

lemald:
Okay sorry one minor note that I do think we should actually err on the side of piping this route through `:ensure_auth`, but otherwise everything is looking great. 👍

lib/skate_web/router.ex
firestack: [`794a91e` (#2380)](https://github.com/mbta/skate/pull/2380/commits/794a91ee64780421905f77a81687b50dfbee8cf8)

lemald (approved):

PR #2383, opened 2024-02-10
feat: Add <LoggedInAs /> component

joshlarson:
Screenshot 2024-02-09 at 8 37 48 PM

---
  - https://app.asana.com/0/0/1206526459943484

### DIFF
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -1,9 +1,8 @@
 @use "./color/tokens_2024" as tokens;
 
 .c-logged-in-as__container {
-  display: flex;
   align-items: center;
-  gap: 0.5rem;
+  display: flex;
 }
 
 .c-logged-in-as__icon {
@@ -12,9 +11,10 @@
 }
 
 .c-logged-in-as__text {
-  font-weight: var(--font-weight-reg);
-  font-size: var(--font-size-s);
   color: tokens.$gray-700;
+  font-size: var(--font-size-s);
+  font-weight: var(--font-weight-reg);
+  margin-left: 0.5rem;
 
   p {
     line-height: 1.5;
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -1,0 +1,32 @@
+.c-logged-in-as__container {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
+.c-logged-in-as__icon {
+  background-color: #6d39ac;
+  color: #f8f8f9;
+  font-size: 20px;
+  font-weight: 600;
+  height: 32px;
+  width: 32px;
+  border-radius: 16px;
+  text-align: center;
+  line-height: 32px;
+}
+
+.c-logged-in-as__text {
+  font-weight: 400;
+  font-size: 14px;
+  color: #6e7175;
+
+  p {
+    line-height: 21px;
+    margin: 0px;
+  }
+}
+
+.c-logged-in-as__email {
+  font-weight: 600;
+}
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -5,15 +5,8 @@
 }
 
 .c-logged-in-as__icon {
-  background-color: #6d39ac;
-  color: #f8f8f9;
-  font-size: 20px;
-  font-weight: 600;
   height: 32px;
   width: 32px;
-  border-radius: 16px;
-  text-align: center;
-  line-height: 32px;
 }
 
 .c-logged-in-as__text {
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -1,12 +1,12 @@
 .c-logged-in-as__container {
   display: flex;
   align-items: center;
-  gap: 8px;
+  gap: 0.5rem;
 }
 
 .c-logged-in-as__icon {
-  height: 32px;
-  width: 32px;
+  height: 2rem;
+  width: 2rem;
 }
 
 .c-logged-in-as__text {
@@ -16,7 +16,7 @@
 
   p {
     line-height: 1.5;
-    margin: 0px;
+    margin: 0;
   }
 }
 
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -1,3 +1,5 @@
+@use "./color/tokens_2024" as tokens;
+
 .c-logged-in-as__container {
   display: flex;
   align-items: center;
@@ -12,7 +14,7 @@
 .c-logged-in-as__text {
   font-weight: var(--font-weight-reg);
   font-size: var(--font-size-s);
-  color: #6e7175;
+  color: tokens.$gray-700;
 
   p {
     line-height: 1.5;
--- a/assets/css/_logged_in_as.scss
+++ b/assets/css/_logged_in_as.scss
@@ -10,16 +10,16 @@
 }
 
 .c-logged-in-as__text {
-  font-weight: 400;
-  font-size: 14px;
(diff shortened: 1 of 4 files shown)


assets/src/components/loggedInAs.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+
+export const LoggedInAs = ({ email }: { email: string }) => (
+  <div className="c-logged-in-as__container">
+    <div className="c-logged-in-as__icon">{email[0]?.toUpperCase() ?? ""}</div>
```
joshlarson, line 5: This hacks in the username icon; whenever that's ready, we should use that here instead.

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,32 @@
+.c-logged-in-as__container {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
+.c-logged-in-as__icon {
+  background-color: #6d39ac;
```
lemald, line 8: **question:** if I'm not mistaken this is the new eggplant-500 that we're not actually using yet, so I guess there's a question of which to use for now. I'm asking design in the context of my own ticket FWIW.

lemald: Oh heh true.

assets/src/components/loggedInAs.tsx
joshlarson: No more! https://github.com/mbta/skate/pull/2383/commits/8fb447e8adc48f1158038538c65f40796cb520eb

assets/css/_logged_in_as.scss
joshlarson: 👋 https://github.com/mbta/skate/pull/2383/commits/8fb447e8adc48f1158038538c65f40796cb520eb

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,25 @@
...
+
+.c-logged-in-as__icon {
+  height: 32px;
+  width: 32px;
+}
+
+.c-logged-in-as__text {
+  font-weight: 400;
+  font-size: 14px;
+  color: #6e7175;
```
joshlarson, line 15: As @lemald pointed out about a different color (that got `git`-ed away), this color may want to be replaced with `$color-gray-700`, although if we do that before [this PR](https://github.com/mbta/skate/pull/2365) is merged, then the color will be slightly wrong, since `$color-gray-700` is currently #3c3f4c

joshlarson:
```suggestion
  color: $color-gray-700;
```

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.c-logged-in-as__icon {
+  height: 32px;
+  width: 32px;
+}
+
+.c-logged-in-as__text {
+  font-weight: 400;
+  font-size: 14px;
```
lemald, line 14: **suggestion:** rather than hardcoding these you can use values defined in `_skate_ui.scss`.

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,25 @@
+.c-logged-in-as__container {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
+.c-logged-in-as__icon {
+  height: 32px;
+  width: 32px;
```
lemald, line 9: **suggestion:** (here and a few other places) we usually use `rem` for sizing, except within Leaflet maps because those already use `px` for a lot of the sizing.

lemald:
I forgot to add yesterday - I'm not approving yet because of some of my comments on the CSS, but aside from that I've looked over the code and it all looks good to me. 👍

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.c-logged-in-as__text {
+  font-weight: 400;
+  font-size: 14px;
+  color: #6e7175;
+
+  p {
+    line-height: 21px;
```
firestack, line 18: question: can unitless values be used instead here?
[developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values)

assets/css/_logged_in_as.scss
joshlarson: On further thought - the color is more than slightly wrong - the other links on this page are `gray-600`, not `gray-700`, so the text winds up looking pretty out of place. I think we should keep this as-is and follow up after the bootstrap colors PR is merged.

assets/css/_logged_in_as.scss
joshlarson: Ahh good. I was wondering about that, and planning to ask.

assets/css/_logged_in_as.scss
joshlarson: Screenshot 2024-02-13 at 10 51 43 AM

Nailed it

assets/css/_logged_in_as.scss
joshlarson: Ahhh cool - https://github.com/mbta/skate/pull/2383/commits/e40fec99962d5201d3358ef704403ebc094acb1c

assets/css/_logged_in_as.scss
joshlarson: Okay but here we go... doing it correctly! https://github.com/mbta/skate/pull/2383/commits/e40fec99962d5201d3358ef704403ebc094acb1c

firestack (approved):

assets/css/_logged_in_as.scss
lemald: Per Slack discussion: let's go with gray-700 for this PR since we're doing other work to integrate the new color palette changes.

joshlarson: If I'm synthesizing correctly from all of the other discussions, I'm going to wait for https://github.com/mbta/skate/pull/2365 to get merged, and then use the new_v2 (or 2024) gray-700.

joshlarson: Aaaand here we go: https://github.com/mbta/skate/pull/2383/commits/9c5ea5b2c3bc03fbf1381c07477966eb00e5421c

assets/css/_logged_in_as.scss
```diff
@@ -0,0 +1,27 @@
+@use "./color/tokens_2024" as tokens;
+
+.c-logged-in-as__container {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
```
lemald, line 6: **nitpick:** this is probably less of an issue two years on, but we have run into problems with `gap` in flexbox in the past. See #1541.

lemald (approved):
You may want to change how you accomplish the spacing in the flexbox to avoid the use of `gap` per my above comment, but it's been a couple of years since that was last an issue so I'll leave it up to your discretion.

assets/css/_logged_in_as.scss
firestack: We really should get caniuse in CI (again?)

joshlarson: Oof! Browser incompatibility strikes again!

Would you recommend switching to something like `margin-right: 0.5rem` on the icon?

joshlarson: Or (I think equivalently) `margin-left: 0.5rem` on the text?

joshlarson: https://github.com/mbta/skate/pull/2383/commits/04458340325ba72ec0daefe42ae7a89574684d13

PR #2388, opened 2024-02-12
CircleButton and UserAvatar components

lemald:
Asana ticket: [⚙️ Create component for email address initial icon](https://app.asana.com/0/1148853526253420/1206526457565579/f)

### DIFF
--- a/assets/css/_circle_button.scss
+++ b/assets/css/_circle_button.scss
@@ -1,0 +1,11 @@
+.c-circle-button {
+  width: 2rem;
+  height: 2rem;
+
+  background-color: $color-gray-50;
+  border-radius: 50%;
+}
+
+.c-circle-button__content {
+  margin: 0.25rem;
+}
--- a/assets/css/_circle_button.scss
+++ b/assets/css/_circle_button.scss
@@ -19,7 +19,7 @@
     }
   }
 
-  &--active {
+  &.c-circle-button--active {
     background-color: tokens.$gray-200;
   }
 }
--- a/assets/css/_circle_button.scss
+++ b/assets/css/_circle_button.scss
@@ -9,7 +9,7 @@
     background-color: $color-gray-100;
   }
 
-  &:focus-within {
+  &:focus-visible {
     border: 0.0625rem solid $color-eggplant-500;
 
     .c-circle-button__content {
--- a/assets/css/_circle_button.scss
+++ b/assets/css/_circle_button.scss
@@ -4,6 +4,22 @@
 
   background-color: $color-gray-50;
   border-radius: 50%;
+
+  &:hover {
+    background-color: $color-gray-100;
+  }
+
+  &:focus-within {
+    border: 0.0625rem solid $color-eggplant-500;
+
+    .c-circle-button__content {
+      margin: calc(0.25rem - 0.0625rem);
+    }
+  }
+
+  &--active {
+    background-color: $color-gray-200;
+  }
 }
 
 .c-circle-button__content {
--- a/assets/css/_circle_button.scss
+++ b/assets/css/_circle_button.scss
@@ -1,16 +1,18 @@
+@use "./color/tokens_2024" as tokens;
+
 .c-circle-button {
   width: 2rem;
   height: 2rem;
 
-  background-color: $color-gray-50;
+  background-color: tokens.$gray-50;
   border-radius: 50%;
 
   &:hover {
-    background-color: $color-gray-100;
+    background-color: tokens.$gray-100;
   }
 
   &:focus-visible {
-    border: 0.0625rem solid $color-eggplant-500;
+    border: 0.0625rem solid var(--bs-primary);
 
     .c-circle-button__content {
       margin: calc(0.25rem - 0.0625rem);
@@ -18,7 +20,7 @@
   }
 
   &--active {
-    background-color: $color-gray-200;
+    background-color: tokens.$gray-200;
   }
 }
 
--- a/assets/src/components/circleButton.tsx
+++ b/assets/src/components/circleButton.tsx
@@ -19,6 +19,7 @@ export const CircleButton = ({
     ])}
     onClick={onClick}
     title={title}
+    aria-pressed={isActive}
   >
     <div className="c-circle-button__content">{children}</div>
   </button>
--- a/assets/src/components/circleButton.tsx
+++ b/assets/src/components/circleButton.tsx
@@ -1,0 +1,13 @@
+import React, { ReactNode } from "react"
+
+export const CircleButton = ({
+  children,
+  onClick,
+}: {
+  children: ReactNode
+  onClick: () => void
+}) => (
+  <button className="c-circle-button" onClick={onClick}>
+    <div className="c-circle-button__content">{children}</div>
+  </button>
+)
--- a/assets/src/components/circleButton.tsx
+++ b/assets/src/components/circleButton.tsx
@@ -1,13 +1,22 @@
 import React, { ReactNode } from "react"
+import { joinClasses } from "../helpers/dom"
 
 export const CircleButton = ({
   children,
   onClick,
+  isActive,
 }: {
   children: ReactNode
   onClick: () => void
+  isActive: boolean
 }) => (
-  <button className="c-circle-button" onClick={onClick}>
+  <button
+    className={joinClasses([
+      "c-circle-button",
+      isActive && "c-circle-button--active",
+    ])}
+    onClick={onClick}
+  >
     <div className="c-circle-button__content">{children}</div>
   </button>
 )
--- a/assets/src/components/circleButton.tsx
+++ b/assets/src/components/circleButton.tsx
@@ -5,10 +5,12 @@ export const CircleButton = ({
   children,
   onClick,
   isActive,
+  title,
 }: {
-  children: ReactNode
-  onClick: () => void
+  children?: ReactNode
+  onClick?: () => void
   isActive: boolean
+  title?: string
 }) => (
   <button
     className={joinClasses([
@@ -16,6 +18,7 @@ export const CircleButton = ({
       isActive && "c-circle-button--active",
     ])}
     onClick={onClick}
+    title={title}
   >
     <div className="c-circle-button__content">{children}</div>
   </button>
--- a/assets/stories/skate-components/circleButton.stories.tsx
+++ b/assets/stories/skate-components/circleButton.stories.tsx
@@ -9,8 +9,30 @@ const meta = {
 } satisfies Meta<typeof CircleButton>
 
 export default meta
-type Story = StoryObj<typeof meta>
+type Story = StoryObj<typeof CircleButton>
 
 export const Default: Story = {
-  args: { children: <UserAvatar userName="foo" />, isActive: false },
+  args: { isActive: false },
+}
+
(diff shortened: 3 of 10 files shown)


assets/src/components/circleButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children,
+  onClick,
+  isActive,
+  title,
+}: {
+  children?: ReactNode
+  onClick?: () => void
+  isActive: boolean
+  title?: string
+}) => (
```
joshlarson, line 14: Should this take arguments to specify its size? Either `small`/`medium`/`large` or `height`/`width`? Or is there a way to do that that I missed?

lemald: From the Figma, there are numerous sizes of the user avatar, but only the one size of the circle button. After discussion with design I decided to make the avatar fill its container and make the button a fixed size.

assets/stories/skate-components/userAvatar.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <Story />
+      </div>
+    ),
+  ],
+} satisfies Meta<typeof UserAvatar>
+
+export default meta
+type Story = StoryObj<typeof UserAvatar>
+
+export const Default: Story = { args: { userName: "fake@test.com" } }
```
joshlarson, line 20: Nit: It might be helpful to nest these (along with the upcoming `<LoggedInAs />` component) into a subfolder, rather than having them all live top-level. I called it `identity`, but I'm not overly attached to that name.


Screenshot 2024-02-12 at 1 48 51 PM

Just thinking that the top level of ☝️ will get a bit cluttered.

lemald: I'd be open to moving UserAvatar, though CircleButton isn't necessarily tied to the user avatar (and I think design envisions using it in other places in the future).

assets/stories/skate-components/userAvatar.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import React from "react"
+
+import { UserAvatar } from "../../src/components/userAvatar"
+
+const meta = {
+  component: UserAvatar,
```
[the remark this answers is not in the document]
lemald: I'm sort of confused - you should be able to just type in an email in the UserAvatar Storybook entry. Are you thinking of CircleIcon? I don't really want to put stuff about the inner workings of the UserAvatar component in the stories for CircleIcon.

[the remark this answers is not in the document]
joshlarson: Yes, I'm sorry - I mixed up which was which between `CircleButton` and `UserAvatar`, and put my comments in the wrong place. 

I'll migrate over to `circleAvatar.stories.tsx` to continue this :)

assets/stories/skate-components/circleButton.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const meta = {
+  component: CircleButton,
+} satisfies Meta<typeof CircleButton>
+
+export default meta
+type Story = StoryObj<typeof CircleButton>
+
+export const Default: Story = {
+  args: { isActive: false },
```
joshlarson, line 15: I originally wrote my comment (on the wrong story) when the `children` component was exposed as an argument, which made the arguments table kind of a mess. I don't feel as strongly about it right now, but there are two things about the UX here that kind of throw me off, and I imagine that they'd throw a non-technical user off as well.

1. The default story is something you'd never actually see in production. Right now, the only thing we'll ever see live is a `CircleButton` with a `UserAvatar` inside, and even in the future, we'll always put _something_ inside.
2. There's no way to change the letter rendered inside the icon on this page. Not the biggest deal on this page, because of the `UserAvatar` stories, but I also don't think it's the worst thing in the world to allow this to be more customizable.

joshlarson: To take my two piddly comments above, and cast them in a wildly overblown way, both of ☝️ points reflect the question of whether Storybook is a tool for dev, or a tool for people other than dev.

If we place boundaries around what types of things we allow in our Storybook that really only make sense if you know the inner workings of those icons (for instance, the fact that the user icon thing is implemented using a `children` attribute), then that places Storybook as a dev-only tool. 

Another way of putting it is that if an object has any sort of nontrivial interface, then either the story code needs to be hyper-aware of that nontrivial interface (by using a decorator), or the person clicking around in Storybook needs to be aware (by seeing an interface that either lacks options or displays things about JSXNodes). I would prefer to have the Story code itself be the thing that's aware, and let people interacting with Storybook have a better experience.

lemald: So for what it's worth I was originally conceiving of this all as one component but after talking to design I realized that they actually saw it as two different components, so I think in this case it does make sense to break it apart like this (though I agree in general that UX of Storybook for non-devs should at least be a consideration).

joshlarson: Huh, fair enough.

firestack: Maybe we should start organizing stories according to [atomic design](https://bradfrost.com/blog/post/atomic-web-design/)?
This seems like it'd be a molecule?

firestack: It'd maybe make it more clear on how granular a component is if it were organized according to atomic design, which I see as addressing point 1

> 1. The default story is something you'd never actually see in production. Right now, the only thing we'll ever see live is a `CircleButton` with a `UserAvatar` inside, and even in the future, we'll always put _something_ inside.

assets/css/_circle_button.scss
```diff
@@ -0,0 +1,27 @@
...
+
+  background-color: $color-gray-50;
+  border-radius: 50%;
+
+  &:hover {
+    background-color: $color-gray-100;
+  }
+
+  &:focus-within {
+    border: 0.0625rem solid $color-eggplant-500;
```
joshlarson, line 13: I could be wrong, but I think this is applying the focus border when you click on the button, which ... I don't _think_ that's what focus state is supposed to do? I think the behavior is supposed to be that we get only the "active" state (the darker circle around the avatar) on click, not the focus outline as well.

lemald: The thing is that the active state is actually orthogonal from the hover / focus states, and it just so happens that clicking on the button both focuses it and toggles the active state. I've asked design how they want the precedence to work when we're in both states, but for now I'm letting the interaction states override the active / inactive state since the other way around seems like it would be worse for usability.

assets/css/_circle_button.scss
```diff
@@ -0,0 +1,27 @@
...
+
+    .c-circle-button__content {
+      margin: calc(0.25rem - 0.0625rem);
+    }
+  }
+
+  &--active {
+    background-color: $color-gray-200;
```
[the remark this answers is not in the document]
lemald: That is a good point about the grays, I'll bring this up when I talk with design about the purples as well.

assets/stories/skate-components/circleButton.stories.tsx
lemald: If we feel that the solution is to pursue a large reorganization of our Storybook stories according to atomic design, is that something we can look into in a followup task?

firestack: @lemald https://app.asana.com/0/1148853526253420/1206580891011553/f

lemald: Thanks!

joshlarson:
I'm not approving yet because of the two design-related unresolved questions around focus state and the background gray-200 color of the active state. The rest looks good to me.

I do think that it would be good for us to get a clearer sense as a team of what our storybook stories should look like, but that discussion doesn't need to block this PR. Plus, design should probably be involved in that anyway, since our goal is for this to be a tool at least partly for them.

assets/css/_circle_button.scss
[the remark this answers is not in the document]
lemald: See 850d2029 (changed to use the new grays).

assets/css/_circle_button.scss
lemald: Changed to use `:focus-visible` so the focus state only appears when tabbing.

assets/css/_circle_button.scss
```diff
@@ -0,0 +1,29 @@
...
+
+.c-circle-button {
+  width: 2rem;
+  height: 2rem;
+
+  background-color: tokens.$gray-50;
+  border-radius: 50%;
+
+  &:hover {
+    background-color: tokens.$gray-100;
```
joshlarson, line 11: One additional nit - it looks like the hover state is overriding the active state, which I doubt is what design intended (it feels pretty weird to me). I can't quite wrap my mind around _why_ it's happening, since the nav menu looks like its CSS is set up pretty similarly, but it's not exhibiting this behavior.

That said, I don't think we should block merging for more than a very short time on this, so if we can get design feedback quickly on it, and/or if there's an obvious fix, then hurrah, but otherwise I'm good to merge and follow up with a fix (if indeed a fix is required)

assets/css/_circle_button.scss
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  &:focus-visible {
+    border: 0.0625rem solid var(--bs-primary);
```
firestack, line 15: nit(non-blocking): just wondering if the border utilities would've worked here? https://getbootstrap.com/docs/5.3/utilities/borders/ seems like it would've been classes `border` and `border-primary`?

(comment originally authored at https://github.com/mbta/skate/commit/850d20296ea5d5dcc22375afcf1a7277c458f58a#r138593057 )

firestack: Though I guess this would've had to have been 
```suggestion
    @extend .border,.border-primary;
```
because of the `:focus-visible` state, unless bootstrap has like, "is focus visible" utilities that generate a `border-focus-visible` and `border-primary-focus-visible`

lemald: Yeah... maybe possible (I'd like to make it work) but seems like it might be too much effort for the moment.

assets/css/_circle_button.scss
joshlarson: Ooooh I think I see the problem... sort of. It has to do with how CSS decides how to interpret conflicting rules.

assets/css/_circle_button.scss
```diff
@@ -0,0 +1,29 @@
...
+
+    .c-circle-button__content {
+      margin: calc(0.25rem - 0.0625rem);
+    }
+  }
+
+  &--active {
```
joshlarson:
```suggestion
  &.c-circle-button--active {
```

When there are conflicting rules, like `:hover` versus a specific class, CSS will pick whichever rule trigger is more specific. So without applying this, the rule trigger was `.c-circle-button:hover` (class plus hover state) versus `.c-circle-button--active` (just a class). With applying this change, now it would be `.c-circle-button .c-circle-button--active` (two classes), which I guess CSS sees as more specific than `.c-circle-button:hover`

joshlarson: lol great minds

joshlarson (approved):

PR #2384, opened 2024-02-10
feat: Add logged-in-as element to mobile nav menu

joshlarson:
Screenshot 2024-02-09 at 8 36 51 PM

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206526459943489

PR #2394, opened 2024-02-14
 feat(ts/compmonents/navMenu): add `<LoggedInAs/>` to Mobile `<NavMenu/>`

firestack:
Supersedes: #2384 

---
  - https://app.asana.com/0/0/1206526459943489

### DIFF
--- a/assets/src/components/loggedInAs.tsx
+++ b/assets/src/components/loggedInAs.tsx
@@ -1,8 +1,15 @@
-import React from "react"
+import React, { ComponentPropsWithoutRef } from "react"
 import { UserAvatar } from "./userAvatar"
+import { joinClasses } from "../helpers/dom"
 
-export const LoggedInAs = ({ email }: { email: string }) => (
-  <div className="c-logged-in-as__container">
+export const LoggedInAs = ({
+  email,
+  ...divProps
+}: { email: string } & ComponentPropsWithoutRef<"div">) => (
+  <div
+    {...divProps}
+    className={joinClasses(["c-logged-in-as__container", divProps.className])}
+  >
     <div className="c-logged-in-as__icon">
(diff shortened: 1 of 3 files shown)


joshlarson (approved):

assets/src/components/loggedInAs.tsx
```diff
@@ -1,8 +1,15 @@
...
-export const LoggedInAs = ({ email }: { email: string }) => (
-  <div className="c-logged-in-as__container">
+export const LoggedInAs = ({
+  email,
+  ...divProps
+}: { email: string } & ComponentPropsWithoutRef<"div">) => (
+  <div
+    {...divProps}
+    className={joinClasses(["c-logged-in-as__container", divProps.className])}
+  >
```
firestack, line 12: @mbta/skate-developers what do you think about this approach to adding the classes needed for padding?

I _could_ wrap this in a `<div/>` in `<NavMenu/>`, but where possible, I prefer to not introduce more `div`'s if it can be avoided.

joshlarson: I'm still new enough to React-land that if this is somehow evil, I wouldn't necessarily catch it, but I really like the effect this has, of being able to add classes (and other attributes if needed) to the outer div here, so I'm all for it.

joshlarson (approved):
This time I mean it for real!

assets/src/components/loggedInAs.tsx
firestack: This _seems_ to be fairly normal in React (see, react-bootstrap making basically every HTMLElement attribute available, and more available via `to={}`) But I'm not sure where it's too much, and if this is even the right way to go about it, but I think it's fine so far.

PR #2397, opened 2024-02-14
feat: Switch name of Keycloak test group to recover from Keycloak rollout issues

joshlarson:
This exists as an escape hatch in case the Keycloak rollout runs into issues that cause us to not be able to access `/admin` in order to roll it back.

If we do run into that kind of issue, then we merge this PR and immediately deploy it to production.


joshlarson:
This was useful, and is thankfully no longer needed!

joshlarson:
Apologies for the spam, but Github was still showing this on my "Open PR's tab", and I'm wondering if that's because it went straight from "Draft" to "Closed"

PR #2398, opened 2024-02-14
chore/fix: Upgrade node to 20.11.0

joshlarson:
This fixes an issue where running individual tests sometimes fails with this error:

node:internal/assert:14 throw new ERR_INTERNAL_ASSERTION(message);

More info here: https://github.com/nodejs/node/issues/48763
And here: https://github.com/nodejs/node/issues/50383

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 elixir 1.15.7-otp-26
 erlang 26.1.2
-nodejs 20.7.0
+nodejs 20.11.0
 adr-tools 3.0.0
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN mix local.hex --force && \
   mix local.rebar --force && \
   mix do deps.get --only prod
 
-FROM node:20.7.0-alpine3.17 as assets-builder
+FROM node:20.11.0-alpine3.19 as assets-builder
 
 WORKDIR /root
 ADD . .


lemald:
Really hyped if this ends up fixing it, this error has bugged me for a while!

Question: is this fix available anywhere in the 20.x branch or do we have to go to 21? I just ask because the even-numbered releases get an LTS period so 21 will actually become unsupported sooner than 20 per [the schedule](https://nodejs.org/en/about/previous-releases).

joshlarson:
I will dig some more - I just picked the most recent version, but I think it did get resolved somewhere in 20.x.

joshlarson:
@lemald I replaced it with the latest node 20.x version and that still worked! Thanks for catching that!

lemald:
@joshlarson sorry, the one other place you'll need to update is the `Dockerfile`, since that references a particular version of the Node image. See #2193 (but you shouldn't need any changes to `package.json` or the lockfile for this upgrade).

joshlarson:
@lemald Updated, I believe!

lemald (approved):
Confirmed locally that the Docker build works.

joshlarson:
Oh - I was wrong about 20.11.0 fixing it, and I couldn't find a 20.11.1 alpine docker image (I'm also not sure if 20.11.1 even fixes it)

I'll probably merge this as-is, because while it didn't 100% fix it, it did seem to reduce the probability that it was happening 🤯 (maybe that's just confirmation bias)

lemald:
> Oh - I was wrong about 20.11.0 fixing it, and I couldn't find a 20.11.1 alpine docker image (I'm also not sure if 20.11.1 even fixes it)
> 
> I'll probably merge this as-is, because while it didn't 100% fix it, it did seem to reduce the probability that it was happening 🤯 (maybe that's just confirmation bias)

Ah, that's unfortunate but I agree it's still worth merging this, doesn't hurt to keep our dependencies up-to-date.

PR #2399, opened 2024-02-15
 feat(ex/skate_web/AuthController): redirect to mbta.com on logout 

firestack:


firestack:
@lemald we'll need to make sure each environment has `https://www.mbta.com/` as allowed redirect uri's

lemald (approved):

PR #2400, opened 2024-02-15
feat: "roll over" to new guardian tokens when enabling Keycloak

lemald:
**Asana ticket:** [Better handle transition state](https://app.asana.com/0/1148853526253420/1206613654510834/f)

This will help avoid unusual states around the time that we roll out Keycloak by requiring users to log back in via the new SSO flow as opposed to continuing to use their old token.

A note on Guardian terminology: it may seem weird that we weren't calling `Guradian.Plug.EnsureAuthenticated` before, but that mainly calls the optional `verify_claims` callback. The default `verify_claims` implementation just always accepts the claims so the plug wouldn't have done anything previously.

### DIFF
--- a/lib/skate_web/auth_manager.ex
+++ b/lib/skate_web/auth_manager.ex
@@ -38,19 +38,17 @@ defmodule SkateWeb.AuthManager do
   def resource_from_claims(_), do: {:error, :invalid_claims}
 
   def verify_claims(%{"sub" => subject} = claims, _options) do
-    keycloak_enabled? =
-      "keycloak-sso" in Enum.map(Skate.Settings.TestGroup.get_override_enabled(), & &1.name)
+    auth_provider =
+      if "keycloak-sso" in Enum.map(Skate.Settings.TestGroup.get_override_enabled(), & &1.name) do
+        :keycloak
+      else
+        :cognito
+      end
 
-    if keycloak_enabled? do
-      case subject do
-        @v3_resource_prefix <> _user_ -> {:ok, claims}
-        _ -> {:error, :invalid_claims}
-      end
-    else
-      case subject do
-        @v2_resource_prefix <> _user_id -> {:ok, claims}
-        _ -> {:error, :invalid_claims}
-      end
+    case {auth_provider, subject} do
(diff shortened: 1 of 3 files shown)


lib/skate_web/auth_manager.ex
```diff
@@ -10,20 +10,51 @@ defmodule SkateWeb.AuthManager do
...
+      case subject do
+        @v3_resource_prefix <> _user_ -> {:ok, claims}
+        _ -> {:error, :invalid_claims}
+      end
+    else
+      case subject do
+        @v2_resource_prefix <> _user_id -> {:ok, claims}
+        _ -> {:error, :invalid_claims}
+      end
+    end
```
firestack, line 54: suggestion&nit(non-blocking): I like this syntax better (though I wish that `keycloak_enabled?` was a `auth_provider` variable that was `:cognito` or `:keycloak`) thoughts?
```suggestion
    case {keycloak_enabled?, subject} do
        {true, @v3_resource_prefix <> _user_id} -> {:ok, claims}
        {false, @v2_resource_prefix <> _user_id} -> {:ok, claims}
        _ -> {:error, :invalid_claims}        
    end
```

lemald: I like this a lot better, and I think we can do the `auth_provider` thing! If you want to get credit, you can issue a larger suggestion that also changes lines 41-42 to instead do `auth_provider = if "keycloak-sso" in ...`.

firestack: I could commit it directly if you'd like @lemald

lemald: Works for me.

firestack: 7ae373ea283d366192e35450effb628f8884e267

firestack (dismissed):

lib/skate_web/auth_manager.ex
lemald: Excellent, LGTM.

joshlarson (approved):
Looks great! A major improvement over what was there before!

PR #2402, opened 2024-02-16
fix: Fix text color on the detour dropdown

joshlarson:
From the design review for the detour modal - I missed this when doing the original detour dropdown work.

[Design Review Asana Ticket](https://app.asana.com/0/1148853526253420/1206545095976211/f)

> The default text should use the body font color (#252627, Gray-900), not blue.

### DIFF
--- a/assets/src/components/map/dropdown.tsx
+++ b/assets/src/components/map/dropdown.tsx
@@ -1,27 +1,16 @@
-import React, { PropsWithChildren } from "react"
+import React, { ComponentProps } from "react"
 import { Dropdown } from "react-bootstrap"
 
-interface DropdownMenuProps extends PropsWithChildren {}
-
-export const DropdownMenu = ({ children }: DropdownMenuProps) => {
+export const DropdownMenu = (props: ComponentProps<typeof Dropdown.Menu>) => {
   return (
     <Dropdown.Menu
       className="c-dropdown-popup-menu border-box inherit-box"
       show
-    >
-      {children}
-    </Dropdown.Menu>
+      {...props}
+    />
   )
 }
 
-interface DropdownItemProps extends PropsWithChildren {
-  onClick?: () => void
-}
-
-export const DropdownItem = ({ children, onClick }: DropdownItemProps) => {
-  return (
-    <Dropdown.Item as="button" onClick={onClick}>
-      {children}
-    </Dropdown.Item>
-  )
+export const DropdownItem = (props: ComponentProps<typeof Dropdown.Item>) => {
+  return <Dropdown.Item as="button" {...props} />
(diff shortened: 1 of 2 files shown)


firestack:
Could we add `as="button"` on the `DropdownItem` instead? I think _a_ part of this problem is that it's a link and it shouldn't be (IMO). 
When I added `as="button"` locally, the `color` value changed to `var(--bs-dropdown-link-color)`, which references `var(--bs-body-color)`, which seems more correct if the text color is supposed to be the body text color.

joshlarson:
@firestack how does [this](https://github.com/mbta/skate/pull/2402/commits/03b31f8533c67f8fdaceb1906a4782637fbe50b5) look?

firestack:
> @firestack how does [this](https://github.com/mbta/skate/pull/2402/commits/03b31f8533c67f8fdaceb1906a4782637fbe50b5) look?

I personally think it's a little premature to add it as a non-overridable default. But I'm more okay with this. Maybe rename `DropdownItem` to `DropdownButton`?

joshlarson:
Hmmm - I'm a little torn. I figure in the unlikely event we need to override this, we can add an optional override then.

I'm open to it if there's an easy way to just make skate/`DropdownItem` support _all_ of the fields that react-bootstrap/`Dropdown.Item` does.

assets/src/components/map/dropdown.tsx
```diff
@@ -14,4 +14,14 @@ export const DropdownMenu = ({ children }: DropdownMenuProps) => {
...
+  onClick?: () => void
+}
+
+export const DropdownItem = ({ children, onClick }: DropdownItemProps) => {
+  return (
+    <Dropdown.Item as="button" onClick={onClick}>
+      {children}
+    </Dropdown.Item>
+  )
+}
```
firestack, line 27:
```suggestion
export const DropdownItem = (props: ComponentProps<typeof Dropdown.Item>) => (
  <Dropdown.Item as="button" {...props} />
)
```

firestack:
> I'm open to it if there's an easy way to just make skate/`DropdownItem` support _all_ of the fields that react-bootstrap/`Dropdown.Item` does.

https://github.com/mbta/skate/pull/2402/files#r1492963708

joshlarson:
@firestack Kewl kewl! [Here's](https://github.com/mbta/skate/pull/2402/commits/7519d4241c947c19ba6f0e69ac729a27dad8ebd4) my attempt at that (`DropdownMenu` got the same treatment)

lemald (approved):
As far as I can tell this incorporates @firestack's feedback, so I'm going to go ahead and approve.

PR #2409, opened 2024-02-21
fix: call `onOpen` before `connect` and add useEffect cleanup

lemald:


firestack (approved):
Seems simple enough to at least find out if it changes anything for Joan.

assets/src/hooks/useSocket.ts
```diff
@@ -36,11 +36,11 @@ const useSocket = (): SocketStatus => {
     const initialSocket = new Socket("/socket", {
       params: { token: userToken },
     })
-    initialSocket.connect()
     initialSocket.onOpen(() => setConnectionStatus(ConnectionStatus.Connected))
     initialSocket.onClose(() =>
       setConnectionStatus(ConnectionStatus.Disconnected)
     )
+    initialSocket.connect()
```
joshlarson, line 43: This isn't complete yet, right? I think we're still not returning a cleanup function. Or if a cleanup function isn't actually needed, should we update the title of the PR?

Also, just to make sure I'm understanding this correctly - this resolves a potential race condition where `.connect()` is called _and resolves_ before `.onOpen` is called, which means that `setConnectionStatus(ConnectionStatus.Connected))` is never called?

assets/tests/hooks/useSocket.test.ts
```diff
@@ -11,6 +11,7 @@ jest.mock("phoenix", () => ({
     connect: jest.fn(),
     onOpen: jest.fn(),
     onClose: jest.fn(),
+    disconnect: jest.fn(),
```
joshlarson, line 14: Does this need to be here? Is there a test that relies on this that hasn't made it into the PR yet?

assets/src/hooks/useSocket.ts
lemald: Yeah sorry this PR is sort of left in a weird state in a lot of ways. I think this functionality could use some cleanup but I'm no longer convinced that this `useEffect` hook is key to the issue:
1. Even if there is a race condition the worst case is that the connection status gets left as "loading," which doesn't trigger the refresh modal (only the "disconnected" state does that).
2. Even if we add a cleanup function (which we should for the sake of completeness), it would never get called in practice because the token comes from the server-rendered HTML that's only updated on a full page load so I don't think it would have any impact on the state. My worry was that the `onClose` callback was setting the connection to "disconnected" and then we weren't able to get back to connected, but then I remembered that we just trigger a full refresh anyway.

assets/tests/hooks/useSocket.test.ts
lemald: This is left over from when I was experimenting with the actual cleanup function - I had the cleanup call `disconnect`, so any test involving this hook would attempt to call the `disconnect` method on unmount.

lemald:
Given what I said above about this likely no longer being relevant, I'm going to close the PR. I still think we want to revisit this logic, but this change on its own won't really help anything as far as I can tell.

PR #2411, opened 2024-02-22
Log user information during socket connection

lemald:
This adds as much logging as we can when accepting / rejecting a new socket connection, including the user ID when a well-formed but expired token is presented.

I could have also loaded the actual username or other information from the database, but I opted to instead make it easier to pull that information in CSV format and not add the overhead to every socket authentication.

`with` continues to be my favorite bit of Elixir syntax.


lib/skate_web/channels/user_socket.ex
```diff
@@ -22,9 +23,26 @@ defmodule SkateWeb.UserSocket do
   # See `Phoenix.Token` documentation for examples in
   # performing token verification on connect.
   def connect(%{"token" => token}, socket, _connect_info) do
+    user_id =
+      with %{claims: claims} <- Guardian.peek(SkateWeb.AuthManager, token),
```
firestack, line 27:
```suggestion
      with %{claims: claims} <- SkateWeb.AuthManager.peek(token),
```

lib/skate_web/channels/user_socket.ex
```diff
@@ -22,9 +23,26 @@ defmodule SkateWeb.UserSocket do
   # See `Phoenix.Token` documentation for examples in
   # performing token verification on connect.
   def connect(%{"token" => token}, socket, _connect_info) do
+    user_id =
```
firestack, line 26: suggestion(non-blocking): What do you think about naming this variable differently so it's obvious that this is not safe to use for anything but logging?
Maybe `unreliable_user_id`, or `unconfirmed_user_id`. I was thinking `unvalidated_user_id` which is maybe better, but I'm not sure `unvalidated` is a proper word?

lemald: How about `unverified`? I think that matches up with Guardian terminology.

firestack: works for me 👍🏻

lemald: cd5e4940

firestack (approved):

PR #2412, opened 2024-02-22
 feat: add user-guide vanity URL

firestack:
---
  - https://app.asana.com/0/0/1206673444071866

### DIFF
--- a/lib/skate_web/controllers/redirect_controller.ex
+++ b/lib/skate_web/controllers/redirect_controller.ex
@@ -25,5 +25,9 @@ defmodule SkateWeb.Redirect do
   @aup_url Application.compile_env(:skate, :acceptable_use_policy)
   defp allowed_url_from_atom(:aup), do: {:ok, @aup_url}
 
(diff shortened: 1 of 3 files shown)


lib/skate_web/controllers/redirect_controller.ex
```diff
@@ -25,5 +25,9 @@ defmodule SkateWeb.Redirect do
   @aup_url Application.compile_env(:skate, :acceptable_use_policy)
   defp allowed_url_from_atom(:aup), do: {:ok, @aup_url}
 
+  @user_guide_url Application.compile_env(:skate, :user_guide_url)
+  defp allowed_url_from_atom(:user_guide),
+    do: {:ok, @user_guide_url}
+
```
firestack, line 31: I think I'd mind less if we didn't have to do this part. I like that it's configurable, but having to do this extra step is somewhat annoying.

lemald: Yeah, it's not entirely ideal but I think for now the amount of static redirects we'll need to do is pretty small.

lemald (approved):

PR #2415, opened 2024-02-23
feat: missed stops endpoint

lemald:
Asana ticket: [Missed stops API endpoint](https://app.asana.com/0/1148853526253420/1206680523793247/f)

### DIFF
--- a/lib/skate_web/controllers/detour_missed_stops_controller.ex
+++ b/lib/skate_web/controllers/detour_missed_stops_controller.ex
@@ -46,7 +46,7 @@ defmodule SkateWeb.DetourMissedStopsController do
     ).(trip_id)
   end
 
-  defp missed_stops(cfg) do
-    Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1).(cfg)
+  defp missed_stops(args) do
+    Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1).(args)
   end
 end
--- a/lib/skate_web/controllers/detour_missed_stops_controller.ex
+++ b/lib/skate_web/controllers/detour_missed_stops_controller.ex
@@ -10,29 +10,16 @@ defmodule SkateWeb.DetourMissedStopsController do
         "connection_start" => connection_start,
         "connection_end" => connection_end
       }) do
-    route_pattern_fn =
-      Application.get_env(:skate_web, :route_pattern_fn, &Schedule.route_pattern/1)
-
-    shape_with_stops_fn =
-      Application.get_env(
-        :skate_web,
-        :shape_with_stops_fn,
-        &Schedule.shape_with_stops_for_trip/1
-      )
-
-    missed_stops_fn =
-      Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1)
-
-    with route_pattern <- route_pattern_fn.(route_pattern_id),
+    with route_pattern <- route_pattern(route_pattern_id),
          false <- is_nil(route_pattern),
          shape_with_stops <-
-           shape_with_stops_fn.(route_pattern.representative_trip_id),
+           shape_with_stops(route_pattern.representative_trip_id),
          false <- is_nil(shape_with_stops) do
       connection_start_location = Location.new(connection_start["lat"], connection_start["lon"])
       connection_end_location = Location.new(connection_end["lat"], connection_end["lon"])
 
       missed_stops =
-        missed_stops_fn.(%MissedStops{
+        missed_stops(%MissedStops{
           connection_start: connection_start_location,
           connection_end: connection_end_location,
           stops: shape_with_stops.stops,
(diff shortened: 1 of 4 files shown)


lib/skate_web/controllers/detour_missed_stops_controller.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    shape_with_stops_fn =
+      Application.get_env(
+        :skate_web,
+        :shape_with_stops_fn,
+        &Schedule.shape_with_stops_for_trip/1
+      )
+
+    missed_stops_fn =
+      Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1)
```
firestack, line 24: nit/suggestion(non-blocking): I have a little trouble focusing on the content of the function itself when having to also try and mentally ignore these setup env calls.
How would you feel about doing something similar to what I did in the `auth_controller` and instead define these as private functions?
https://github.com/mbta/skate/blob/a3c3e2495de9d27ee618dcb0db9359eda623af04/lib/skate_web/controllers/auth_controller.ex#L85-L90

I could also see `Mox` being a solution here? But I can't say I really understand it well enough yet to know how it'd be used.

lemald: What do you think of 28f1c160?

firestack: perfect! Thank you!

lib/skate_web/controllers/detour_missed_stops_controller.ex
```diff
@@ -0,0 +1,52 @@
...
+  defp shape_with_stops(trip_id) do
+    Application.get_env(
+      :skate_web,
+      :shape_with_stops_fn,
+      &Schedule.shape_with_stops_for_trip/1
+    ).(trip_id)
+  end
+
+  defp missed_stops(cfg) do
+    Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1).(cfg)
```
joshlarson, line 50:
```suggestion
  defp missed_stops(args) do
    Application.get_env(:skate_web, :missed_stops_fn, &MissedStops.missed_stops/1).(args)
```

**Nit**: I got confused at first and thought that `cfg` was some kind of application-level config.

lemald: I borrowed the `cfg` terminology from the `MissedStops` module itself, though I agree that `args` is probably clearer. The `MissedStops` struct is a little weird IMO because it's only used as the argument to the `missed_stops` function but maybe @firestack could speak to why it's implemented that way?

firestack: I probably should've used something that would accept either a keyword list or map, but  I didn't want to make it overly complex. I was finding that it was easy to get the args confused and it ended up being difficult to understand at a glance when reading the code back, so for readability's sake, I went with a struct. 

I'm fine with changing it, though I think the arguments should continue to be named in the calling code. I liked using a struct because it allows for requiring all fields to be present and making that somewhat visible at "writing time" and compile time.

lemald: My perhaps controversial opinion is that I'd just write it as a `/4` function, though I think we should hold off on any refactoring on that and do it in another ticket / PR if we want to. In the meantime I like @joshlarson's naming change so I'll apply the suggestion.

test/skate_web/controllers/detour_missed_stops_controller_test.exs
```diff
@@ -0,0 +1,72 @@
...
+      missed_stop = Enum.at(shape_with_stops.stops, 0)
+
+      connection_start = shape_with_stops.points |> Enum.at(0) |> Util.Location.as_location!()
+      connection_end = shape_with_stops.points |> Enum.at(1) |> Util.Location.as_location!()
+
+      reassign_env(:skate_web, :route_pattern_fn, fn _ -> route_pattern end)
+      reassign_env(:skate_web, :shape_with_stops_fn, fn _ -> shape_with_stops end)
+
+      reassign_env(:skate_web, :missed_stops_fn, fn %MissedStops{
+                                                      connection_start: ^connection_start,
```
joshlarson, line 23: How come you mocked the algorithm, rather than just providing some values for `route_pattern` and `shape_with_stops`, etc, that have simple enough expected results?

I'm partly asking because mocking out the internal function makes this test kind of brittle, and makes the whole PR a bit tougher to review. If we update `missed_stops_fn` to return data in a different way, then this test will still pass, and won't tell us anything about how well this controller works in context.

lemald: See if this was TypeScript I'd just make sure that it was set up such that the mocked function has to have the same type signature as the original, meaning that any changes to the interface without updating the mocked function would trigger a type error...

In any event I'm not sure we could elegantly do that in Elixir / Dialyzer except for _maybe_ via behaviours. My concern with leaving the test to call `missed_stops/1` directly is that it's a very complicated function and mocking out input data and expected output data would involve a lot of geospatial reasoning that seems pretty disconnected from the controller itself. I think overall the current approach is the lesser evil given that the return type of the function (list of stops) seems like it will be fairly stable.

joshlarson: It's not exactly blocking my review of this (it does make reviewing it a bit more complicated, because I have to cross-reference this mock with the behavior of `missed_stops`), but at some point, would you object if I tried refactoring this test to use "realistic" test data instead? Getting reasonable fake data that abstracts away a lot of the geospatial reasoning is a good use case for factories, and we're starting to build the beginnings of a good factory system.

Basically, I'm not as convinced that this is the lesser of two evils, but I don't think it's worth blocking this PR on, and that's kind of a philosophical point anyway.

joshlarson (approved):

test/skate_web/controllers/detour_missed_stops_controller_test.exs
lemald: Yeah I'd certainly check out the result if you want to try that!

PR #2419, opened 2024-02-26
Fetch and display missed stops

lemald:
Asana ticket: [Fetch and display missed stops](https://app.asana.com/0/0/1206680523793255/f)

The count of the missed stops and any other style enhancements will go in a follow-up PR, since I want to start getting feedback on these changes quickly. I'm not entirely satisfied with using `instantPromise` in the `<DiversionPage>` test, but I wasn't able to get it to work with `mockResolvedValue` no matter what combination of things I tried, probably just because of some React rendering and test mocking subtleties that I haven't figured out. Ideas for things to try are welcome.


assets/src/components/detours/diversionPage.tsx
```diff
@@ -39,7 +39,7 @@ export const DiversionPage = ({
       <div className="l-diversion-page__panel bg-light">
         <DiversionPanel
           directions={directions}
-          missedStops={missedStops}
+          missedStops={missedStops || undefined}
```
joshlarson, line 42: This is sort of a philosophical thing, but `DiversionPage` doesn't feel like the right place to put this. As far as `missedStops` is concerned, `DiversionPage` is just a wire that transmits data from `useDetour` to `DiversionPanel`, so I'd rather have the logic for translating _generic-falsy-value_ into `undefined` live in `useDetour` or `DiversionPanel`.

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -1,11 +1,12 @@
...
+import React from "react"
 import { RoutePill } from "../routePill"
 import { DetourShape } from "../../models/detour"
 import { ListGroup } from "react-bootstrap"
+import { Stop } from "../../schedule"
 
 export interface DiversionPanelProps {
   directions?: DetourShape["directions"]
-  missedStops?: ReactNode
+  missedStops?: Stop[]
```
joshlarson, line 9: **praise:** I like this change. Makes it a lot easier to think about if `missedStops` is _just data_, rather than rendering logic as well.

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -67,7 +68,13 @@ export const DiversionPanel = ({
       {missedStops && (
         <section className="pb-3">
           <h2 className="c-diversion-panel__h3">Missed Stops</h2>
-          {missedStops}
+          <ListGroup as="ol">
+            {missedStops.map((missedStop) => (
+              <ListGroup.Item key={missedStop.id}>
```
firestack, line 73: issue: Are these ID's unique?

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
lemald: I'm worried about that too, but I think I'm going to attempt rebasing this onto your changes once you merge them and I'll see how hard (or not) it ends up being.

assets/stories/skate-components/detours/diversionPanel.stories.tsx
```diff
@@ -43,15 +42,6 @@ export const WithDirections: Story = {
...
-      <>
-        <ListGroup variant="flush" as="ol">
-          <ListGroup.Item as="li">Stop 1</ListGroup.Item>
-          <ListGroup.Item as="li">Stop 2</ListGroup.Item>
-          <ListGroup.Item as="li">Stop 3</ListGroup.Item>
-          <ListGroup.Item as="li">Stop 4</ListGroup.Item>
-        </ListGroup>
-      </>
-    ),
+    missedStops: stopFactory.buildList(4),
```
joshlarson, line 45: Are these deterministic?

lemald: Good question - the sequences might not be depending on the order in which this runs relative to other factories. The thing that actually gets rendered is the name, so I think I'll explicitly specify the names so as to avoid spurious diffs in the future.

joshlarson: Lol - yes they are

Screenshot 2024-02-26 at 12 00 23 PM

assets/src/components/detours/diversionPage.tsx
lemald: How would you feel about just changing the type of `missedStops` returned from `useDetour()` from `Stop[] | null` to `Stop[] | undefined` much like what `useDetourDirections()` does?

assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -5,9 +5,13 @@ import { DetourMap } from "../../../src/components/detours/detourMap"
 import "@testing-library/jest-dom/jest-globals"
 import { defaultCenter } from "../../../src/components/map"
 import { latLngLiteralFactory } from "../../factories/latLngLiteralFactory"
+// import { fetchDetourMissedStops } from "../../../src/api"
```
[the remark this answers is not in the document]
lemald: 7deba436

assets/src/components/detours/diversionPage.tsx
joshlarson: I'd be fine with that

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
joshlarson: Hahaha good luck!

assets/tests/hooks/useDetour.test.ts
```diff
@@ -179,12 +190,37 @@ describe("useDetour", () => {
...
+    const { result } = renderHook(() => useDetour("routePatternId"))
+
+    jest
+      .mocked(fetchDetourMissedStops)
+      .mockResolvedValue(stopFactory.buildList(3))
 
     act(() => result.current.addConnectionPoint({ lat: 0, lon: 0 }))
     act(() => result.current.addConnectionPoint({ lat: 0, lon: 0 }))
 
+    await waitFor(() => {
```
joshlarson, line 202: Is this (and the mock-resolved-value above) necessary as part of this test? It almost looks like setup meant for a different set of assertions?

assets/tests/hooks/useDetour.test.ts
```diff
@@ -169,7 +180,7 @@ describe("useDetour", () => {
   })
 
   test("when `waypoints` is not empty, `canUndo` is `true`", async () => {
-    const { result } = renderHook(useDetour)
+    const { result } = renderHook(() => useDetour("routePatternId"))
```
joshlarson, line 183: **Nit (non-blocking):** This `renderHook(() => useDetour("routePatternId"))` is duplicated a lot throughout this test - I think in every single test in this file. Is there a way to extract it?

assets/src/components/detours/diversionPage.tsx
lemald: cf06a98a

assets/src/components/detours/diversionPanel.tsx
firestack: But a stop could be visited twice, so that would mean these are not unique, correct?

assets/tests/hooks/useDetour.test.ts
lemald: I just added it to make the test more realistic since the hook is now doing that API logic when an endpoint is added, but I could be convinced to remove it.

assets/src/components/detours/diversionPanel.tsx
lemald: We don't have stop sequence in the data we get back, but I think the correct way to handle this is actually to just remove duplicated stops from the list, since the list is basically "which stops on this route are somehow affected by the detour."

assets/tests/hooks/useDetour.test.ts
joshlarson: I'd vote in favor of removing it, since I think it makes it a bit more confusing to tell exactly _when_ the bits that the test asserts on happen, and makes it seem like the test relies on this setup.

assets/tests/hooks/useDetour.test.ts
lemald: f3509893

assets/src/components/detours/diversionPanel.tsx
firestack: I disagree, In my mind, this UI element represents the ordered list of stops that are missed. So to me, that means a stop that is missed twice would appear twice.

But I think we're both making assumptions about what should be here, so maybe we should ask the larger team?

If we do the "which stops on this route are somehow affected by the detour." version, I think we should switch this `<ListGroup>` to a `ul`

assets/tests/hooks/useDetour.test.ts
lemald: Ahh, I remember now - without this, the API logic still fires with the default mock resolved value of `null` and then triggers a `Warning: An update to TestComponent inside a test was not wrapped in act(...).` error. That said, to actually make sure I'm avoiding a race condition I should update the `waitFor` to look for `.no.toBeUndefined()`.

joshlarson:
Current status: I'm mostly holding off on approving in order to make room for a resolution to the stop-ID-uniqueness conversation. Otherwise I think this looks great!

 - I'd love to see [that extra stuff in the tests removed](https://github.com/mbta/skate/pull/2419#discussion_r1503017949); I think that is approval-blocking for me, but I could probably be persuaded otherwise
 - [Hard-coding nicer stop names](https://github.com/mbta/skate/pull/2419#discussion_r1503007689) for storybook is a nice-to-have (won't block approval)

assets/tests/hooks/useDetour.test.ts
lemald: 6f9e27d1

assets/src/components/detours/diversionPanel.tsx
lemald: The purpose of this data is to be used in alerts, and alerts can only select stop IDs, not stop IDs combined with stop sequences: https://gtfs.org/realtime/reference/#message-entityselector

That said I agree that if this is the case we should change to a `ul`. I'll do that along with the other updates and we can ask the larger group, but I don't want to block this PR on that.

assets/tests/hooks/useDetour.test.ts
joshlarson: That's actually... kind of surprising, since the update is triggered by `act(() => result.current.addConnectionPoint({ lat: 0, lon: 0 }))`, which _is_ wrapped in `act`.

But I'll let this go. It's annoying, but not a huge deal, and we can take a more holistic approach to cleaning these up later if we decide that's a thing we should do.

lemald: That's what triggers the `useEffect` to fire, but the promise from `fetchDetourMissedStops` still has to resolve, hence it happening outside of that `act`.

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
lemald: It actually ended up not being a bad rebase, there were a bunch of manual conflicts to resolve but none of them were actually challenging and the file rename wasn't an issue at all.

assets/stories/skate-components/detours/diversionPanel.stories.tsx
lemald: ffe26ef0

assets/tests/hooks/useDetour.test.ts
joshlarson: Oooooh - that's annoying! I wish there were a way to make promises and things like that in tests resolve synchronously.

assets/src/components/detours/diversionPanel.tsx
lemald: I still have to actually make those updates, but one interesting thing is that on the back-end we rely on `shape_with_stops_for_trip`, which in turn calls `stops_for_trip`, which in turn doesn't do anything special to handle stops that are visited twice on the same trip. So just flagging that if we do want to be sensitive to stop sequence here, we would need to revisit some of that logic as well.

lemald:
@firestack @joshlarson okay I believe I have now addressed the various items I had outstanding with both of you.

assets/tests/hooks/useDetour.test.ts
```diff
@@ -28,7 +33,11 @@ describe("useDetour", () => {
     const start = { lat: 0, lon: 0 }
     const end = { lat: 1, lon: 1 }
 
-    const { result } = renderHook(useDetour)
+    const { result } = renderHook(useDetourWithFakeRoutePattern)
+
+    jest
+      .mocked(fetchDetourMissedStops)
+      .mockImplementation(() => instantPromise(null))
```
firestack, line 40: suggestion: Can this be changed to use `mockResolvedValue` and `waitFor` like the other tests?
I really don't like `instantPromise`, especially when we have `mockResolvedValue` and `Promise.resolve`, as this sidesteps some of the react rendering things (I'll detail that more later).

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -147,4 +150,48 @@ describe("DetourMap", () => {
...
+
+    await fireEvent.click(
+      container.querySelector(".c-detour_map--original-route-shape")!
+    )
+
+    await waitFor(() =>
+      expect(screen.queryByText(stop.name)).toBeInTheDocument()
+    )
+
+    expect(screen.getByText(stop.name)).toBeInTheDocument()
```
firestack, line 172: suggestion(non-blocking): AFAIK, `waitFor` will retry a `expect` until it succeeds, or it will fail the test, so the second assert is redundant

```suggestion
    await waitFor(() =>
      expect(screen.getByText(stop.name)).toBeInTheDocument()
    )
   
```

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -147,4 +150,48 @@ describe("DetourMap", () => {
...
+    await fireEvent.click(
+      container.querySelector(".c-detour_map--original-route-shape")!
+    )
+
+    await waitFor(() =>
+      expect(screen.queryByText(stop.name)).toBeInTheDocument()
+    )
+
+    expect(screen.getAllByText(stop.name)).toHaveLength(1)
+  })
```
firestack, line 196: https://github.com/mbta/skate/pull/2419#discussion_r1503316396

```suggestion
    await waitFor(() =>
      expect(screen.getAllByText(stop.name)).toHaveLength(1)
    )
  })
```

firestack (approved):
I added a few comments that are nice-to-haves, but this LGTM

assets/tests/hooks/useDetour.test.ts
lemald: I also want to avoid `instantPromise`, but I've been having a really hard time getting `mockResolvedValue` to work for many of these tests, especially if there isn't a change in the ultimate result of the hook that I can `waitFor`. Maybe I'm missing something, though.

PR #2422, opened 2024-02-26
feat: Add ability to clear entire detour in one go

joshlarson:
---
  - https://app.asana.com/0/0/1206660467722364
  - Parent Ticket: https://app.asana.com/0/home/1205718271156544/1206412398464686

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -36,6 +36,7 @@ export const DiversionPage = ({
 
     canUndo,
     undo,
+    clear,
   } = useDetour()
 
   return (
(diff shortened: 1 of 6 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -59,6 +60,7 @@ export const DiversionPage = ({
           onClickOriginalShape={addConnectionPoint}
           undoDisabled={canUndo === false}
           onUndo={undo}
+          onClear={clear}
```
joshlarson, line 63: **Thought:** This is exactly the kind of annoying little diff that we can get away from if we pursue @firestack's idea and make this whole thing more composable.

lemald (approved):

PR #2424, opened 2024-02-26
feat: Highlight and widen clickable area on original route shape for detour map

joshlarson:
https://github.com/mbta/skate/assets/912020/819e8756-37d0-4685-91aa-35a544a7a381

---
  - https://app.asana.com/0/0/1206698934433015
  - https://app.asana.com/0/0/1206423114607620

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,4 +1,4 @@
-.c-detour_map--original-route-shape-skeleton {
+.c-detour_map--original-route-shape-core {
   stroke: $color-kiwi-500;
   opacity: 50%;
 }
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,5 +1,15 @@
 .c-detour_map--original-route-shape {
   stroke: $color-kiwi-500;
+  opacity: 50%;
+}
+
+.c-detour_map--original-route-shape-outline {
+  stroke: $color-kiwi-500;
+  opacity: 0%;
+
+  &:hover {
+    opacity: 35%;
+  }
 }
 
 .c-detour_map--original-route-shape__unstarted {
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,9 +1,9 @@
-.c-detour_map--original-route-shape {
+.c-detour_map--original-route-shape-skeleton {
   stroke: $color-kiwi-500;
   opacity: 50%;
 }
 
-.c-detour_map--original-route-shape-outline {
+.c-detour_map--original-route-shape {
   stroke: $color-kiwi-500;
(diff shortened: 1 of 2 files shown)


assets/src/components/detours/detourMap.tsx
```diff
@@ -211,18 +211,44 @@ const OriginalRouteShape = ({
...
+    },
+    mouseout: () => {
+      setHovered(false)
+    },
+  }
+
+  return (
+    <>
+      <Polyline
+        key={`outline-hovered-${hovered}`}
```
joshlarson, line 230: This `key` is here because Leaflet doesn't React that well - without it, Leaflet doesn't re-render the Polyline with the updated `opacity`. 

I'm not convinced that this is the best way to do this.

assets/css/detours/_detour_map.scss
```diff
@@ -1,5 +1,15 @@
+.c-detour_map--original-route-shape-skeleton {
```
joshlarson, line 1: 💀 I'd love to come up with a better name for this.

My original idea was to call this `original-route-shape` and the other one `original-route-shape-outline`, but the other one is really standing in for the _actual_ route shape, since that's where all the actual behavior is. This one is just a visual marker showing where the route shape is when it's not highlighted.

assets/css/detours/_detour_map.scss
```diff
@@ -1,5 +1,15 @@
...
+  stroke: $color-kiwi-500;
+  opacity: 50%;
```
joshlarson, line 3: *Note: This doesn't quite match the hi-fi designs, I think. According to Figma, when the route is highlighted, this should be completely opaque, rather than effectively at 85% opacity (or however 50% opacity and 35% opacity stack up. I don't really want to fix that right now though, because keeping it as-is means that it can be fully CSS, while making it fully opaque would involve at least a bit more complexity*

assets/css/detours/_detour_map.scss
lemald: Just a random thought, how about "core" as opposed to "skeleton?" Not sure if that's any better.

joshlarson: I think it's better than "skeleton" at least!

assets/src/components/detours/detourMap.tsx
```diff
@@ -212,17 +212,25 @@ const OriginalRouteShape = ({
...
-      click: onClick,
-    }}
-  >
-    {children}
-  </Polyline>
+  <>
+    <Polyline
+      weight={6}
+      positions={positions}
+      className={joinClasses(["c-detour_map--original-route-shape-skeleton"])}
```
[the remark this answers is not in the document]
joshlarson: Good catch, thanks!

firestack (approved):

lemald (approved):
LGTM as well!

PR #2429, opened 2024-02-27
feat(ts/components/diversionPage): add share detour panel 

firestack:

PR #2430, opened 2024-02-27
 refactor: extract original route query into byRole like API 

firestack:
This depends on #2429 because I added some tests which also needed this, but to avoid merge conflicts I'm basing this off of #2429 

This follows the pattern in https://github.com/mbta/skate/blob/main/assets/tests/testHelpers/selectors/components/map/markers/stopIcon.ts#L4-L11

### DIFF
--- a/assets/tests/testHelpers/selectors/components/detours/diversionPage.tsx
+++ b/assets/tests/testHelpers/selectors/components/detours/diversionPage.tsx
@@ -1,3 +1,14 @@
+import { expect } from "@jest/globals"
 import { byRole } from "testing-library-selector"
 
 export const finishDetourButton = byRole("button", { name: "Finish Detour" })
+
+export const originalRouteShape = {
+  get(container: HTMLElement): Element {
+    const maybeShape = container.querySelector(
+      ".c-detour_map--original-route-shape"
+    )
+    expect(maybeShape).not.toBeNull()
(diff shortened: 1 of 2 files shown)


assets/tests/testHelpers/selectors/components/detours/diversionPage.tsx
```diff
@@ -1,3 +1,14 @@
...
 
 export const finishDetourButton = byRole("button", { name: "Finish Detour" })
+
+export const originalRouteShape = {
+  get(container: HTMLElement): Element {
+    const maybeShape = container.querySelector(
+      ".c-detour_map--original-route-shape"
+    )
+    expect(maybeShape).not.toBeNull()
+    return maybeShape as Element
```
lemald, line 12: Ugh, I really wish that the `expect(...).no.toBeNull()` somehow worked with TypeScript's type narrowing such that you didn't need the `as Element` on the following line.

lemald (approved):

assets/tests/testHelpers/selectors/components/detours/diversionPage.tsx
firestack: > Ugh, I really wish that the `expect(...).no.toBeNull()` somehow worked with TypeScript's type narrowing

:same: :1000: 💯

PR #2432, opened 2024-02-27
fix(css+sb/diversion-panel): fix panels not expanding in stories

firestack:


assets/css/_diversion_page.scss
```diff
@@ -56,7 +56,20 @@
   }
 }
 
+
 .l-diversion-page-panel {
+  // Try to fill space and expand children when not in
+  // the `l-diversion-page` grid
+  :not(.l-diversion-page__panel) > & {
```
firestack, line 63: I wanted this to be the default behavior when not used in a `.l-diversion-page` component, but the selector was much more complicated and I'm not sure of the performance implications, so I instead opted to make an assumption that we'll only be in a `l-diversion-page` grid if we're the immediate child of the `__panel` slot.

The more expensive query looked like this
```suggestion
  &:not(.l-diversion-page *) {]
```

firestack: I considered hardcoding this to happen only in storybook contexts, but then the behavior of the component would be different when not used in storybook.

lemald: Yeah, after thinking through the options I agree that this appears to be a good overall balance. 👍

lemald (approved):

joshlarson (approved):

PR #2433, opened 2024-02-27
feat: missed stops count badge

lemald:
Asana ticket: [Show count of missed stops in a badge](https://app.asana.com/0/1148853526253420/1206705505932414/f)

I was really torn about how to handle the new color and I'm open to any and all suggestions about how to implement it. Adding it as a theme color makes it accessible via Bootstrap's auto-generated utility classes, which is super useful. On the other hand, it seems really weird to add something like this that's probably a one-off to the theme. Also, putting it in `_map_overrides.scss` seems weird, but it didn't feel like it fit particularly well into any of the other places we store color tokens.

As for the result, it looks mostly good to my eyes except for the sizing, which is a bit different from what's in the designs. I want to avoid too much with custom sizing so I'm going to try to get clarity from design about how it's supposed to relate to the sizing of its parent component and I'll adjust if needed.


assets/css/bootstrap/_map_overrides.scss
```diff
@@ -9,4 +9,5 @@ $theme-colors: (
   "service-alert": semantic.$service-alert,
   "light": new_tokens.$gray-50,
   "dark": new_tokens.$gray-900,
+  "missed-stop": #e8002a,
```
firestack, line 12: suggestion(non-blocking): I hope this isn't too annoying, but maybe we could put this in `tokens_2024` as `missed-stop-1`, and then define `missed-stop` in the `semantic` file via that token?

I think the color belongs in the tokens file because it's become a raw value we're using, but I think it's important that we pull from `semantic` where possible, so it's unfortunate that it'll look like the name is duplicated because the token name and semantic name are the same, but I think it's better in terms of organization in the long run.

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -74,7 +74,12 @@ export const DiversionPanel = ({
 
         {missedStops && (
           <section className="pb-3">
-            <h2 className="c-diversion-panel__h3">Missed Stops</h2>
+            <h2 className="c-diversion-panel__h2">
+              Missed Stops{" "}
+              <Badge bg="missed-stop" className="c-diversion-panel__h3">
```
firestack, line 79: **praise**: I love the way that this ended up being usable via the `bg` parameter!

assets/css/bootstrap/_map_overrides.scss
lemald: No, this is exactly the sort of comment I was hoping before because I was personally undecided about where to put it. 😅 I think I'll go with that option unless I suddenly come up with a different counterproposal.

lemald: Is this what you had in mind: 50565ddf.

firestack: Love it! Thanks!

All the rest of this comment is non-blocking feedback ATM 👍🏻 

---

I think the only thing I'm not sure about is which "color groups" this should be added to

image
As shown in the above image, it's currently in the `Theme Colors` group, but I'm not really sure it belongs there? But I don't think it belongs in service colors or mbta colors either...

Maaaaaybe it should be in the `Misc Colors > System` group?..
image

Hard choices 😅

firestack (approved):

assets/css/bootstrap/_map_overrides.scss
lemald: Hmm... I agree that in a sense it doesn't make sense as a "theme color" semantically, but it is in fact implemented as a "theme color" in Bootstrap. Maybe this means I shouldn't have done it that way after all? 🙃 But I could probably also do some overrides in `assets/stories/skate-design/colors.ts` to change where the color actually shows up in Storybook.

firestack: I think it's being added to bootstrap's theme colors, because [I really like how we can use the `bg` parameter due to that choice](https://github.com/mbta/skate/pull/2433#discussion_r1504995254). 

What if we open a ticket to review and potentially fix this with design and merge this PR?

firestack: I think adding `missed_stop` to the bootstrap theme colors map is a good decision, because [I really like how we can use the `bg` parameter due to that choice](https://github.com/mbta/skate/pull/2433#discussion_r1504995254). But I don't think it necessarily needs to be in that export group though.

I tried playing with it locally and I came up with d1a61c79b5b6978294db15c2c1bef9455851f307, thoughts? (I also realized that my recommendation of `missed-stop-1` was not a great suggestion)

lemald: Okay, yeah, that seems like a nice balance, I'll go ahead and push it.

PR #2434, opened 2024-02-27
fix: Do not highlight and widen original route shape after the detour…

joshlarson:
… is finished

---
  - https://app.asana.com/0/0/1206708177134307

### DIFF
--- a/assets/tests/components/detours/detourMap.test.tsx
+++ b/assets/tests/components/detours/detourMap.test.tsx
@@ -20,6 +20,7 @@ const DetourMapWithDefaults = (
     endPoint={undefined}
     waypoints={[]}
     undoDisabled={false}
+    originalShapeClickable={true}
     onClickMap={() => {}}
     onClickOriginalShape={() => {}}
     onUndo={() => {}}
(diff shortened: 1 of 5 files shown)


joshlarson:
Oops

assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -20,6 +20,7 @@ const DetourMapWithDefaults = (
     endPoint={undefined}
     waypoints={[]}
     undoDisabled={false}
+    originalShapeClickable={true}
```
lemald, line 23: **suggestion:** is it possible to add a couple of tests for what happens when this prop is enabled / disabled?

firestack (approved):

assets/tests/components/detours/detourMap.test.tsx
joshlarson: [How about these?](https://github.com/mbta/skate/pull/2434/commits/e205afd809a220afdd339af1d747bc7ab8a0c30b)

lemald (approved):

PR #2436, opened 2024-02-28
fix: clear missed stops when detour is no longer in a finished state

lemald:
Thanks to @firestack for catching.

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -55,6 +55,8 @@ export const useDetour = (routePatternId: RoutePatternId) => {
           }
         }
       )
+    } else {
+      setMissedStops(undefined)
     }
 
     return () => {
--- a/assets/tests/hooks/useDetour.test.ts
+++ b/assets/tests/hooks/useDetour.test.ts
@@ -295,4 +295,25 @@ describe("useDetour", () => {
 
     expect(result.current.missedStops).toBe(missedStops)
   })
+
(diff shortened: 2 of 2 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -55,6 +55,8 @@ export const useDetour = (routePatternId: RoutePatternId) => {
           }
         }
       )
+    } else {
```
joshlarson, line 58: Should this also be gated by `shouldUpdate`? It feels like the odds of a race condition here are pretty low, but not 0?

lemald: `shouldUpdate` is necessary because of the race condition in the promise. This `if` / `else` statement will only be evaluated when the `useEffect` hook re-renders, at which point `shouldUpdate` in the previous promise will have been set to `false` by the cleanup of the old version of the hook. So I think it's fine.

joshlarson: Yeah I wasn't sure if the race condition could also just happen if the browser resolved `useEffect`s in a funny order or something. But maybe that's not how it works? I do imagine you'd have to try in order to make this cause a problem (if it's even possible)

joshlarson (approved):

assets/src/hooks/useDetour.ts
lemald: My understanding is that when a given `useEffect` hook re-renders, you at least know that it has already run the cleanup for the previous iteration.

PR #2437, opened 2024-02-28
 feat(ts/hooks/useDetour): move detour finished state into useDetour hook

firestack:
---
  - https://app.asana.com/0/0/1206710026837334

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -6,7 +6,7 @@ import React, {
 } from "react"
 import { DiversionPanel } from "./diversionPanel"
 import { DetourMap } from "./detourMap"
-import { useDetour } from "../../hooks/useDetour"
+import { DetourState, useDetour } from "../../hooks/useDetour"
 import { Alert, CloseButton } from "react-bootstrap"
 import * as BsIcons from "../../helpers/bsIcons"
 import { OriginalRoute } from "../../models/detour"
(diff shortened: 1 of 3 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -40,6 +41,8 @@ export const DiversionPage = ({
     canUndo,
     undo,
     clear,
+    finishDetour,
+    editDetour,
```
joshlarson, line 45: **Non-blocking:** This name confused me for a moment - I thought that `editDetour` was a function that places waypoints or whatever, not one that returns to the edit screen. Maybe ... `returnToEditScreen` or something like that? Or `unfinishDetour`?

joshlarson (approved):

assets/src/components/detours/diversionPage.tsx
firestack: I usually don't like the style of untyped/string event messages (like `send({type: "finish-detour"})`), but I'm imagining us ending up with a state machine in here at some point, so I modeled these names after what I thought the transistion event name may be called. I agree that it's a bit difficult to understand what it's going to do from name alone. Given that, I could be convinced, but I'm of the opinion that the state machine should not know anything about what screen is showing based on the state. So I could locally alias this name to the following
```suggestion
    editDetour: returnToEditScreen,
```

But I would assume that's not a suitable answer.

I don't really like `unfinishDetour`. I did consider moving all functions under a `fn` object or _something_, but that felt a bit too far. I could sequester the transistion functions under something like a `send` object?
so `send.finishDetour` and `send.editDetour`?

firestack: Open to other answers though. I just have a preference for keeping the function name similar to what the event name for the transistion would look like.

PR #2438, opened 2024-02-28
feat: Add algorithm for segmenting a route based on connection points

joshlarson:


lib/skate/detours/route_segments.ex
```diff
@@ -0,0 +1,68 @@
+defmodule Skate.Detours.RouteSegments do
+  @moduledoc """
+  Break a route into segments based on `onnection_start` and `connection_end` points
```
lemald, line 3: **nitpick:**
```suggestion
  Break a route into segments based on `connection_start` and `connection_end` points
```

lib/skate/detours/route_segments.ex
```diff
@@ -0,0 +1,68 @@
...
+  defmodule Params do
+    @moduledoc false
+    @type t :: %__MODULE__{
+            connection_start: Util.Location.From.t(),
+            connection_end: Util.Location.From.t(),
+            shape: nonempty_list(Util.Location.From.t())
+          }
+    @enforce_keys [:connection_start, :connection_end, :shape]
+    defstruct [:connection_start, :connection_end, :shape]
+  end
```
lemald, line 15: **suggestion:** given that all three of these are required, I think it would be more idiomatic to just define `route_segments` as a function that takes 3 arguments rather than 1 and not define an entirely separate `Params` struct.

lib/skate/detours/route_segments.ex
```diff
@@ -0,0 +1,68 @@
...
+        detour: [%Util.Location{latitude: 0, longitude: 1}, %Util.Location{latitude: 0, longitude: 2}],
+        after_detour: [%Util.Location{latitude: 0, longitude: 2}, %Util.Location{latitude: 0, longitude: 3}]
+      }
+  """
+  @spec route_segments(Params.t()) :: Result.t()
+  def route_segments(%__MODULE__.Params{
+        connection_start: connection_start,
+        connection_end: connection_end,
+        shape: shape
+      }) do
```
lemald, line 51: **suggestion:** if you include a guard here I think it would provide a clearer and more up-front state when attempting to pass in an empty list, rather than having `Enum.min_by` to raise an exception.
```suggestion
      }) when length(shape) > 0 do
```

lib/skate/detours/route_segments.ex
```diff
@@ -0,0 +1,68 @@
...
+
+    {_, end_index} =
+      shape
+      |> Enum.with_index()
+      |> Enum.min_by(fn {point, _index} -> Util.Location.distance(point, connection_end) end)
+
+    %__MODULE__.Result{
+      before_detour: Enum.slice(shape, 0..start_index),
+      detour: Enum.slice(shape, start_index..end_index),
+      after_detour: Enum.slice(shape, end_index..-1)
```
lemald, line 65: **question:** probably not a big concern in practice, but it's probably worth briefly thinking over how this would play out with small lists, or edge cases like when the start and end point share the same closest point on the shape, or when they're out of order. Looking at it so far I don't have immediate concerns about any of those, but probably worth considering.

lemald:
Okay @joshlarson I've completed a first-pass review, I flagged a couple of things one of which is more a general question, but overall it looks good.

lib/skate/detours/route_segments.ex
joshlarson: That's a good call. Even though I don't think the logic will change, I'll add test cases for those use cases, so that it's super explicit what happens in those cases.

lemald: Sounds good to me!

lib/skate/detours/route_segments.ex
joshlarson: I think the way I did it is the only way to make the lexical order of `shape`, `connection_start`, and `connection_end` not matter. If I used a keyword list, then they have to be in the right order (which makes keyword lists super annoying to work with! 😠), and with just a regular arg list, they still have to be in the right order and it's even harder to remember.

I don't feel that strongly about it though, so happy to do whatever makes sense.

lemald: There are plenty of arity-3 functions so I don't think it's particularly hard to remember (I'd feel differently if we were talking 5+ arguments or something). Also, thanks to `@doc` and `@spec` you can get good inline documentation in your editor via LSP when you go to use the function. Plus, using a struct rather than individual arguments makes it harder to use in pipelines.

joshlarson: https://github.com/mbta/skate/pull/2438/commits/434618f81afa36e2eaff1bafade928165ffd6e88

lib/skate/detours/route_segments.ex
joshlarson: Oh I accidentally lumped a few different things together into one commit, but https://github.com/mbta/skate/pull/2438/commits/434618f81afa36e2eaff1bafade928165ffd6e88 should address this

lemald (approved):
👍

PR #2439, opened 2024-02-28
feat: Add DetoursController, which steals the logic from the missed stops controller

joshlarson:


lemald (approved):
I feel like we could still find a better name for the endpoint, but I don't have any ideas off the top of my head and I don't want to waste time bikeshedding it. Maybe as we actually add more functionality a better name will suggest itself.

PR #2445, opened 2024-02-29
init: lockout controls initial work

firestack:
This does the bare minimum to get the controls locked out in the finished state and get us started down the path of conditionally returning functions from `useDetour`. I'm expecting to open another PR which refactors the type that `useDetour` returns and collapse duplicated tests (e.g., tests with "does nothing" conditions, and removing "permission" booleans like `canUndo`).

### DIFF
--- a/assets/tests/hooks/useDetour.test.ts
+++ b/assets/tests/hooks/useDetour.test.ts
@@ -289,7 +289,7 @@ describe("useDetour", () => {
   test("when `startPoint` is set, `canAddPoints` is `true`", async () => {
     const { result } = renderHook(useDetourWithFakeRoutePattern)
 
-    act(() => result.current.addConnectionPoint({ lat: 0, lon: 0 }))
+    act(() => result.current.addConnectionPoint?.({ lat: 0, lon: 0 }))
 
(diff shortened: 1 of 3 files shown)


assets/tests/hooks/useDetour.test.ts
```diff
@@ -422,4 +422,49 @@ describe("useDetour", () => {
 
     await waitFor(() => expect(result.current.state).toBe(DetourState.Edit))
   })
+
+  describe("when `state` is `Finished`, controls are locked out", () => {
```
lemald, line 426: **nitpick (non-blocking):** since these tests are all pretty similar maybe you could rewrite them with `test.each`? Not a big deal though because it's only a few of them and there isn't that much repeated code.

firestack: Good point! Normally I'd agree, but there's a few refactors in the works, so I'll keep that in mind when working on them, but for now I think I'm going to leave it as is.

lemald: Totally fair, I marked it non-blocking for a reason. 🙂

lemald (approved):

PR #2447, opened 2024-02-29
fix: Remove useless Params

joshlarson:
Per some [feedback](https://github.com/mbta/skate/pull/2438#discussion_r1507643206) in the PR where this was introduced, I removed the usage of this Params struct, but I forgot to remove the struct itself.


lemald (approved):
Heh, I should have noticed this before approving. 🤦‍♂️

PR #2451, opened 2024-03-01
feat: Read route segment info in api.ts

joshlarson:


assets/src/models/finishedDetour.ts
```diff
@@ -1,9 +1,29 @@
...
 import { StopData, stopsFromData } from "./stopData"
 
 export const FinishedDetourData = type({
   missed_stops: array(StopData),
+  route_segments: type({
+    before_detour: array(
+      type({
+        lat: number(),
+        lon: number(),
+      })
```
lemald, line 12: **suggestion (non-blocking):** since this is repeated a few times you could pull it out into its own non-`export`ed value.

PR #2454, opened 2024-03-01
chore: Upgrade various workflow actions to get rid of Node 16 deprecation warnings

joshlarson:
No ticket.

There's still one deprecation warning in the `mbta/actions/dialyzer` action, but that will involve (I think) upgrading [this](https://github.com/mbta/actions/blob/119bf30dc6bdb9f7fc7264e3814462f397147689/dialyzer/action.yml#L17), which will take a bit more effort, just to make sure we do it right.

[Example of some of the warnings](https://github.com/mbta/skate/actions/runs/8113620719?pr=2451)

Screenshot 2024-03-01 at 11 29 31 AM

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           fetch-depth: 0 # 👈 Required to retrieve git history
       - name: ASDF cache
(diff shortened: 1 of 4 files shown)


lemald (approved):
Thank you!

PR #2455, opened 2024-03-01
feat: Draw original and diverted route segments on the map

joshlarson:
Screenshot 2024-03-04 at 2 35 32 PM

---
  - https://app.asana.com/0/0/1206733203010774

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -12,6 +12,13 @@
   }
 }
 
+.c-detour_map--original-route-shape-diverted {
+  stroke: $black;
+  stroke-dasharray: 14 28;
+  stroke-linecap: square;
+  opacity: 50%;
+}
+
 .c-detour_map--original-route-shape__unstarted {
   cursor: url("data:image/svg+xml,%3Csvg width='22' height='22' viewBox='0 0 22 22' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cstyle%3E .color %7B fill: %2303A900; %7D %3C/style%3E%3Ccircle cx='11' cy='11' r='10' class='color' style='filter: opacity(0.5) blur(0.5px)'/%3E%3Ccircle cx='11' cy='11' r='7' class='color' stroke='white' stroke-width='2'/%3E%3C/svg%3E")
       11 11,
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -16,7 +16,7 @@
 
 .c-detour_map--original-route-shape-diverted {
   stroke: semantic.$missed-stop;
-  stroke-dasharray: 10 20;
+  stroke-dasharray: 10 10;
   opacity: 50%;
 }
 
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,3 +1,5 @@
+@use "../color/definitions" as semantic;
+
 .c-detour_map--original-route-shape-core {
   stroke: $color-kiwi-500;
   opacity: 50%;
@@ -13,9 +15,8 @@
 }
 
 .c-detour_map--original-route-shape-diverted {
-  stroke: $black;
-  stroke-dasharray: 14 28;
-  stroke-linecap: square;
+  stroke: semantic.$missed-stop;
+  stroke-dasharray: 10 20;
   opacity: 50%;
 }
 
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -14,6 +14,7 @@ import {
 } from "../../util/pointLiterals"
 import { MapTooltip } from "../map/tooltip"
 import { joinClasses } from "../../helpers/dom"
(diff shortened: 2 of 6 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -59,11 +62,13 @@ export const useDetour = (routePatternId: RoutePatternId) => {
         (result) => {
           if (shouldUpdate) {
             setMissedStops(result?.missedStops || undefined)
+            setRouteSegments(result?.routeSegments || undefined)
```
firestack, line 65: suggestion(non-blocking): what do you think about combining the `missedStops` and `routeSegments` states into a single `finishedDetour` state?

assets/src/components/detours/detourMap.tsx
```diff
@@ -254,3 +271,32 @@ const OriginalRouteShape = ({
...
+}
+
+const DivertedRouteShape = ({
+  segments: { beforeDetour, detour, afterDetour },
+}: DivertedRouteShapeProps) => (
+  <>
+    <Polyline
+      weight={6}
+      interactive={false}
+      positions={beforeDetour.map(shapePointToLatLngLiteral)}
```
firestack, line 286: nit(non-blocking): I can't pull this down and check right now, but since there's no referential identity preserved here, do the polylines continuously rerender and have the zoom issue we saw the other time? We may need to memo these if so.

joshlarson: I didn't notice any issue when I was playing around with this, but.... maybe this is like two bugs that cancel each other out?

firestack (approved):
I would _like_ to chat about the open comments, but otherwise LGTM

joshlarson:
> I would like to chat about the open comments, but otherwise LGTM

Agreed! I agree about the finished detour thing - I was contemplating _just doing it_, but then I forgot.

The other thing... we should definitely double-check whether that actually works.

assets/css/detours/_detour_map.scss
```diff
@@ -12,6 +12,13 @@
   }
 }
 
+.c-detour_map--original-route-shape-diverted {
+  stroke: $black;
```
joshlarson, line 16: I think I may have gotten this from the wrong part of the Figma.

joshlarson: ... so I think I need to update this to match the designs in the component section of the Figma, rather than on the map itself.

assets/src/hooks/useDetour.ts
joshlarson: @firestack how does [this](https://github.com/mbta/skate/pull/2455/commits/d5ff1dffbf0d540f1f725dc0c536720aa574044b) look? 

It goes halfway - it uses `finishedDetour` as internal state, but still exposes `missedStops` and `routeSegments` separately. I kind of prefer this, especially with the refactor that you and I were working on a few days ago, but I'm open to discussion about it.

firestack: > but still exposes `missedStops` and `routeSegments` separately

That's exactly what I was intended/thinking. LGTM thanks!

assets/css/detours/_detour_map.scss
joshlarson: [this should fix that](https://github.com/mbta/skate/pull/2455/commits/cd0448cee0609a8b08d9e85a9e108c297f660b73)

assets/css/detours/_detour_map.scss
```diff
@@ -12,6 +14,12 @@
   }
 }
 
+.c-detour_map--original-route-shape-diverted {
+  stroke: semantic.$missed-stop;
```
joshlarson, line 18: @mbta/skate-developers just want to sanity-check that I'm doing this in the way that makes the most sense. I just sort of flailed around until I was able to reference the `missed-stop` color in a way that worked, but maybe there's a better way?

assets/src/components/detours/detourMap.tsx
joshlarson: @firestack I'm not seeing a user-facing issue, but intuitively it feels like this is _wrong_.

Do you think it's worth it to `useMemo` here anyway, just to be on the safe side? (And to avoid making people's browsers do extra work?)

assets/css/detours/_detour_map.scss
firestack: Looks correct to me

assets/src/components/detours/detourMap.tsx
firestack: the `useMemo` docs say to only use it as a performance optimization, so if we're not seeing issues, then I think we're fine? Easy to do later on if it's an issue

assets/css/detours/_detour_map.scss
joshlarson: Allllmost! [this](https://github.com/mbta/skate/pull/2455/commits/53b2368e0e5f149e9588dc21f18e7ca9773a2234) should be the rest of the fix

joshlarson: Okay [there we go](https://github.com/mbta/skate/pull/2455/commits/62e4d0f86061344d7695b0fca0f5a7c1ba731f3f)

PR #2458, opened 2024-03-01
feat: Update detour-shape weight and color to match hi-fi design

joshlarson:
Screenshot 2024-03-04 at 3 38 13 PM

---
  - https://app.asana.com/0/0/1206741046767439

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -27,7 +27,8 @@
 }
 
 .c-detour_map--detour-route-shape {
-  stroke: $color-lemon-500;
+  stroke: $color-eggplant-400;
+  opacity: 50%;
 }
 
 .c-detour_map-circle-marker--start {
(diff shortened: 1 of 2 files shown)


assets/css/detours/_detour_map.scss
```diff
@@ -26,7 +26,7 @@
 }
 
 .c-detour_map--detour-route-shape {
-  stroke: $color-lemon-500;
+  stroke: $color-eggplant-400;
```
lemald, line 29: **suggestion:** I think if you add `stroke-opacity: 0.5` it will look even more like the hi-fi designs, as it is it still looks a bit different from the hi-fi designs for me.

joshlarson: Good catch - thanks!

joshlarson: [this should fix it](https://github.com/mbta/skate/pull/2458/commits/c15d0cf32c43da0cfc778f7ca0192f738f7bb00c)

lemald (approved):

PR #2463, opened 2024-03-04
feat(ts/components/diversionPage): add share screen text stub

firestack:
This gets the information we have into the text box on the share screen. This does not include connection points as that information is only available in the backend right now, I'm intending to open another PR which adds connection points to the API and adds it to this text.

---
  - https://app.asana.com/0/0/1206710026837345

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -2,6 +2,7 @@ import React, {
   ComponentProps,
   ComponentPropsWithoutRef,
   PropsWithChildren,
+  useEffect,
   useState,
 } from "react"
 import { DiversionPanel } from "./diversionPanel"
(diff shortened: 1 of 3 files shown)


assets/src/hooks/useNearestIntersection.ts
lemald, line 1: **question:** are these changes from a previous iteration where you were using intersections as "connection points?" It doesn't appear that this is needed for the rest of the changes as they are currently and I'm not sure that allowing undefined values makes sense to do as a refactor without a specific use case.

firestack: Oh yes! thanks for catching this I completely forgot 😅

assets/src/components/detours/diversionPage.tsx
```diff
@@ -47,6 +48,24 @@ export const DiversionPage = ({
...
+        `${originalRoute.routeName} ${originalRoute.routeDescription} from`,
+        originalRoute.routeOrigin,
+        originalRoute.routeDirection,
+        ,
+        "Turn-by-Turn Directions:",
+        ...(directions?.map((v) => v.instruction) ?? []),
+        ,
+        `Missed Stops (${missedStops?.length}):`,
+        ...(missedStops?.map(({ name }) => name) ?? ["no stops"]),
+      ].join("\n")
```
lemald, line 65: **suggestion (non-blocking):** I think there's enough logic here that it _might_ make sense to put in a private helper function, but I do realize it would have to take quite a few arguments so it might be kind of unwieldy.

lemald (approved):
Just one suggestion left but it's non-blocking.

assets/src/components/detours/diversionPage.tsx
firestack: I'll keep this in mind for after I add connection points 👍🏻

PR #2464, opened 2024-03-04
general fixes for routing and instructions

firestack:


lib/skate/open_route_service_api/directions_request.ex
```diff
@@ -8,8 +8,9 @@ defmodule Skate.OpenRouteServiceAPI.DirectionsRequest do
     Type that represents a request made to OpenRouteService's Directions API
   """
   @type t() :: %__MODULE__{
-          coordinates: [[float()]]
+          coordinates: [[float()]],
+          continue_straight: boolean()
         }
 
-  defstruct coordinates: []
+  defstruct coordinates: [], continue_straight: true
```
joshlarson, line 15: This looks like it's not being used. Can we remove it?

firestack: Nope! this is intentional as it sets `continue_straight: true` as the default value.

lib/skate/open_route_service_api.ex
```diff
@@ -86,7 +86,9 @@ defmodule Skate.OpenRouteServiceAPI do
        directions:
          segments
          |> Enum.flat_map(& &1["steps"])
-         |> Enum.filter(fn %{"type" => type} -> map_type(type) not in [:goal, :error] end)
```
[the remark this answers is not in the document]
firestack: I'll look into it 👍🏻 A little unsure since we don't have tests for this at all it seems

lib/skate/open_route_service_api/directions_request.ex
firestack: I did this to set this as the default any time this struct is created. So this is technically doing _something_ and is used. That said, I _could_ be convinced that we should instead be putting this explicitly in the code instead of what is essentially, a constructor, but I prefer this for the simplicity right now.

joshlarson: I meant that the field itself isn't being referenced anywhere, unless I missed something? Nobody is ever referencing `continue_straight`.

firestack: Here's _a_ place it's being created(via defaults) & consumed https://github.com/mbta/skate/blob/b314c701d61726776d07cbecedcccc56235ca4aa/lib/skate/open_route_service_api.ex#L57-L62

joshlarson: Talked offline - I get it now! `continue_straight` is a field that's being serialized by the ORS client into a JSON blob as part of the request for directions. 

Thanks for the explanation!

firestack: I _could_ add tests for this, but I'm not sure it's really worth it

```diff
diff --git a/test/skate/open_route_service_api_test.exs b/test/skate/open_route_service_api_test.exs
index 0ed37645..6a44aefc 100644
--- a/test/skate/open_route_service_api_test.exs
+++ b/test/skate/open_route_service_api_test.exs
@@ -132,6 +132,21 @@ defmodule Skate.OpenRouteServiceAPITest do
                %{"lat" => 0, "lon" => 0},
                %{"lat" => 0, "lon" => 1}
              ])
+  end
 
+  test "is called with `continue_straight`" do
+    expect(
+      Skate.OpenRouteServiceAPI.MockClient,
+      :get_directions,
+      fn %DirectionsRequest{continue_straight: true} ->
+        {:ok, build(:ors_directions_json)}
+      end
+    )
+
+    assert {:ok, %DirectionsResponse{}} =
+             Skate.OpenRouteServiceAPI.directions([
+               %{"lat" => 0, "lon" => 0},
+               %{"lat" => 0, "lon" => 1}
+             ])
   end
 end
```

lemald (approved):
Tests look good to me!

lib/skate/open_route_service_api/directions_request.ex
joshlarson: What about something like this? 👇 

```diff
modified   test/skate_web/controllers/detour_route_controller_test.exs
@@ -35,6 +35,20 @@ defmodule SkateWeb.DetourRouteControllerTest do
                json_response(conn, 200)
     end

+    @tag :authenticated
+    test "formats request correctly", %{conn: conn} do
+      expect(Skate.OpenRouteServiceAPI.MockClient, :get_directions, fn args ->
+        assert args.coordinates == [[0, 0], [1, 1]]
+        assert args.continue_straight == true
+
+        {:ok, build(:ors_directions_json)}
+      end)
+
+      post(conn, ~p"/api/detours/directions",
+        coordinates: [%{"lat" => 0, "lon" => 0}, %{"lat" => 1, "lon" => 1}]
+      )
+    end
+
     @tag :authenticated
     test "returns directions as a flat list", %{conn: conn} do
       expect(Skate.OpenRouteServiceAPI.MockClient, :get_directions, fn _ ->
```

firestack: 15c5d4994dd5cd41090ea6a83d6e6e739b2ad81e

joshlarson (approved):

PR #2465, opened 2024-03-04
feat: add jest linter

firestack:
I almost committed a `test.only` in #2463 and was surprised that the linter didn't prevent me from doing so. Commits should be reviewable in order and auditable via `npm run lint:check` at each point in time

### DIFF
--- a/assets/tests/components/propertiesList.test.tsx
+++ b/assets/tests/components/propertiesList.test.tsx
@@ -213,26 +213,9 @@ describe("vehicleProperties", () => {
   })
 
   test("operator information gives last name if that's all that's available", () => {
-    const vehicleSansOperator = {
-      ...vehicle,
-      operatorId: "1234",
-      operatorFirstName: null,
-      operatorLastName: "SMITH",
-    } as unknown
-
-    const properties = vehicleProperties(
-      vehicleSansOperator as VehicleInScheduledService
-    )
-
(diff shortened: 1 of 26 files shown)


assets/tests/components/propertiesList.test.tsx
```diff
@@ -212,27 +212,10 @@ describe("vehicleProperties", () => {
...
-    expect(properties.find((prop) => prop.label === "Operator")!.value).toEqual(
-      "SMITH #1234"
-    )
-  })
-
   test("operator information gives last name if that's all that's available", () => {
     const properties = vehicleProperties(
       vehicleFactory.build({
-        operatorFirstName: "JOHN",
+        operatorFirstName: null,
```
lemald, line 218: I think you want to keep both tests, revert this particular change, and rename this one to something like "only includes operator last name when`operatorLastNameOnly` is set" because what this is doing is testing the argument in the second position of the `vehicleProperties` function.

lemald: I suspect I introduced this with some bad copy-pasting in #1887, and looking at that PR may help piece together what the original intent was.

firestack: does this work? 0748b8250cfb0b353e726137cac22d5d334f43dd

lemald: Yes, perfect.

lemald (approved):

PR #2466, opened 2024-03-04
feat: Add cursor for placing a waypoint

joshlarson:
Cursor in action:

Screenshot 2024-03-04 at 3 36 08 PM

---
  - https://app.asana.com/0/0/1206759503438416

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,6 +1,16 @@
 @use "../color/definitions" as semantic;
 @use "../color/tokens_2024" as tokens;
 
+.c-detour_map--map__clickable {
+  cursor: url("data:image/svg+xml,%3Csvg width='14' height='14' viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cstyle%3E .color %7B fill: %236D39AC; %7D %3C/style%3E%3Ccircle cx='7' cy='7' r='6' class='color' style='filter: blur(0.5px)'/%3E%3Ccircle cx='7' cy='7' r='4.5' class='color' stroke='white'/%3E%3C/svg%3E")
+      11 11,
+    pointer;
+
+  .c-vehicle-map {
+    cursor: unset;
+  }
+}
+
 .c-detour_map--original-route-shape-core {
   stroke: $color-kiwi-500;
   opacity: 50%;
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -56,7 +56,7 @@ interface DetourMapProps {
    * Callback fired when the map is clicked.
    * @param point
    */
-  onClickMap: (point: ShapePoint) => void
+  onClickMap?: (point: ShapePoint) => void
 
   /**
    * User signal to describe the state of the undo button.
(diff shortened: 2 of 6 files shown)


assets/src/components/detours/detourMap.tsx
```diff
@@ -119,7 +127,7 @@ export const DetourMap = ({
 
       <MapEvents
         click={(e) => {
-          onClickMap(latLngLiteralToShapePoint(e.latlng))
+          onClickMap && onClickMap(latLngLiteralToShapePoint(e.latlng))
```
firestack, line 130:
```suggestion
          onClickMap?.(latLngLiteralToShapePoint(e.latlng))
```

assets/src/components/map.tsx
```diff
@@ -106,11 +107,33 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
...
+      for (const [className, shouldBePresent] of Object.entries(
+        props.extraClasses
+      )) {
+        if (shouldBePresent && !classList.contains(className)) {
+          classList.add(className)
+        }
+        if (!shouldBePresent && classList.contains(className)) {
+          classList.remove(className)
+        }
+      }
```
firestack, line 122:
```suggestion
      Object.entries(props.extraClasses)
        .map(([className, shouldBePresent]) => 
          classList.toggle(className, shouldBePresent))
```

joshlarson: That does look like a much simpler way of doing that - thanks!

assets/src/components/map.tsx
```diff
@@ -106,11 +107,33 @@ const Map = (props: Props): ReactElement<HTMLDivElement> => {
...
+        .filter(([_className, shouldBePresent]) => shouldBePresent)
+        .map(([className, _shouldBePresent]) => className)
+    : []
+
   return (
     <>
       <div className={stateClasses} />
       <MapContainer
-        className="c-vehicle-map"
+        className={joinClasses(["c-vehicle-map", ...classNames])}
```
firestack, line 129: question: why are you using both `className` and `element.classList`?

joshlarson: `className` here gives it the right classes when it first loads. `classList` lets it change those classes interactively in response to events.

User-facing effects:
- If I only use `className` and don't use `classList`, then when a user clicks to start a detour, the new fancy cursor doesn't show up.
- If I only use `classList` and don't use `className`, then it doesn't render correctly in storybook (and wouldn't render correctly if we ever started out with a detour in a partially-created state, say, after someone navigated back to an auto-saved draft)

firestack: would you be open to/could you use `stateClasses` or wrap `<Map/>` in `<DetourMap/>` with a `div` which has these classes?

joshlarson: Honestly yes - I don't love the solution I came up with here.

When I wrapped the `<Map />` in `<DetourMap />` in a div, the whole map disappeared, I think because it got a size of zero. That's easy to fix (I think), but when I saw that I got scared and ran the other way.

No idea what `stateClasses` are; happy to try them out though!

firestack: > When I wrapped the `<Map />` in `<DetourMap />` in a div, the whole map disappeared, I think because it got a size of zero. That's easy to fix (I think), but when I saw that I got scared and ran the other way.

I _think_ if you added `h-100` to it, it'd not be size zero anymore


> No idea what `stateClasses` are; happy to try them out though!

It's a prop on `<Map/>` which adds classes on a `div` right next to the map. See https://github.com/mbta/skate/blob/main/assets/css/_leaflet.scss#L9-L10 for existing usages via sibling selectors

joshlarson: I see I see!

[How does this look?](https://github.com/mbta/skate/pull/2466/commits/9510fdf56d6854c85df1a51a17ab31bc292ab241)

assets/src/components/detours/detourMap.tsx
```diff
@@ -100,69 +100,78 @@ export const DetourMap = ({
...
-          Undo
-        </Button>
-        <Button variant="primary" disabled={undoDisabled} onClick={onClear}>
-          Clear
-        </Button>
-      </CustomControl>
+    <div
+      className={joinClasses([
+        "h-100",
+        onClickMap ? "c-detour_map--map__clickable" : null,
```
firestack, line 106: nit(non-blocking)
```suggestion
        onClickMap && "c-detour_map--map__clickable",
```

assets/css/detours/_detour_map.scss
```diff
@@ -1,5 +1,15 @@
 @use "../color/definitions" as semantic;
 
+.c-detour_map--map__clickable {
+  cursor: url("data:image/svg+xml,%3Csvg width='14' height='14' viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cstyle%3E .color %7B fill: %236D39AC; %7D %3C/style%3E%3Ccircle cx='7' cy='7' r='6' class='color' style='filter: blur(0.5px)'/%3E%3Ccircle cx='7' cy='7' r='4.5' class='color' stroke='white'/%3E%3C/svg%3E")
+      11 11,
+    pointer;
+
+  .c-vehicle-map {
+    cursor: unset;
+  }
```
firestack, line 10: question: what is un-setting the cursor doing here? It seems like it'd overwrite the cursor property on ln4

firestack (commented):
One question but otherwise LGTM!

assets/src/components/map.tsx
firestack: LGTM!

assets/css/detours/_detour_map.scss
firestack: Ah! I was mistaking `unset` for `revert`, thanks for explaining!!

firestack (approved):

PR #2468, opened 2024-03-04
feat: Update start, end, and way- point markers on the detour map

joshlarson:
Screenshot 2024-03-04 at 3 39 19 PM

---
  - https://app.asana.com/0/0/1206759503438415

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -1,4 +1,5 @@
 @use "../color/definitions" as semantic;
+@use "../color/tokens_2024" as tokens;
 
 .c-detour_map--original-route-shape-core {
   stroke: $color-kiwi-500;
@@ -48,6 +49,6 @@
 }
 
 .c-detour_map-circle-marker--detour-point {
-  stroke: $color-eggplant-500;
+  stroke: tokens.$eggplant-500;
   fill: white;
 }
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -39,8 +39,8 @@
 }
 
 .c-detour_map-circle-marker--start {
-  stroke: #03a900;
-  fill: #03a900;
+  stroke: tokens.$detour-start-point;
+  fill: tokens.$detour-start-point;
 }
 
 .c-detour_map-circle-marker--end {
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -38,11 +38,16 @@
 }
 
 .c-detour_map-circle-marker--start {
-  stroke: $color-kiwi-400;
-  fill: $color-kiwi-400;
+  stroke: #03a900;
+  fill: #03a900;
 }
 
 .c-detour_map-circle-marker--end {
-  stroke: $color-strawberry-400;
-  fill: $color-strawberry-400;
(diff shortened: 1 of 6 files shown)


assets/css/detours/_detour_map.scss
```diff
@@ -31,11 +31,16 @@
 }
 
 .c-detour_map-circle-marker--start {
-  stroke: $color-kiwi-400;
-  fill: $color-kiwi-400;
+  stroke: #03a900;
+  fill: #03a900;
```
lemald, line 35: **suggestion (non-blocking):** I wonder if it would be possible to find a home for this in `_tokens_2024.scss`. Perhaps it could be sort of like `$missed-stop`, except it doesn't need to be added to the Bootstrap theme colors. But, this is probably a genuine one-off so if there's no good place for it I wouldn't sweat it.

assets/src/components/detours/detourMap.tsx
```diff
@@ -177,61 +177,84 @@ const MapEvents = (props: Leaflet.LeafletEventHandlerFnMap) => {
...
   />
 )
 
-const EndMarker = ({ position }: { position: LatLngLiteral }) => (
-  <StartOrEndMarker classSuffix="end" title="Detour End" position={position} />
+export const EndMarker = ({ position }: { position: LatLngLiteral }) => (
+  <StartOrEndMarker
+    title="Detour End"
+    position={position}
+    icon={<StartOrEndIcon classSuffix={"end"} />}
```
lemald, line 192: **suggestion:**
```suggestion
    icon={<EndIcon />}
```

lemald:
@joshlarson I've left a couple of comments, but everything else looks 👍 to me.

assets/css/detours/_detour_map.scss
```diff
@@ -31,11 +31,16 @@
...
 
 .c-detour_map-circle-marker--end {
-  stroke: $color-strawberry-400;
-  fill: $color-strawberry-400;
+  stroke: $color-strawberry-700;
+  fill: $color-strawberry-700;
+}
+
+.c-detour_map-circle-marker--detour-point {
+  stroke: $color-eggplant-500;
```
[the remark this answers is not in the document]
joshlarson: ooh I probably want to do all of those, don't I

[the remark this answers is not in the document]
joshlarson: Okay this makes sense to me. The strawberry colors did not change, but the eggplant ones did.

I think that means that I goofed the detour route shape color as well.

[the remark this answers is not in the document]
joshlarson: [here we go](https://github.com/mbta/skate/pull/2468/commits/4bc7c0f78d973bbd9c2b9778ec7ce497590b0c33)

assets/css/detours/_detour_map.scss
joshlarson: [how does this look?](https://github.com/mbta/skate/pull/2468/commits/253edeb35ff96c08ad05317757be14579a0086ab)

I agree that this is a good change, since things that are one-offs often don't stay one-offs 😅 I just didn't know the best way to do it!

lemald: Yeah, I'm not sure this is the best long-term solution if that list grows but I think for now it's a good starting point.

lemald (approved):

PR #2469, opened 2024-03-04
feat: map button components

lemald:
Asana ticket: [Create underlying button components](https://app.asana.com/0/1148853526253420/1206753759080074/f)

The main things I'm looking for feedback on here are:
1. The use of the `title` prop for the text in the button at large sizes and both how that should ultimately render in the HTML and how it should be exposed to users of the component.
2. The `size` property and whether or not the way it's used is weird (though nobody will be setting it explicitly to `undefined` except in Storybook so I don't think it's much of an issue).
3. The way I actually customized the Bootstrap buttons via overriding CSS variables in `_map_button.scss` wherever possible.

Also I would like to add at least a couple of basic automated tests, I think those will mostly be around the functionality for question 1 above.

### DIFF
--- a/assets/css/map/controls/_map_button.scss
+++ b/assets/css/map/controls/_map_button.scss
@@ -1,0 +1,30 @@
+@use "../../color/tokens_2024" as tokens;
+
+.c-map-button {
+  box-shadow: 0 0 0 2px rgb($black, 0.15);
+  --bs-btn-border-radius: 0.2rem;
+  --bs-btn-color: #{$black};
+  --bs-btn-hover-color: #{$black};
+  --bs-btn-hover-bg: #{tokens.$eggplant-50};
+  --bs-btn-active-color: #{tokens.$eggplant-500};
+  --bs-btn-active-bg: #{$white};
+  --bs-btn-focus-box-shadow: 0 0 0 2px #{rgb(tokens.$eggplant-500, 0.5)};
+  --bs-btn-disabled-color: #{tokens.$gray-700};
+  --bs-btn-disabled-bg: #{tokens.$gray-100};
+  --bs-btn-disabled-opacity: 1;
+}
+
+.c-map-button--s {
+  width: 2rem;
+  height: 2rem;
+}
+
+.c-map-button--m {
+  width: 2.375rem;
+  height: 2.375rem;
+}
+
+.c-map-button--l {
+  width: 3rem;
+  height: 3rem;
+}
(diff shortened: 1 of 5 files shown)


firestack (approved):
So far this looks good, I'm still mulling it over on the CSS side. I think the `title` prop is fine for now, it's not an issue and we can revisit later if needed.

assets/css/map/controls/_map_button.scss
```diff
@@ -0,0 +1,30 @@
...
+  --bs-btn-color: #{$black};
+  --bs-btn-bg: #{$white};
+  --bs-btn-hover-color: #{$black};
+  --bs-btn-hover-bg: #{tokens.$eggplant-50};
+  --bs-btn-active-color: #{tokens.$eggplant-500};
+  --bs-btn-active-bg: #{$white};
+  --bs-btn-focus-box-shadow: 0 0 0 2px #{rgb(tokens.$eggplant-500, 0.5)};
+  --bs-btn-disabled-color: #{tokens.$gray-700};
+  --bs-btn-disabled-bg: #{tokens.$gray-100};
+  --bs-btn-disabled-opacity: 1;
```
firestack, line 16: note(non-blocking): Just so we're on the same page, what I'm mulling over, is whether we should be using css variables here or scss variables here. And additionally, whether tokens is the right level to pull from.

I'm not sure if I have an opinion with any good reasons yet though.

firestack: But, I'm okay with this being approved and merged since It Works™️

lemald: > note(non-blocking): Just so we're on the same page, what I'm mulling over, is whether we should be using css variables here or scss variables here. And additionally, whether tokens is the right level to pull from.
> 
> I'm not sure if I have an opinion with any good reasons yet though.

Basically, yes - though I'm not sure if scss variables would really work because I'm not sure how we would localize the change to just the map button as opposed to all buttons. I was mainly wondering if there was some other option that I had missed, or if it would be better to just write our own CSS for this. But, if you're okay with this I think it works well so I'll go ahead and merge!

lemald:
I'm going to merge this so that I can move on to another branch to integrate the buttons with the detour page, but I'll put up some design review tickets with links to Storybook so that design can take a look.

assets/css/map/controls/_map_button.scss
firestack: Oh the scss variables vs css variables I was referring to was
```
  --bs-btn-hover-bg: #{tokens.$eggplant-50};
```
and something like this instead
```
  --bs-btn-hover-bg: var(--color-eggplant-50);
```


Part of me wishes `tokens.$eggplant-50 === "var(--color-eggplant-50)"` so we didn't even have to think about it and we'd get the "Good DX™️" for "free"

PR #2473, opened 2024-03-05
feat: add connection points to finished detour panel

firestack:
This is a little on the larger size. I can break this up into separate frontend and backend PR's if that'd help but this should be reviewable by commit to help with that.

---
  - https://app.asana.com/0/0/1206694541310034

### DIFF
--- a/assets/tests/api.test.ts
+++ b/assets/tests/api.test.ts
@@ -400,7 +400,7 @@ describe("fetchShapeForRoute", () => {
 })
 
 describe("fetchFinishedDetour", () => {
-  test("fetches missed stops in finished detour", () => {
+  test("fetches a finished detour given a Route Pattern and connection points", () => {
     const stopData = stopDataFactory.buildList(3)
     const [connection_stop_start, connection_stop_end] =
       stopDataFactory.buildList(2)
--- a/assets/tests/api.test.ts
+++ b/assets/tests/api.test.ts
@@ -34,7 +34,12 @@ import routeTabFactory from "./factories/routeTab"
 import stopFactory from "./factories/stop"
 import * as browser from "../src/models/browser"
 import { string, unknown } from "superstruct"
-import { LocationType, RouteType, stopsFromData } from "../src/models/stopData"
+import {
+  LocationType,
+  RouteType,
+  stopFromData,
+  stopsFromData,
+} from "../src/models/stopData"
 import * as Sentry from "@sentry/react"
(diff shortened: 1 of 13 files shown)


assets/tests/api.test.ts
```diff
@@ -397,6 +402,50 @@ describe("fetchShapeForRoute", () => {
 describe("fetchFinishedDetour", () => {
   test("fetches missed stops in finished detour", () => {
     const stopData = stopDataFactory.buildList(3)
+    const [connection_stop_start, connection_stop_end] =
```
joshlarson, line 405: **Nit (non-blocking)**: I goofed the last time I edited this, but it might be good to update the test name above to reflect that we care about more than just missed stops. I would almost say that the test should just be `test("fetches all relevant fields in finished detour", /* ... */)` or something like that.

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -45,6 +49,57 @@ defmodule SkateWeb.DetoursControllerTest do
       assert %{"data" => %{"missed_stops" => [^expected_missed_stop]}} = json_response(conn, 200)
     end
 
+    @tag :authenticated
+    test "returns connection points", %{conn: conn} do
```
joshlarson, line 53: It sort of worries me that there's no test validating that this works when `connection_start`/`connection_end` are `nil`, but I did test it by hand (at least for `connection_end`) and it worked just fine.... and also the test itself could be kind of annoyingly repetitive. 

I'm kind of on the fence about this whether I would even add a test for this, because of ☝️ points, so I'm definitely not going to block approval on it! 😅

test/skate/detours/missed_stops_test.exs
```diff
@@ -37,9 +38,100 @@ defmodule Skate.Detours.MissedStopsTest do
         ]
       }
 
+      assert %MissedStops.Result{missed_stops: missed_stops} =
```
joshlarson, line 41: **Nit (non-blocking)**: Could this assertion and the one below it be combined into a single assertion?

joshlarson (approved):
I made a few comments, but they're all non-blocking. This looks great!

test/skate_web/controllers/detours_controller_test.exs
firestack: Going to followup with another PR to add test coverage for this

PR #2476, opened 2024-03-06
feat: show stop markers on detour map

joshlarson:


assets/src/components/detours/diversionPage.tsx
```diff
@@ -124,7 +124,8 @@ export const DiversionPage = ({
             undoDisabled={canUndo === false}
             onUndo={undo ?? (() => {})}
             onClear={clear ?? (() => {})}
-          />
+            stops={originalRoute.shape.stops || []}
+          ></DetourMap>
```
lemald, line 128: **suggestion:** you can keep this is a self-closing tag since it doesn't have any children.

joshlarson: [Good call, thanks!](https://github.com/mbta/skate/pull/2476/commits/c32481eebf2bcf0c7faf9f0d0f3ca25534a43002)

PR #2478, opened 2024-03-06
refactor: use react-bootstrap components for LayersControl

lemald:
There's no ticket for this, but given that the layers control was on my brain as we're about to add it to the detour map I decided to quickly revisit it to refactor it to use react-bootstrap components. The only change to the markup was that I didn't carry over the `position-relative` and `stretch-link` classes that were used on the layers radio buttons. These were introduced in #2242 but now when I remove them it doesn't appear to make a visual or functional difference, but let me know if you can think of a reason why they might still be needed and I'll add them back.

### DIFF
--- a/assets/src/components/map/controls/layersControl.tsx
+++ b/assets/src/components/map/controls/layersControl.tsx
@@ -149,18 +149,19 @@ const VehicleLayerOptions = ({
   return (
     <div className="c-layers-control__vehicle_layers_control">
       <h2 id={sectionLabelId}>Vehicles</h2>
-      <Form.Check
-        type="switch"
-        id={formCheckId}
-        checked={pullbackLayerEnabled}
-        onChange={onTogglePullbackLayer}
-        onKeyDown={(event) => {
-          if (event.key === "Enter") {
-            onTogglePullbackLayer?.()
-          }
-        }}
-        label="Show pull-backs"
-      />
+      <Form.Check type="switch" id={formCheckId}>
+        <Form.Check.Input
+          role="switch"
+          checked={pullbackLayerEnabled}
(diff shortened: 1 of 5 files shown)


joshlarson (dismissed):
I'm not thaaaat familiar with the layers control part of the world, but the snapshot diff looks like a highly decorated noop and clicking around, everything still seemed to work, so I think this LGTM

firestack:
> The only change to the markup was that I didn't carry over the `position-relative` and `stretch-link` classes that were used on the layers radio buttons.

I can't check right now, but you should ensure that you can select a radio button by clicking it's "row" and not only the text or the radio icon.

joshlarson:
> I can't check right now, but you should ensure that you can select a radio option by clicking it's "row" and not only the text or the radio icon.

Ahh is that what `stretch-link` does? 

I just checked and selecting in that way is not possible anymore.

firestack:
(just giving implementation detail background while it's on my mind) `position-relative` makes it so the containing row becomes the reference container for any absolutely positioned elements, `stretch-link` then adds a pseudo-class element(::before or ::after I don't remember) which absolutely positions itself to fill the container, so any clicks on the pseudo element bubble up into the `<label>` or `<input>` to register the change.

lemald:
> > I can't check right now, but you should ensure that you can select a radio option by clicking it's "row" and not only the text or the radio icon.
> 
> Ahh is that what `stretch-link` does?
> 
> I just checked and selecting in that way is not possible anymore.

Okay I thought I had tested that out but apparently I hadn't. Please have a look at 4521c6ba and d4fa17b9.

joshlarson (approved):

PR #2479, opened 2024-03-06
refactor: Separate out detour return statements and types by DetourState

joshlarson:


assets/src/components/detours/diversionPage.tsx
```diff
@@ -41,11 +40,27 @@ export const DiversionPage = ({
...
     canUndo,
-    undo,
-    clear,
-    finishDetour,
-    editDetour,
-  } = useDetour(originalRoute.routePatternId)
+  } = detourInfo
+
+  const addWaypoint =
+    detourInfo.state === DetourState.Edit ? detourInfo.addWaypoint : undefined
```
joshlarson, line 46: After chewing on this over the evening, I'm going back and forth on whether I like this approach.

On the one hand, it offends my sensibilities to have a bunch of fields that may or may not be `undefined`.

On the other hand, I'm not sure how much we gain by forcing `diversionPage` (or `DetourMap`) to do this sort of type checking, which creates a lot of syntactic noise and is brittle in its own way, rather than just letting `useDetour` "undefine" the things that it needs to.

I do want to see a move towards things like `addWaypoint` just being defined in the places closest to where it's used, but I'm less convinced than I was earlier that this is the right way to get there.

joshlarson:
Closing this because it's out of date enough that there are a bunch of conflicts with `main` and because I'm not as convinced that this was the right refactor to do here. We might be better off with a collection of smaller refactors instead.

PR #2482, opened 2024-03-07
feat: Don't draw missed stops as regular stops on the detour map

joshlarson:


assets/src/hooks/useDetour.ts
```diff
@@ -38,6 +38,11 @@ const useDetourDirections = (shapePoints: ShapePoint[]) => {
   }
 }
 
+const getVisitedStops = (stops: Stop[], missedStops: Stop[]) => {
+  const missedStopIds = missedStops.map((stop) => stop.id)
```
lemald, line 42: **suggestion (non-blocking):** probably not super relevant given the size of the inputs here, but you could consider using a `Set` since order doesn't matter, just membership.

joshlarson: Yeah probably a good call. I made the same judgment call about size-of-inputs with the intent to see how easy it would be to use a `Set`... and then forgot. Thanks for catching!

joshlarson: [Turns out it was pretty easy!](https://github.com/mbta/skate/pull/2482/commits/3db7d8aab8301b67c19a90963b54dc3ad708988a)

joshlarson:
Made this a draft based on https://github.com/mbta/skate/pull/2484#discussion_r1517908005. I'm experimenting with a different approach that will make this PR obsolete. 

Feel free to hold off on reviewing this.

joshlarson:
Confirmed - different approach is much nicer. Closing this as obsolete

PR #2483, opened 2024-03-07
feat: Add new icons for missed stops

joshlarson:


assets/css/map/markers/_missed_stop_icon.scss
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    .c-missed-stop-icon__line {
+      fill: $color-eggplant-700;
+    }
+  }
+
+  &:focus .c-missed-stop-icon {
+    .c-missed-stop-icon__focus-circle {
+      stroke: $color-eggplant-500;
```
lemald, line 62: **question:** do we want to be using the old 2021 color palette, or the 2024 palette? I think for the detour stuff in general we've been using the 2024 colors because in some places the the 2021 colors look noticeably off.

firestack (approved):
At _some_ point, I'd like to potentially figure out what css between these markers and the stop icons could be shared, but this LGTM as is 👍🏻

lemald:
> At _some_ point, I'd like to potentially figure out what css between these markers and the stop icons could be shared, but this LGTM as is 👍🏻

I haven't looked over it completely to get a sense of how complicated this would be, but I'm also curious how much of the SVG code can be factored out between the two types of stop markers.

lemald (approved):
See my question about colors but I guess that ultimately comes down to what design wants. 👍

assets/css/map/markers/_missed_stop_icon.scss
joshlarson: WUPS! Yeah let's try to use the right color here.

...and maybe for regular stop icons too?

joshlarson: Actually, thinking this through for a moment... our options are:
1. Leave it as-is for now (it may look a bit out of place because of the 2021/2024 colors)
2. Change it only for missed stops (that might look out of place because then regular stops and missed stops will look different)
3. Change it for regular stops AND missed stops (now we're changing existing stuff, which is... probably fine because we want to move towards 2024 colors anyway, but still a bit risky)

Doesn't really seem like we have a great answer here. I think my preference is to do option 1, and have a quick design/dev discussion about whether (and when) we should change both missed stops and regular stops to the 2024 color.

firestack: I like 1, agree that changing existing stuff is not something we should _"just do"_

lemald: Agreed that 1 is fine, 2 would also be fine but either way agreed that a quick design / dev discussion is warranted.

joshlarson: [Asana ticket to follow up on this](https://app.asana.com/0/1148853526253420/1206798268447431/f)

PR #2484, opened 2024-03-07
feat: Draw missed stops on the detour map

joshlarson:
---
  - https://app.asana.com/0/0/1206770478865875

### DIFF
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -25,6 +25,7 @@ import {
 import { TileType } from "../../tilesetUrls"
 import { setTileType } from "../../state/mapLayersState"
 import { StateDispatchContext } from "../../contexts/stateDispatchContext"
+import { uniqBy } from "../../helpers/array"
 
 interface DetourMapProps {
   /**
@@ -39,7 +40,7 @@ interface DetourMapProps {
   /*
    * Stops along the original route shape
    */
-  stops: Stop[]
+  stops: (Stop & { missed: boolean })[]
 
   /**
    * Coordinate to display as the beginning connection point.
@@ -230,12 +231,13 @@ export const DetourMap = ({
         <ZoomLevelWrapper>
           {(zoomLevel: number) => (
             <>
-              {stops.map((stop) => (
+              {uniqBy(stops, (stop) => stop.id).map((stop) => (
                 <StopMarkerWithStopCard
                   key={stop.name}
                   stop={stop}
                   zoomLevel={zoomLevel}
                   interactionStatesDisabled={false}
+                  missed={stop.missed}
                 />
               ))}
             </>
--- a/assets/tests/components/detours/detourMap.test.tsx
+++ b/assets/tests/components/detours/detourMap.test.tsx
@@ -11,6 +11,10 @@ import { layersControlButton } from "../../testHelpers/selectors/components/map"
 import { mockTileUrls } from "../../testHelpers/mockHelpers"
 import { tilesetUrlForType } from "../../../src/tilesetUrls"
 import stopFactory from "../../factories/stop"
+import {
+  missedStopIcon,
+  stopIcon,
+} from "../../testHelpers/selectors/components/map/markers/stopIcon"
 
 beforeEach(() => {
   jest.mocked(tilesetUrlForType).mockReturnValue(undefined)
@@ -235,7 +239,7 @@ describe("DetourMap", () => {
 
     const { container } = render(<DetourMapWithDefaults stops={stops} />)
 
-    expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(3)
+    expect(stopIcon.getAll(container)).toHaveLength(3)
   })
 
   test("stops are drawn as missed stop icons when they are missed", () => {
@@ -247,8 +251,8 @@ describe("DetourMap", () => {
 
     const { container } = render(<DetourMapWithDefaults stops={stops} />)
 
-    expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(1)
-    expect(container.querySelectorAll(".c-missed-stop-icon")).toHaveLength(2)
+    expect(stopIcon.getAll(container)).toHaveLength(1)
+    expect(missedStopIcon.getAll(container)).toHaveLength(2)
   })
 
   test("duplicate stops are only drawn once", () => {
@@ -262,7 +266,7 @@ describe("DetourMap", () => {
 
     const { container } = render(<DetourMapWithDefaults stops={stops} />)
 
-    expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(1)
-    expect(container.querySelectorAll(".c-missed-stop-icon")).toHaveLength(2)
+    expect(stopIcon.getAll(container)).toHaveLength(1)
+    expect(missedStopIcon.getAll(container)).toHaveLength(2)
   })
 })
--- a/assets/tests/components/detours/detourMap.test.tsx
+++ b/assets/tests/components/detours/detourMap.test.tsx
@@ -10,6 +10,7 @@ import { RealDispatchWrapper } from "../../testHelpers/wrappers"
 import { layersControlButton } from "../../testHelpers/selectors/components/map"
 import { mockTileUrls } from "../../testHelpers/mockHelpers"
 import { tilesetUrlForType } from "../../../src/tilesetUrls"
+import stopFactory from "../../factories/stop"
 
 beforeEach(() => {
   jest.mocked(tilesetUrlForType).mockReturnValue(undefined)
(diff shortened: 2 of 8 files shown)


joshlarson:
This needs two things before it can be merged:
1. The build is failing because stop deduplication is done at the `DetourMap` layer, which is possibly not the right place for it.
2. This requires more tests.

joshlarson:
Pulling this back into draft for a little bit because I want to move stop-deduplication into `useDetour`, which means I want to sneak in another PR before this one.

assets/src/components/detours/detourMap.tsx
```diff
@@ -196,6 +198,15 @@ export const DetourMap = ({
...
             ))}
+            {missedStops.map((stop) => (
+              <StopMarkerWithStopCard
+                key={stop.name}
+                stop={stop}
+                zoomLevel={zoomLevel}
+                interactionStatesDisabled={false}
+                missed
+              />
+            ))}
```
firestack, line 209: note(non-blocking): Just pointing this out so we're aware of it, but I imagine we'll get to this at a later point in this epic. If my understanding is correct, the tab order will be somewhat weird because the stops are rendered in two separate chunks, so it'll first tab over all the non-missed stops, and then it'd tab over the missed stops, and I'm not sure that's what we want (but I don't actually know what we want)

We'd probably need to re-merge these stop lists based on stop sequence and render them in order.

joshlarson: Ah yeah that's a good point. 

Do you think it would make sense for `useDetour` to return a collection of things that are like `Stop[]`, but also have a `missed` attribute?

joshlarson: Thinking about it more, this is a good idea, and I think there's an easy way to accomplish this. Stay tuned!

joshlarson: @firestack better approach achieved, and that meant we can get rid of https://github.com/mbta/skate/pull/2482!

assets/src/components/detours/detourMap.tsx
```diff
@@ -32,7 +33,7 @@ interface DetourMapProps {
   /*
    * Stops along the original route shape
    */
-  stops: Stop[]
+  stops: (Stop & { missed: boolean })[]
```
joshlarson, line 36: I'm on the fence about this. On the one hand, I _really_ hate inlined type definitions, since they get complicated and tough to read. 

On the other hand, this isn't actually that complicated, and creating a bespoke type for this (like `DetourMapStop` or something like that), feels a bit like overkill.

assets/tests/hooks/useDetour.test.ts
```diff
@@ -508,18 +508,60 @@ describe("useDetour", () => {
...
+
+      const expectedStops = [
+        { ...stop1, missed: false },
+        { ...stop2, missed: false },
+      ]
+
+      expect(result.current.stops).toStrictEqual(expectedStops)
+    })
+
+    test("when the detour is finished, `stops` excludes any missed stops", async () => {
```
firestack, line 531: note: this test name seems like it may be out of date?

assets/tests/hooks/useDetour.test.ts
```diff
@@ -508,18 +508,60 @@ describe("useDetour", () => {
...
+      const expectedStops = [
+        { ...stop1, missed: false },
+        { ...stop2, missed: true },
+        { ...stop3, missed: true },
+        { ...stop4, missed: false },
+      ]
 
-      expect(result.current.stops).toBe(stops)
+      expect(result.current.stops).toStrictEqual(expectedStops)
     })
```
firestack, line 565: question: is this assertion in the `waitFor` important?
It seems like you could instead do

```suggestion
      const expectedStops = [
        { ...stop1, missed: false },
        { ...stop2, missed: true },
        { ...stop3, missed: true },
        { ...stop4, missed: false },
      ]

      await waitFor(() => {
        expect(result.current.stops).toStrictEqual(expectedStops)
      })      
    })
```

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -437,8 +437,39 @@ describe("DiversionPage", () => {
...
+      />
     )
+
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
```
firestack, line 467: question: do these need to be async?
```suggestion
    await act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })
    await act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })

```

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -437,8 +437,39 @@ describe("DiversionPage", () => {
...
+    waitFor(() => {
+      expect(screen.getByText(stop2.name)).toBeInTheDocument()
+    })
+
+    expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(3)
+    expect(container.querySelectorAll(".c-missed-stop-icon")).toHaveLength(1)
   })
```
firestack, line 474: question: Similar to https://github.com/mbta/skate/pull/2484#discussion_r1520288912 could this be simplified?
```suggestion
    waitFor(() => {
      expect(screen.getByText(stop2.name)).toBeInTheDocument()
      expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(3)
      expect(container.querySelectorAll(".c-missed-stop-icon")).toHaveLength(1)
    })
  })
```

I'm not sure if the first assertion is needed?
```suggestion
    waitFor(() => {
      expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(3)
      expect(container.querySelectorAll(".c-missed-stop-icon")).toHaveLength(1)
    })
  })
```

assets/tests/hooks/useDetour.test.ts
joshlarson: [Woops!](https://github.com/mbta/skate/pull/2484/commits/e8905231c87f0d7ab712c1879eb244ac25c6734b)

assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -224,4 +225,44 @@ describe("DetourMap", () => {
...
+    const stops = [
+      { ...stopFactory.build(), missed: true },
+      duplicateStop,
+      { ...stopFactory.build(), missed: false },
+      duplicateStop,
+    ]
+
+    const { container } = render(<DetourMapWithDefaults stops={stops} />)
+
+    expect(container.querySelectorAll(".c-stop-icon")).toHaveLength(1)
```
firestack, line 265: suggestion: what do you think about wrapping this and the missed stop icon selectors up as a reusable query (`stopIcons.get`?) like https://github.com/mbta/skate/blob/e8905231c87f0d7ab712c1879eb244ac25c6734b/assets/tests/testHelpers/selectors/components/mapPage/map.ts#L10-L15 and putting it next to `stopIcon.get` in https://github.com/mbta/skate/blob/e8905231c87f0d7ab712c1879eb244ac25c6734b/assets/tests/testHelpers/selectors/components/map/markers/stopIcon.ts#L4-L11 ?

firestack: Oh! I just read the return value from [`byRole`](https://github.com/domasx2/testing-library-selector) and I think it'd be better to add this as `stopIcon.getAll`, not `stopIcons.get()`

firestack (commented):
I've left a few comments but otherwise this looks good 👍🏻

assets/tests/components/detours/detourMap.test.tsx
joshlarson: Love it!

Would you recommend doing missed stops as `stopIcon.getAllMissed()` or as `missedStopIcon.getAll()`?

firestack: I'd prefer `missedStopIcon.getAll()`, in the case that we can replace this with a  selector from the `byRole` library in the future, the API would be consistent if we did it this way IMO

joshlarson: Agreed - it also _feels_ more _right_ to do `missedStopIcon.getAll()`.

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: [They do not](https://github.com/mbta/skate/pull/2484/commits/1e65828bb35b13d0fbccc3ac9a5d0217b9ec45a4)

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: [Yeah the first assertion isn't really what this test is about](https://github.com/mbta/skate/pull/2484/commits/c2163793b71d78153585c5fca0155197e64c28b3)

assets/tests/components/detours/detourMap.test.tsx
joshlarson: [This is a lot nicer](https://github.com/mbta/skate/pull/2484/commits/332c63ee924269f3704ee5ececbd597eaf741659)

assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -224,4 +229,44 @@ describe("DetourMap", () => {
...
+  })
+
+  test("duplicate stops are only drawn once", () => {
+    const duplicateStop = { ...stopFactory.build(), missed: true }
+    const stops = [
+      { ...stopFactory.build(), missed: true },
+      duplicateStop,
+      { ...stopFactory.build(), missed: false },
+      duplicateStop,
+    ]
```
firestack, line 265: suggestion(non-blocking): [I thought I posted this already but I can't find it, so sorry if this is a duplicate] I'm not sure if it's important that this test data match real world data, but since we can only have contiguous missed stops right now, maybe this test data should match?


```suggestion
    const duplicateStop = { ...stopFactory.build(), missed: true }
    const stops = [
      duplicateStop,
      { ...stopFactory.build(), missed: true },
      duplicateStop,
      { ...stopFactory.build(), missed: false },
    ]
```

OR


```suggestion
    const duplicateStop = stopFactory.build()
    const stops = [
      { ...stopFactory.build(), missed: true },
      { ...duplicateStop, missed: true },
      { ...stopFactory.build(), missed: false },
      { ...duplicateStop, missed: false },
    ]
```

assets/src/components/detours/detourMap.tsx
firestack: I was also on the fence about this, but it works and it's not that difficult to read, so I think it's fine.

firestack (approved):
LGTM!

PR #2485, opened 2024-03-07
feat: make detour modal receive focus on opening

lemald:
Asana ticket: [⚙️ Have the detour modal grab focus when it renders](https://app.asana.com/0/1148853526253420/1206648035312400/f)

### DIFF
--- a/assets/css/detours/_detour_modal.scss
+++ b/assets/css/detours/_detour_modal.scss
@@ -1,11 +1,11 @@
 .c-detour-modal {
   inset: 0;
   position: fixed;
-  transition: all 0.3s;
   z-index: map-get($z-page-layout-context, "detour-modal");
 }
 
-.c-detour-modal__hidden {
+.c-detour-modal__transition-exit-active {
+  transition: all 0.3s;
   transform: scale(1.02) translate(0px, -50px);
   opacity: 0;
   z-index: map-get($z-page-layout-context, "detour-modal-hidden");
--- a/assets/css/detours/_detour_modal.scss
+++ b/assets/css/detours/_detour_modal.scss
@@ -8,5 +8,4 @@
   transition: all 0.3s;
   transform: scale(1.02) translate(0px, -50px);
   opacity: 0;
-  z-index: map-get($z-page-layout-context, "detour-modal-hidden");
 }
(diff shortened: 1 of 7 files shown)


assets/tests/components/mapPage.test.tsx
```diff
@@ -1913,9 +1913,7 @@ describe("<MapPage />", () => {
 
       await userEvent.keyboard("{Escape}")
 
-      expect(container.querySelector(".c-detour-modal")).toHaveClass(
-        "c-detour-modal__hidden"
-      )
+      expect(container.querySelector(".c-detour-modal")).not.toBeInTheDocument()
```
joshlarson, line 1916: This is great! It really annoyed me that the modal was always in the document, just hidden

assets/css/detours/_detour_modal.scss
```diff
@@ -1,11 +1,11 @@
 .c-detour-modal {
   inset: 0;
   position: fixed;
-  transition: all 0.3s;
   z-index: map-get($z-page-layout-context, "detour-modal");
 }
 
-.c-detour-modal__hidden {
+.c-detour-modal__transition-exit-active {
```
joshlarson, line 7: The exit transition works perfectly! The enter one doesn't seem to be the same as it was before. I was playing around a little bit with the `...-enter`/`-enter-active`/etc classes, and I couldn't get it to work, and it might not be worth the effort, but I just wanted to flag that.

assets/css/detours/_detour_modal.scss
```diff
@@ -1,11 +1,11 @@
...
+  transition: all 0.3s;
   transform: scale(1.02) translate(0px, -50px);
   opacity: 0;
   z-index: map-get($z-page-layout-context, "detour-modal-hidden");
```
joshlarson, line 11: Now that the modal gets removed from the DOM instead of just hidden with 0 opacity, I think we can get rid of this z-index as well (and probably the `detour-modal-hidden` thing from `$z-page-layout-context`)

joshlarson:
This all looks really good! I'm holding off on approving for now because of the enter animation - I feel like there has to be a way to preserve the behavior it had before.

assets/css/detours/_detour_modal.scss
lemald: ea834db3

assets/css/detours/_detour_modal.scss
lemald: Turns out in this case it's actually `appear` and `appear-active` - see ef10b0dc.

joshlarson (approved):

PR #2486, opened 2024-03-08
refactor: Move de-duplication logic into useDetour and out of render components

joshlarson:


assets/tests/hooks/useDetour.test.ts
```diff
@@ -363,13 +385,13 @@ describe("useDetour", () => {
     act(() => result.current.addConnectionPoint?.({ lat: 0, lon: 0 }))
 
     await waitFor(() => {
-      expect(result.current.missedStops).not.toBeUndefined()
+      expect(result.current.missedStops).not.toStrictEqual([])
```
firestack, line 388: I'm surprised the linter didn't catch this, can we use `toHaveLength(0)` instead? https://jestjs.io/docs/next/expect#tohavelengthnumber

(same for a few other instances of this)

joshlarson: Ah yeah - I wanted to use `toBeEmpty()`, but for some reason that claimed not to be a function, even thought Typescript was happy with it.

Will fix.

joshlarson: [Done](https://github.com/mbta/skate/pull/2486/commits/ae850c1973c49100f4c057e6c1474e17263bc806)

firestack: > Ah yeah - I wanted to use `toBeEmpty()`, but for some reason that claimed not to be a function, even thought Typescript was happy with it.

Ah, yeah, that's a testing-library jest-dom matcher for DOM stuff, unfortunately sounds very similar to what you'd expect to use for empty arrays(at least it's been deprecated). You were probably getting the error because _likely_ (I didn't check) you don't have the jest-dom matchers imported.

https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeempty

assets/tests/hooks/useDetour.test.ts
```diff
@@ -519,7 +541,26 @@ describe("useDetour", () => {
...
+        useDetour(
+          originalRouteFactory.build({
+            shape: shapeFactory.build({
+              stops: [stop1, stop2, stop3, stop2, stop4],
+            }),
+          })
+        )
+      )
+
+      expect(result.current.stops).toStrictEqual([stop1, stop2, stop3, stop4])
```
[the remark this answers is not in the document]
joshlarson: Hmmmm...

So if we don't de-dup, then the effect will be that we draw icons over each other. I can't imagine that that's the behavior we want.

Also, and admittedly this is easily solvable, but React yells at us if there are duplicates, because the `key` value in the array under `ZoomLevelWrapper` is duplicated.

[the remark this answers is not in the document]
firestack: > Also, and admittedly this is easily solvable, but React yells at us if there are duplicates, because the `key` value in the array under `ZoomLevelWrapper` is duplicated.

My opinion is that we're doing something wrong by treating `stop_id` as the primary key here.

> So if we don't de-dup, then the effect will be that we draw icons over each other. I can't imagine that that's the behavior we want.

Yeah, I get why we're doing it, and I know we don't have enough information to actually do anything about this right now, so again, this is non-blocking, but I wanted to point this out because this now not only applies to the panel text, but also the map[though I do acknowledge that the map would've failed on duplicated stops anyway, because we're using `stop_id` as the primary key].

firestack (approved):

PR #2491, opened 2024-03-11
fix: make modal and tooltip appear over rest of application

firestack:
By changing the main application container to start it's own stacking context, the modals and tooltips created by react-bootstrap then by default cover the application contained in the `main` element without interference because of the new stacking context created. So now modals and tooltips show up in the order they're defined (as long as they don't have wildly huge z-index's themselves, or we'd need to apply this same isolation to modals as well)

---
  - https://app.asana.com/0/0/1206808449589844

### DIFF
--- a/assets/css/_u_isolation.scss
+++ b/assets/css/_u_isolation.scss
@@ -1,0 +1,3 @@
+.isolation-isolate {
+  isolation: isolate;
+}
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -120,6 +120,7 @@ $vpp-location-padding: 1rem;
 @import "vehicle_route_summary";
 @import "view_header";
 
+@import "u_isolation";
 @import "u_no_overscroll";
 
 // #region Effects
@@ -168,4 +169,8 @@ body {
   position: relative;
   flex: 1 1 0;
   min-height: 0;
+}
+
+main.container {
+  @extend .isolation-isolate;
 }
--- a/assets/css/_z_indexes.scss
+++ b/assets/css/_z_indexes.scss
@@ -4,8 +4,6 @@ $z-page-layout-context: (
   "view": 1100,
   "navbar": 1200,
   "user-info-popover": 1201,
-  "modal-backdrop": 2100,
-  "modal": 2101,
 );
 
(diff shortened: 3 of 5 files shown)


assets/css/_u_isolation.scss
lemald, line 1: **question (non-blocking):** I like the `_u` naming convention for utilities, and I see we already use it for `_u_no_overscroll.scss`, but how would you feel about making a separate `utilities` directory to put these under instead?

firestack: Sure!

lemald (approved):
Looks good! Relatedly, I recently created [a ticket](https://app.asana.com/0/1148853526253420/1206770630583768/f) to standardize on Restart UI for all of the various things called "modals" in Skate, so hopefully that will allow removing more `z-index` properties.

assets/css/_u_isolation.scss
firestack: Thoughts? https://github.com/mbta/skate/pull/2492

assets/css/app.scss
```diff
@@ -169,3 +170,7 @@ body {
   flex: 1 1 0;
   min-height: 0;
 }
+
+main.container {
+  @extend .isolation-isolate;
```
joshlarson, line 175: Question (non-blocking): Why create the utility class rather than just put `isolation: isolate` here?

joshlarson (approved):
Getting rid of the z-indexes makes me sooooooooo happy!

assets/css/app.scss
firestack: I started with this, but I've been needing to create stacking contexts in different parts of Skate so I figured I'd just create it as a utility class. Not entirely attached to the utility class method.

lemald: FWIW I think this goes along nicely with the conventions of the Bootstrap utility classes we've been using more and more, so I'm in favor.

Semi-relatedly, I'm not sure if this is necessarily better, but you could just add this class directly to the `<main>` element in `app.html.heex` rather than using an SCSS mixin. That might break out expectations that styling-related stuff can all be found in the front-end code though.

PR #2492, opened 2024-03-11
 refactor: move css utilities into own folder 

firestack:
https://github.com/mbta/skate/pull/2491#discussion_r1519848626


assets/css/_utilities.scss
```diff
@@ -0,0 +1,4 @@
+@import "utilities/utility_hideable";
```
lemald, line 1: **suggestion:** while making this change, it's probably also worth standardizing on `u_` vs. `utility_` as a prefix, or simply dropping the prefix entirely given that the directory structure should now establish that these are utilities.

firestack: [`ac1d2c9` (#2492)](https://github.com/mbta/skate/pull/2492/commits/ac1d2c9f6e1528f14f3894a4dae1d6c0e55595a1)

lemald (approved):

PR #2495, opened 2024-03-12
refactor: apiCallWithError function

lemald:
Asana ticket: [Version of apiCall that surfaces error states](https://app.asana.com/0/0/1206816132107678/f)

I'm opening this up for review as its own PR just because it's a pretty impactful change for the future of how we do API requests. I have a followup subtask to actually pass this result up through the hooks. I may or may not also do some work on [factoring out the hook logic into a general-purpose helper](https://app.asana.com/0/1148853526253420/1205240803395719/f) if it makes sense to.

### DIFF
--- a/assets/src/util/fetchResult.ts
+++ b/assets/src/util/fetchResult.ts
@@ -19,3 +19,9 @@ export const isOk = <T>(r: FetchResult<T>): r is Ok<T> => "ok" in r
 
 export const isFetchError = <T>(r: FetchResult<T>): r is FetchError =>
   "is_error" in r && r.is_error === true
+
+export const loading = <T>(): FetchResult<T> => ({ is_loading: true })
+
+export const ok = <T>(r: T): FetchResult<T> => ({ ok: r })
+
+export const fetchError = <T>(): FetchResult<T> => ({ is_error: true })
--- a/assets/src/util/fetchResult.ts
+++ b/assets/src/util/fetchResult.ts
@@ -20,8 +20,18 @@ export const isOk = <T>(r: FetchResult<T>): r is Ok<T> => "ok" in r
 export const isFetchError = <T>(r: FetchResult<T>): r is FetchError =>
   "is_error" in r && r.is_error === true
 
+/**
+ * Creates a {@link FetchResult} representing a loading state
+ */
 export const loading = <T>(): FetchResult<T> => ({ is_loading: true })
 
+/**
+ * Creates a {@link FetchResult} representing a success state, along
+ * with loaded data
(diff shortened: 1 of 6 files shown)


lemald:
In the future I think we'll also definitely want the ability to return additional data with an error, but since we don't have an immediate need for that as part of this ticket I'm going to punt on it for now.

assets/tests/api.test.ts
```diff
@@ -230,6 +232,97 @@ describe("checkedApiCall", () => {
...
+    const parse = jest.fn(() => "parsed")
+
+    return apiCallWithError({
+      url: "/",
+      dataStruct: string(),
+      parser: parse,
+    }).then((parsed) => {
+      expect(parse).toHaveBeenCalledWith("raw")
+      expect(parsed).toEqual(ok("parsed"))
+    })
```
firestack, line 260:
```suggestion
    await expect(apiCallWithError({
      url: "/",
      dataStruct: string(),
      parser: parse,
    })).resolves.toEqual(ok("parsed"))
    expect(parse).toHaveBeenCalledWith("raw")
```

lemald: Have a look at d73dfc27 - I also removed one test because I realized we no longer have a default return value.

assets/src/util/fetchResult.ts
```diff
@@ -19,3 +19,9 @@ export const isOk = <T>(r: FetchResult<T>): r is Ok<T> => "ok" in r
 
 export const isFetchError = <T>(r: FetchResult<T>): r is FetchError =>
   "is_error" in r && r.is_error === true
+
+export const loading = <T>(): FetchResult<T> => ({ is_loading: true })
```
joshlarson, line 23:
```suggestion
/*
 * Constructor functions to enable a user to create FetchResult's easily
 * For instance, something like:
 *   const result = loading()
 *   const result = ok(successfulData)
 */
export const loading = <T>(): FetchResult<T> => ({ is_loading: true })
```

Nit (non-blocking): It took me a minute to figure out what these were. An explanatory comment might be helpful

lemald: How about 22d0737a? It's in JSDoc format so it provides inline documentation in editors and stuff.

joshlarson: Even better! I've never fully understood the JSDoc thing, but that's pretty neat!

firestack (approved):
I don't have anymore feedback and this seems good as it is now. I figure if anything comes to mind later we can just fix it in another PR.

PR #2496, opened 2024-03-12
feat: Drop start and end points exactly where the cursor is

joshlarson:


assets/tests/components/detours/detourMap.test.tsx
```diff
@@ -47,19 +47,15 @@ const DetourMapWithDefaults = (
...
-      />
+      <DetourMapWithDefaults onClickOriginalShape={onClickOriginalShape} />
     )
 
     fireEvent.click(
       container.querySelector(".c-detour_map--original-route-shape")!
     )
 
-    expect(onClickOriginalShape).toHaveBeenNthCalledWith(1, shapePoint)
+    expect(onClickOriginalShape).toHaveBeenCalled()
```
joshlarson, line 58: This test was giving me some grief. I have a mild preference for being able to test more directly that the point that was actually clicked is the point passed into `onClickOriginalShape`, but I can't figure out how to simulate what point was clicked. (Passing `{ latlng: {lat: 0, lng: 0} }` as a second argument to `fireEvent` didn't seem to work ☹️)

assets/src/util/math.ts
```diff
@@ -9,46 +7,3 @@ import { LatLng, LatLngLiteral } from "leaflet"
...
- * @param point Point of reference to check distance to
- * @returns
- * if {@link positions} is empty, returns `undefined`
- *
- * Otherwise, returns a {@link ClosestPosition} object containing
- * - {@link ClosestPosition.position}: the closest element in {@link positions} to {@link point}
- * - {@link ClosestPosition.index}: the index of the element in {@link positions}
- * - {@link ClosestPosition.distance}: the distance to {@link point}
- */
-export const closestPosition = (
```
joshlarson, line 31: I'm on the fence about actually removing this, but the other part of this PR removes the only invocation of this function, so it's not strictly needed. Part of me thinks that we might keep it around "just in case", and I also worry that that's how code-hoarding happens.

firestack (approved):

PR #2499, opened 2024-03-13
feat: useDetourDirections error state

lemald:
Asana ticket: [Bubble up error state into useDetour results](https://app.asana.com/0/0/1206816132107679/f)

I was thinking of also adding the UI element to this, but that's turning out to be more complex (needing to do some stuff with theme colors). The main thing I'm thinking over with the hook logic is how we're going to handle removing the last waypoint on failure of an API call. Part of me thinks we can add another `useEffect` hook in `useDetour` that invokes the `undo` callback directly if there's an error, and this approach would work well with that.


assets/tests/factories/detourShapeFactory.ts
```diff
@@ -1,8 +1,8 @@
 import { Factory } from "fishery"
-import { DetourShape } from "../../src/models/detour"
+import { DetourDirection, DetourShape } from "../../src/models/detour"
 import { shapePointFactory } from "./shapePointFactory"
 
-export const directionsFactory = Factory.define<DetourShape["directions"][0]>(
+export const directionsFactory = Factory.define<DetourDirection>(
```
joshlarson, line 5: Love this change - I'd been struggling to wrap my mind around the array-indexed type for a while!

assets/src/hooks/useDetour.ts
```diff
@@ -2,29 +2,37 @@ import { useEffect, useMemo, useState } from "react"
...
-      setDirections(undefined)
+      setDetourShape(ok({ coordinates: [], directions: undefined }))
       return
     }
 
     fetchDetourDirections(shapePoints).then((detourInfo) => {
       if (isOk(detourInfo) && shouldUpdate) {
-        setDetourShape(detourInfo.ok.coordinates)
-        setDirections(detourInfo.ok.directions)
+        setDetourShape(ok(detourInfo.ok))
```
joshlarson, line 33: Can we just replace this whole if statement with 👇?
```jsx
setDetourShape(detourInfo)
```

It looks like right now we're just destructuring the two possibilities and then restructuring them back to exactly what they were before.

(I can't do it as a suggestion because you didn't modify line 32, so Github won't let me include that in the diff ☹️)

lemald: Ah good call, see 1b5fa7a8!

joshlarson (approved):

PR #2500, opened 2024-03-13
feat: Add function to find the nearest point on a line segment to a given point and the associated distance

joshlarson:


lib/util/location.ex
```diff
@@ -97,6 +97,89 @@ defmodule Util.Location do
...
+    x = longitude_scale_factor * (longitude - start_longitude)
+    y = latitude_scale_factor * (latitude - start_latitude)
+
+    squared_segment_length = segment_end_x ** 2 + segment_end_y ** 2
+
+    # Transform the coordinates of the non-segment point to (u, v),
+    # where segment_start is at (0, 0) and segment_end is at (1, 0).
+    # We don't bother computing the v coordinate because we don't need
+    # it.
+    u = (x * segment_end_x + y * segment_end_y) / squared_segment_length
```
joshlarson, line 157: I decided to structure it this way because I found it easier to understand as a series of two transformations (and then reversing those two transformations at the end).

I could also consolidate it into a single formula if that would be easier for others to understand.

Also, @firestack mentioned that this would use a dot product, and I was like "noooo", but lo and behold - this formula is indeed a dot product!

firestack: I haven't had the chance to really read over this and understand it, though I've been trying. Given that this ended up being the dot product, could the dot product be extracted into a function to hide the operation? I personally think it'd read easier, but I'm not sure what it'll look like after that and could easily be wrong on this.

lib/util/location.ex
```diff
@@ -97,6 +97,89 @@ defmodule Util.Location do
...
+      iex> Util.Location.distance_from_segment(
+      ...>   Util.Location.new(42.00004, -71.00014),
+      ...>   {Util.Location.new(42, -71.0002), Util.Location.new(42.0001, -71.0001)}
+      ...> )
+      %{
+        closest_point: %Util.Location{latitude: 42.00004711559566, longitude: -71.00015288440434},
+        distance: 1.3264961988869075
+      }
+  """
+  @spec distance_from_segment(
```
joshlarson, line 115: Bikeshed time!

Is this the right name for this function? I'm leaning towards renaming it to `nearest_point_to_segment` or something like that.

Also, is this the right file for this? It does have to do with `Location`'s, which is why I put it here, but maybe this is bespoke enough that it should get its own file?

firestack: I like `nearest_point_to_segment`, but I think that maybe this function shouldn't return distance, it's not that hard to compute it later, so I'd prefer that instead, but I'd like to hear why this should return the distance value.

firestack (approved):

PR #2502, opened 2024-03-14
chore: Reduce number of allowable Dependabot PR's to 1

joshlarson:
We seem to be struggling to keep up with Dependabot PR's, which is causing us to leave older Dependabot PR's open for a long time. Reducing the number of PR's we allow to be open at a time might help us keep these under control more easily.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,13 +5,13 @@ updates:
   schedule:
     interval: daily
     time: "10:00"
-  open-pull-requests-limit: 10
+  open-pull-requests-limit: 1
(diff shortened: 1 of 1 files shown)


firestack (approved):
I'm not sure I agree with
> Reducing the number of PR's we allow to be open at a time will help us keep these under control more easily.

But I'm interested in trying!

joshlarson:
@firestack haha that's a valid point, and while the "problem" is buried in the sub-message part of a commit message that almost certainly nobody will ever read, I'm still gonna change it to "might" before merging 😂

PR #2503, opened 2024-03-14
Routing error alert

lemald:
Asana ticket: [Surface error state in UI](https://app.asana.com/0/0/1206816132107680/f)

I'm particularly interested in feedback on how I added to the `ui-alert` theme color palette (or whatever the proper terminology is). It didn't quite fit into the existing map overrides or variable overrides, so I had to add a new override file, put I added a comment explaining why. I only bothered with the particular colors that are relevant for alerts (see [the Bootstrap documentation](https://getbootstrap.com/docs/5.3/components/alerts/#sass-loops) on how they generate CSS variables for each theme color).

### DIFF
--- a/assets/css/bootstrap/_theme_color_additions.scss
+++ b/assets/css/bootstrap/_theme_color_additions.scss
@@ -1,0 +1,25 @@
+@use "../color/tokens_2024" as new_tokens;
+
+$custom-theme-text: (
+  "ui-alert": new_tokens.$strawberry-800,
+);
+
+$theme-colors-text: map-merge($theme-colors-text, $custom-theme-text);
+
+$custom-theme-bg-subtle: (
+  "ui-alert": new_tokens.$strawberry-50,
+);
+
+$theme-colors-bg-subtle: map-merge(
+  $theme-colors-bg-subtle,
+  $custom-theme-bg-subtle
+);
+
+$custom-theme-border-subtle: (
+  "ui-alert": new_tokens.$strawberry-100,
+);
+
+$theme-colors-border-subtle: map-merge(
+  $theme-colors-border-subtle,
+  $custom-theme-border-subtle
+);
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -36,6 +36,7 @@ export const DiversionPage = ({
 
     detourShape,
     directions,
+    routingError,
 
     stops,
     missedStops,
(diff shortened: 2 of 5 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -258,6 +260,25 @@ const DiversionPagePanelFooter = ({
...
+      variant="ui-alert"
+      className="position-absolute top-0 left-0 mt-3 start-50 translate-middle-x icon-link z-1"
+      show={show}
+    >
+      <BsIcons.ExclamationTriangleFill />
+      Something went wrong. Please try again.
+      <CloseButton onClick={() => setShow(false)} />
+    </Alert>
+  )
+}
```
firestack, line 280: suggestion/general-feedback(non-blocking): I'd prefer to try and make the `dismissible` prop work, but I'm open to making this feedback non-blocking and addressing it in a separate PR. I did some looking locally, and it seems there are two issues causing the close button to look weird

1. classes `border-box` and `inherit-box` are overriding bootstrap's preferred values.
	- Because `inherit-box` indiscriminately sets the box sizing value, when the close button requests a `box-sizing: content-box;` it's being ignored due to this. We'll likely solve this as apart of the CSS reboot changes. We could provide an override to `inherit-box` to make it stop applying to children of the alert.
2. Font size
	- When I was experimenting locally, I was still seeing some strange behaviour after I added `content-box` back. I think the issue here is that my browser's font size is set to 15, but somewhere on the page is overriding it to 16, but somehow both are coexisting and conflicting. This makes the close button overflow the alert container, looking a little strange. I'm pretty sure this only happens for me though because of my custom default user agent font size. This would also be fixed as apart of the reboot changes.
3. Another issue, I was finding that the close button's focus border was intentionally overflowing the alert, I believe for visual emphasis. The `contain: paint` value I set on the `__map` class seems to be the wrong approach, and should instead be converted to use the `isolation-isolate` utility class.

lemald: I would also prefer just using `dismissible` if it's not too much effort to get it working! Right now I'm mainly focused on the interactions around automatically clearing the last waypoint on error, but I'll keep this in mind.

assets/css/bootstrap/_theme_color_additions.scss
firestack, line 1: I really wished that bootstrap would've just let us do all this via the `_map_overrides`, but since it doesn't, this seems like the right approach and something we'll defintely need as we continue to tweak bootstrap.

lemald: It is kind of frustrating that none of the default override files work, especially since the documentation [does mention](https://getbootstrap.com/docs/5.3/customize/sass/#add-to-map) the general use case of adding values to existing maps but doesn't go into this gotcha.

firestack (approved):

PR #2504, opened 2024-03-14
fix: using escape key on detour modal prompts for close confirmation

lemald:
Asana ticket: [⚙️ [extra] Pressing escape key on detour modal should ask confirmation before closing](https://app.asana.com/0/1148853526253420/1206845632479144/f)

This ended up involving more refactoring than I thought it would but I like that we can now rely on the Restart modal's handling of the escape key event itself. I think it does sort of expose a gap in our testing of the `<DetourModal>` component so I might additionally go in and add some tests for that.

### DIFF
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -33,14 +33,15 @@ import { ok, loading } from "../../../src/util/fetchResult"
 const DiversionPage = (
   props: Omit<
     Partial<ComponentProps<typeof DiversionPageDefault>>,
-    "originalRoute"
+    "originalRoute" | "showConfirmCloseModal"
   > & {
     originalRoute?: Partial<
       ComponentProps<typeof DiversionPageDefault>["originalRoute"]
     >
+    showConfirmCloseModal?: boolean
   }
 ) => {
-  const { originalRoute, ...otherProps } = props
+  const { originalRoute, showConfirmCloseModal, ...otherProps } = props
   return (
     <DiversionPageDefault
       originalRoute={{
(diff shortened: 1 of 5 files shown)


assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -33,14 +33,15 @@ import { ok, loading } from "../../../src/util/fetchResult"
 const DiversionPage = (
   props: Omit<
     Partial<ComponentProps<typeof DiversionPageDefault>>,
-    "originalRoute"
+    "originalRoute" | "showConfirmCloseModal"
   > & {
     originalRoute?: Partial<
       ComponentProps<typeof DiversionPageDefault>["originalRoute"]
     >
+    showConfirmCloseModal?: boolean
```
firestack, line 41: question(non-blocking): is this needed? it seems like the `Partial` on ln35 would already do this. The override on ln38 is needed because `Partial` doesn't imply `DeepPartial`, but it seems like that wouldn't matter for this type?

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -54,6 +55,9 @@ const DiversionPage = (
         zoom: 16,
         ...originalRoute,
       }}
+      showConfirmCloseModal={
+        showConfirmCloseModal === undefined ? false : showConfirmCloseModal
```
firestack, line 59:
```suggestion
        showConfirmCloseModal ?? false
```

firestack (commented):
A few comments but this otherwise LGTM

assets/tests/components/detours/diversionPage.test.tsx
lemald: Ah I went back and forth a bunch getting this type to work, but you're right that it can be simplified. See 0c4e8865.

firestack (approved):

PR #2506, opened 2024-03-14
feat: Update route segments to connect closer to given start and end points

joshlarson:
---
  - https://app.asana.com/0/0/1206823255274991

### DIFF
--- a/lib/skate/detours/route_segments.ex
+++ b/lib/skate/detours/route_segments.ex
@@ -69,11 +69,16 @@ defmodule Skate.Detours.RouteSegments do
 
     {:ok,
      %__MODULE__.Result{
-       before_detour: Enum.slice(shape, 0..start_index) ++ [nearest_start_point],
+       before_detour:
+         Enum.slice(shape, 0..start_index) ++
+           [nearest_start_point],
        detour:
          [nearest_start_point] ++
-           Enum.slice(shape, (start_index + 1)..end_index) ++ [nearest_end_point],
-       after_detour: [nearest_end_point] ++ Enum.slice(shape, (end_index + 1)..-1)
+           Enum.slice(shape, (start_index + 1)..end_index) ++
+           [nearest_end_point],
+       after_detour:
+         [nearest_end_point] ++
+           Enum.slice(shape, (end_index + 1)..-1)
      }}
   end
 
--- a/lib/skate/detours/route_segments.ex
+++ b/lib/skate/detours/route_segments.ex
@@ -80,7 +80,7 @@ defmodule Skate.Detours.RouteSegments do
   defp nearest_point_to_shape(shape, point) do
     {%{nearest_point: nearest_point}, index} =
       shape
-      |> Enum.zip(Enum.drop(shape, 1))
+      |> Enum.chunk_every(2, 1, :discard)
       |> Enum.map(fn segment -> Util.Location.nearest_point_to_segment(point, segment) end)
       |> Enum.map(fn nearest_point ->
         %{nearest_point: nearest_point, distance: Util.Location.distance(nearest_point, point)}
--- a/lib/skate/detours/route_segments.ex
+++ b/lib/skate/detours/route_segments.ex
@@ -21,15 +21,31 @@ defmodule Skate.Detours.RouteSegments do
   ## Examples
       iex> alias Util.Location
       iex> Skate.Detours.RouteSegments.route_segments(
-      ...>   [Location.new(0, 0), Util.Location.new(0, 1), Util.Location.new(0, 2), Util.Location.new(0, 3)],
-      ...>   Util.Location.new(0, 1),
-      ...>   Util.Location.new(0, 2)
+      ...>   [
+      ...>     Location.new(0.0, 0.0),
+      ...>     Location.new(0.0, 1.0),
+      ...>     Location.new(0.0, 2.0),
+      ...>     Location.new(0.0, 3.0)
+      ...>   ],
+      ...>   Location.new(0.0, 1.3),
+      ...>   Location.new(0.0, 2.6)
       ...> )
       {:ok,
         %Skate.Detours.RouteSegments.Result{
-          before_detour: [%Util.Location{latitude: 0, longitude: 0}, %Util.Location{latitude: 0, longitude: 1}],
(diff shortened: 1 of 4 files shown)


lib/skate/detours/route_segments.ex
```diff
@@ -21,49 +21,73 @@ defmodule Skate.Detours.RouteSegments do
...
+         [nearest_start_point] ++
+           Enum.slice(shape, (start_index + 1)..end_index) ++ [nearest_end_point],
+       after_detour: [nearest_end_point] ++ Enum.slice(shape, (end_index + 1)..-1)
      }}
   end
+
+  defp nearest_point_to_shape(shape, point) do
+    {%{nearest_point: nearest_point}, index} =
+      shape
+      |> Enum.zip(Enum.drop(shape, 1))
```
firestack, line 83: question(non-blocking): is there a reason to prefer `zip` over [`chunk_every/4`](https://hexdocs.pm/elixir/main/Enum.html#chunk_every/2)?
```suggestion
      |> Enum.chunk_every(2, 1, :discard)
```

firestack: I do see that `nearest_point_to_segment` takes a `Tuple` and not a `List` so that's one change, but I'm not sure it really matters?

lib/skate/detours/route_segments.ex
```diff
@@ -21,49 +21,73 @@ defmodule Skate.Detours.RouteSegments do
...
+      |> Enum.map(fn segment -> Util.Location.nearest_point_to_segment(point, segment) end)
+      |> Enum.map(fn nearest_point ->
+        %{nearest_point: nearest_point, distance: Util.Location.distance(nearest_point, point)}
+      end)
+      |> Enum.with_index()
+      |> Enum.min_by(fn {%{distance: distance}, _} -> distance end)
```
firestack, line 89: question(non-blocking): forgive me if I'm missing something, but do any consumers of this function use the `:distance` return value?

This may be worse for performance, but I was curious if the distance value ever needed to make it into the list at all.

```suggestion
      |> Enum.with_index()
      |> Enum.min_by(fn {%{nearest_point: nearest_point}, _} -> Util.Location.distance(nearest_point, point) end)

```

lib/skate/detours/route_segments.ex
joshlarson: Huh - I didn't realize that `chunk_every` behaved that way. 

I imagine that I could do `chunk_every` and then transform the resulting array into a tuple before passing it into `nearest_point_to_segment`

lib/skate/detours/route_segments.ex
joshlarson: Oh yeah that's a good call. I don't think it's worse for performance, because we have to compute the distance that number of times anyway. And if it turns out that we do wind up needing it, then we can start storing it then.

lib/skate/detours/route_segments.ex
joshlarson: [Use `chunk_every`](https://github.com/mbta/skate/pull/2506/commits/2c97c4513e23ed10422ff27e5120d44ba7923ef3)
[Convert from array to tuple](https://github.com/mbta/skate/pull/2506/commits/323dc824b1bb319ea78a8dc3965847858789e24d)

lib/skate/detours/route_segments.ex
joshlarson: [That's much nicer!](https://github.com/mbta/skate/pull/2506/commits/84638ddf08338df9b2a4cc5acc46b0cb3e6f8b6d)

firestack: > I don't think it's worse for performance, because we have to compute the distance that number of times anyway.

Oh! I just read this wrong, I saw `min_by` and thought `sort` 😅 
Glad this turned out well!

lib/skate/detours/route_segments.ex
```diff
@@ -21,49 +21,69 @@ defmodule Skate.Detours.RouteSegments do
...
     {:ok,
      %__MODULE__.Result{
-       before_detour: Enum.slice(shape, 0..start_index),
-       detour: Enum.slice(shape, start_index..end_index),
-       after_detour: Enum.slice(shape, end_index..-1)
+       before_detour: Enum.slice(shape, 0..start_index) ++ [nearest_start_point],
+       detour:
+         [nearest_start_point] ++
+           Enum.slice(shape, (start_index + 1)..end_index) ++ [nearest_end_point],
+       after_detour: [nearest_end_point] ++ Enum.slice(shape, (end_index + 1)..-1)
```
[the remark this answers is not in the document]
firestack: FWIW, I spent a bit of time trying to figure out what I wanted from a "clean up" and I didn't come to any satisfying solutions.

The one thing that made it _marginally_ better, was pulling out the `slice` results into their own values before concatenating them, but I'm not convinced it's worth.

I also tried using `Enum.split` instead, since the current behaviour is a little wrong, but that didn't make anything cleaner [somewhat expected though] :/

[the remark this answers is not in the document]
firestack: if I could have _anything_, I'd align these via newlines and have the `++` operator be at the beginning, but I'm pretty sure the formatter would blow these changes away.

```suggestion
       before_detour: 
         Enum.slice(shape, 0..start_index)
         ++ [nearest_start_point],
       detour:
         [nearest_start_point]
         ++ Enum.slice(shape, (start_index + 1)..end_index)
         ++ [nearest_end_point],
       after_detour:
       	[nearest_end_point]
       	++ Enum.slice(shape, (end_index + 1)..-1)
```

firestack (approved):
A nit but it's non blocking 👍🏻

lib/skate/detours/route_segments.ex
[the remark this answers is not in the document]
joshlarson: The formatter _partly_ allows what you're looking for, but it shoved the `++`'s to the end of the previous line rather than the beginning of the next line.

I agree with you that having the `++` at the beginning of the next line, but I think breaking it into separate lines is an improvement over what's currently there, so I'm gonna roll with it.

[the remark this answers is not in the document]
joshlarson: [Improvement!](https://github.com/mbta/skate/pull/2506/commits/c0da15e26097a634ee4cb5c29031b2eae49644ec)

PR #2507, opened 2024-03-15
feat(ex/skate/open_route_service_api): send 40ft bus info to ORS for HGV profile routing

firestack:
The the first commit is the actual functionality and tests, the other commits are just me trying to make it a bit cleaner. I can easily drop the later two commits if they're disliked.

---
  - https://app.asana.com/0/0/1206826435176362

### DIFF
--- a/lib/skate/open_route_service_api/directions_request.ex
+++ b/lib/skate/open_route_service_api/directions_request.ex
@@ -9,8 +9,19 @@ defmodule Skate.OpenRouteServiceAPI.DirectionsRequest do
   """
   @type t() :: %__MODULE__{
           coordinates: [[float()]],
-          continue_straight: boolean()
+          continue_straight: boolean(),
+          options: map()
         }
 
-  defstruct coordinates: [], continue_straight: true
+  defstruct coordinates: [],
+            continue_straight: true,
+            options: %{
+              profile_params: %{
+                restrictions: %{
+                  length: 12.192,
+                  width: 3.2004,
+                  height: 3.5052
+                }
+              }
+            }
 end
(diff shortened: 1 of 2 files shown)


lib/skate/open_route_service_api/directions_request.ex
```diff
@@ -4,13 +4,46 @@ defmodule Skate.OpenRouteServiceAPI.DirectionsRequest do
...
+      end
+
+      @type t :: %{restrictions: HgvRestrictions.t()}
+      defstruct [:restrictions]
+    end
+
+    @type t :: %{profile_params: ProfileParams.t()}
+    defstruct [:profile_params]
+  end
+
```
firestack, line 31: I thought doing this inline would be easy, but it doesn't look great. I'm open to other options for getting this typed and having a good place for bus profiles to live. I personally liked having all this in a single file, but I could be convinced to move these into separate files.
One issue, I had to move the definitions to before their first use, otherwise the compiler complains.

lib/skate/open_route_service_api/directions_request.ex
```diff
@@ -4,13 +4,46 @@ defmodule Skate.OpenRouteServiceAPI.DirectionsRequest do
...
+          options: Options.t()
         }
 
-  defstruct coordinates: [], continue_straight: true
+  defstruct coordinates: [],
+            continue_straight: true,
+            options: %{
+              profile_params: %{
+                restrictions:
+                  Skate.OpenRouteServiceAPI.DirectionsRequest.Options.ProfileParams.HgvRestrictions.bus_40ft()
```
firestack, line 46: this name is so long lol
1. Should we move the `OpenRouteServiceAPI` to the root? so there's no `Skate.` prefix anymore?
2. should we alias this so this call isn't so long?

joshlarson (approved):

PR #2508, opened 2024-03-15
feat(ex/skate/open_route_service_api): add `vehicle_type` to ORS API call

firestack:
I realized we hadn't set the bus type on our HGV profile, so this adds that to the API request.


hannahpurcell (approved):
Ok so, total newbie here! But the changes seem logical, legible, and a test was added. So I'm confident to approve now 👍

PR #2510, opened 2024-03-18
refactor: Update nearest-point algorithm to more explicitly use vector math

joshlarson:


lib/util/location.ex
```diff
@@ -115,39 +196,16 @@ defmodule Util.Location do
...
-    # segment and the point in question to (x, y) coordinates, where
-    # the origin is the segment start point, x is meters east from
-    # there, and y is meters north
-    segment_end_x = longitude_scale_factor * (end_longitude - start_longitude)
-    segment_end_y = latitude_scale_factor * (end_latitude - start_latitude)
+    segment_vector = displacement(start_loc, end_loc)
 
-    x = longitude_scale_factor * (longitude - start_longitude)
-    y = latitude_scale_factor * (latitude - start_latitude)
+    point_vector = displacement(start_loc, point)
```
joshlarson, line 204: `point_vector` seems like a silly name, but I'm not sure of a better one. It's the vector pointing from the `start` point of the segment to the non-segment point.

lib/util/vector2d.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Util.Vector2d do
+  defstruct [:x, :y]
+
+  def dot_product(%__MODULE__{x: x1, y: y1}, %__MODULE__{x: x2, y: y2}) do
```
joshlarson, line 4: This needs docs and tests

lib/util/location.ex
```diff
@@ -97,6 +99,85 @@ defmodule Util.Location do
...
+
+  ## Example
+      iex> Util.Location.displacement(
+      ...>    Util.Location.new(42.0002, -71.0001),
+      ...>    Util.Location.new(42.0001, -71.0002)
+      ...> )
+      %Util.Vector2d{x: -8.26337887771932, y: -11.119508022906418}
+  """
+  @spec displacement(from :: __MODULE__.From.t(), to :: __MODULE__.From.t()) :: Vector2d.t()
+  def displacement(%__MODULE__{latitude: from_latitude, longitude: from_longitude}, %__MODULE__{
```
firestack, line 130: just a note, I'm having difficulty understanding this in context. I understand what it's doing, but when I read `segment_vector = displacement(start_loc, end_loc)` it just doesn't make sense to me from reading it, and I've gotta do extra "mental 'math'" to figure out what I'm supposed to understand. Do you think there could be a different word to accurately describe this, or a way I should be thinking about this word?

joshlarson: Let's ideate tomorrow. I've been steeped in this for a while now, so it makes probably _too much sense_ to me 😅

lib/util/vector2d.ex
joshlarson: This has docs and tests!

lib/util/location.ex
joshlarson: I'm thinking maybe... `displacement_from`, and swap the order of the arguments. That way it would be something like:

```elixir
end_loc |> displacement_from(start_loc)
```

And then a similar treatment to `displace`, so that it would become

```elixir
start_loc |> displace_by(vector)
```

(I know that credo won't let us do it _exactly_ like that, but it still might make it easier to tell what's going on.)

firestack: I like the idea of reordering the arguments 👍🏻

firestack (approved):
Maybe @hannahpurcell should have a look over this before merging but I'm good with this 👍🏻 (would be nice to close out the last comment thread though)

lib/util/location.ex
joshlarson: [How does this look?](https://github.com/mbta/skate/pull/2510/commits/7bfb9a68b22a5daa540ba86ca751548cf115595a)

joshlarson: [And this to fix some formatting](https://github.com/mbta/skate/pull/2510/commits/4c7295e35b67940a1d354a80b9a6e4bd7cbc4bed)

hannahpurcell (commented):
Thank you so much for your patience!! So far, the logic is definitely complex, but making sense with a slow comb-through. I paused before completely finishing because I have a big question that might help things make a little more sense to me.

lib/util/location.ex
```diff
@@ -115,76 +223,126 @@ defmodule Util.Location do
...
+    #
+    #                      P <-- off-segment point
+    #                     /
+    #                    /
+    #                   /
+    #  start-point --> S--------E <-- end-point
+    #
+    # segment_vector would be the vector pointing from S to E
+    # point_vector would be the vector pointing from S to P
+    #
```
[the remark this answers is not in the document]
joshlarson: Yes - this is a great point!

lib/util/location.ex
```diff
@@ -115,76 +223,126 @@ defmodule Util.Location do
...
+        # on forever.
+        #
+        #           P                           P
+        #          /                             \
+        #         /                               \
+        #        /                                 \
+        #       S---Q----E                      Q   S--------E
+        #
+        squared_segment_length ->
+          projected_vector = Vector2d.project_onto(point_vector, segment_vector)
```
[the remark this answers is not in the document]
joshlarson: You're right - I should have said something like `S -> Q is what you get when you project S -> P onto the line containing S -> E`. 

Literally _everything_ except for the lat/long coordinates is vectors, not points (you can kind of tell that in the code, because almost all of the variables inside the function are `%Vector2d{...}`'s, but that's only immediately obvious if you look at the `@spec`'s of the various functions called), but I was fuzzy in my language in the explanation. I shall clean that up before merging.

lib/util/location.ex
```diff
@@ -97,6 +99,112 @@ defmodule Util.Location do
     distance(as_location!(lhs), as_location!(rhs))
   end
 
+  @doc """
+  Returns the displacement to `to` from `from` as a `Util.Vector2d` given in meters.
```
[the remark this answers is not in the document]
joshlarson: For sure! 

Do you think that "between `to` and `from`". I guess the variable names already make it pretty clear that we're going FROM `from` TO `to` 😂

lib/util/location.ex
```diff
@@ -115,76 +223,126 @@ defmodule Util.Location do
...
+        #      P
+        #     /
+        #    /
+        #   /
+        #  SE
+        #
+        0.0 ->
+          Vector2d.zero()
+
+        # Otherwise we project the off-segment vector onto the line
```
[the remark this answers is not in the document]
joshlarson: In the paragraph where I refer to `off-segment point`, I'm actually referring to the point itself:
```
Convert both the segment end-point and the off-segment point
into vectors representing their displacements from the segment
start-point.
```

joshlarson:
@hannahpurcell and @firestack I re-worked some of the explanation. Can you give it a once-over and let me know if it's any clearer?

lib/util/location.ex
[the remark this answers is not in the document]
hannahpurcell: Ahhh yes, tracking now. I see my later question about the projection did make this part a little clearer to me

lib/util/location.ex
```diff
@@ -97,6 +99,112 @@ defmodule Util.Location do
...
+  def displace_by(
+        %__MODULE__{latitude: from_latitude, longitude: from_longitude},
+        %Vector2d{x: x, y: y}
+      ) do
+    latitude_scale_factor =
+      1000 *
+        distance(
+          new(from_latitude, from_longitude),
+          new(from_latitude + 0.001, from_longitude)
+        )
```
hannahpurcell, line 193: It seems that the choice of `1000` versus another multiplier might be a bit arbitrary, except that this function needs to be the mathy inverse of the `displacement_from` function. Like, `displacement_from` gets a vector by calculating a scale_factor and multiplying the lat-lng by that factor, and this function undoes that scaling by dividing the lat-lng by the same scale factor. I think this is readable and comprehensible enough, but I can also see these two functions being refactored so that the scale_factor is calculated in a separate function that both `displacement_from` and `displace_by` use. Up to you if you like that or not!

hannahpurcell (approved):
This is great! Thank you! One nonblocking idea:

lib/util/location.ex
```diff
@@ -286,62 +306,64 @@ defmodule Util.Location do
...
-            # side closer to the end-point. Notice that in the diagram
-            # below, E is the cloest point on the segment S--E to both
-            # P and Q, which means that S--E, or our original
-            # segment_vector, is our nearest vector.
+            # squared_segment_length, then projected_vector is longer
+            # than the segment, which means it falls off on the side
+            # closer to the end-point. Notice that in the diagram
+            # below, N (the nearest point) sticks to E so that it
+            # stays on the segment S--E, so S->N (nearest_vector) is
+            # the same as S->E (segment_vector)
```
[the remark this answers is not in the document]
joshlarson: At first I was all like "ugh, not another letter to remember", but `N` and `Q` are different, and trying to _not_ introduce `N` got even more confusing, I thought.

lib/util/location.ex
joshlarson: Oh yeah good call. That had bothered me a bit, but then I completely forgot.

joshlarson: I believe that [this](https://github.com/mbta/skate/pull/2510/commits/9d07fd4bcdee050de2ea6720a15d1e42e9e98e94) should address that. I went with `small_increment` to represent `0.001` (and divided by `small_increment` instead of multiplying by `1000`) to make it clearer that it's just a random small number, and also because now you don't have to guess at the relationship between `1000` and `0.001`.

PR #2515, opened 2024-03-21
feat: (tiny) Add "return to route" message after turn-by-turn directions

hannahpurcell:
[Asana task](https://app.asana.com/0/1205385723132845/1206754739214649)

Just a lil phrase addition, to round off the turn-by-turn directions.

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -107,3 +107,7 @@
 .c-diversion-panel__help-text {
   line-height: 1.5;
 }
+
+.c-diversion-panel__conclude-detour-text {
+  font-weight: 500;
+}
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -107,7 +107,3 @@
 .c-diversion-panel__help-text {
   line-height: 1.5;
 }
-
-.c-diversion-panel__conclude-detour-text {
-  font-weight: 500;
-}
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -67,12 +67,8 @@ export const DiversionPanel = ({
                 </ListGroup.Item>
               ))}
               {detourFinished && (
-                <ListGroup.Item
-                  as="li"
-                >
-                  <strong className="fw-medium">
-                    Return to Regular Route
-                  </strong>
+                <ListGroup.Item as="li">
+                  <strong className="fw-medium">Return to Regular Route</strong>
                 </ListGroup.Item>
               )}
             </ListGroup>
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -67,7 +67,11 @@ export const DiversionPanel = ({
                 </ListGroup.Item>
               ))}
               {detourFinished && (
-                <ListGroup.Item key="return-to-route" as="li">
+                <ListGroup.Item
+                  key="return-to-route"
+                  as="li"
+                  className="c-diversion-panel__conclude-detour-text"
+                >
                   Return to Regular Route
                 </ListGroup.Item>
               )}
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -66,6 +66,11 @@ export const DiversionPanel = ({
                   {d.instruction}
                 </ListGroup.Item>
               ))}
+              {detourFinished && (
+                <ListGroup.Item key="return-to-route" as="li">
+                  Return to Regular Route
+                </ListGroup.Item>
+              )}
             </ListGroup>
           ) : (
             <DirectionsHelpText />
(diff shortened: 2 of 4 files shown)


firestack (approved):
This looks good!

It seems unspecified in the ticket, but my understanding was we also wanted to add this to the directions in the `detourPanel` directions when the route is finished. Do you wanna merge this PR and open a second for the other part?

hannahpurcell:
> we also wanted to add this to the directions in the `<DetourPanel/>` directions when the route is finished

Ah that makes sense! Since it's a 2-line change and ticket-relevant, I'd be happy to make that change here. I'll re-request review when it's done

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -66,6 +66,15 @@ export const DiversionPanel = ({
                   {d.instruction}
                 </ListGroup.Item>
               ))}
+              {detourFinished && (
+                <ListGroup.Item
+                  key="return-to-route"
```
firestack, line 71: suggestion: would this still work if we dropped the `key` prop? AFAIK I don't think we need it since we're not deriving this from a list.
```suggestion
```

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -66,6 +66,15 @@ export const DiversionPanel = ({
...
+                  as="li"
+                  className="c-diversion-panel__conclude-detour-text"
+                >
+                  Return to Regular Route
+                </ListGroup.Item>
```
firestack, line 76: suggestion: would using `<strong/>` work instead?
```suggestion
                >
                  <strong>Return to Regular Route</strong>
                </ListGroup.Item>
```

hannahpurcell: Not quite, because `<strong/>` maps to `font-weight: bold`, while the [Figma](https://www.figma.com/file/XGmmwVROXdOvnWtJeUFC50/Skate-%7C-Unplanned-Detours-%7C-Hand-Off?type=design&node-id=252-3825&mode=design&t=N13hE4ik6aMXmEX9-0) calls for medium font-weight

firestack: I personally think that's okay, but we should loop in design. My thinking is either the designs should be updated to accommodate `<strong>`'s default value, or we update the styling of `<strong>` to match what Figma says.

For now, so we can merge this, can we use `<strong>` but set a class on it to set the font weight to what it says in Figma? Otherwise I think this PR would be blocked by a round trip through design.

```suggestion
                >
                  <strong className="c-diversion-panel__conclude-detour-text">
                    Return to Regular Route
                  </strong>
                </ListGroup.Item>
```

joshlarson:
This is a good candidate to throw the `chromatic` tag on, in order to validate that the changes look right in Storybook.

This should create a diff - will link once Chromatic finished spinning

assets/css/_diversion_page.scss
```diff
@@ -107,3 +107,7 @@
 .c-diversion-panel__help-text {
   line-height: 1.5;
 }
+
+.c-diversion-panel__conclude-detour-text {
+  font-weight: 500;
```
joshlarson, line 112: Nit (non-blocking): I'm surprised that we don't already have [a font weight for](https://github.com/mbta/skate/blob/a981349f8d87f62d095c35d0d3c6816b9ddc2a60/assets/css/_skate_ui.scss#L43-L50) medium (if that's what font-weight 500 is called).

If we add it there, then this could become

```suggestion
  font-weight: var(--font-weight-medium);
```

This might conflict with @firestack's comment though. I'm still unsteady on my CSS feet 😅

firestack: I don't think there's any conflict here! They should coexist well

assets/src/components/detours/diversionPanel.tsx
joshlarson: *Question*: What's the benefit of using `<strong />` if we're overriding its styling anyway? Does it provide an accessibility assist?

firestack: ooh I found a better solution!
```suggestion
                >
                  <strong className="fw-medium">
                    Return to Regular Route
                  </strong>
                </ListGroup.Item>
```

firestack: IMO `<strong>` is semantically correct for this context _and_ communicates that the text is emphasized via the accessibility tree

hannahpurcell: Ay I like this! Makes sense to me. Thanks 👏

firestack (approved):
🎉
We don't have any test's for directions it seems, and nothing for the "Return to Regular Route" text in the Create panel, I would say that we merge this and then add tests in a follow-up PR [or open a ticket(?)].

joshlarson:
> This is a good candidate to throw the chromatic tag on, in order to validate that the changes look right in Storybook.
> 
> This should create a diff - will link once Chromatic finished spinning

Argh! I was wrong - there's no story validating what I wanted to validate. Oh well!

joshlarson (approved):

PR #2517, opened 2024-03-22
feat: Add more flexible OpenRouteService configuration to support self-hosting

joshlarson:
---
  - https://app.asana.com/0/0/1206910009181796

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -14,6 +14,7 @@ config :skate,
 config :skate, Skate.OpenRouteServiceAPI,
   api_base_url: System.get_env("OPEN_ROUTE_SERVICE_API_URL"),
   api_key: System.get_env("OPEN_ROUTE_SERVICE_API_KEY"),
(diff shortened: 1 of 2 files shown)


joshlarson:
Converted back to draft because we need the OPEN_ROUTE_SERVICE_DIRECTIONS_PATH to be defined in all environments before we merge this!

firestack:
> Converted back to draft because we need the OPEN_ROUTE_SERVICE_DIRECTIONS_PATH to be defined in all environments before we merge this!

Would it be worth falling back to the old directions path for now in the config file and then remove that code in a later PR?

joshlarson:
>> Converted back to draft because we need the `OPEN_ROUTE_SERVICE_DIRECTIONS_PATH` to be defined in all environments before we merge this!
>
> Would it be worth falling back to the old directions path for now in the config file and then remove that code in a later PR?

Good call! Now this doesn't need to be blocked on https://github.com/mbta/devops/pull/1642

firestack (approved):

PR #2521, opened 2024-03-26
fix: Moved "Return to Regular Route" into root directions list

hannahpurcell:
The little things I like about this approach is that
- "Return to Regular Route" is put with other directions, which I think it should be!
- The direction is in the root hook instead of scattered in multiple components
- The test for `diversionPage` now sufficiently covers the check "was 'Return to Regular Route' added to the directions sequence"
- We can still choose to make directions more structured with typing later (per Josh's suggestion)

Also included, updating the Storybook story for this!

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -92,7 +92,7 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   const directions = isOk(detourShape)
     ? finishedDetour
       ? detourShape.ok.directions?.concat({
-          instruction: "Return to Regular Route",
+          instruction: "Regular Route",
         })
       : detourShape.ok.directions
     : undefined
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -88,10 +88,13 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   )
 
   const coordinates = isOk(detourShape) ? detourShape.ok.coordinates : []
+  // Only append direction "Regular Route" after detour is finished
   const directions = isOk(detourShape)
-    ? detourShape.ok.directions?.concat({
-        instruction: "Return to Regular Route",
-      })
+    ? finishedDetour
+      ? detourShape.ok.directions?.concat({
+          instruction: "Return to Regular Route",
+        })
+      : detourShape.ok.directions
     : undefined
 
   const canAddWaypoint = () => startPoint !== null && endPoint === null
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -88,7 +88,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   )
 
   const coordinates = isOk(detourShape) ? detourShape.ok.coordinates : []
-  const directions = isOk(detourShape) ? detourShape.ok.directions : undefined
+  const directions = isOk(detourShape)
+    ? detourShape.ok.directions?.concat({
+        instruction: "Return to Regular Route",
+      })
+    : undefined
 
   const canAddWaypoint = () => startPoint !== null && endPoint === null
   const addWaypoint = canAddWaypoint()
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -261,11 +261,11 @@ describe("DiversionPage", () => {
 
     const { container } = render(<DiversionPage />)
 
-    await act(async () => {
+    act(() => {
       fireEvent.click(originalRouteShape.get(container))
     })
 
-    await act(async () => {
+    act(() => {
       fireEvent.click(originalRouteShape.get(container))
     })
 
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -410,7 +410,7 @@ describe("DiversionPage", () => {
           "Turn left on Main Street",
           "Turn right on High Street",
           "Turn sharp right on Broadway",
-          "Return to Regular Route",
+          "Regular Route",
           ,
           "Connection Points:",
           start.name,
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -242,6 +242,74 @@ describe("DiversionPage", () => {
     expect(screen.queryByTitle("Detour End")).toBeNull()
   })
 
+  test("shows 'Regular Route' text when the detour is finished", async () => {
+    jest.mocked(fetchDetourDirections).mockResolvedValue(
+      ok(
+        detourShapeFactory.build({
+          directions: [
+            { instruction: "Turn left on Main Street" },
+            { instruction: "Turn right on High Street" },
+            { instruction: "Turn sharp right on Broadway" },
+          ],
+        })
+      )
+    )
+
+    jest
+      .mocked(fetchFinishedDetour)
+      .mockResolvedValue(
+        finishedDetourFactory.build()
+      )
+    
+    const { container } = render(<DiversionPage />)
+
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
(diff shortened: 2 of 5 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -88,7 +88,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   )
 
   const coordinates = isOk(detourShape) ? detourShape.ok.coordinates : []
-  const directions = isOk(detourShape) ? detourShape.ok.directions : undefined
+  const directions = isOk(detourShape)
+    ? detourShape.ok.directions?.concat({
+        instruction: "Return to Regular Route",
```
firestack, line 93: question: this text doesn't match the copy in the Figma, was it updated somewhere else that I missed?

assets/src/hooks/useDetour.ts
```diff
@@ -88,7 +88,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
...
+      })
+    : undefined
```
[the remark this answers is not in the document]
hannahpurcell: Ahhh yes! Thank you

assets/src/hooks/useDetour.ts
hannahpurcell: Ohhhhh!! I got so locked onto the ticket title that I glossed over the diff with the figma

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
joshlarson: This is enough of a tricky edge case that I would recommend having a test for it. Ideally at the `diversionPage` level (to protect against refactors), but I would be okay with the `useDetour` level as well or instead.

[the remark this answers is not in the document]
hannahpurcell: Paired with @firestack to write the tests, and I understand things so much better! (Thanks, Kayla 😄 )

Added a test at the `diversionPage` level to verify the text appears at the right times. (Is the second non-existence test too large?) I was not able to include a test to verify the `<strong>` attribute. I think [this release](https://github.com/testing-library/dom-testing-library/pull/1269) of `testing-library` could resolve how `findByRole` ignores the `strong` role.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -242,6 +242,70 @@ describe("DiversionPage", () => {
...
+
+    expect(screen.queryByText("Regular Route")).toBeNull()
+
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
```
firestack, line 302: suggestion: I don't think the inner blocks need `async`?
```suggestion
    await act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })

    await act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })
```

firestack (dismissed):
Two comments but otherwise this looks good!

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -242,6 +242,70 @@ describe("DiversionPage", () => {
...
+
+    fireEvent.click(screen.getByRole("button", { name: "Undo" }))
+
+    expect(screen.queryByText("Regular Route")).toBeNull()
```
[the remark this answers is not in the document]
hannahpurcell: Ok! What are the things you think about when choosing between those options? And why choose one over the other here? I saw a lot of `.toBeNull()` matchers in this test suite for elements that should not be present after some actions, but I've also seen `.not.toBeInTheDocument()`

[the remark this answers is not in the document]
firestack: AFAIK the `toBeNull` matchers on DOM elements are just legacy code. I prefer to use the jest-dom matchers wherever possible when working with DOM elements, just so that the code both reads better and is clearer about what it's asserting.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -242,6 +242,70 @@ describe("DiversionPage", () => {
...
+    })
+
+    await act(async () => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
+    expect(await screen.findByText("Regular Route")).toBeVisible()
+  })
+
+  test("does not show 'Regular Route' when detour not finished", async () => {
```
joshlarson, line 275:
```suggestion
  test("does not show 'Regular Route' when detour is not finished", async () => {
```

Nit: grammar

assets/tests/components/detours/diversionPage.test.tsx
firestack: No because `act` is `async`

hannahpurcell: Ah that makes sense! Had to [read up](https://medium.com/@bmb21/reacts-sync-and-async-act-caa297b658b0) a tad on act async, and it does work like it sounds. (There are some other functions here that do use `async` to wrap an identical click action, like `missed stops are filled in when detour is complete`, so I initially was copying those. Seems like `async` is unneeded there, too)

firestack: ^I'm confirming this assumption though!

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: 🤔 Twas not a typo, but if this makes it meaningfully more readable to you, might as well fix.

assets/tests/components/detours/diversionPage.test.tsx
firestack:
```suggestion
    act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })

    act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })
```

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: WELP!

I don't feel strongly about it, but I _think_ most of the test titles read more like full sentences?

hannahpurcell: I'm not sure we can/should aim for unanimity in sentence structure across devs for test titles, but I'll commit this one since all I have to do is click a button!

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -242,6 +242,70 @@ describe("DiversionPage", () => {
...
+
+    act(() => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
+    act(() => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
+    fireEvent.click(screen.getByRole("button", { name: "Undo" }))
```
firestack, line 304: nit(non-blocking): hmm, these seem to be undoing each-other, and we're not waiting on the results of the `Undo` click, could we remove both of these?

```suggestion
```

firestack (approved):
LGTM, Thanks @hannahpurcell!!

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: Oh derp, embarrassing. My thought process was that I want to verify that the "regular route" message is absent in both situations where a detour is "unfinished" -- pre-completion of detour, and post-undo. I can tuck a `expect(await screen.findByText("Regular Route")).toBeVisible()` before the undo, but then we have repeated tests, essentially

joshlarson: It _sort of_ feels like this is actually two tests hiding in one. First, testing that `Regular Route` doesn't appear before connecting a detour, and second, testing that `Regular Route` disappears after undoing a connected detour. 

Please treat ☝️ as musings, not things to act on right now, unless you want to :). When we come to a clearer testing strategy overall, that'll be a good opportunity to rework some of these tests.

hannahpurcell: It's true, I did want to lump two tests together, heh. they are "similar vibe" but different things 😅

PR #2524, opened 2024-03-28
feat: return `:no_route` error from ORS

firestack:
This returns errors from ORS and processes it into a smaller set of error values. The frontend is currently ignoring the return value of this so this can be made out of step with the frontend and therefore reviewed in a smaller chunk.

---
  - https://app.asana.com/0/0/1206941188727470

### DIFF
--- a/test/skate_web/controllers/detour_route_controller_test.exs
+++ b/test/skate_web/controllers/detour_route_controller_test.exs
@@ -177,7 +177,24 @@ defmodule SkateWeb.DetourRouteControllerTest do
           coordinates: [%{"lat" => 1, "lon" => 100}, %{"lat" => 2, "lon" => 101}]
         )
 
-      assert json_response(conn, 500)
+      assert %{"error" => %{"type" => "unknown"}} = json_response(conn, 500)
+    end
+
+    @tag :authenticated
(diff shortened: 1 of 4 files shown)


hannahpurcell (approved):
Fab!

firestack:
I'm going to go ahead and merge this so that trunk stays current, but @joshlarson I'd love to chat about the tests in here if you have any thoughts.

test/skate_web/controllers/detour_route_controller_test.exs
```diff
@@ -177,7 +177,24 @@ defmodule SkateWeb.DetourRouteControllerTest do
           coordinates: [%{"lat" => 1, "lon" => 100}, %{"lat" => 2, "lon" => 101}]
         )
 
-      assert json_response(conn, 500)
+      assert %{"error" => %{"type" => "unknown"}} = json_response(conn, 500)
+    end
+
+    @tag :authenticated
+    test "returns a 500-level response with type: :no_route when ORS reports point not found error",
```
joshlarson, line 184: Swooping in at later-than-the-last-minute!

I appreciate this test because it makes it abundantly clear what the relationship between state-of-the-world and output is. 

It also exposes a question: Would this particular use case make more sense as a 4xx-level error, because it's a problem with user input rather than a problem with the server?

PR #2525, opened 2024-03-29
feat: load skate variables via 1password

firestack:


.envrc.template
```diff
@@ -4,35 +4,35 @@
...
-export POSTGRES_USERNAME=
-export POSTGRES_PASSWORD=
-export POSTGRES_HOSTNAME=
+# export POSTGRES_USERNAME=
+# export POSTGRES_PASSWORD=
+# export POSTGRES_HOSTNAME=
 
 ## API key from [Swiftly Transitime API](https://swiftly-inc.stoplight.io/docs/realtime-standalone/YXBpOjI4NDM2MDU3-swiftly-api-reference)
-export SWIFTLY_AUTHORIZATION_KEY=
+# export SWIFTLY_AUTHORIZATION_KEY=
```
hannahpurcell, line 20: Would you mind moving the variables that are now covered in 1password to a new section separate from `Required` and `Optional`, maybe `Imported from `1Password`?

hannahpurcell (approved):

.envrc.template
firestack: Love it! I still like the idea of having everything categorized into `Required` and `Optional` sections though, what do you think of this change with a subsection within each section for 1password variables?

[`e48cb34` (#2525)](https://github.com/mbta/skate/pull/2525/commits/e48cb348d7780664a1c2bc30992317a24be26cc0)

README.md
```diff
@@ -24,6 +24,14 @@ To avoid having to set these manually in your local development environment, [di
...
 The environment variables are documented in the [Skate `.envrc.template` file](.envrc.template).
 
+> [!NOTE]
+> Some of these configuration values are shared between Skate team members.
+> While there are still some values which are not shared and need to be configured in
+> `.envrc.private`, to facilitate easier setup and to reduce the amount of work done when cloning
+> for team members, the [`.envrc`](./.envrc) file is configured to source the shared
+> configuration values from 1Password using the 
+> [.env.1p.skate file](./.env.1p.skate) and the 1Password CLI (which you must also have on your system).
+
```
firestack, line 34: [context: github comment](https://github.com/mbta/skate/pull/2522#discussion_r1544546586) @hannahpurcell: Here's the docs you "asked" for!! I'd appreciate any feedback you have, especially on the "copy". I'm not sold on using a block quote and the `[!NOTE]` callout syntax, but I think it helps when reading it.

.env.1p.skate
```diff
@@ -0,0 +1,22 @@
+# MBTA Skate Secret Environment Variables
```
joshlarson, line 1: Nit: It's throwing me off a bit that all of the other `.envrc` files are called `.envrc...`, but this one is `.env.1p...`. Could we call this `.envrc.1p.skate` instead?

firestack: this isn't a `.envrc` file though, this is specifically a `.env` file, which does not use shell functions (i.e., `export`) and is only key=value mappings [as I currently understand it].

joshlarson: That is a fair point that I hadn't noticed, and I withdraw my request

joshlarson (approved):

README.md
joshlarson: The `1Password CLI` that we need installed is `op`, right? Should we include a link here to installation instructions? 

This link, I think?
https://developer.1password.com/docs/cli/get-started/

firestack: @joshlarson updated!

PR #2526, opened 2024-03-29
 feat(direnv/template): source secrets directly from `mix phx.gen.secret`

firestack:
Instead of having users fill this in manually, we can source it directly from mix when we load direnv.


hannahpurcell (approved):
Oh sure! Yeah this helps 👍

PR #2527, opened 2024-03-29
fix: Await directions fetch before adding "Regular Route" to list

hannahpurcell:
[Asana task](https://app.asana.com/0/1205385723132845/1206941178860494)

### DIFF
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -61,8 +61,8 @@ export const DiversionPanel = ({
           <h2 className="c-diversion-panel__h2">Detour Directions</h2>
           {directions ? (
             <ListGroup as="ol">
-              {directions.map((d) => (
-                <ListGroup.Item key={d.instruction} as="li">
+              {directions.map((d, i) => (
+                <ListGroup.Item key={i} as="li">
                   {d.instruction == "Regular Route" ? (
                     <strong className="fw-medium">{d.instruction}</strong>
                   ) : (
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -13,33 +13,40 @@ import {
 
 const useDetourDirections = (
   shapePoints: ShapePoint[]
-): FetchResult<DetourShape> => {
+): { finishedDirections: boolean; detourShape: FetchResult<DetourShape> } => {
   const [detourShape, setDetourShape] = useState<FetchResult<DetourShape>>(
     loading()
   )
+  const [finishedDirections, setFinishedDirections] = useState(false)
 
   useEffect(() => {
     let shouldUpdate = true
 
-    if (shapePoints.length < 2) {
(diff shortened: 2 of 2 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -13,33 +13,40 @@ import {
 
 const useDetourDirections = (
   shapePoints: ShapePoint[]
-): FetchResult<DetourShape> => {
+): { finishedDirections: boolean; detourShape: FetchResult<DetourShape> } => {
   const [detourShape, setDetourShape] = useState<FetchResult<DetourShape>>(
     loading()
   )
+  const [finishedDirections, setFinishedDirections] = useState(false)
```
firestack, line 20: suggestion: instead of adding a new state, could we derive this status from the `isLoading` function in `fetchResult`?

I think that would mean replacing the `setFinishedDirections(false)` on ln26 to `setDetourShape(loading())`

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -61,8 +61,8 @@ export const DiversionPanel = ({
           <h2 className="c-diversion-panel__h2">Detour Directions</h2>
           {directions ? (
             <ListGroup as="ol">
-              {directions.map((d) => (
-                <ListGroup.Item key={d.instruction} as="li">
+              {directions.map((d, i) => (
+                <ListGroup.Item key={i} as="li">
```
joshlarson, line 65: `i` in this context is the index, right?

If so, then I _think_ React recommends against using the index as a key. I'm not an expert here, so I could easily see this exact use case not really being a problem, but it's still something that I've sort of been trained to shy away from.

https://react.dev/learn/rendering-lists#why-does-react-need-keys

assets/src/hooks/useDetour.ts
joshlarson: I might be wrong here, but I think that this suggestion will lead to a situation where the directions blink in and out of existence when the user adds or removes waypoints. 

But the existing logic will involve the `Regular Route` instruction disappearing a split second before any other directions when you undo. 

I think the _actual_ thing that's making this hard to think about is the fact that we're doing a lot of this logic in the frontend, when it really belongs in the backend. Let the server decide what the directions are and whether to add in "Regular Route" and whatever else needs to be shown, and let the frontend handle as few separate pieces of state as possible.

To be fair, I don't think we should do that in this PR, or even soon/immediately. But it's starting to look like having detour logic spread between the frontend and the backend is tech debt that's generating higher and higher interest payments.

hannahpurcell: Ah! Ok I was having issues with my envrc that were invisible to me for a while and made me think it was a separate issue. Ok, so I've tried the derived `isLoading` state and saw the blinking directions that Josh noted. I've also seen the lag in `Regular Route` instructions disappearing on undo, as noted. Given the preferred solution may be a refactor, I'll pause on this work and bring that up during stand-up tomorrow so we can weigh as a group whether this tech debt needs doing now or later.

PR #2529, opened 2024-04-02
fix: set deploy to only happen on full releases (not pre-releases)

firestack:
One tiny change because I got the `type` event wrong


hannahpurcell (approved):
Ah! Sorry I didn't help catch that before

PR #2531, opened 2024-04-05
feat: add api helper function to return `Result<T, E>` from API requests

firestack:
This creates a new `Result` type and also uses it immediately in `api.ts` to show an example of it's usage. 

---

You can see this code in use on the PR https://github.com/mbta/skate/pull/2534

---
  - https://app.asana.com/0/0/1207056301082043

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -20,7 +20,16 @@ import {
   TripId,
 } from "./schedule.d"
 import { RouteTab } from "./models/routeTab"
-import { array, assert, Struct, StructError } from "superstruct"
+import {
+  any,
+  array,
+  assert,
+  create,
+  is,
+  object,
+  Struct,
+  StructError,
+} from "superstruct"
 import { ShapeData, shapeFromData, shapesFromData } from "./models/shapeData"
 import { StopData, stopsFromData } from "./models/stopData"
 import {
@@ -46,6 +55,7 @@ import {
   finishedDetourFromData,
 } from "./models/finishedDetour"
 import { FetchResult, ok, fetchError } from "./util/fetchResult"
+import { Ok, Err, Result } from "./util/result"
 
 export interface RouteData {
   id: string
@@ -142,6 +152,32 @@ export const apiCallWithError = <T, U>({
       }
 
       return fetchError()
+    })
+
+export const apiCallResult = async <T, E>(
+  url: Parameters<typeof fetch>[0],
+  OkStruct: Struct<T, unknown>,
+  ErrStruct: Struct<E, unknown>,
+  requestInit?: Parameters<typeof fetch>[1]
+): Promise<Result<T, E>> =>
+  fetch(url, requestInit)
+    .then(async (response) => {
+      const json: unknown = await response.json()
+
+      if (response.ok && is(json, object({ data: any() }))) {
+        return Ok(create(json.data, OkStruct))
+      } else {
+        assert(json, object({ error: any() }))
+        return Err(create(json.error, ErrStruct))
+      }
+    })
+    .catch((error) => {
+      if (error instanceof StructError) {
+        Sentry.captureException(error)
+      }
+      // We throw within the returned promise because downstream consumers
+      // should handle errors themselves
+      throw error
     })
 
 export const parseRouteData = ({
--- a/assets/src/util/result.ts
+++ b/assets/src/util/result.ts
@@ -1,0 +1,56 @@
+/**
+ * The discriminator value used to determine if a {@linkcode Result<T, E>} is
+ * {@linkcode Ok<T>} or {@linkcode Err<E>}.
+ */
+enum ResultTag {
+  Ok,
+  Err,
+}
+
+/**
+ * The key of the {@linkcode Result<T, E>} discriminator value.
+ *
+ * This is not exported so that usages of the objects cannot be
+ * accessed or modified manually outside of this module.
+ */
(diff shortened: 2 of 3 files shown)


assets/src/util/result.ts
```diff
@@ -0,0 +1,56 @@
...
+export const Err = <E>(v: E): Err<E> => ({
+  [tag]: ResultTag.Err,
+  err: v,
+})
+
+/**
+ * If {@linkcode r} is {@linkcode Ok<T>}, it applies {@linkcode fn} to
+ * the value contained in {@linkcode r}. Otherwise returns {@linkcode r}.
+ */
+export const map = <T, U, E>(r: Result<T, E>, fn: (v: T) => U): Result<U, E> =>
```
joshlarson, line 55: I'm kind of torn on this, but I'm leaning towards thinking that this is the sort of thing that does deserve its own tests.

I'm also wondering if there's a way to build this so that usage looks like:
```ts
result.map((foo) => bar)
```

rather than
```ts
map(result, (foo) => bar)
```

but that's more of a syntax nitpick than anything else.

firestack: > I'm kind of torn on this, but I'm leaning towards thinking that this is the sort of thing that does deserve its own tests.
> 
> I'm also wondering if there's a way to build this so that usage looks like:
> 
> ```ts
> result.map((foo) => bar)
> ```
> 
> rather than
> 
> ```ts
> map(result, (foo) => bar)
> ```
> 
> but that's more of a syntax nitpick than anything else.

Agreed, but then I had to consider whether I put the `map` function on the object during object 'construction' or if I create a JS class and I ended up spending more than enough time thinking about that that I went with a function instead for simplicity.

firestack: > I'm kind of torn on this, but I'm leaning towards thinking that this is the sort of thing that does deserve its own tests.

Tests! [`20716d8` (#2531)](https://github.com/mbta/skate/pull/2531/commits/20716d8de35ffe6838907d38f89319187a503d59)

assets/src/api.ts
```diff
@@ -144,6 +154,32 @@ export const apiCallWithError = <T, U>({
...
+    })
+    .catch((error) => {
+      if (error instanceof StructError) {
+        Sentry.captureException(error)
+      }
+      // We throw within the returned promise because downstream consumers
+      // should handle errors themselves
+      throw error
+    })
+
```
firestack, line 182: @mbta/skate-developers any thoughts on keeping this in `api.ts` vs it's own file?

If voting for moving it out, what would you name the file?

joshlarson: Hmmm - I think it makes sense where it is, but I'm not _opposed_ to moving it?

HOWEVER, thinking about what file I would put this in made me start bikeshedding the name of the function. `apiCallResult` makes me think that it's transforming some data into a nicely formatted result, which it IS doing, but I would want the name to also indicate that it's actually doing the API call. I'm not sure what TO call it though - `fetchFromApi` or something like that?

This is non-blocking, obviously. We can rename it later if a name idea speaks to us.

assets/src/util/result.ts
joshlarson: I'll play around with changing it as a potential follow-up PR, possibly!

assets/tests/util/result.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  test("when `isErr`, allows accessing the internal value", () => {
+    const value = Err(10) as Result<unknown, number>
+
+    expect(isErr(value) && value.err).toEqual(10)
+  })
+})
+
+describe("Result<T, E>::map", () => {
+  test("applies function if Ok<T>", () => {
+    const fn = jest.fn(() => true)
```
joshlarson, line 36: Nit, non-blocking

Hmmm - any reason not to just test this with a real function? So something like

```typescript
const result: Result<number, unknown> = Ok(3)

expect(map(result, (x) => x * 2)).toEqual(Ok(6))
```

It's not like... some big objective truth or anything, just that it feels more natural to me to test this way (I'm not a huge fan of mocks, especially when non-mocked things work just fine).

joshlarson (dismissed):
This looks good!

I did make a comment that I'd love to get your thoughts on, but since I'm out tomorrow, blocking approval on it isn't reasonable, and I was going to approve this PR either way - https://github.com/mbta/skate/pull/2531/files#r1561591516

assets/tests/util/result.test.ts
firestack: IMO, this is "real function" it just also comes with tracking abilities so we can assert what it's called with. It's just a personal preference to be able to assert that the function was called so I'm okay with leaving it off.

What do you think of [`d8230eb` (#2531)](https://github.com/mbta/skate/pull/2531/commits/d8230eba8735ed5ba1a4d551cc24192ce46776a8)?
I like it because it's completely impossible for `map` to know the value of `value` ahead of time, and this also tests that `map` allows for transforming the _type_ contained within `Ok<T>`

joshlarson: Works for me! And I do like the idea of changing the type 👍

joshlarson (approved):

joshlarson:
Sorry for being slow at re-approving! It took me a minute to realize that you'd re-requested approval!

assets/tests/util/result.test.ts
joshlarson: I'd be tempted to try giving the other test a similar treatment, but as before - this is non-blocking.

PR #2533, opened 2024-04-05
init(ts/hooks/useApiCall): add hook for consuming api promises

firestack:
This adds a generic hook for consuming Promise's, we currently only have a need for this with API results.

This will be combined with #2531 to make implementing API endpoints which may return an error easier.

---

You can see this code in use on the PR https://github.com/mbta/skate/pull/2534

---
  - https://app.asana.com/0/0/1207056301082042
  - https://app.asana.com/0/0/1207091097322621

### DIFF
--- a/assets/src/hooks/useApiCall.ts
+++ b/assets/src/hooks/useApiCall.ts
@@ -1,0 +1,65 @@
+import { useEffect, useState } from "react"
+
+/**
+ * Return type of {@linkcode useApiCall}
+ */
+interface ApiLoadingResult<T> {
+  /**
+   * Indicates if {@linkcode useApiCall} is in the process of computing a new
+   * value
+   */
+  isLoading: boolean
+
+  /**
+   * The last value that {@linkcode useApiCall} resolved.
+   *
+   * Is `undefined` when it has not resolved or it resolves to `undefined`
+   */
+  result: T | undefined
+}
+
+/** Hook prop arguments for {@linkcode useApiCall} */
+interface UseApiCallProps<T> {
+  /**
+   * Function which returns a promise that is called every time it changes.
+   * @param abortSignal An {@linkcode AbortSignal} that signals if the promise
+   *    should attempt to cancel early
+   */
+  apiCall: (abortSignal: AbortSignal) => Promise<undefined | T>
+}
+
+/**
+ * Calls the provided function returning a {@linkcode Promise} and tracks the
+ * loading state and the previously resolved value.
+ *
+ * @returns {ApiLoadingResult<T>}
+ */
+export const useApiCall = <T>({
+  apiCall,
+}: UseApiCallProps<T>): ApiLoadingResult<T> => {
+  const [apiResult, setApiResult] = useState<T | undefined>(undefined)
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const controller = new AbortController()
+
+    setIsLoading(true)
+    apiCall(controller.signal).then((value) => {
+      if (controller.signal.aborted) {
+        return
+      }
+
+      setApiResult(value)
+      setIsLoading(false)
+    })
+
+    return () => {
+      controller.abort()
+    }
+  }, [apiCall, setApiResult, setIsLoading])
+
+  return {
+    isLoading,
+    result: apiResult,
+  }
+}
--- a/assets/tests/hooks/useApiCall.test.ts
+++ b/assets/tests/hooks/useApiCall.test.ts
@@ -1,0 +1,189 @@
+import { describe, expect, jest, test } from "@jest/globals"
+import { act, renderHook, waitFor } from "@testing-library/react"
+import { useApiCall } from "../../src/hooks/useApiCall"
+
+const renderUseApiCall = (initialProps: Parameters<typeof useApiCall>[0]) =>
+  renderHook(useApiCall, {
+    initialProps,
+  })
+
+const PromiseWithResolvers = <T>(): {
+  promise: Promise<T>
+  resolve: (value: T | PromiseLike<T>) => void
+  reject: (reason?: any) => void
+} => {
+  let resolve: ((value: T | PromiseLike<T>) => void) | undefined = undefined
+  let reject: ((reason?: any) => void) | undefined = undefined
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res
+    reject = rej
+  })
+
+  if (resolve === undefined || reject === undefined) {
+    throw Error("Promise failed to assign `resolve` and/or `reject`")
+  }
+
+  return { promise, resolve, reject }
+}
+
+describe("when first rendered", () => {
+  test("should return loading state", () => {
+    const { result } = renderUseApiCall({
+      apiCall: () => new Promise(() => {}),
+    })
+
+    expect(result.current?.isLoading).toBe(true)
+  })
+
+  test("should call the input function", () => {
+    const fn = jest.fn(() => new Promise(() => {}))
+
+    renderUseApiCall({
+      apiCall: fn,
+    })
+
+    expect(fn).toHaveBeenCalledTimes(1)
+  })
+})
+
+describe("when the input function changes", () => {
+  test("should change to loading state", async () => {
+    const fn = jest.fn(() => Promise.resolve(undefined))
+    const fn2 = jest.fn(() => Promise.resolve(undefined))
+
+    const { result, rerender } = renderUseApiCall({
+      apiCall: fn,
+    })
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false), {
+      interval: 1,
+    })
+
+    rerender({ apiCall: fn2 })
+
+    await waitFor(() => expect(result.current.isLoading).toBe(true), {
+      interval: 1,
+    })
+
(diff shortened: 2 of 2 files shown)


assets/src/hooks/useApiCall.ts
```diff
@@ -0,0 +1,56 @@
...
+   * value
+   */
+  isLoading: boolean
+
+  /**
+   * The last value that {@linkcode useApiCall} resolved.
+   *
+   * Is `undefined` when it has not resolved or it resolves to `undefined`
+   */
+  result: T | undefined
```
firestack, line 18: I could change this to conditionally return `result` only after it's loaded. 
```suggestion
  result?: T
```

That would then allow changing 
```
ln29: Promise<undefined | T>
```

to

```
ln29: Promise<T>
```

firestack: I'd then need a sentinel value that isn't `undefined` for tracking the not loaded state, but I could take ideas from the result type in https://github.com/mbta/skate/pull/2531 to hide that value

assets/src/hooks/useApiCall.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @returns {ApiLoadingResult<T>}
+ */
+export const useApiCall = <T>(
+  apiCall: (abortSignal: AbortSignal) => Promise<undefined | T>
+): ApiLoadingResult<T> => {
+  const [apiResult, setApiResult] = useState<T | undefined>(undefined)
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const controller = new AbortController()
```
firestack, line 35: I found out this exists!
I'm not sure if this is better or should be preferred over using a simple `let cancel = false` definition, but I liked how this made the code a bit cleaner to understand, also allowing me to pass in a read only reference to this to the `apiCall` parameter as well!

assets/src/hooks/useApiCall.ts
joshlarson: I'm conflicted about this. If we keep it exactly as is, I would recommend calling it `useCachedApiCall`, since it's doing an API call (actually it's doing ... anything that returns a promise?), and then caching the result through the next full API round-trip.

But I sort of think that that responsibility should live in the calling code instead. Let the API code be responsible for fetching and returning data, and if I want to keep track of a result after that, then it's on me, not on the API code.

And then the type I would want to see here is:

```
interface Loading {
  loading: true
}

interface Loaded<T> {
  loading: false
  result: T
}

type ApiCallResult<T> = Loading | Loaded<T>
```

(Yeah that does lead into the annoying moment where typescript won't let you load `acr.result` unless you've somehow convinced it that it's `Loaded` instead of `Loading`...)

joshlarson: The annoying moment, which will be incredibly helpful in production code, but annoying in tests.

assets/src/hooks/useApiCall.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Calls the provided function returning a {@linkcode Promise} and tracks the
+ * loading state and the previously resolved value.
+ *
+ * @returns {ApiLoadingResult<T>}
+ */
+export const useApiCall = <T>({
+  apiCall,
+}: UseApiCallProps<T>): ApiLoadingResult<T> => {
```
firestack, line 39: I'm actually really conflicted on merging this now. We've needed something like this for a while, but if we can wait a little longer, once `use` gets out of the canary channel, we'll have a much better way of handling promises and loading states via react's `Suspense` features.
If I could figure out a way to polyfill `use` or make this interact with `Suspense` well I'd be more inclined to merge it, but I now am not sure that this is worth it, because I would prefer to rip out any usages of this afterward for `use` once it's available

https://react.dev/reference/react/use
https://react.dev/reference/react/useDeferredValue
https://react.dev/reference/react/Suspense

hannahpurcell: How long does it usually take for features to get through the canary channel?

firestack: I don't know how long it takes, but it does look like it's encouraged to use canary if you want to be ready for new features when they land or try them out early. It's more stable than experimental branches and bugs are treated the same as release channels. So maybe that's a good reason to try the canary release?

firestack: https://react.dev/blog/2023/05/03/react-canaries

firestack: [for posterity]: we decided we're uncomfortable with depending on a canary channel in an ops tool at this moment.

assets/src/hooks/useApiCall.ts
firestack: We talked about keeping it as is, do you still want a rename to `useCachedApiCall` @joshlarson ?

joshlarson: Nah I think this is fine. 

Can we explicitly have a separate state for `Loading` and `CachedLoading` though?

I think the actual thing that's throwing me off is your comment 👇 
```ts
/* 
 * Is `undefined` when it has not resolved or it resolves to `undefined`
 */
```

Having a separate type for `Loading` versus `CachedLoading` would allow for a distinction between "this is the first time we've used this API call" versus "we're loading, but the previously-cached value is `undefined`".

HOWEVER, we can proooooobably refactor to that later, so let's consider this comment non-blocking.

assets/tests/hooks/useApiCall.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    reject = rej
+  })
+
+  if (resolve === undefined || reject === undefined) {
+    throw Error("Bad")
+  }
+
+  return { promise, resolve, reject }
+}
+describe("when first rendered", () => {
```
joshlarson, line 24:
```suggestion

describe("when first rendered", () => {
```

Nit: formatting

assets/tests/hooks/useApiCall.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const PromiseWithResolvers = <T>() => {
+  let resolve: (value: T) => void = () => {}
+  let reject: (reason?: any) => void = () => {}
+  const promise = new Promise((res, rej) => {
+    resolve = res
+    reject = rej
+  })
+
+  if (resolve === undefined || reject === undefined) {
```
joshlarson, line 18: I'm trying to understand this. Will this non-deterministically detect a race condition (or something) where `resolve` or `reject` get set to `undefined`?

joshlarson (approved):

assets/tests/hooks/useApiCall.test.ts
firestack: This was just to make typescript happy since I didn't have access to [Promise.withResolvers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)

firestack: I don't actually know why I don't have access to that, especially in a _test_ environment, but I wasn't getting anywhere and didn't want to spend more time on it. I assume it has something to do with the package.json or something.

assets/src/hooks/useApiCall.ts
firestack: I tried doing this and then all of a sudden tests started acting super weird 😅

assets/tests/hooks/useApiCall.test.ts
firestack: Oh uh, I don't think I need this anymore since I tried to give `resolve` and `reject` default values to try and get rid of this `if` statement. Not sure what would be better.

assets/src/hooks/useApiCall.ts
joshlarson: WELP! Guess that moves that into refactor-land then!

joshlarson: Sorry - I meant follow-up-PR-refactor-land!

assets/tests/hooks/useApiCall.test.ts
joshlarson: Oooh I see - it's to get Typescript to recognize those values as not-`undefined`!!

This is wounding my brain, but is there any way to know that the function passed in to `Promise`'s constructor will get called before the if statement happens?

I guess I dunno whether the if statement is better, or whether the default values are better... if there's some surprise and `resolve`/`reject` don't get initialized, then the test will be easier to debug with the if statement (cause it'll fail there), although I would perhaps update the message to be something more verbose than `Error("Bad")` 😂

firestack: 😭 
```
const PromiseWithResolvers = <T>() => {
  let resolve: undefined | ((v: T) => void) = undefined
  let reject: undefined | ((v: any) => void) = undefined
  const promise = new Promise<T>((res, rej) => {
    resolve = res
    reject = rej
  })

  if (resolve === undefined || reject === undefined) {
    throw Error("Bad")
  }

  return { promise, resolve, reject }
}
```
```
    tests/hooks/useApiCall.test.ts:130:7 - error TS2349: This expression is not callable.
      Type 'never' has no call signatures.

    130       resolve("first result")
```

firestack: Gosh I wish we had `Promise.withResolvers` 😞

PR #2534, opened 2024-04-05
 refactor(ts/api/fetchDetourDirections): use `apiCallResult`

firestack:
This switches `fetchDetourDirections` over to use `apiCallResult` so that we can return errors from the backend. Switching to `apiCallResult` sets us up for https://github.com/mbta/skate/pull/2543 .

There were a few unrelated test issues due to introducing promises into the mix, so the commits should be reviewable in order to help discombobulate some of the changes.

Depends On:
- #2531
- #2533 
- #2539
- #2544

---
  - https://app.asana.com/0/0/1207091097322625
  - https://app.asana.com/0/0/1206816132107683

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -25,6 +25,8 @@ import {
   array,
   assert,
   create,
+  enums,
+  Infer,
   is,
   object,
   Struct,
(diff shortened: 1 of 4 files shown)


hannahpurcell (approved):
Ok! No concerns here 👍

PR #2537, opened 2024-04-10
inv: log ueberauth failiure struct

firestack:


lib/skate_web/controllers/auth_controller.ex
```diff
@@ -65,7 +67,14 @@ defmodule SkateWeb.AuthController do
     |> redirect(to: ~p"/")
   end
 
-  def callback(%{assigns: %{ueberauth_failure: %{provider: :keycloak}}} = conn, _params) do
+  def callback(
+        %{assigns: %{ueberauth_failure: %{provider: :keycloak} = auth_struct}} = conn,
+        _params
+      ) do
+    Logger.error(
```
joshlarson, line 74: Perhaps medium/long-term, this should go to Sentry?

Unless this is the sort of thing that happens every time you go to `skate.mbta.com` but have been logged out, in which case we probably don't want this to go to Sentry.

PR #2539, opened 2024-04-12
fix:test(ts/hooks/useDetour): add missing awaits for async

firestack:
Turns out, `waitFor` accepts either a promise or regular function, but always returns a promise. Ignoring that promise essentially means that the `expect`'s are ignored. This fixes that so our tests can be relied on.

https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

```ts
function waitFor<T>(
  callback: () => T | Promise<T>,
  [...]
): Promise<T>
```


hannahpurcell (approved):
Ohhhhh! I see! I'll try to keep this in mind. Thank you 👍 👍

PR #2542, opened 2024-04-16
feat(ts/jest): move jest config to ts file

firestack:
It's a little annoying when you have to reference docs without autocomplete to edit or modify jest config. This moves it into it's own config file that's also typescript so we can have a typed config :)


joshlarson (approved):
🎉

PR #2543, opened 2024-04-16
feat(ts/components/diversionPage): show specific error when route cannot get to point

firestack:


assets/src/components/detours/diversionPage.tsx
```diff
@@ -123,7 +123,13 @@ export const DiversionPage = ({
               Detour is not editable from this screen.
             </Alert>
           )}
-          {routingError && <RoutingErrorAlert />}
+          {routingError?.type === "no_route" && (
+            <RoutingErrorAlert>
+              You can&apos;t route to this location. Please try a different
+              point.
```
firestack, line 129: nit(non-blocking): I almost want to wrap this like
```suggestion
              You can&apos;t route to this location.
              Please try a different point.
```

Just to get github to display it nicely 😭 
image

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -121,27 +125,40 @@ describe("DiversionPage", () => {
...
+      title: "No Route",
+      directionsResult: Err({ type: "no_route" }),
+      alertError:
+        "You can't route to this location. Please try a different point.",
+    },
+    {
+      title: "Unknown",
+      directionsResult: Err({ type: "unknown" }),
+      alertError: "Something went wrong. Please try again.",
+    },
```
hannahpurcell, line 143: Ooooo nifty!

hannahpurcell (approved):
Looks great!

assets/src/components/detours/diversionPage.tsx
joshlarson: If it has to wrap in the browser, it doesn't look good.

Screenshot 2024-04-17 at 2 20 44 PM

Probably not a huge deal, especially since I imagine this will mostly be used on screens that are large enough to avoid text-wrapping, but still.

assets/src/components/detours/diversionPage.tsx
```diff
@@ -123,7 +123,13 @@ export const DiversionPage = ({
...
+            </RoutingErrorAlert>
+          )}
+          {routingError?.type === "unknown" && <RoutingErrorAlert />}
```
joshlarson, line 132: Nit: What if the logic for what message to display lived in the `<RoutingErrorAlert />` instead of the outer level? So then at the top level, it would just look like:

```tsx
{routingError && <RoutingErrorAlert error={routingError} />}
```

joshlarson (approved):

assets/src/components/detours/diversionPage.tsx
firestack: I'm not sure eager wrapping is much better... 
image

assets/src/components/detours/diversionPage.tsx
firestack: I prefer to leave it as is for now, but we could address this in a follow up PR if desired.

PR #2544, opened 2024-04-16
 fix:test(ts/detours): fix tests acting outside of test boundry 

firestack:
In #2534 I discovered quite a few tests fell over once I switched to using promises. These are changes that weren't specific to any changes I made in other PR's extracted for review (thanks `jj`!!).


hannahpurcell (approved):
❤️

PR #2547, opened 2024-04-22
feat: Prepend intersection to turn-by-turn directions

hannahpurcell:
[Asana task](https://app.asana.com/0/1205385723132845/1206879951774577)

The nearest intersection (used in the vehicle details panel to identify current bus location) will be reused in the turn-by-turn detour directions. 

This feature needs to be tested with users to decide whether that intersection direction is useful in all cases (and not misleading, if the vehicle's current orientation doesn't align with that intersection direction)


assets/src/hooks/useDetour.ts
```diff
@@ -54,6 +61,14 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
     }
   }, [routePatternId, startPoint, endPoint])
 
+  useEffect(() => {
+    if (startPoint) {
+      fetchNearestIntersection(startPoint.lat, startPoint.lon).then((result) =>
+        setNearestIntersection(result)
+      )
+    }
+  }, [startPoint])
```
firestack, line 70: suggestion: could you use `useNearestIntersection` instead of `fetchNearestIntersection`?

hannahpurcell: I considered it, but was first running into issues, and subsequently figured the hook wasn't what I needed here anyway.

First the issues: fetching the nearest intersection isn't valid until the first startPoint is placed. But conditionally running the hook is not permissible. So I could change the innards of the custom hook to skip running if the lat/lon are null.

Then I figured that the benefit of this hook is that it forms a sort of continual listener, given a dynamic vehicle position. Since placed waypoint is not dynamic, I decided that just doing a standalone fetch was preferred. What do you think?

hannahpurcell: This might be a super basic question, but I want to learn! I'm still a bit stuck on the second part of my comment, on why a hook is better here than the singular fetch? In this case, `nearestIntersection` is only dependent on startPoint and won't be dynamic, similar to how `finishedDetour` works. And in that `useEffect`, we directly fetch the finished detour status without a hook

firestack: So we don't _have_ to use `useNearestIntersection` but you've essentially just implemented it[`useNearestIntersection`] here, but with less guard rails around `useEffect` and having the state now be managed by `useDetour` instead of `useNearestIntersection`.

hannahpurcell: I thought this would be incredibly straightforward, because I actually had a branch with much of this work done (hence I planned to push the changes first thing yesterday.) Until I realized I needed a lot of test fixes, and sadly thats what I've been pulling my hair out over, yet again. 😓 Pushing up my progress thus far to see if at least the hook restructuring is on the right track. And putting it in a separate PR: https://github.com/mbta/skate/pull/2556

assets/src/components/detours/diversionPage.tsx
```diff
@@ -95,7 +103,7 @@ export const DiversionPage = ({
         <div className="l-diversion-page__panel bg-light">
           {state === DetourState.Edit && (
             <DiversionPanel
-              directions={directions}
+              directions={extendedDirections}
```
joshlarson, line 106: There is a test that validates the other `directions` -> `extendedDirections` change, but not this one. Would it be possible to add this test as well?

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -486,6 +494,10 @@ describe("DiversionPage", () => {
       routeSegments: routeSegmentsFactory.build(),
     })
 
+    const intersection = "Avenue 1 & Street 2"
+    ;(useNearestIntersection as jest.Mock)
+        .mockReturnValueOnce({ ok: intersection })
```
joshlarson, line 499: This test is failing because of `mockReturnValueOnce` - the hook is called a fairly large number of times, but after the first call, it's returning `null`, which is why the test is failing.

Also, and I'm aware that our current styles are inconsistent and nobody mentioned (until now 😇), but we're preferring to move towards the `jest.mocked` style of mocks, rather than `;(...)`. 

```suggestion
jest.mocked(useNearestIntersection).mockReturnValue({
  ok: intersection,
})
```

hannahpurcell: Ohhhhhhhhhhhhhhhhhhh!!!!! Ahh!!! Thank you so much, this just clears all the fog right up

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -70,6 +71,13 @@ beforeEach(() => {
 
 jest.mock("../../../src/api")
 
+jest.mock("../../../src/hooks/useNearestIntersection", () => ({
+  __esModule: true,
+  useNearestIntersection: jest.fn(() => {
+    return { is_loading: true }
+  }),
+}))
```
firestack, line 79: suggestion: I believe this should work without the extra options?(at least if combined with adding the mock to `beforeEach`)
(yet another piece of tech debt that makes it hard to contribute due to how often it's used)

```suggestion
jest.mock("../../../src/hooks/useNearestIntersection")
```

assets/src/hooks/useNearestIntersection.ts
```diff
@@ -17,15 +17,17 @@ export const useNearestIntersection = (
...
+          if (result) {
+            setResult({ ok: result })
+          } else {
+            setResult({ is_error: true })
+          }
         }
-      }
-    })
+      })
+    } else setResult({ ok: null })
```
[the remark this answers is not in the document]
hannahpurcell: Hmm... I see why `ok: null` is odd, but I also didn't consider this an error, because nothing has gone wrong at this point. We're just calling `useNearestIntersection` too soon, before startPoint has been set. What if I set back to initial state of `{is_loading: true}`?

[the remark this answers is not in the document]
firestack: > What if I set back to initial state of `{is_loading: true}`?

👍🏻 I think this works for me! I'd personally prefer to use `loading()` from `fetchResult` instead of the written out object.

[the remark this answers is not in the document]
hannahpurcell: I wish there was a 💡 emoji reaction on PRs, I'd use that all the time

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: The api mock was maybe accidentally removed and is still needed, but otherwise yes! It does still work without all the extra options, thank you

firestack: > The api mock was maybe accidentally removed and is still needed
Whoops yeah didn't mean to catch `api` in this suggestion

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -85,10 +90,37 @@ describe("DiversionPage", () => {
...
+    jest.mocked(useNearestIntersection).mockReturnValue({
+      ok: intersection,
+    })
+
+    const { container } = render(<DiversionPage />)
+
+    fireEvent.click(originalRouteShape.get(container))
+
+    fireEvent.click(originalRouteShape.get(container))
+
```
firestack, line 120: issue: I think both of these need `act`?
```suggestion
    act(() => fireEvent.click(originalRouteShape.get(container)))

    act(() => fireEvent.click(originalRouteShape.get(container)))
```

firestack (commented):
LGTM, one last thing about `act` but otherwise 🍰

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: Ahhhh I see! I needed to read up, and [this example](https://github.com/threepointone/react-act-examples/blob/master/sync.md#events) helped it connect for me.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -486,6 +522,11 @@ describe("DiversionPage", () => {
       routeSegments: routeSegmentsFactory.build(),
     })
 
+    const intersection = "Avenue 1 & Street 2"
+    jest.mocked(useNearestIntersection).mockReturnValue({
+      ok: intersection,
+    })
```
firestack, line 528: nit(non-blocking): would prefer the `ok` constructor from `fetchResult` instead of the object
```suggestion
    jest.mocked(useNearestIntersection).mockReturnValue(ok(intersection))
```

firestack (approved):

joshlarson (approved):

PR #2549, opened 2024-04-22
feat: Add endpoint showing the currently-deployed version

joshlarson:


lib/skate_web/controllers/version_controller.ex
```diff
@@ -0,0 +1,7 @@
+defmodule SkateWeb.VersionController do
+  use SkateWeb, :controller
+
+  def version(conn, _params) do
+    send_resp(conn, 200, Application.get_env(:skate, :version))
```
firestack, line 5: suggestion(non-blocking): IMO it's more useful to get this at compile time than runtime. I dunno about testing here though...
```suggestion
    send_resp(conn, 200, Application.compile_env(:skate, :version))
```

firestack, line 5:
```suggestion
    send_resp(conn, :ok, Application.get_env(:skate, :version))
```

joshlarson: I got 👇 error 😭 

```
(RuntimeError) Application.compile_env/3 cannot be called inside functions, only in the module body
```

joshlarson:
Seems to work - that's cool!

https://skate-dev-blue.mbtace.com/version

lib/skate_web/controllers/version_controller.ex
```diff
@@ -0,0 +1,7 @@
...
+  end
```
firestack: https://github.com/mbta/skate/pull/2549/files/cf37dce0f0efb749a5224ce7f4f892e7717dd15b#r1575376527
```suggestion
  @version Application.compile_env(:skate, :version)

  def version(conn, _params) do
    send_resp(conn, 200, @version)
  end
```

lib/skate_web/controllers/version_controller.ex
firestack: https://github.com/mbta/skate/pull/2549/files/cf37dce0f0efb749a5224ce7f4f892e7717dd15b#r1575385654

lib/skate_web/controllers/version_controller.ex
joshlarson: Yeah... that does mess up the tests though. I guess the value of testing this endpoint at all is pretty minimal though 🙃

config/config.exs
```diff
@@ -89,6 +89,7 @@ config :skate,
       }
     ]
   },
+  version: System.get_env("SENTRY_RELEASE"),
```
firestack, line 92: suggestion: I think that the version endpoint would error when [`send_resp/3`](https://hexdocs.pm/plug/Plug.Conn.html#send_resp/3) receives `nil` when this is undefined. I'd prefer to either solve this with a static value or force the variable to be set. (Unless there's other reasoning for letting this be `nil`)

```suggestion
  version: System.get_env("SENTRY_RELEASE", "dirty"),
```

or

```suggestion
  version: System.get_env("SENTRY_RELEASE", "versioned"),
```

If we simply require this all the time now then we should do
```suggestion
  version: System.fetch_env!("SENTRY_RELEASE"),
```

firestack (approved):

config/config.exs
joshlarson: My preference is to simply always require it now, so I just pulled in your last suggestion

joshlarson: Aaaand that broke the build 🤦‍♂️

PR #2551, opened 2024-04-23
fix: Make detour map respect street view mode

joshlarson:


assets/src/components/detours/detourMap.tsx
```diff
@@ -125,20 +132,29 @@ export const DetourMap = ({
...
   ] = useContext(StateDispatchContext)
 
+  const [streetViewEnabled, setStreetViewEnabled] = useState(false)
+
   return (
     <div
       className={joinClasses([
         "h-100",
-        onClickMap && "c-detour_map--map__clickable",
+        !streetViewEnabled && onClickMap && "c-detour_map--map__clickable",
```
joshlarson, line 141: I am _very_ much not a fan of this boolean goop. But this worked, I have high confidence in it, and I don't want the fix to be blocked on a refactor, which is why I PR'ed it as-is.

The refactor I'm thinking is to have `onPlaceWaypoint` be an input directly into `DetourMap` and then have `onClickMap` be computed based on `streetViewEnabled` and the rest of the detour state.

assets/src/components/detours/detourMap.tsx
```diff
@@ -103,7 +110,7 @@ export const DetourMap = ({
   waypoints,
 
   onClickOriginalShape,
-  onClickMap,
+  onAddWaypoint,
```
joshlarson, line 113: Soooooo I was a bit torn about this, but here's some of my thinking:

- Originally, we were treating this as `onClickMap`, because we didn't want `<DetourMap />` to be overly concerned with what it means to actually click the map. Basically, we were letting the mapping of "click the map" <-> "add waypoint" be `<DiversionPage />`'s responsibility.
- BUT
- There's more complicated logic now, which means that to keep `<DetourMap />` agnostic to what clicking the map means would force that logic into `<DiversionPage />` (including pushing street view stuff up into `<DiversionPage />`, which ...I don't want to do.
- And it doesn't bother me that much to have `<DetourMap />` be responsible for that. I get that having the `<Map />` be in charge of both "how do I recognize a click?" and "what should I do when you click?" is an SRP violation, and it's better to have some additional layer to separate those questions, but `<DetourMap />` IS that layer.
- Basically, I don't think it's a problem to let `<DetourMap />` know things about detours.
- That said, it would be good to give `onClickOriginalShape` the same treatment as `onClick` -> `onAddWaypoint`. Right now, `onAddWaypoint` and `onClickOriginalShape` are sort of living at two separate layers of abstraction, which is not the best thing in the world.
- I didn't do that here for two reasons. First, I didn't need to. And second, `onClick` was a bit more nicely encapsulated, which made this change easy. `onClickOriginalShape` (and its friends `startPoint === null` / `endPoint === null`) is (are) more tangled, and I'd rather muck around with them as a follow up.

assets/src/components/detours/detourMap.tsx
```diff
@@ -70,7 +77,7 @@ interface DetourMapProps {
    * Callback fired when the map is clicked.
    * @param point
```
hannahpurcell, line 78: Your description of the considerations you made while renaming this function was very helpful. Thanks for outlining it all. I may be easily swayed, but it does seem sound to me! I'd adjust this comment to suit the new purpose.

hannahpurcell (approved):
Works well 👍 👍

assets/src/components/detours/detourMap.tsx
```diff
@@ -67,10 +74,10 @@ interface DetourMapProps {
   onClickOriginalShape: (point: ShapePoint) => void
 
   /**
-   * Callback fired when the map is clicked.
+   * Callback fired when the map tries to add a waypoint.
```
joshlarson, line 77: @hannahpurcell updating this comment is a good call! What do you think of this?

PR #2554, opened 2024-04-24
feat: Redirect to the page you navigated to on successful auth

joshlarson:
---
  - https://app.asana.com/0/0/1207128168281466


lib/skate_web/plugs/capture_auth_return_path.ex
```diff
@@ -0,0 +1,27 @@
+defmodule SkateWeb.Plugs.CaptureAuthReturnPath do
```
firestack, line 1: question(non-blocking): Any thoughts on how to document this?

lib/skate_web/plugs/capture_auth_return_path.ex~
firestack, line 1: issue: This seems accidental

test/skate_web/controllers/auth_controller_test.exs
```diff
@@ -43,6 +43,36 @@ defmodule SkateWeb.AuthControllerTest do
...
+            }
+          }
+        }
+      }
+
+      conn =
+        conn
+        |> assign(:ueberauth_auth, mock_auth)
+        |> init_test_session(%{post_auth_return_to: ~p"/detours"})
+        |> get(~p"/auth/keycloak/callback")
```
firestack, line 70: suggestion(non-blocking): I wonder if there's a way to test this without "mocking" the redirect path, and instead attempt to navigate to `/detours`, get redirected, go though login, and then assert we end up at `/detours`?

lib/skate_web/plugs/capture_auth_return_path.ex
firestack: love it!

lib/skate_web/plugs/capture_auth_return_path.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  use SkateWeb, :verified_routes
+
+  def init(default), do: default
+
+  def call(conn, _default) do
+    conn
+    |> fetch_session()
+    |> put_session(:post_auth_return_to, request_url(conn))
+  end
```
firestack, line 17: suggestion: can we add the `@impl` attrs and `@behavoir` for this?

```suggestion
  @behaviour Plug

  @impl Plug
  def init(default), do: default

  @impl Plug
  def call(conn, _default) do
    conn
    |> fetch_session()
    |> put_session(:post_auth_return_to, request_url(conn))
  end
```

I'm not sure if this will compile or pass tests okay, but if it does I'd prefer it.

firestack (commented):
a few comments to address but otherwise LGTM!

lib/skate_web/plugs/capture_auth_return_path.ex
```diff
@@ -0,0 +1,55 @@
...
+  @moduledoc """
+  Set a cookie value `:post_auth_return_to` with the current request url,
+  and allow a mechanism to get that url easily.
+  """
+
+  import Plug.Conn
+
+  use SkateWeb, :verified_routes
+
+  def init(default), do: default
```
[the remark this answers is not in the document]
firestack: Good to know! I was hoping they'd provide a default impl, too bad.

firestack (approved):

test/skate_web/controllers/auth_controller_test.exs
joshlarson: Follow-up task: https://app.asana.com/0/1152340551558956/1207199406216195/f

I assigned it to myself so that it stays on my radar, but I'm not attached to _me_ being the one to do it.

PR #2559, opened 2024-04-26
fix: Drop waypoints in the right place

joshlarson:
---
  - https://app.asana.com/0/0/1207103959249126

### DIFF
--- a/assets/css/detours/_detour_map.scss
+++ b/assets/css/detours/_detour_map.scss
@@ -3,7 +3,7 @@
 
 .c-detour_map--map__clickable {
   cursor: url("data:image/svg+xml,%3Csvg width='14' height='14' viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cstyle%3E .color %7B fill: %236D39AC; %7D %3C/style%3E%3Ccircle cx='7' cy='7' r='6' class='color' style='filter: blur(0.5px)'/%3E%3Ccircle cx='7' cy='7' r='4.5' class='color' stroke='white'/%3E%3C/svg%3E")
-      11 11,
+      7 7,
     pointer;
 
   .c-vehicle-map {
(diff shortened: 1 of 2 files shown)


assets/src/components/detours/detourMap.tsx
```diff
@@ -320,7 +320,7 @@ const WaypointMarker = ({ position }: { position: LatLngLiteral }) => (
     position={position}
     divIconSettings={{
       iconSize: [10, 10],
-      iconAnchor: new Leaflet.Point(5, 5),
+      iconAnchor: new Leaflet.Point(5, 10),
```
joshlarson, line 323: I'm gonna be honest - I don't fully understand why this is necessary, and I think there's a chance that this is actually just masking (or correcting for?) an SVG error elsewhere, but I did a (-n admittedly small) bit of digging and wasn't able to figure it out.

assets/css/detours/_detour_map.scss
```diff
@@ -3,7 +3,7 @@
 
 .c-detour_map--map__clickable {
   cursor: url("data:image/svg+xml,%3Csvg width='14' height='14' viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cstyle%3E .color %7B fill: %236D39AC; %7D %3C/style%3E%3Ccircle cx='7' cy='7' r='6' class='color' style='filter: blur(0.5px)'/%3E%3Ccircle cx='7' cy='7' r='4.5' class='color' stroke='white'/%3E%3C/svg%3E")
-      11 11,
```
joshlarson, line 6: Oops!

assets/src/components/detours/detourMap.tsx
firestack: Based on the docs, I think we should get rid of `iconAnchor` altogether (maybe leaving a comment that it's not specified because we want it centered

> [...] Centered by default if size is specified [...]
>
> -- https://leafletjs.com/reference.html#divicon-iconanchor

firestack (approved):

PR #2560, opened 2024-04-26
cleanup:test: rewrite jest mocks to use `jest.mocked`

firestack:
There's more to do, but as an incremental fix, I think this is good enough for 15m of time.

I ran these commands, and then reverted tests that started failing.

```
, ast-grep run -p '($V as jest.Mock)' -r 'jest.mocked($V)' ./assets/ -U

, ast-grep run -p '($V as jest.Mock<$_>)' -r 'jest.mocked($V)' ./assets/ -i
```


joshlarson (approved):
I basically just looked through to see if I could find any obvious mistakes, which I couldn't, and also the builds pass.

Thanks for doing this!

PR #2562, opened 2024-04-27
feat: add mbta asana workflow

firestack:
I don't really have a way to experiment or test this till this file exists on main, I could strip this back so I can do more testing in a PR first.

This workflow is used on a few of our projects: https://github.com/search?q=org%3Ambta+asana.yml&type=code

---

Asana Ticket: https://app.asana.com/0/0/1207185409019452/f

### DIFF
--- a/asana.yml
+++ b/asana.yml
@@ -1,0 +1,20 @@
+name: Asana integration for GitHub PRs
+on:
+  workflow_dispatch:
+  pull_request:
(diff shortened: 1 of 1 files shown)


hannahpurcell (approved):
Oooo nice! I hadn't seen the workflow after it went through a lot of changes. This is cool

PR #2570, opened 2024-05-01
feat: Consolidate 3 modals into BaseNotificationModal

hannahpurcell:
The JSX of the Chelsea Bridge modals seemed structurally copy-pasted from the Inactive Notification Modal (the classnames still referred to the Inactive Notif Modal, which seemed misleading), so it seemed logical to consolidate into BaseNotificationModal before converting that modal to the Restart UI component.

Asana Ticket: https://app.asana.com/0/1205385723132845/1206770630583768


assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
```diff
@@ -1,10 +1,18 @@
...
 import { initialState, setNotification } from "../../../src/state"
 import ChelseaLoweredNotificationModal from "../../../src/components/notificationModals/chelseaLoweredNotificationModal"
+import ReactDOM from "react-dom"
+
+jest.mock("react-dom", () => {
+  return {
+    ...(jest.requireActual("react-dom") as typeof ReactDOM),
+    createPortal: (node: ReactNode): ReactPortal => node as ReactPortal,
+  }
+})
```
firestack, line 15: question: what happens without this mock?

assets/src/components/notificationModals/basicNotificationModal.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div className="c-basic-notification-modal">
+        {/* TODO: Close button IS tab-able, but not visually given focus */}
+        <button
+          title="Close"
+          className="c-basic-notification-modal__close-button"
+          onClick={closeModal}
+        >
+          <OldCloseIcon />
+        </button>
+        <div className="c-notification__title">{title}</div>
```
firestack, line 37: question: is this intended to be `.c-notification__title`? If this style is apart of the `.c-basic-notification-modal` component, I think we should move this styling into a `.c-basic-notification-modal`, but if the restart ui modal ends up throwing away any of this, I'd be open to making this non-blocking.

firestack: Oh I didn't realize the other modals did this too :melt:

firestack: Rethinking this some more, consider this non-blocking for now.

assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
firestack: Hmm, seems the issue is related to our use of `import renderer from "react-test-renderer"` in some of these tests?
It doesn't seem like all of the tests which had this added need it.

[undocumented :c] We're trying to move off `react-test-renderer` fwiw, so if you wanted to go ahead and do that conversion I'd be in support of that. Happy to pair if you'd like. I can also fix this if you'd prefer.

hannahpurcell: Oh good catch! I missed that the chelseaLoweredNotificationModal didn't need it. Removed it there, and now I'm looking into what moving off `react-test-renderer` in the other 3 files will entail.

firestack: for posterity, this is the `render` method we've preferred to use recently `import { render } from "@testing-library/react"`

to get snapshots to work with this, you'll likely need to use `asFragment()` or `baseElement` as the return type is different between the two render methods.

hannahpurcell: Ok so. I did test out using render from "@testing-library/react", but ran into a snag, covered in a SO issue [here](https://stackoverflow.com/questions/69240360/react-jest-snapshot-only-returning-empty-documentfragment). Basically, the usually `asFragment` mechanics don't play nicely with modals, which are rendered separately from the render functions container. So I'm trying out a couple things to figure out if this is something to do now or in another task. I'm happy to do it, but just trying to figure out scope

assets/src/components/notificationModals/basicNotificationModal.tsx
hannahpurcell: Ah I see. I don't see any styles attached to this selector anyway, so I'd be open to just eliminating the className entirely

assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
firestack: Ah, if the content of `asFragment` doesn't have any of the modal in it, then you'd likely want `baseElement` which points to the `<body>` iirc.

assets/tests/components/notificationModals/inactiveStateModal.test.tsx
```diff
@@ -69,18 +69,18 @@ describe("InactiveNotificationModal", () => {
...
   test("renders for a notification with no runs", () => {
     ;(useMinischeduleRuns as jest.Mock).mockImplementationOnce(() => [])
-    const tree = renderer
-      .create(<InactiveNotificationModal notification={notification} />)
-      .toJSON()
-    expect(tree).toMatchSnapshot()
+    const result = render(
+      <InactiveNotificationModal notification={notification} />
+    )
+    expect(result.baseElement).toMatchSnapshot()
```
firestack, line 83: suggestion(non-blocking): I don't think it's worth blocking over, but [Kent C Dodd's "Common mistakes with React Testing Library"](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-wrapper-as-the-variable-name-for-the-return-value-from-render) recommends

> Advice: destructure what you need from render or call it view.

I usually prefer to destructure the needed items

```suggestion
    const { baseElement } = render(
      <InactiveNotificationModal notification={notification} />
    )
    expect(baseElement).toMatchSnapshot()
```

assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
hannahpurcell: Oh I forgot to make the same changes with modal. Might as well!

firestack (approved):

firestack (approved):
Approving again to try and get the asana task to run properly

assets/src/components/notificationModals/basicNotificationModal.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useContext } from "react"
+import { Modal } from "@restart/ui"
+import { setNotification } from "../../state"
+import { StateDispatchContext } from "../../contexts/stateDispatchContext"
+import { OldCloseIcon } from "../../helpers/icon"
+
+const BasicNotificationModal = ({
```
joshlarson, line 7: Commenting a bit late here, but did you mean for this to be `Basic` or `Base`?

I only ask because the title of the PR says `Base`, and that's what I would normally have expected.

PR #2571, opened 2024-05-01
feat: Part 2 of converting modals to Restart UI - Disconnected & Loading modals

hannahpurcell:
Directed to merge onto the first branch

https://app.asana.com/0/1205385723132845/1206770630583768


assets/src/components/disconnectedModal.tsx
```diff
@@ -1,18 +1,22 @@
 import React from "react"
 import { reload } from "../models/browser"
+import { Modal } from "@restart/ui"
 
 const DisconnectedModal = () => (
-  <>
-    <div className="c-modal">
+  <Modal
```
joshlarson, line 6: Suggestion: Each place where we're using restart's `Modal`, all of the props are duplicated. What if we created our own custom modal component that abstracted this?

I would recommend calling it `<SkateModal />` or something like that, not because I think that's a good name, but because [we already have a component](https://github.com/mbta/skate/blob/9ffd12145c842c46b1a4f9058a1e2e28634a5e62/assets/src/components/modal.tsx) called `<Modal />`.


(Honestly, the existing `<Modal />` should probably be called something else, because its role in the app is to _always_ be there, and to conditionally take over the app, rendering a modal on top of Skate, but coming up with better names for that is beyond the scope of this PR, IMO)

joshlarson:
Like this so far!

I'm holding off on approving for now because I'm interested in your thoughts on [this comment](https://github.com/mbta/skate/pull/2571/files#r1588266682), and also because the build is currently failing (there are a few test files that need `ReactPortal` and `ReactNode` imported, I think)

assets/src/components/disconnectedModal.tsx
hannahpurcell: In the modals converted so far, this is true that the props are duplicated. But not so for a few additional modals I'm converting now, though they are very close to the same. That said, I think given the slight prop variation between them, I'd prefer to keep the component invocations as-is, or at least defer a consolidation effort

hannahpurcell: Oh darnit, this gave me an idea 🫠

hannahpurcell: Ok I was wondering how many modals I should split out to a separate generic modal. Here's my thinking: there are three categories of modal. Notification modals, which use the basic Restart modal, basic class stylings for "c-modal", and extra styling from "_basic_notification_modal". Preset modals, which use a generic InputModal component that can be converted to the Restart modal, but separately from the other modals because it uses a special "c-input-modal" styling. And then "Other modals" to include disconnected and loading modals, which are just styled with bare bones "c-modal". I think there's some work that can be done here to consolidate the notification and "other modals" because I don't think the styling of those are very different, but that could lead to visual changes and a conversation with design. So for now, I'd kinda like to defer that refactor.

For the time being: if I make a new parent component to represent a sort of `<SkateModal />`, then `<InactiveNotificationModal ... />` would include `<BasicNotificationModal ... />`, which would include `<SkateModal .../>` which would be an extremely simple component (simple is good, but that's an extra layer of nesting than I prefer.) And `<SkateModal />` would only be used in 3 places atm. I dunno, maybe these are fine, the file navigation would just get bothersome to me?

joshlarson: Part of me thinks that we might be able to avoid at least some of the indirection you mentioned with judicious use of props - `<SkateModal inactive />`, for instance - but I agree that we don't need to block this particular PR on that.

joshlarson (approved):

assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
```diff
@@ -1,10 +1,18 @@
...
 import { initialState, setNotification } from "../../../src/state"
 import ChelseaLoweredNotificationModal from "../../../src/components/notificationModals/chelseaLoweredNotificationModal"
+import ReactDOM from "react-dom"
+
+jest.mock("react-dom", () => {
+  return {
+    ...(jest.requireActual("react-dom") as typeof ReactDOM),
+    createPortal: (node: ReactNode): ReactPortal => node as ReactPortal,
+  }
+})
```
firestack, line 15: question: merge conflict?

assets/tests/components/__snapshots__/modal.test.tsx.snap
```diff
@@ -143,10 +153,6 @@ exports[`Modal renders loading modal when appropriate 1`] = `
         Loading...
       </div>
     </div>
-    <div
-      aria-hidden="true"
-      class="c-modal-backdrop"
-    />
```
firestack, line 149: The [new backdrop](https://github.com/mbta/skate/pull/2571/files#diff-a5d8b215b8b92f35520456915b39141b327d3f33bf71425344cbb88248bb3f8fR129-R131) seems to be missing `aria-hidden=true`?

hannahpurcell: Ooooooo! Ok fantastic catch. This is important and I'm surprised Restart UI doesn't have the nice aria features built-in.  Do you know why we moved towards using Restart UI as opposed to component libraries that maybe have more of the accessibility framework? It's been a while since I've worked on keeping an app accessible, so it might be so easy to incorporate that we don't care to have a library do it, but I would guess that it'd be nice to let a library do some of the lifting?

firestack: IIRC We want to move to Restart UI to get rid of the management code of opening and closing a modal and managing focus states without needing to fully turn ourselves over to a specific modal yet. I'm not sure why the library doesn't add `aria-hidden` here, I'm assuming there may be some info I'm not aware of and maybe aria-hidden isn't needed, but I assume that since we're in control of how the backdrop is rendered it's up to us and not really the library.

assets/tests/components/notificationModals/chelseaLoweredNotificationModal.test.tsx
hannahpurcell: toooootally 🫠

assets/tests/components/__snapshots__/modal.test.tsx.snap
hannahpurcell: That's fair. I checked the accessibility tree (had to learn how to do that) and sure enough, the backdrop is not getting hidden at all, so great catch, and easy enough inline fix

firestack (approved):

PR #2577, opened 2024-05-03
fix: Block dummy detour page unless the test group is enabled

joshlarson:
What's here is only a partial solution, because navigating to `/detours` will still give you a nav menu and a blank screen (rather than an actual, fully functioning dummy detour page). I think an "Actual Solution" would involve fiddling with the routes in `router.ex`, or would involve creating a real 404 page in the frontend, and I'm not sure that's worth the effort right now.


firestack (approved):
Perfect, this is what we did in the past for search maps too.

firestack:
Feel free to merge, no changes since I reviewed.

PR #2578, opened 2024-05-03
feat: Part 3 of converting to Restart modals: Preset modals

hannahpurcell:
I canNOT figure out this test, once again. I've tried changing the target to the result container, baseElement, and asFragment() with no luck. The escape key should trigger the onHide, which should trigger closeInputModal...


assets/src/components/mapPage/vehiclePropertiesCard.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { ReactNode, useId } from "react"
 import { joinTruthy, joinClasses } from "../../helpers/dom"
 import { useCurrentTimeSeconds } from "../../hooks/useCurrentTime"
-import { useNearestIntersection } from "../../hooks/useNearestIntersection"
+import { useNearestIntersectionFetchResult } from "../../hooks/useNearestIntersection"
```
joshlarson, line 4: Did this PR pull in some unrelated changes? Would a rebase fix this?

hannahpurcell: Therrrrrrre we go. Not sure how I got things entangled

joshlarson: Git'll do that to ya!

assets/src/components/inputModal.tsx
```diff
@@ -9,22 +10,18 @@ const InputModal = ({
...
-      <div className="c-input-modal-backdrop" aria-hidden={true} />
-    </>
+    <Modal
+      className="c-input-modal"
+      onHide={() => {
+        dispatch(closeInputModal())
+      }}
+      show
+      renderBackdrop={(props) => (
+        <div {...props} className="c-modal-backdrop" aria-hidden />
```
joshlarson, line 20: This may be a silly question, but the `className` changed from `c-input-modal-backdrop` to `c-modal-backdrop`. Does that make a difference?

assets/tests/components/inputModal.test.tsx
```diff
@@ -1,24 +1,25 @@
...
-import { render, fireEvent } from "@testing-library/react"
+import { render, waitFor } from "@testing-library/react"
 import InputModal from "../../src/components/inputModal"
 import { StateDispatchProvider } from "../../src/contexts/stateDispatchContext"
 import { initialState, closeInputModal } from "../../src/state"
+import userEvent from "@testing-library/user-event"
 
 describe("InputModal", () => {
-  test("can be dismissed with escape key", () => {
+  test("can be dismissed with escape key", async () => {
```
joshlarson, line 10: Non-blocking nit: (I actually prefer for this to be addressed in a follow-up PR, so _very_ non-blocking) This test asserts something that we don't care about, which is that a particular function gets called. If it can be done without too much trouble, I'd rather have the test assert that the modal disappears from the screen.

... part of why I would rather have that be a follow-up is that this is a good change, and my suggestion ☝️ is the sort of thing that... usually is pretty easy, but occasionally winds up sending one down a horrible rabbit hole, and I don't want this PR blocked on said rabbit hole.

joshlarson (approved):

assets/src/components/inputModal.tsx
hannahpurcell, line 20:
```suggestion
        <div {...props} className="c-input-modal-backdrop" aria-hidden />
```

PR #2580, opened 2024-05-03
tests: Lift useDetour tests to diversionPage

joshlarson:
~Mostly creating this PR (for now) to provide visibility into _how_ I'm approaching this.~

~I'll also use it at the end in order to keep the diff as small as possible (for instance, I don't really want to merge re-ordered imports or random newline creations or deletions)~

🎉 This is ready for review now 🎉 

Asana Ticket: https://app.asana.com/0/1148853526253420/1206972823536566/f


assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -129,30 +213,57 @@ describe("DiversionPage", () => {
...
+    fireEvent.click(originalRouteShape.get(container))
+
+    await waitFor(() => {
+      expect(
+        container.querySelector(".c-detour_map--map__clickable")
+      ).not.toBeInTheDocument()
+    })
+  })
+
+  test("when end point has been set, finish detour button is visible", async () => {
```
joshlarson, line 255: Sometimes Git gets things amusingly wrong. This test isn't new.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -90,7 +90,82 @@ describe("DiversionPage", () => {
...
+    act(() => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+    act(() => {
+      fireEvent.click(container.querySelector(".c-vehicle-map")!)
+    })
+
+    await waitFor(() => {
+      expect(
+        container.querySelectorAll(".c-detour_map-circle-marker--detour-point")
```
joshlarson, line 147: I think a lot of these tests could be cleaned up quite a lot with more judicious use of [test selectors](https://github.com/mbta/skate/blob/eb1aa60ba2f73eb5d227eee61c07823635491f59/assets/tests/testHelpers/selectors/components/detours/diversionPage.tsx)

Then this, for instance, could read

```tsx
espect(waypointMarker.getAll(container)).toHaveLength(1)
```

The more boilerplate, like these hard-coded selectors, we get out of the meat of the tests, the easier they'll be to work with, and the more reliable their coverage will be.

---

I'd like to do that, but I'd like to do it as a follow-up task.

assets/src/hooks/useDetour.ts
```diff
@@ -105,7 +105,7 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   const canUndo = startPoint !== null && state === DetourState.Edit
 
   const undo = () => {
-    if (!canUndo) return
+    // if (!canUndo) return
```
joshlarson, line 108: This was dead code - `undo` (and `clear`) was never possible to call if `canUndo` wasn't `true`.

assets/src/hooks/useDetour.ts
```diff
@@ -79,9 +79,9 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
     detourShape.result && isOk(detourShape.result)
       ? detourShape.result.ok.directions
       : undefined
-
-  // Only append direction "Regular Route" after detour is finished
-  if (!detourShape.isLoading && directions && finishedDetour) {
+  // // Only append direction "Regular Route" after detour is finished
+  // if (!detourShape.isLoading && directions && finishedDetour) {
+  if (directions && finishedDetour) {
```
joshlarson, line 84: This is the sort of thing that's tricky enough to test that I could be persuaded to add the `!detourShape.isLoading` condition back in untested at least for now, but I want to think a bit harder about what kind of test or tests might lead to that condition mattering.

assets/src/hooks/useDetour.ts
```diff
@@ -153,8 +153,8 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
      * Sets {@link startPoint} if unset.
      * Otherwise sets {@link endPoint} if unset.
      */
-    addConnectionPoint:
-      state === DetourState.Finished ? undefined : addConnectionPoint,
+    addConnectionPoint,
+    //     state === DetourState.Finished ? undefined : addConnectionPoint,
```
joshlarson, line 157: I _think_ this was dead code. `addConnectionPoint` is only callable when the original route shape is clickable, which isn't true in `DetourState.Finished`.

assets/src/hooks/useDetour.ts
```diff
@@ -213,14 +213,18 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
     /**
      * Removes the last waypoint in {@link waypoints} if {@link canUndo} is `true`.
      */
-    undo: state === DetourState.Finished ? undefined : undo,
+    undo,
```
joshlarson, line 216: Dead code: `canUndo` is `false` in `DetourState.Finished`, which means that these functions are locked behind disabled buttons.

assets/src/hooks/useDetour.ts
```diff
@@ -213,14 +213,18 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
...
      */
-    clear: state === DetourState.Finished ? undefined : clear,
+    clear,
+    //   clear: state === DetourState.Finished ? undefined : clear,
+
     /** When present, puts this detour in "finished mode" */
     finishDetour: endPoint !== null ? finishDetour : undefined,
     /** When present, puts this detour in "edit mode" */
-    editDetour: state === DetourState.Finished ? editDetour : undefined,
+    editDetour,
```
joshlarson, line 227: Dead code: `editDetour` is hidden behind a button that's only visible in `DetourState.Finished`.

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -57,7 +57,7 @@ export const DiversionPanel = ({
           </div>
         </section>
 
-        <section className="pb-3">
+        <section id="diversion-panel__directions" className="pb-3">
```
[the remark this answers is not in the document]
joshlarson: Can we put a `title` prop on the section itself? Would that enable `getByTitle` to work?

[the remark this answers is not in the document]
joshlarson: Would a `title` prop be better than `data-test-id`?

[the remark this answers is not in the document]
firestack: fwiw, I try to avoid any of the other queries provided by RTL except in specific circumstances where it's unavoidable. I specifically try to avoid test id's and `getByTitle`
> `getByTitle`: The `title` attribute is not consistently read by screenreaders, and is not visible by default for sighted users
> 
> -- https://testing-library.com/docs/queries/about/#priority

I **_always_** reach for `byRole` first (double quoting for emphasis)
> _This should be your top preference for just about everything. There's not much you can't get with this (if you can't, it's possible your UI is inaccessible)._
>
> > `getByRole`: This can be used to query every element that is exposed in the [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/AOM). With the `name` option you can filter the returned elements by their [accessible name](https://www.w3.org/TR/accname-1.1/). This should be your top preference for just about everything. There's not much you can't get with this (if you can't, it's possible your UI is inaccessible). Most often, this will be used with the `name` option like so: `getByRole('button', {name: /submit/i})`. Check the [list of roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles).
> >
> >-- https://testing-library.com/docs/queries/about/#priority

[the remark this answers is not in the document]
firestack: A counterpoint to that option, you often mention that refactoring code should not result in failing tests. If all we wanted was an extra container to change the styling on the header, the `parentElement` approach would immediately break.

```
<h2 className="c-diversion-panel__h2">Detour Directions</h2>
```
=>
```
<div><h2 className="c-diversion-panel__h2">Detour Directions</h2></div>
```

You could get around this with a `.closest("section")`, but then if we changed `section` to say `article`, it would also fail.

Not necessarily trying to persuade you away from that, I don't mind updating tests in a case like this, but I want to make sure we have all the "facts" on the table.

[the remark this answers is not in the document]
joshlarson: I'm super torn here. I don't think there's an answer that _fully_ works the way I want it to, and I'm still new enough to RTL that I don't have great instincts on this question. I get the sense that the two "best" (least bad?) approaches are:
 - Use a `data-test-id`, which will likely _work_, but leaks some test stuff into the code.
 - Use `.parentElement` or `.closest("section")`, which will mostly work, but does have the risk of non-user-facing DOM-adjustments breaking these tests

@mbta/skate-developers do you all have a preference between the two options above? (Or option 3: "other")?

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -479,6 +720,31 @@ describe("DiversionPage", () => {
     expect(screen.getByRole("button", { name: "Edit Detour" })).toBeVisible()
   })
 
+  test("'Share Detour Details' screen returns to editing screen when edit detour button is clicked", async () => {
```
hannahpurcell, line 723: Any other validation needed here to ensure that returning to the Edit Detour page does indeed allow users to edit the detour? Maybe verify the undo and clear buttons are active again?

hannahpurcell (commented):
Cool! Thoughts so far

assets/tests/hooks/useDetour.test.ts
```diff
@@ -149,70 +81,11 @@ describe("useDetour", () => {
...
-    act(() => result.current.undo?.())
-
-    await waitFor(() => {
-      expect(jest.mocked(fetchDetourDirections)).toHaveBeenCalledTimes(3)
-      expect(jest.mocked(fetchDetourDirections)).toHaveBeenNthCalledWith(3, [
-        start,
-        mid,
-      ])
-    })
-  })
```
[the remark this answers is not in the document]
joshlarson: Ooooh good catch! I do appear to have dropped that one.

assets/tests/hooks/useDetour.test.ts
```diff
@@ -293,65 +127,10 @@ describe("useDetour", () => {
...
   })
 
-  test("when `startPoint` is null, `canUndo` is `false`", async () => {
-    const { result } = renderHook(useDetourWithFakeRoutePattern)
-
-    await waitFor(() => {
-      expect(result.current.startPoint).toBeNull()
-      expect(result.current.canUndo).toBe(false)
-    })
-  })
```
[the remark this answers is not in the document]
joshlarson: Agreed - thanks for catching that!

assets/tests/hooks/useDetour.test.ts
```diff
@@ -149,70 +81,11 @@ describe("useDetour", () => {
...
-      expect(jest.mocked(fetchDetourDirections)).toHaveBeenCalledTimes(3)
-      expect(jest.mocked(fetchDetourDirections)).toHaveBeenNthCalledWith(3, [
-        start,
-        mid,
-      ])
-    })
-  })
-
-  test("when `undo` removes the last waypoint, `detourShape` and `directions` should be empty", async () => {
+  test("clears `detourShape` when `undo` removes the last waypoint", async () => {
```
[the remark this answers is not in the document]
joshlarson: This test case got moved to `"clicking on 'Undo' restores the placeholder text when appropriate"` in `diversionPage.test.tsx`.

I didn't move assertions about `routeSegments` or `detourShape` because I couldn't figure out how to make those assertions at the `diversionPage` level. I think it's _gotta_ be possible, but I spun my wheels for a bit and decided that it was a problem for a different day.

https://github.com/mbta/skate/pull/2580/files#diff-1535a72a38a9500ad346d1b3d9735aff2c788a2d97cfbf49f2aeaac7aefb26cfR366-R399

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -320,21 +474,23 @@ describe("DiversionPage", () => {
...
-    jest.mocked(fetchDetourDirections).mockResolvedValue(
-      Ok(
-        detourShapeFactory.build({
-          directions: [
-            { instruction: "Turn left on Main Street" },
-            { instruction: "Turn right on High Street" },
-            { instruction: "Turn sharp right on Broadway" },
-          ],
-        })
+    jest.mocked(fetchDetourDirections).mockReturnValue(
```
joshlarson, line 477: Oh I should go through and change these back. I had a misconception about `mockResolvedValue`, which is why I changed these, but this is just diff noise at this point.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -90,7 +90,82 @@ describe("DiversionPage", () => {
...
+      ).toBeInTheDocument()
+    })
+  })
+
+  test("has no waypoints at the beginning or when placing the start point", async () => {
+    const { container } = render(<DiversionPage />)
+
+    expect(
+      container.querySelectorAll(".c-detour_map-circle-marker--detour-point")
+    ).toHaveLength(0)
```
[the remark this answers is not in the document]
joshlarson: I think it would be valuable to test that (and thus it shall be done), but I don't think that behavior is governed by the existence of `finishDetour`, so I'm not sure this is coverage we lost (although I agree that it's coverage we should have)

I think the existence of `finishDetour` governs whether the `"Finish Detour"` button appears, which is tested (implicitly) [here](https://github.com/mbta/skate/pull/2580/files#diff-1535a72a38a9500ad346d1b3d9735aff2c788a2d97cfbf49f2aeaac7aefb26cfR261-R262)

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: I'm not opposed!

I think this is the sort of thing that shows my ... sort of point 2 of this type of work, that we may want to break this test file up into smaller ones that validate different use cases and/or different parts of the screen. Then it becomes clearer which bits of user-facing coverage are still missing, and makes the test files a lot less cluttered.

assets/src/components/detours/diversionPanel.tsx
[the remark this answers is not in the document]
firestack: No preference here

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: https://app.asana.com/0/1152340551558956/1207337422166095/f

assets/src/hooks/useDetour.ts
joshlarson: Going to add this back in and follow up with improving the coverage [in a later ticket](https://app.asana.com/0/1152340551558956/1207337422166099/f)

assets/src/components/detours/diversionPanel.tsx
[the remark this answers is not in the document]
joshlarson: I went with `.closest("section")`. Hopefully that's not too brittle!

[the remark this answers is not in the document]
joshlarson: [ Feedback: Remove id's from DiversionPanel in favor of more sophisticated selectors](https://github.com/mbta/skate/pull/2580/commits/83c46b58515f162614b887fdbeef33e697545372)

assets/src/hooks/useDetour.ts
joshlarson: [Feedback: Restore check for !detourShape.isLoading ](https://github.com/mbta/skate/pull/2580/commits/e851bd6b972ee3688aacca0efd8fac3ea752cba2)

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -129,30 +213,57 @@ describe("DiversionPage", () => {
...
+      expect(screen.getByTitle("Detour End")).not.toBeNull()
+    })
 
     fireEvent.click(originalRouteShape.get(container))
 
-    expect(screen.getByRole("button", { name: "Finish Detour" })).toBeVisible()
+    await waitFor(() => {
+      expect(screen.getAllByTitle("Detour Start")).toHaveLength(1)
+      expect(screen.getAllByTitle("Detour End")).toHaveLength(1)
+    })
```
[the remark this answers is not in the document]
joshlarson: Good call - just pushed [Feedback: Add missing coverage around clicking after the detour is finished](https://github.com/mbta/skate/pull/2580/commits/44d802e5c4e6492947e5fd48c8a6e87791b13f78)

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: [Feedback: Restore (correct) usages of mockResolvedValue](https://github.com/mbta/skate/pull/2580/commits/d34c0a0dfa11cdb8e586f6110f5e685e47fb701f)

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -479,6 +763,31 @@ describe("DiversionPage", () => {
...
+    await userEvent.click(screen.getByRole("button", { name: "Edit Detour" }))
+
+    await waitFor(() => {
+      expect(
+        screen.queryByRole("button", { name: "Edit Detour" })
+      ).not.toBeInTheDocument()
+      expect(
+        screen.getByRole("button", { name: "Finish Detour" })
+      ).toBeVisible()
+    })
```
firestack, line 788: suggestion: does this work?
```suggestion
    await userEvent.click(await screen.findByRole("button", { name: "Edit Detour" }))

```

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -433,7 +670,39 @@ describe("DiversionPage", () => {
...
+    await waitFor(() => expect(screen.getAllByText(stop.name)).toHaveLength(1))
+  })
+
+  test("missed stops are absent after a completed detour is 'undo'ne", async () => {
+    const stop1 = stopFactory.build()
+    const stop2 = stopFactory.build()
+    jest
+      .mocked(fetchFinishedDetour)
+      .mockResolvedValue(
+        finishedDetourFactory.build({ missedStops: [stop1, stop2] })
```
firestack, line 682: nit(non-blocking)
```suggestion
    jest
      .mocked(fetchFinishedDetour)
      .mockResolvedValue(
        finishedDetourFactory.build({ missedStops: stopFactory.buildList(2) })
```

assets/tests/hooks/useDetour.test.ts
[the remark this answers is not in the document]
joshlarson: [Feedback: Add missing coverage validating when we call the API](https://github.com/mbta/skate/pull/2580/commits/3793bb5d1904a7d70674ad717a547a5b534f206f)

assets/tests/hooks/useDetour.test.ts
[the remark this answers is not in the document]
joshlarson: [Feedback: Add missing coverage that 'Undo' and 'Clear' get disabled when the detour is fully cleared out](https://github.com/mbta/skate/pull/2580/commits/e8b491b850dfb2ad8d0f3f7b5aa8ddf91f7e90bc)

☝️ adds both the use case you mentioned, and also a similar one when the detour is cleared because of the `Clear` button.

assets/tests/components/detours/diversionPage.test.tsx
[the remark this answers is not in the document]
joshlarson: [Feedback: Add test that explicitly validates the initial state](https://github.com/mbta/skate/pull/2580/commits/d296898b1aa0d6c4a0ecb3854d066e1d8ad01be9)

It shall be done, and thusly, it has been done!

(It also enabled me to remove an assertion from a different test, which was also contorting itself to test this, since this explicit coverage was missing!)

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: [Feedback: Add assertions about undo and clear buttons when returning to edit screen](https://github.com/mbta/skate/pull/2580/commits/9cd5dcc5dca5519e09b529a91578004547a2864b)

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -90,7 +102,118 @@ describe("DiversionPage", () => {
...
+        )
+      ).toBeVisible()
+    })
+  })
+
+  test("calls the API when the first waypoint is added", async () => {
+    const { container } = render(<DiversionPage />)
+
+    act(() => {
+      fireEvent.click(originalRouteShape.get(container))
```
firestack, line 184: question(non-blocking): why `not.toBeNull` here?

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: I think Github may have gotten confused about what you were actually suggesting - can you re-suggest based on the diff as it currently is?

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: I think Github got confused again :(

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -479,6 +763,31 @@ describe("DiversionPage", () => {
...
+
+    fireEvent.click(originalRouteShape.get(container))
+
+    await userEvent.click(finishDetourButton.get())
+
+    await waitFor(() => {
+      expect(screen.getByRole("button", { name: "Edit Detour" })).toBeVisible()
+    })
+
+    await userEvent.click(screen.getByRole("button", { name: "Edit Detour" }))
```
firestack, line 893:
```suggestion
    await userEvent.click(await screen.findByRole("button", { name: "Edit Detour" }))
```

assets/tests/components/detours/diversionPage.test.tsx
firestack: https://github.com/mbta/skate/pull/2580/files/d34c0a0dfa11cdb8e586f6110f5e685e47fb701f#r1603782294

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -129,30 +263,90 @@ describe("DiversionPage", () => {
...
-    expect(
-      screen.getByRole("button", { name: "Finish Detour", hidden: true })
-    ).not.toBeVisible()
+    fireEvent.click(container.querySelector(".c-vehicle-map")!)
 
     fireEvent.click(originalRouteShape.get(container))
 
-    expect(screen.getByRole("button", { name: "Finish Detour" })).toBeVisible()
+    await waitFor(() => {
+      expect(screen.getByTitle("Detour End")).not.toBeNull()
```
firestack, line 276: https://github.com/mbta/skate/pull/2580#discussion_r1603769317

joshlarson: [Feedback: Replace instances of not.toBeNull() with toBeVisible()](https://github.com/mbta/skate/pull/2580/commits/e9ce55cc8a39b533b7fac7d761f6c942c1651bc8)

firestack: > Maybe a copy-pasta from somewhere?
It certainly seems to be prevalent in this file from before these changes

firestack (approved):

assets/tests/hooks/useDetour.test.ts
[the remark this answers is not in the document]
hannahpurcell: I can imagine a regression where while the placeholder text is restored when "undo" is clicked, the directions are not removed as expected. Does that seem to warrant a check here?

hannahpurcell (approved):
Just making sure you see the one comment I left unresolved, but I'm good with making that "non-blocking", as most of my critical comments were addressed. Feel free to address at your availability.

assets/tests/hooks/useDetour.test.ts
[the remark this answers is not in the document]
joshlarson: Yeah probably. Will add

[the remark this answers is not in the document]
joshlarson: The fact that there's no way to trigger that without contorting the code is itself an implementation detail, so yes, this should be added.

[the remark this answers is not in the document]
joshlarson: [Feedback: Validate that directions are removed when placeholder text is visible](https://github.com/mbta/skate/pull/2580/commits/1810874fbbb8c2f906664c60cf13b54064e92e41)

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -126,33 +260,93 @@ describe("DiversionPage", () => {
...
-  test("when end point has been set, finish detour button is visible", async () => {
+  test("clicking on route shape after the detour is ended doesn't do anything", async () => {
     const { container } = render(<DiversionPage />)
 
     fireEvent.click(originalRouteShape.get(container))
 
-    expect(
-      screen.getByRole("button", { name: "Finish Detour", hidden: true })
-    ).not.toBeVisible()
+    fireEvent.click(container.querySelector(".c-vehicle-map")!)
```
firestack, line 271: issue(non-blocking): I'm working off your branch for the state machine stuff and the tests started failing I _think_ because these aren't wrapped in `act`, I can do so in my PR, but if you're up for it, could you maybe take a pass at the file and wrap fire event's in them?

joshlarson: I think that can be arranged. I had been wondering whether `fireEvent` should _always_, _never_, or _sometimes_ be wrapped in `act`.

I'm thinking ... probably _always_?

firestack: Okay I'm not actually sure that this is the cause, still figuring that out I guess, but I think wrapping fireEvent in act still stands? (unless I'm mis-remembering something

firestack: I thought always too but I'm looking at the docs and I'm just confused now 🫠

joshlarson: Well, I'll do it for now, and see what happens!

joshlarson: [Feedback: Wrap all fireEvent calls in act(...)](https://github.com/mbta/skate/pull/2580/commits/7295eebd2d6298a10e9bb0388fa3c40f4a83f471)

PR #2564, opened 2024-04-29
feat(ts/diversionPage): change copy details heading format

firestack:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207103683979738/f
Asana Ticket: https://app.asana.com/0/1148853526253420/1207152763000093/f

---

I opened this as a test for [⚙️ Integrate Github and Asana](https://app.asana.com/0/1148853526253420/1207152763000093/f)

This is ready for review, but I'm waiting on #2565 and fixing Asana access before merging this

PR #2582, opened 2024-05-06
fix(ts/diversionPage/share-detour): move turn-by-turn directions below missed stops

firestack:
This was intended to be done as apart of #2564 but the Asana ticket formatting threw me off when doing the original work.

Fixes: https://github.com/mbta/skate/pull/2564

Asana Ticket: https://app.asana.com/0/0/1207248614404482/f


hannahpurcell (approved):
👍 👍

PR #2587, opened 2024-05-09
feat(ex/application): add deployment version to startup log

firestack:
I was working on [Splunk CSRF Login Issue Dashboard](https://mbta.splunkcloud.com/en-US/app/search/skate_login_csrf_issue_tracking) and didn't have access to deployment version info in Splunk, so this felt like it should be easy to add.

Asana Ticket: https://app.asana.com/0/0/1207280328006578/f


hannahpurcell (approved):
🎉

PR #2588, opened 2024-05-09
feat: Select preset for minimal page

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207162343766764


lib/skate_web/router.ex
```diff
@@ -106,6 +106,7 @@ defmodule SkateWeb.Router do
     get "/settings", PageController, :index
     get "/detours", PageController, :index
     get "/minimal", PageController, :index
+    get "/minimal/*id", PageController, :index
```
firestack, line 109: question(non-blocking): what does the `*` mean here? I'm used to something like `:id` but not `*id`, is this a way to get phoenix to ignore the parameter but know it's there?

assets/src/components/minimalLadder.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {
+  deselectRouteInTab,
+  flipLadderInTab,
+  toggleLadderCrowdingInTab,
+} from "../state"
+import { useParams } from "react-router-dom"
+
+export const MinimalLadder = () => {
+  const [{ routeTabs }, dispatch] = useContext(StateDispatchContext)
+  const id = useParams().id as string
```
firestack, line 14: Does typescript complain if we do it like this?

```suggestion
  const { id } = useParams()
```

> https://reactrouter.com/en/main/hooks/use-params

firestack (commented):
praise: loving this so far! Not able to finish review today but I've left a few comments.

assets/src/components/app.tsx
```diff
@@ -72,7 +73,10 @@ export const AppRoutes = () => {
         <div className="l-app__main">
           <Routes>
             {inTestGroup(TestGroups.MinimalLadderPage) && (
-              <BrowserRoute path="/minimal" element={<MinimalLadderPage />} />
+              <>
+                <BrowserRoute path="/minimal" element={<MinimalLadderPage />} />
```
joshlarson, line 77: Question: At this point, do we need both of these?

assets/src/components/minimalLadder.tsx
joshlarson: It would type `id` as a `string | undefined`, which then means that it wouldn't be passable into `routeTabById`.

This is an interesting one, because `id` _should_ always be there because of how our `ReactRouter` is configured, but I wonder how to convince Typescript of that fact.

In theory, 👇 would work, but I'm not sure I love it, given that it's more verbose
```typescript
const { id } = useParams() as { id: string }
```

firestack: Or re-stated, create a `<MinimalLadder.FromRouterParam>` which wraps `<MinimalLadder>` and does all the `useParams` stuff there (like asserting the id is specified or showing an error message or redirecting otherwise) and then calls (condensed, since you _may_ know via typescript that `id` is defined at this point.)
```
{id && <MinimalLadder id={id} />}
```

joshlarson: Yeah that makes sense to me!

lib/skate_web/router.ex
hannahpurcell: Ohh huh! I didn't realize I was using both in different situations without a lot of nuance. It is more accurate to use `:` here. The `*` allows the rest of the matching url to be collected into the `id` param in the form `%{"id" => ["list", "of", "separated", "params"]}`. [Doc](https://hexdocs.pm/phoenix/Phoenix.Router.html#module-routing)

assets/src/components/minimalLadder.tsx
hannahpurcell: Ha! I'm not sure why I took the awkward route here. Yes, that's great

joshlarson:
This makes a lot of sense to me! Consider me approving once Kayla approves.

I have some thoughts and questions about UX, but these are thoughts that should _maybe_ be brought up to design, and make the most sense as follow-ups:
- I still don't think that the minimal view should have any of the interactive components. I can add a follow-up task to get rid of them if you all agree.
- Right now it's not particularly discoverable. Should there be a way to navigate to `/minimal`? Like a new nav link in the nav bar gated behind the same test group? I'm torn about this because it raises some thorny questions about users/roles/flags/etc (basically this would be perpetually available only to one user), but still.
- Should there be some kind of error state in case the preset that you tried to refresh gets deleted? (Or in case you navigate to `/minimal/oops` or something like that)

assets/src/components/minimalLadderPage.tsx
```diff
@@ -1,36 +1,26 @@
...
+    <div className="c-minimal-ladder-page">
+      <h3>Select a preset to display</h3>
+      {presets.map((preset) => (
+        <Button
+          className="c-minimal-ladder-page__button"
+          key={preset.uuid}
+          href={"/minimal/" + preset.uuid}
+        >
+          {preset.presetName}
+        </Button>
```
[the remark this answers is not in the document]
hannahpurcell: Yes, I can do this with some variation: the reactstrap Button doesn't like the prop `as`, so instead I can wrap the Button in a NavLink.

If I use the navless Button, the UX is that the page is blank for a moment before the route ladders render fully loaded. If I use the NavLink approach, the page renders kinda wonky before filling out. What is the benefit to going with NavLink here?

May-13-2024 17-21-44

assets/src/components/minimalLadder.tsx
hannahpurcell: Omg once again, I didn't refresh to get all the comments. Ignore my last one, I'm re-reading

hannahpurcell:
@joshlarson agree 1 and 2 are questions for both design & product lead. While the polish would be nice, I'd guess that once this "minimal page" is shipped, we have other detour priorities to work on before finessing this page

assets/src/components/minimalLadder.tsx
hannahpurcell: Looking at this with fresh eyes. While creating the `.FromRouterParam` may be the proper way, I don't naturally prefer this approach. I'll do it if I must! But it adds a layer of, like, unneeded padding around the MinimalLadder component that I think won't actually come in handy down the road? It's a layer of extra stuff around MinimalLadder that will add complexity in the App component, to fix one line in MinimalLadder that could be a little tighter. I suppose I find the existing one line more readable. Is that a bad take?

assets/src/components/minimalLadderPage.tsx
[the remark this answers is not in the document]
firestack: 1. I realize we don't _need_ `NavLink` specifically, but I'd like to use the [`Link` from `reactrouter`](https://reactrouter.com/en/main/components/link) since this is a url that is managed by reactrouter.
2. Is the difference in UX shown in the video posted above? I'm having a hard time seeing the difference

assets/src/components/minimalLadder.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const routeTab = routeTabById(routeTabs, id)
+  if (!routeTab) window.location.href = "/minimal"
```
firestack, line 17: suggestion: can we use [`useNavigate`](https://reactrouter.com/en/main/hooks/use-navigate)?

assets/src/components/minimalLadderPage.tsx
```diff
@@ -1,36 +1,26 @@
...
+      ))}
```
firestack, line 23: suggestion: Can we use bootstrap to style and lay these out?
```suggestion
      <Nav>
        {presets.map((preset) => (
          <Nav.Item key={preset.uuid}>
            <Nav.Link as={Link} to={"/minimal/" + preset.uuid}>
              {preset.presetName}
            </Nav.Link>
          </Nav.Item>
        ))}
      </Nav>
```

assets/src/components/minimalLadder.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const { selectedRouteIds, ladderDirections, ladderCrowdingToggles } =
+    routeTab || {
+      selectedRouteIds: [] as string[],
+      ladderDirections: {},
+      ladderCrowdingToggles: {},
+    }
```
firestack, line 24: nit(non-blocking)
```suggestion
  const { selectedRouteIds = [], ladderDirections = {}, ladderCrowdingToggles = {} } =
    routeTab || {}
```

assets/src/components/minimalLadder.tsx
firestack: I'm not okay with the type coersion (the `as string`), as I feel that's somewhat defeating the usefulnes of typescript, so I think you should handle the `undefined` case explicitly, but however you want to handle that is fine (if you don't feel like making a wrapper component to handle that logic)

firestack: In terms of reability, I don't find it harder to read, and I find it easier to reason about when the required parameter is just a parameter declared the correct type, but I this this is really just style and taste and nothing to worry about any specific way of doing it.

assets/css/app.scss
```diff
@@ -70,6 +70,7 @@ $vpp-location-padding: 1rem;
 @import "map/markers/missed_stop_icon";
 @import "map/markers/stop_icon";
 @import "map/markers/user_location_marker";
+@import "minimal_ladder_page";
```
firestack, line 73: note: I don't think we'll need this if you do https://github.com/mbta/skate/pull/2588#discussion_r1600363159

assets/src/components/minimalLadderPage.tsx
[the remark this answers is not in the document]
firestack: Ended up sketching out an idea https://github.com/mbta/skate/pull/2588#discussion_r1600363159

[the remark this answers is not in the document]
firestack: Also I tried the UX of a regular link vs a react router link, and I feel that the loading is smoother because it's a SPA transition when using `<Link>`, but it's a regular HTTP request when using `<Button href={} />` which ends up doing full page layout and first time rendering again.

assets/src/components/minimalLadder.tsx
hannahpurcell: Ohhhkay finally implemented, and I understand it better! Thanks for your patience 🙏

assets/src/components/minimalLadderPage.tsx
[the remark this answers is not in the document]
hannahpurcell: Ahhhh ok. I see why the react router link choppiness is still a positive tradeoff, then. Thanks! 👍

assets/src/components/minimalLadderPage.tsx
hannahpurcell: Hmmmm I've waffled a bit. My default inclination was to use buttons, and I'm not married to that. But they _are_ bootstrap, and I aesthetically like them. It feels nitpicky to not just commit the above, but since I have to add the imports anyway, I think I'm going to commit a blend that I have, as noted and seen in [this comment](https://github.com/mbta/skate/pull/2588#discussion_r1599100828), but I'll keep the Link

assets/css/_minimal_ladder_page.scss
```diff
@@ -0,0 +1,7 @@
+.c-minimal-ladder-page {
+  margin: 30px;
+}
```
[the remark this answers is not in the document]
hannahpurcell: It's used in MinimalLadderPage, as container padding

assets/tests/components/__snapshots__/minimalLadderPage.test.tsx.snap
```diff
@@ -1,173 +1,35 @@
...
-          </span>
-          Reverse
-        </button>
-      </div>
-      <div
-        class="c-ladder"
-        style="width: 184px;"
+        Preset 1
+      </button>
+    </a>
```
firestack, line 21: issue: Acording to the anchor element's "Permitted content":

> Permitted content:
>> Transparent, except that no descendant may be interactive content or an <a> element, and no descendant may have a specified tabindex attribute. 
>
> -- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#technical_summary

I believe a `button` is ["interactive content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#interactive_content) and I think we're really only using this for styling?

What if you put `btn btn-primary` on the `<a>`?

assets/css/_minimal_ladder_page.scss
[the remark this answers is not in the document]
firestack: Thanks! Sorry I missed it!

assets/src/components/minimalLadderPage.tsx
firestack: I think this is related? https://github.com/mbta/skate/pull/2588#discussion_r1603608483

firestack (approved):

PR #2593, opened 2024-05-10
feat: Direct `About Skate` clicks to the /user-guide endpoint

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1206673444071869/f

### DIFF
--- a/assets/src/components/nav/leftNav.tsx
+++ b/assets/src/components/nav/leftNav.tsx
@@ -178,18 +178,15 @@ const LeftNav = ({
               </button>
             </li>
             <li>
-              <NavLink
-                className={({ isActive }) =>
-                  "c-left-nav__link" +
-                  (isActive ? " c-left-nav__link--active" : "")
-                }
+              <a
+                className="c-left-nav__link"
                 title="About Skate"
                 target="_blank"
-                to="/user-guide"
+                href="/user-guide"
               >
                 <QuestionMarkIcon className="c-left-nav__icon" />
                 About Skate
-              </NavLink>
+              </a>
             </li>
             <li>
               <NavLink
--- a/assets/src/components/nav/leftNav.tsx
+++ b/assets/src/components/nav/leftNav.tsx
@@ -1,7 +1,6 @@
 import React, { useState, useContext } from "react"
-import { NavLink, useLocation } from "react-router-dom"
+import { NavLink } from "react-router-dom"
 import { StateDispatchContext } from "../../contexts/stateDispatchContext"
-import { displayHelp } from "../../helpers/appCue"
 import { openDrift } from "../../helpers/drift"
 import { tagManagerEvent } from "../../helpers/googleTagManager"
 import NotificationBellIcon from "../notificationBellIcon"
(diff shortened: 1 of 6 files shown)


assets/tests/components/nav/leftNav.test.tsx
```diff
@@ -309,19 +308,6 @@ describe("LeftNav", () => {
     expect(openDrift).toHaveBeenCalled()
   })
 
-  test("clicking About Skate button displays help", async () => {
```
joshlarson, line 312: I wasn't sure how to test that a new browser window opened, but I also didn't think that this was an unstable enough part of the system that it was worth putting too much effort into figuring out how to write that test.

This is worth a discussion though!

assets/src/components/nav/leftNav.tsx
```diff
@@ -180,14 +178,18 @@ const LeftNav = ({
...
             </li>
             <li>
-              <button
-                className="c-left-nav__link"
-                onClick={() => displayHelp(location)}
+              <NavLink
+                className={({ isActive }) =>
+                  "c-left-nav__link" +
+                  (isActive ? " c-left-nav__link--active" : "")
+                }
```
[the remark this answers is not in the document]
joshlarson: Hmmm - probably not much benefit, huh.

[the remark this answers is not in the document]
joshlarson: Screenshot 2024-05-13 at 1 42 43 PM

Flawless victory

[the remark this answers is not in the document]
joshlarson: Sooo I think we gain _some_ benefit in consistency. Even if we don't need _all_ of the features of `NavLink`, we still get the benefit of it effortlessly looking like the other links in this section!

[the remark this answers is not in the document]
joshlarson: [I guess using the correct attributes also helps](https://github.com/mbta/skate/pull/2593/commits/3768de7d8c3e4caf3356c5b9b670c3a533cd7a11) 🤦‍♂️

assets/tests/components/nav/leftNav.test.tsx
joshlarson: It seems like it _could_ be, but I'm struggling a decent amount with getting the types to behave. I wonder if there's an easier way that I'm missing. Would you be up for pairing on it?

firestack: Happy to pair! Does tomorrow morning work?

joshlarson: Right after standup? I'm not usually online before standup

firestack: We paired and looked into this, because the assertion we want to make is that a new tab opened, and that the browser context doesn't use `window.open` to do so, we couldn't easily find a way to get a spy in to detect the call, so we're going to leave this test out and wait for a future when we figure out/can detect this.

firestack (approved):

PR #2594, opened 2024-05-10
feat: Add a banner on the admin pages indicating what environment you're in

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1206593159272689/f

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,7 +8,8 @@ config :skate,
   restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS") == "true",
   base_tileset_url: System.get_env("BASE_TILESET_URL"),
   satellite_tileset_url: System.get_env("SATELLITE_TILESET_URL"),
-  aws_place_index: System.get_env("AWS_PLACE_INDEX")
+  aws_place_index: System.get_env("AWS_PLACE_INDEX"),
+  environment_name: System.get_env("SENTRY_ENV", "missing-env")
 
 if System.get_env("SECRET_KEY_BASE") do
   config :skate, SkateWeb.Endpoint, secret_key_base: System.get_env("SECRET_KEY_BASE")
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,8 +8,7 @@ config :skate,
   restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS") == "true",
   base_tileset_url: System.get_env("BASE_TILESET_URL"),
   satellite_tileset_url: System.get_env("SATELLITE_TILESET_URL"),
-  aws_place_index: System.get_env("AWS_PLACE_INDEX"),
-  environment_name: System.get_env("SENTRY_ENV", "missing-env")
+  aws_place_index: System.get_env("AWS_PLACE_INDEX")
 
 if System.get_env("SECRET_KEY_BASE") do
   config :skate, SkateWeb.Endpoint, secret_key_base: System.get_env("SECRET_KEY_BASE")
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,7 +8,8 @@ config :skate,
   restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS") == "true",
   base_tileset_url: System.get_env("BASE_TILESET_URL"),
   satellite_tileset_url: System.get_env("SATELLITE_TILESET_URL"),
-  aws_place_index: System.get_env("AWS_PLACE_INDEX")
+  aws_place_index: System.get_env("AWS_PLACE_INDEX"),
+  environment_name: System.get_env("ENVIRONMENT_NAME", "missing-env")
 
 if System.get_env("SECRET_KEY_BASE") do
   config :skate, SkateWeb.Endpoint, secret_key_base: System.get_env("SECRET_KEY_BASE")
--- a/lib/skate_web/components/core_components.ex
+++ b/lib/skate_web/components/core_components.ex
@@ -36,8 +36,8 @@ defmodule SkateWeb.CoreComponents do
   defp banner_background_color(_), do: "var(--color-mbta-bus)"
 
   @envs ["skate-local", "skate-dev", "skate-dev-blue", "skate-dev-green", "skate-prod"]
-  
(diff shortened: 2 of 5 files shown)


config/runtime.exs
```diff
@@ -8,7 +8,8 @@ config :skate,
   restrict_environment_access?: System.get_env("RESTRICT_ENVIRONMENT_ACCESS") == "true",
   base_tileset_url: System.get_env("BASE_TILESET_URL"),
   satellite_tileset_url: System.get_env("SATELLITE_TILESET_URL"),
-  aws_place_index: System.get_env("AWS_PLACE_INDEX")
+  aws_place_index: System.get_env("AWS_PLACE_INDEX"),
+  environment_name: System.get_env("SENTRY_ENV", "missing-env")
```
firestack, line 12: Can we use [`ENVIRONMENT_NAME`](https://github.com/mbta/devops/blob/master/terraform/modules/app-skate-ecs/main.tf#L18C15-L18C31)?

joshlarson: Yes, but just out of curiosity, why is that preferred? Just so that we're not sort of "leaking" sentry into non-sentry parts of the app?

firestack: I don't really think of Sentry as the authoritative source, so I'd personally rather keep sentry things sentry things, and have environment be the source.

firestack: I did mark this as a nit and non-blocking though, sorry I didn't do so sooner.

lib/skate_web/components/core_components.ex
```diff
@@ -27,4 +27,41 @@ defmodule SkateWeb.CoreComponents do
     <% end %>
     """
   end
+
+  defp banner_background_color("skate-local"), do: "gray"
+  defp banner_background_color("skate-dev"), do: "orange"
+  defp banner_background_color("skate-dev-blue"), do: "blue"
+  defp banner_background_color("skate-dev-green"), do: "green"
+  defp banner_background_color("skate-prod"), do: "red"
+  defp banner_background_color(_), do: "gray"
```
firestack, line 36: nit/suggestion(non-blocking): do you think you could replace these with the [mbta colors from `tokens-2024`](https://github.com/mbta/skate/blob/main/assets/css/color/_tokens_2024.scss#L170-L180)?

It's fine if not, I'm considering a PR to move most of this to CSS and make our admin CSS better.

lib/skate_web/components/layouts/barebones.html.heex
```diff
@@ -29,14 +29,33 @@
...
+                  font-style: italic;
+                }
+
+                .banner {
+                  width: 100vw;
+            text-align: center;
+            justify-content: center;
+      height: 48px;
+      line-height: 48px;
+                }
```
firestack, line 51: suggestion: some of these attrs are redundant and we should specify line height as a unitless number[^1][^2]

```suggestion
                .banner {
                    text-align: center;
                    line-height: 3;
                }
```
OR
```suggestion
                .banner {
                    text-align: center;
                    line-height: 4;
                }
```

[^1]:
    > If the page is zoomed to increase the text size, using a unitless value ensures that the line height will scale proportionately.
    > 
    > -- https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#accessibility_concerns

[^2]: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values

firestack (changes requested):
I would like to address [those CSS things](https://github.com/mbta/skate/pull/2594#discussion_r1599047860) but otherwise LGTM 👍🏻

lib/skate_web/components/layouts/barebones.html.heex
joshlarson: RIGHT I remember this! Thank you!

Also I dunno what's up with the HEEx formatting - `mix format` uesd to work, but it seems like it's totally forgotten how HEEx code is supposed to look 😑

config/runtime.exs
joshlarson: Ah it's all good - I was curious about your reasoning, but it was easy enough to do (and it is done now 😄)

lib/skate_web/components/core_components.ex
joshlarson: Does [this](https://github.com/mbta/skate/pull/2594/commits/b76bf5a9b1a84db3a99c0b8cff99b7eaee3daeae) do what you were hoping? I feel like this is a partial move towards better CSS, but only partial.

lib/skate_web/components/layouts/barebones.html.heex
joshlarson: [Here we go - this should fix both of our problems](https://github.com/mbta/skate/pull/2594/commits/6d2f2261a5f9f8c2f8163f03e69137360f2b9f1f)

lib/skate_web/components/core_components.ex
```diff
@@ -27,4 +27,41 @@ defmodule SkateWeb.CoreComponents do
...
+  defp banner_background_color("skate-dev-green"), do: "var(--color-mbta-green)"
+  defp banner_background_color("skate-prod"), do: "var(--color-mbta-red)"
+  defp banner_background_color(_), do: "var(--color-mbta-bus)"
+
+  defp banner_font_color("skate-local"), do: "white"
+  defp banner_font_color("skate-dev"), do: "white"
+  defp banner_font_color("skate-dev-blue"), do: "white"
+  defp banner_font_color("skate-dev-green"), do: "white"
+  defp banner_font_color("skate-prod"), do: "white"
+  defp banner_font_color(_), do: "black"
```
hannahpurcell, line 43: non-blocking!

```suggestion
  @envs ["skate-local", "skate-dev", "skate-dev-blue", "skate-dev-green", "skate-prod"]
  
  defp banner_font_color(env) when env not in envs, do: "white"
  defp banner_font_color(_), do: "black"
```

lib/skate_web/components/core_components.ex
firestack: Great thanks!

One nit is that `skate-prod` should probably be `bus` because it's _possible_ for us in the future to have a `dev-red`, but this works for me as is.

firestack (approved):
I like @hannahpurcell's [suggestion](https://github.com/mbta/skate/pull/2594/files#r1599150662), but otherwise this LGTM!

PR #2595, opened 2024-05-14
feat: Change desktop nav icons to bootstrap

joshlarson:
## Before

Screenshot 2024-05-14 at 1 36 18 PM

## After

Screenshot 2024-05-14 at 1 37 24 PM

Asana Ticket: https://app.asana.com/0/1148853526253420/1207319758165228/f


assets/src/components/nav/leftNav.tsx
```diff
@@ -173,7 +171,7 @@ const LeftNav = ({
                 onClick={openDrift}
                 title="Support"
               >
-                <SpeechBubbleIcon className="c-left-nav__icon" />
+                <ChatFill className="c-left-nav__icon" />
```
firestack, line 174: suggestion: what do you think about trying to get rid of `c-left-nav__icon` in favor of [`icon-link`](https://github.com/mbta/skate/blob/main/assets/src/components/nav/navMenu.tsx#L67) ?
IMO, the point of using the bootstrap icons here is to get access to the `icon-link` and `bi` classes which manage icon layout for us so that we can do less bespoke css in the navs.

assets/css/nav/_left_nav.scss
```diff
@@ -8,6 +8,9 @@
   border-right: 1px solid $color-gray-300;
 
   .c-left-nav__icon {
+    height: 1.25rem;
+    width: 1.25rem;
+    flex-shrink: 0;
```
firestack, line 13: https://github.com/mbta/skate/pull/2595/files#r1603567089

firestack (commented):
Almost ready to approve, but I'd really like to chat about https://github.com/mbta/skate/pull/2595/files#r1603567089 first.

assets/src/components/nav/leftNav.tsx
joshlarson: Yeah that's a good call. I'll try it out on my own and will possibly bug you to pair if it turns out that I don't know what I'm doing?

joshlarson: @firestack are you thinking something like this?

```suggestion
                <ChatFill className="icon-link pe-3 flex-shrink-0" width="1.25rem" height="1.25rem" />
```

(note: `pe-3` sets the padding to be `1rem`, but the existing padding is `0.75rem`, so this isn't exactly right... and I'm not sure how to make it exactly right, but I'm wondering if this is the _type_ of thing you had in mind?)

firestack: put `icon-link` on the `button`, then you might not need any classes

firestack: github won't let me comment on unchanged lines, sorry that was unclear due to that!

joshlarson: Hmmm - the padding and the size still came out too small

firestack: Hmm okay, I guess there's more to do to change this over to bootstrap. If you wouldn't mind, could you create a followup ticket to address this? I'm fine with merging with how it was before and addressing later.

firestack (approved):

assets/src/components/nav/leftNav.tsx
joshlarson: So I think we need some custom CSS or bootstrap utility classes to get the sizing and spacing to come out right, but I'm struggling to figure out how to get the specific values we want out of bootstrap utility classes:

- https://getbootstrap.com/docs/5.0/utilities/spacing/ <-- gives some very specific values (all with respect to `$spacer`, which I can't find defined in our code), none of which gives us the `0.75rem` that we want
  - _Side note: The pattern of how `p-{n}` relates to `p-{n+1}` seems pretty unpredictable - the pattern looks like 0, 1, 2, 4, 6, 12 (in increments of 0.25), which doesn't seem like a pattern I'm familiar with 😵‍💫_
- https://getbootstrap.com/docs/5.0/utilities/sizing/ <-- gives a bunch of sizes proportional to the parent element (I think), that are increments of 25%. Leaving that off seems to produce an icon that's `1rem`x`1rem`, which is just a bit smaller than the `1.25rem`x`1.25rem` that we need.


All of this makes me think that we still need some custom CSS, unless I'm missing some bootstrap utility classes (or icon classes or whatever) that do what we need.

joshlarson: Screenshot 2024-05-17 at 3 53 39 PM

☝️ `Support` uses my attempts at Bootstrap. Everything else uses our bespoke CSS.

firestack: Design has said that they're okay with using bootstrap defaults and not trying to replicate exactly what it looked like before, so I'd like to have a specific ticket to have a conversation around this, since this would also involve switching over to BS's list styling [speculation].

joshlarson: > could you create a followup ticket to address this?

[Indeed I can!](https://app.asana.com/0/1152340551558956/1207348831290026/f)

PR #2597, opened 2024-05-14
refactor(ts/hooks/usePatternsByIdForRoute): use `useApiCall`

firestack:
This is some work I've had sitting around that I did while trying to implement the new detours flow.

The tests get a little bit better due to now having to handle promises correctly.

### DIFF
--- a/assets/tests/hooks/usePatternsByIdForRoute.test.ts
+++ b/assets/tests/hooks/usePatternsByIdForRoute.test.ts
@@ -3,7 +3,6 @@ import { act, renderHook, waitFor } from "@testing-library/react"
 import * as Api from "../../src/api"
 import usePatternsByIdForRoute from "../../src/hooks/usePatternsByIdForRoute"
 import { routePatternFactory } from "../factories/routePattern"
-import { instantPromise } from "../testHelpers/mockHelpers"
 import { PromiseWithResolvers } from "../testHelpers/PromiseWithResolvers"
 import routeFactory from "../factories/route"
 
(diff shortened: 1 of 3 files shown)


assets/tests/hooks/usePatternsByIdForRoute.test.ts
```diff
@@ -45,75 +49,76 @@ describe("usePatternsByIdForRoute", () => {
...
+    })
+
+    expect(jest.mocked(Api.fetchRoutePatterns)).toHaveBeenLastCalledWith("66")
+
     rerender("39")
-    expect(mockFetchRoutePatterns).toHaveBeenLastCalledWith("39")
+
+    expect(jest.mocked(Api.fetchRoutePatterns)).toHaveBeenLastCalledWith("39")
+    // await waitFor(
+    //   () =>
```
joshlarson, line 81: 🧟

firestack: Thanks for catching this! Guess I didn't read over my diff enough before setting it for review, done in https://github.com/mbta/skate/compare/2b4f39595fd199a9345d77445cca4fe0106bcc30..1353e383a98b19ba4bd9ffb65da5fe0a9aa7f18d

assets/tests/hooks/usePatternsByIdForRoute.test.ts
```diff
@@ -45,75 +49,69 @@ describe("usePatternsByIdForRoute", () => {
...
-    expect(patternsFor66DidResolve).toBe(false)
+    expect(jest.mocked(Api.fetchRoutePatterns)).toHaveBeenLastCalledWith("39")
+    await waitFor(() => expect(result.current).toEqual(patternsFor39ById))
+
+    act(() =>
+      resolve(routePatternFactory.buildList(2, { routeId: routeId1.id }))
+    )
 
-    await waitFor(() => expect(patternsFor66DidResolve).toBe(true))
+    await waitFor(() => expect(result.current).toEqual(patternsFor39ById))
```
joshlarson, line 114: If this test finishes executing quickly enough that the promise doesn't have time to resolve (or the hook doesn't have time to respond to the promise resolving), would this still pass?

Also, is it worth having this assertion here followed by the same assertion just below?

firestack: [I think we can drop the assertion after this](https://github.com/mbta/skate/compare/1353e383a98b19ba4bd9ffb65da5fe0a9aa7f18d..2bd018703ea29ec1e59a9d6f44857d3dc8c828ad)

assets/tests/hooks/usePatternsByIdForRoute.test.ts
```diff
@@ -45,76 +49,68 @@ describe("usePatternsByIdForRoute", () => {
...
     rerender("39")
-    expect(mockFetchRoutePatterns).toHaveBeenLastCalledWith("39")
+
+    expect(jest.mocked(Api.fetchRoutePatterns)).toHaveBeenLastCalledWith("39")
   })
 
   test("Keeps most recently requested route patterns if previous resolves finishes later", async () => {
-    const patternFor66 = routePatternFactory.build({ routeId: "66" })
-    // const patternsFor66ById = { [patternFor66.id]: patternFor66 }
+    const { promise, resolve } =
```
joshlarson, line 83: Non-blocking, but calling for a follow-up: It would be nice if this test were clearer about how it tests what it claims to test. 

I think making some of the variable names clearer, and also putting in some explicit literals, like `"39"` and `"66"` would make the use case of this test easier to follow.

assets/tests/hooks/usePatternsByIdForRoute.test.ts
```diff
@@ -32,10 +34,12 @@ describe("usePatternsByIdForRoute", () => {
...
+    await waitFor(() =>
+      expect(result.current).toEqual({ [rp1.id]: rp1, [rp2.id]: rp2 })
+    )
   })
 
-  test("doesn't fetch routes when passed null id", () => {
+  test("doesn't fetch routes when passed null id", async () => {
     const mockFetchRoutePatterns: jest.Mock =
       Api.fetchRoutePatterns as jest.Mock
     mockFetchRoutePatterns.mockReturnValueOnce(
```
joshlarson, line 45: Non-blocking: There are two more instances of `instantPromise` remaining in this file that could be replaced with `mockResolvedValue`.

Happy to do that in a follow-up instead of as part of this PR, but I did want to flag it.

joshlarson (approved):
I turned off the auto-merge because I wanted you to get a chance to see [this comment](https://github.com/mbta/skate/pull/2597/files#r1604051612) and [this comment](https://github.com/mbta/skate/pull/2597/files#r1604045368) before merging, but I did explicitly mark both of those as non-blocking, and I think this looks great!

I'm happy to address both comments myself as follows-up.

assets/tests/hooks/usePatternsByIdForRoute.test.ts
firestack: how do you feel about dropping it entirely and relying on the underlying test from https://github.com/mbta/skate/blob/main/assets/tests/hooks/useApiCall.test.ts#L100-L124 ?

firestack: Happy to keep discussing, I'm going to merge but we can do something as a follow up.

PR #2600, opened 2024-05-16
feat: Updated route-pill with a new size, currently called "large-format"

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207260183650054

Test new format by adding className `c-route-pill--large-format` to any `RoutePill`. Otherwise, `RoutePill` defaults to smaller format.

### DIFF
--- a/assets/css/_route_pill.scss
+++ b/assets/css/_route_pill.scss
@@ -13,7 +13,7 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  
+
   &--large-format {
     width: 4.5rem;
     height: 2.25rem;
--- a/assets/css/_route_pill.scss
+++ b/assets/css/_route_pill.scss
@@ -1,6 +1,5 @@
 .c-route-pill {
-  min-width: 2.875rem;
-  max-width: 2.875rem;
+  width: 2.875rem;
   height: 1.5rem;
 
   border-radius: 0.8125rem;
@@ -10,6 +9,18 @@
   line-height: 1.5rem;
   font-weight: 700;
   font-family: $font-family-route-pill;
+
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  
+  &--large-format {
+    width: 4.5rem;
+    height: 2.25rem;
+    border-radius: 1.875rem;
+    font-size: $h3-font-size;
+    font-weight: 600;
+  }
 }
 
 .c-route-pill--bus {
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -38,10 +38,7 @@ export const DiversionPanel = ({
           <h2 className="c-diversion-panel__h2">Affected route</h2>
 
           <div className="d-flex">
-            <RoutePill
-              className="me-2 align-top"
-              routeName={routeName}
-            />
+            <RoutePill className="me-2 align-top" routeName={routeName} />
 
             <div className="d-inline-block">
               <p className="my-0 c-diversion-panel__description">
--- a/assets/src/components/detours/diversionPanel.tsx
+++ b/assets/src/components/detours/diversionPanel.tsx
@@ -39,7 +39,7 @@ export const DiversionPanel = ({
 
           <div className="d-flex">
             <RoutePill
-              className="d-inline-block me-2 align-top"
+              className="me-2 align-top"
               routeName={routeName}
             />
 
--- a/assets/stories/skate-components/routePill.stories.tsx
+++ b/assets/stories/skate-components/routePill.stories.tsx
@@ -18,5 +18,5 @@ export const BlueLine: Story = {
 }
 
 export const LargeFormat: Story = {
-  args: { routeName: "66", className: "c-route-pill--large-format" },
+  args: { routeName: "66", largeFormat: true },
 }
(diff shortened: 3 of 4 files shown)


assets/css/_route_pill.scss
```diff
@@ -10,6 +9,18 @@
   line-height: 1.5rem;
   font-weight: 700;
   font-family: $font-family-route-pill;
+
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
firestack, line 15: question(non-blocking): why the switch to flex?

hannahpurcell:
I need to make a story for this!

assets/css/_route_pill.scss
joshlarson: Would `line-height: 1.5` work? That way it would auto-adjust based on the font size

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#number

hannahpurcell: The height to font-size ratio is not consistent for both versions of this button. So font-size:height for the default version is 1:1.5 while font-size:height for the large format is 1.25:2.25 or 1:1.8

assets/stories/skate-components/routePill.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+const meta = {
+  component: RoutePill,
+} satisfies Meta<typeof RoutePill>
+
+export default meta
+type Story = StoryObj<typeof RoutePill>
+
+export const Default: Story = {
+  args: { routeName: "66" },
+}
```
joshlarson, line 14: I'd be interested in a few additional stories that show off some of the ways that the route pill can be styled, like 👇 possibly

```suggestion
export const Default: Story = {
  args: { routeName: "66" },
}

export const SilverLine: Story = {
  args: { routeName: "SL3" },
}

export const BlueLine: Story = {
  args: { routeName: "Blue Line" },
}

export const OrangeLine: Story = {
  args: { routeName: "Orange Line" },
}

export const RedLine: Story = {
  args: { routeName: "Red Line" },
}

export const GreenLine: Story = {
  args: { routeName: "D" },
}
```

I'm on the fence though, because we're still under a snapshot limit with Chromatic, which means that adding a bunch of stories will stress that even more.

hannahpurcell: 🤷 unsure! I'd need more details to decide what to do here -- how far are we from the snapshot limit? I didn't think of adding all these stories, given the sizing changes were the focus of the task, but I'd be ok with rolling this in to chip away at tech debt.

assets/css/_route_pill.scss
joshlarson: Ahhh - I had thought it was 1.5 : 2.25 and 1 : 1.5, which would have been consistent.

I'll back out of this conversation then, since I don't have an opinion on this (I would probably have gravitated towards `flex` too, fwiw)

assets/stories/skate-components/routePill.stories.tsx
joshlarson: Eh I'm good either way - I do see your point that adding these is outside the scope of this ticket, so I'm game with punting it.

I'd also be fine with adding just one of those, in order to show off the fact that the route pill changes color when given route names that are not normal bus routes.

hannahpurcell: I like that as a compromise! Low risk to add just one more story without overthinking!

```suggestion
export const Default: Story = {
  args: { routeName: "66" },
}

export const BlueLine: Story = {
  args: { routeName: "Blue Line" },
}
```

assets/stories/skate-components/routePill.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const BlueLine: Story = {
+  args: { routeName: "Blue Line" },
+}
+
+export const LargeFormat: Story = {
+  args: { routeName: "66", className: "c-route-pill--large-format" },
```
joshlarson, line 21: Nit (non-blocking): I wonder if this would be easier to use if we had a boolean flag on the component itself, so that you would render it as

```tsx
<RoutePill />
```

Or

```tsx
<RoutePill largeFormat />
```

And then move the CSS-ifying into the component itself.

assets/css/_route_pill.scss
```diff
@@ -1,6 +1,5 @@
 .c-route-pill {
-  min-width: 2.875rem;
```
joshlarson, line 2: Huh 🤔

assets/src/components/detours/diversionPanel.tsx
```diff
@@ -38,10 +38,7 @@ export const DiversionPanel = ({
           <h2 className="c-diversion-panel__h2">Affected route</h2>
 
           <div className="d-flex">
-            <RoutePill
-              className="d-inline-block me-2 align-top"
-              routeName={routeName}
-            />
+            <RoutePill className="me-2 align-top" routeName={routeName} />
```
joshlarson, line 41: Curiosity: Was `d-inline-block` not doing anything?

hannahpurcell: correct ✅

assets/stories/skate-components/routePill.stories.tsx
hannahpurcell: That's fine! Then if we ever reuse the large format, the className to be used will be enforced by the component.

joshlarson (approved):
🗺️ 💊

PR #2601, opened 2024-05-17
refactor: Extract the render component part of RoutePropertiesCard as its own component

joshlarson:
~Marked as a draft PR for now because I want to add a storybook story for this.~

Storybook story added!

Now it's marked as draft because this might not be the best way to accomplish this goal.

### DIFF
--- a/assets/src/components/mapPage/routePropertiesCard.tsx
+++ b/assets/src/components/mapPage/routePropertiesCard.tsx
@@ -224,14 +224,16 @@ const VariantPicker = ({
   }
 }
 
-const RoutePropertiesCard = ({
+export const RoutePropertiesCardRender = ({
   routePatterns,
-  selectedRoutePatternId,
+  route,
+  selectedRoutePattern,
   selectRoutePattern,
   onClose,
 }: {
   routePatterns: ByRoutePatternId<RoutePattern>
-  selectedRoutePatternId: RoutePatternId
+  route: Route
+  selectedRoutePattern: RoutePattern
   selectRoutePattern: (routePattern: RoutePattern) => void
   onClose?: () => void
 }) => {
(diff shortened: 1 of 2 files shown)


assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -224,27 +224,23 @@ const VariantPicker = ({
   }
 }
 
-const RoutePropertiesCard = ({
+export const RoutePropertiesCardRender = ({
```
joshlarson, line 227: Would love to come up with a better name for this

assets/src/components/mapPage/routePropertiesCard.tsx
```diff
@@ -318,4 +314,33 @@ const RoutePropertiesCard = ({
...
+  return (
+    <RoutePropertiesCardRender
+      routePatterns={routePatterns}
+      route={route}
+      selectedRoutePattern={selectedRoutePattern}
+      selectRoutePattern={selectRoutePattern}
+      onClose={onClose}
+    />
+  )
+}
```
firestack, line 344: suggestion(non-blocking): What do you think about this? I know it'd cause a diff (:/) in the rest of the application (and the `Render` suffix would need to be left off)

```suggestion
const RoutePropertiesCardFromPatternId = ({
  routePatterns,
  selectedRoutePatternId,
  selectRoutePattern,
  onClose,
}: {
  routePatterns: ByRoutePatternId<RoutePattern>
  selectedRoutePatternId: RoutePatternId
  selectRoutePattern: (routePattern: RoutePattern) => void
  onClose?: () => void
}) => {
  const selectedRoutePattern = routePatterns[selectedRoutePatternId]
  const route: Route | null = useRoute(selectedRoutePattern?.routeId)

  if (!route || selectedRoutePattern === undefined) {
    return <></>
  }

  return (
    <RoutePropertiesCardRender
      routePatterns={routePatterns}
      route={route}
      selectedRoutePattern={selectedRoutePattern}
      selectRoutePattern={selectRoutePattern}
      onClose={onClose}
    />
  )
}

RoutePropertiesCard.FromPatternId = RoutePropertiesCardFromPatternId
```

joshlarson: Oh hey - it doesn't actually cause a diff with the rest of the app, because of 👇 

```ts
export default RoutePropertiesCardFromPatternId
```

joshlarson:
Closed in favor of https://github.com/mbta/skate/pull/2609

Turns out the routes are provided by a context, which makes this a lot easier to do without refactoring.

PR #2602, opened 2024-05-20
fix:test(ts/components/diversionPage): add `waitFor` for `expect`

firestack:
This test didn't wait for the final condition, which is needed for moving to state machines.


hannahpurcell (approved):
👍

PR #2603, opened 2024-05-20
[WIP] feat(eslint): no-restricted-exports

firestack:
Default exports have many downsides, from requiring the user to give it a name to only being able to have a single default export. I'd like to formally propose and lint that we don't use default exports.

This unfortunately can't be merged at this moment without fixing all of our existing default usages, or marking them all as allowed for now.


joshlarson:
Unsure of whether this will work, but maybe this? https://eslint.org/docs/latest/use/configure/configuration-files#excluding-files-with-ignores

PR #2605, opened 2024-05-20
cleanup:feat(ts/hooks/useDetour): convert `fetchFinishedDetour` to use `useApiCall`

firestack:
Just a little bit of cleanup and simplification, makes the [transition to state machines](#2607 ) a little cleaner.


hannahpurcell (approved):
💯

PR #2606, opened 2024-05-20
 cleanup(ts/hooks/useDetour): convert compound match to tags

firestack:
Follow on to #2607 which attempts to simplify the conditions to check for finished

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -30,20 +30,18 @@ export enum DetourState {
 
 const selectFinishedState = (
   snapshot: SnapshotFrom<typeof createDetourMachine>
-) => {
+) =>
   /*
-   * There's probably a better way to do this? Tags or maybe context?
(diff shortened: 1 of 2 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -30,20 +30,18 @@ export enum DetourState {
 
 const selectFinishedState = (
   snapshot: SnapshotFrom<typeof createDetourMachine>
-) => {
+) =>
   /*
-   * There's probably a better way to do this? Tags or maybe context?
-   * Tags seem more appropriate, but weird to manage Out-Of-Bounds state via tags.
+   * There's probably a better way to do this? maybe context or sub-machines?
+   * It seems okay to manage this state via tags if we can keep things document,
```
joshlarson, line 36:
```suggestion
   * It seems okay to manage this state via tags if we can keep things documented,
```

hannahpurcell (approved):
Ohhhhhhhhhhh!! I didn't quite understand the tag comment before, but this explains it. I like. 👍

PR #2607, opened 2024-05-20
feat(ts/hooks/useDetour): control waypoints with state machine

firestack:
# Summary
Part 2 of the conversion of `useDetour` to a state machine!
This converts the `waypoint` (including `startPoint` and `endPoint`) to be managed by a state machine, instead of manual logic for managing the state. (Considering revisiting this decision about start and end points!)

Instead of managing separate `startPoint` and `endPoint`s, they're now derived from the waypoints array, this is done so that `startPoint` and `endPoint` are no longer containing implicit state of the status of the machine.

# Review Notes 
This PR involves a few dependencies and steps.

---

Depends on:
- #2605
	- Makes the diff on this PR a bit easier to read 
- #2602 
	- Fixes a test that would otherwise just be noise in this PR 
- #2616
	- First introduces XState and an initial conversion 

---

This PR was done in these steps and such could be reviewed in that order to make the diff a little easier to read.
1. [feat(ts/hooks/useDetour): control waypoints with state machine](https://github.com/mbta/skate/commit/0e77c2f18f461eae19c17df2a04177f644bf1e22)
4. [cleanup(ts/hooks/useDetour): move callback intermediate state into function scope to reduce recalls](https://github.com/mbta/skate/commit/67f6d956e41bf6fa2cdc85d4d5236f222bf9d6b4)

I optionally have [cleanup(ts/hooks/useDetour): convert complex match to tags](https://github.com/mbta/skate/pull/2606) this to simplify the match logic for `finishedDetour`

---

~If desired, [feat(ts/hooks/useDetour): replace state variable with state machine](https://github.com/mbta/skate/commit/3abee9746e18d6c25a7f55efbe849eb739ec2acb) and [feat(ts/hooks/useDetour): control waypoints with state machine](https://github.com/mbta/skate/commit/0e77c2f18f461eae19c17df2a04177f644bf1e22) can be split into two PRs~

**Edit**:
This has been broken into another PR
- https://github.com/mbta/skate/pull/2616

---

Asana Ticket: https://app.asana.com/0/0/1207381767522634/f

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from "react"
+import { useCallback, useMemo } from "react"
 import { ShapePoint } from "../schedule"
 import { fetchDetourDirections, fetchFinishedDetour } from "../api"
 
@@ -29,19 +29,31 @@ export const useDetour = (input: CreateDetourMachineInput) => {
     input,
   })
 
-  const { routePattern } = snapshot.context
-
-  const [startPoint, setStartPoint] = useState<ShapePoint | null>(null)
-  const [endPoint, setEndPoint] = useState<ShapePoint | null>(null)
-  const [waypoints, setWaypoints] = useState<ShapePoint[]>([])
+  const { routePattern, startPoint, endPoint, waypoints } = snapshot.context
+  const allPoints = useMemo(() => {
+    if (!startPoint) {
+      return []
+    } else if (!endPoint) {
+      return [startPoint].concat(waypoints)
+    } else {
+      return [startPoint].concat(waypoints).concat([endPoint])
+    }
+  }, [startPoint, waypoints, endPoint])
 
   const { result: finishedDetour } = useApiCall({
     apiCall: useCallback(async () => {
+      /* Until we have "typegen" in XState,
+       * we need to validate these exist for typescript
+       *
+       * > [Warning] XState Typegen does not fully support XState v5 yet. However,
+       * > strongly-typed machines can still be achieved without Typegen.
+       * > -- https://stately.ai/docs/migration#use-typestypegen-instead-of-tstypes
+       */
       if (routePattern && startPoint && endPoint) {
         return fetchFinishedDetour(routePattern.id, startPoint, endPoint)
-      } else {
-        return null
       }
+
+      return null
     }, [startPoint, endPoint, routePattern]),
   })
 
@@ -50,15 +62,7 @@ export const useDetour = (input: CreateDetourMachineInput) => {
     longitude: startPoint?.lon,
   })
 
-  const detourShape = useDetourDirections(
-    useMemo(
-      () =>
-        [startPoint, ...waypoints, endPoint].filter(
-          (v): v is ShapePoint => !!v
-        ),
-      [startPoint, waypoints, endPoint]
-    ) ?? []
-  )
+  const detourShape = useDetourDirections(allPoints)
 
   const coordinates =
     detourShape.result && isOk(detourShape.result)
@@ -76,38 +80,32 @@ export const useDetour = (input: CreateDetourMachineInput) => {
     })
   }
 
-  const canAddWaypoint = () => startPoint !== null && endPoint === null
+  const canAddWaypoint = () =>
+    snapshot.can({
+      type: "detour.edit.place-waypoint",
+      location: { lat: 0, lon: 0 },
+    })
+
   const addWaypoint = canAddWaypoint()
-    ? (p: ShapePoint) => {
-        setWaypoints((positions) => [...positions, p])
+    ? (location: ShapePoint) => {
+        send({ type: "detour.edit.place-waypoint", location })
       }
     : undefined
 
-  const addConnectionPoint = (point: ShapePoint) => {
-    if (startPoint === null) {
-      setStartPoint(point)
-    } else if (endPoint === null) {
-      setEndPoint(point)
-    }
-  }
+  const addConnectionPoint = (point: ShapePoint) =>
+    send({
+      type: "detour.edit.place-waypoint-on-route",
+      location: point,
+    })
 
-  const canUndo =
-    startPoint !== null && snapshot.matches({ "Detour Drawing": "Editing" })
+  const canUndo = snapshot.can({ type: "detour.edit.undo" })
 
   const undo = () => {
-    if (endPoint !== null) {
-      setEndPoint(null)
-    } else if (waypoints.length > 0) {
-      setWaypoints((positions) => positions.slice(0, positions.length - 1))
-    } else if (startPoint !== null) {
-      setStartPoint(null)
-    }
+    send({ type: "detour.edit.undo" })
   }
 
   const clear = () => {
-    setEndPoint(null)
-    setStartPoint(null)
-    setWaypoints([])
+    send({ type: "detour.edit.clear-detour" })
   }
 
   const finishDetour = () => {
@@ -208,7 +206,9 @@ export const useDetour = (input: CreateDetourMachineInput) => {
     clear,
 
     /** When present, puts this detour in "finished mode" */
-    finishDetour: endPoint !== null ? finishDetour : undefined,
+    finishDetour: snapshot.can({ type: "detour.edit.done" })
+      ? finishDetour
+      : undefined,
     /** When present, puts this detour in "edit mode" */
     editDetour,
   }
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -1,4 +1,5 @@
 import { setup, assign, fromPromise, ActorLogicFrom, InputFrom } from "xstate"
+import { ShapePoint } from "../schedule"
 import { Route, RouteId, RoutePattern } from "../schedule"
 import { fetchRoutePatterns } from "../api"
 
@@ -9,6 +10,10 @@ export const createDetourMachine = setup({
       routePattern?: RoutePattern
 
       routePatterns?: RoutePattern[]
+
+      waypoints: ShapePoint[]
+      startPoint: ShapePoint | undefined
+      endPoint: ShapePoint | undefined
     }
 
     input:
@@ -29,8 +34,6 @@ export const createDetourMachine = setup({
         }
 
     events:
-      | { type: "detour.edit.done" }
-      | { type: "detour.edit.resume" }
       | { type: "detour.route-pattern.open" }
       | { type: "detour.route-pattern.done" }
       | { type: "detour.route-pattern.delete-route" }
@@ -39,6 +42,13 @@ export const createDetourMachine = setup({
           type: "detour.route-pattern.select-pattern"
           routePattern: RoutePattern
         }
+      | { type: "detour.edit.done" }
+      | { type: "detour.edit.resume" }
+      | { type: "detour.edit.clear-detour" }
+      | { type: "detour.edit.place-waypoint-on-route"; location: ShapePoint }
+      | { type: "detour.edit.place-waypoint"; location: ShapePoint }
+      | { type: "detour.edit.undo" }
+      | { type: "detour.share.copy-detour"; detourText: string }
   },
   actors: {
     "fetch-route-patterns": fromPromise<
@@ -63,11 +73,40 @@ export const createDetourMachine = setup({
     "set.route-id": assign({
       route: (_, params: { route: Route }) => params.route,
     }),
+    "detour.add-start-point": assign({
+      startPoint: (_, params: { location: ShapePoint }) => params.location,
+    }),
+    "detour.remove-start-point": assign({
+      startPoint: undefined,
+    }),
+    "detour.add-waypoint": assign({
+      waypoints: ({ context }, params: { location: ShapePoint }) => [
+        ...context.waypoints,
+        params.location,
+      ],
+    }),
+    "detour.remove-last-waypoint": assign({
+      waypoints: ({ context }) => context.waypoints.slice(0, -1),
+    }),
+    "detour.add-end-point": assign({
+      endPoint: (_, params: { location: ShapePoint }) => params.location,
+    }),
+    "detour.remove-end-point": assign({
+      endPoint: undefined,
+    }),
+    "detour.clear": assign({
+      startPoint: undefined,
+      waypoints: [],
+      endPoint: undefined,
+    }),
   },
 }).createMachine({
   id: "Detours Machine",
   context: ({ input }) => ({
     ...input,
+    waypoints: [],
+    startPoint: undefined,
+    endPoint: undefined,
   }),
 
   initial: "Detour Drawing",
@@ -176,19 +215,86 @@ export const createDetourMachine = setup({
           },
         },
         Editing: {
+          initial: "Pick Start Point",
           on: {
             "detour.route-pattern.open": {
               target: "Pick Route Pattern",
             },
-            "detour.edit.done": {
-              target: "Share Detour",
+            "detour.edit.clear-detour": {
+              target: ".Pick Start Point",
+              actions: "detour.clear",
             },
+          },
+          states: {
+            "Pick Start Point": {
+              on: {
+                "detour.edit.place-waypoint-on-route": {
+                  target: "Place Waypoint",
+                  actions: {
+                    type: "detour.add-start-point",
+                    params: ({ event: { location } }) => ({
+                      location,
+                    }),
+                  },
+                },
+              },
+            },
+            "Place Waypoint": {
+              on: {
+                "detour.edit.place-waypoint": {
+                  actions: {
+                    type: "detour.add-waypoint",
+                    params: ({ event: { location } }) => ({
+                      location,
+                    }),
+                  },
+                },
+                "detour.edit.place-waypoint-on-route": {
+                  target: "Finished Drawing",
+                  actions: {
+                    type: "detour.add-end-point",
+                    params: ({ event: { location } }) => ({
+                      location,
+                    }),
+                  },
+                },
+                "detour.edit.undo": [
+                  {
+                    guard: ({ context }) => context.waypoints.length === 0,
+                    actions: "detour.remove-start-point",
+                    target: "Pick Start Point",
+                  },
+                  {
+                    actions: "detour.remove-last-waypoint",
+                    target: "Place Waypoint",
+                  },
+                ],
+              },
+            },
+            "Finished Drawing": {
+              on: {
+                "detour.edit.undo": {
+                  actions: "detour.remove-end-point",
+                  target: "Place Waypoint",
+                },
+                "detour.edit.done": {
+                  target: "Done",
+                },
+              },
+            },
+            Done: {
+              type: "final",
+            },
+          },
+
+          onDone: {
+            target: "Share Detour",
           },
         },
         "Share Detour": {
           on: {
             "detour.edit.resume": {
-              target: "Editing",
+              target: "Editing.Finished Drawing",
             },
           },
         },


assets/src/hooks/useDetour.ts
```diff
@@ -73,45 +104,40 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
     })
   }
 
-  const canAddWaypoint = () => startPoint !== null && endPoint === null
+  const canAddWaypoint = () =>
+    snapshot.can({
+      type: "detour.edit.place-waypoint",
+      location: { lat: 0, lon: 0 },
```
firestack, line 110: note: if the event has guards which check this value, then we'd poentially need to make this an input to the "can" check function.

assets/src/hooks/useDetour.ts
```diff
@@ -139,7 +152,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
 
   return {
     /** The current state of the detour machine */
-    state,
+    state: snapshot.matches({ "Detour Drawing": "Editing" })
+      ? DetourState.Edit
+      : snapshot.matches({ "Detour Drawing": "Share Detour" })
+      ? DetourState.Finished
+      : DetourState.Edit,
```
firestack, line 159: Any desires for a follow up PR to get rid of the `DetourState` type?

joshlarson: Yes please!

joshlarson: Not urgent though IMO

joshlarson, line 159:
```suggestion
      : DetourState.Finished
```

Not a huge deal, but this confused me because it made me think that there was some other state where we'd want `DetourState` to be `Edit` (of course, this is really just an argument for getting rid of `state` entirely, since the state machine can handle that now)

joshlarson: Oh especially because `DetourState` is what `DiversionPage` uses to figure out which panel to plant on the left... it's not as immediately simple as it looks O_o

firestack: Hmm, we could "just do it ™️" in this PR by merging this into this PR https://github.com/mbta/skate/pull/2615

firestack: *nods* I hated this, I just went with something and moved on to other things, I think https://github.com/mbta/skate/pull/2615 addresses this? But I can apply this in the mean time

joshlarson: Yeah if it were going to be long-lived code, then this would matter more, but since this is on its way out, I don't think it matters either way.

joshlarson: I vote nay on that - I think #2615 deserves to be its own chunk of work and should be reviewed separately.

assets/src/hooks/useDetour.ts
joshlarson: I'm not sure I understand this comment?

assets/src/hooks/useDetour.ts
```diff
@@ -26,35 +28,64 @@ export enum DetourState {
...
-    }, [startPoint, endPoint, routePatternId]),
+
+      /* Until we have "typegen" in XState,
+       * we need to validate these exist for typescript
+       *
+       * > [Warning] XState Typegen does not fully support XState v5 yet. However,
+       * > strongly-typed machines can still be achieved without Typegen.
+       * > -- https://stately.ai/docs/migration#use-typestypegen-instead-of-tstypes
+       */
+      if (
```
joshlarson, line 63: Near the bottom, we're still computing `startPoint` and `endPoint` (using a slightly different formula that I _think_ will always give the same answer, assuming the state machine states are set up correctly).

Why not just compute `startPoint` and `endPoint` as values computed from `snapshot`? So something like:


```tsx
const startPoint = snapshot.context.waypoints.at(0)
const endPoint = snapshot.matches({
                   "Detour Drawing": { Editing: "Finished Drawing" },
                 }) && snapshot.context.waypoints.at(-1)

/* ... */

if (startPoint && endPoint) {
  return fetchFinishedDetour(routePatternId, startPoint, endPoint)
}

/* ... */

return {
  startPoint,

  endPoint: endPoint || undefined,
 
  /* ... */
}
```

assets/src/hooks/useDetour.ts
```diff
@@ -26,35 +28,64 @@ export enum DetourState {
...
 
   const detourShape = useDetourDirections(
-    useMemo(
-      () =>
-        [startPoint, ...waypoints, endPoint].filter(
-          (v): v is ShapePoint => !!v
-        ),
-      [startPoint, waypoints, endPoint]
-    ) ?? []
+    // Shouldn't need `useMemo` anymore because the state machine is tracking this ref
```
joshlarson, line 87: Nice! 

We don't need the comment now though, do we?

assets/src/hooks/useDetour.ts
```diff
@@ -73,45 +104,40 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
...
-  const addConnectionPoint = (point: ShapePoint) => {
-    if (startPoint === null) {
-      setStartPoint(point)
-    } else if (endPoint === null) {
-      setEndPoint(point)
-    }
-  }
+  const addConnectionPoint = (point: ShapePoint) =>
+    send({
+      type: "detour.edit.place-connection-point",
```
joshlarson, line 121: On the one hand, we're removing one silly if statement. On the other, this feels like it really shows where the power of the state machine comes in!

assets/src/models/createDetourMachine.ts
```diff
@@ -1,25 +1,114 @@
-import { setup } from "xstate"
+import { setup, assign } from "xstate"
+import { ShapePoint } from "../schedule"
 
 export const createDetourMachine = setup({
   types: {} as {
+    context: {
```
joshlarson, line 6: Gotta be honest - the diff on this file is actually pretty nice!

joshlarson:
This looks so good! I'm so excited to see how this makes the rest of our work on detours easier!

> If desired, [feat(ts/hooks/useDetour): replace state variable with state machine](https://github.com/mbta/skate/commit/3abee9746e18d6c25a7f55efbe849eb739ec2acb) and [feat(ts/hooks/useDetour): control waypoints with state machine](https://github.com/mbta/skate/commit/0e77c2f18f461eae19c17df2a04177f644bf1e22) can be split into two PRs.

I think that splitting this in that way would be a good idea. I have a few reasons for thinking this:
1. This whole thing was a _lot_ easier to review commit-by-commit (well, those two commits - the other commits are mostly fixes and prework that would disappear in a rebase), which means that those commits make sense as standalone work.
2. I'm pretty sure the first PR by itself unblocks the detour route selection panel.
3. I think the first PR is either ready to merge or very very close, while there are a bunch of comments on the second one that I'd be interested in resolving before merging it.

assets/src/hooks/useDetour.ts
firestack: https://github.com/mbta/skate/compare/c9f4c6281fb4cf0f6787dfca9377e401a9155d30..9dd609fc2b042e8f816ff0ffd46d503c651073c0

assets/src/hooks/useDetour.ts
firestack: I mainly just wanted to mention that the values in `location` are _potentially_ significant. `.can` checks any associated `guards` on an event before reporting if you "can" do something or not. Because a guard can read the event parameters, a guard _could_ deny `location` values that don't meet arbitrary conditions

firestack:
> > If desired, [feat(ts/hooks/useDetour): replace state variable with state machine](https://github.com/mbta/skate/commit/3abee9746e18d6c25a7f55efbe849eb739ec2acb) and [feat(ts/hooks/useDetour): control waypoints with state machine](https://github.com/mbta/skate/commit/0e77c2f18f461eae19c17df2a04177f644bf1e22) can be split into two PRs.
> 
> I think that splitting this in that way would be a good idea. I have a few reasons for thinking this:
> 
> 2. I'm pretty sure the first PR by itself unblocks the detour route selection panel.

I've opened up a PR for the first half https://github.com/mbta/skate/pull/2616

assets/src/models/createDetourMachine.ts
```diff
@@ -1,23 +1,114 @@
...
+    }
+    events:
+      | { type: "detour.share" }
+      | { type: "detour.resume-editing" }
+      | { type: "detour.edit.place-waypoint"; location: ShapePoint }
+      | { type: "detour.edit.place-connection-point"; location: ShapePoint }
+      | { type: "detour.edit.clear-detour" }
+      | { type: "detour.edit.undo" }
+      | { type: "detour.edit.done" }
+      | { type: "detour.share.copy-detour"; detourText: string }
```
hannahpurcell, line 17: tiny: would it make sense to alphabetize this list so that the two `share` events are co-located?

firestack: I intentionally sorted them this way, and but I'd like to reconsider some of these event names, so if you have any suggestions!

firestack: Turns out, I don't actually use `detour.share`! So we can delete it altogether!

hannahpurcell: out of curiosity, what was the sorting?

firestack: Upon reflection, It might have been "Just Vibes™️" of what felt right, I'm rethinking this and am preparing to sort these alphabetically, thanks for pointing this out!

firestack: Sorted!

assets/src/hooks/useDetour.ts
firestack: I tried to do this in [`9c4e8a9` (#2607)](https://github.com/mbta/skate/pull/2607/commits/9c4e8a9ba94c1ab0c1861c4659740905903664b7)
Let me know what you think!

assets/src/hooks/useDetour.ts
firestack: I just cleaned this up in https://github.com/mbta/skate/pull/2616 => https://github.com/mbta/skate/compare/c5999e059f2a5db282837cb620a18c2168166476..ed078862225b57c93cc6a8ad7d4bf78fdd5d36b7

hannahpurcell (commented):
Instead of sitting on these comments, I'll go ahead and release "round one" of review, just on the `createDetourMachine.ts`

assets/src/models/createDetourMachine.ts
```diff
@@ -1,23 +1,114 @@
...
+                  ],
+                },
+                "detour.edit.place-connection-point": {
+                  target: "Finished Drawing",
+                  actions: [
+                    {
+                      type: "detour.add-waypoint",
+                      params: ({ event: { location } }) => ({
+                        location,
+                      }),
```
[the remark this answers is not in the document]
hannahpurcell: Ok for some reason, the latest pushes made gh lose track of this comment, so I'm going to find another example and attach a new comment (since I see it's still there)

assets/src/models/createDetourMachine.ts
```diff
@@ -1,28 +1,134 @@
...
+                    }),
+                  },
+                },
+                "detour.edit.place-connection-point": {
+                  target: "Finished Drawing",
+                  actions: {
+                    type: "detour.add-waypoint",
+                    params: ({ event: { location } }) => ({
+                      location,
+                    }),
```
hannahpurcell, line 86: tiny: could the syntax for this function be made identical to the previous references of the action type: detour.add-waypoint? (re-post of [this comment](https://github.com/mbta/skate/pull/2607#discussion_r1610591340))

assets/src/models/createDetourMachine.ts
```diff
@@ -1,23 +1,114 @@
...
+      window.navigator.clipboard?.writeText(detourText)
+    },
   },
 }).createMachine({
   id: "Detours Machine",
+  context: () => ({
+    waypoints: [],
+  }),
 
   initial: "Detour Drawing",
```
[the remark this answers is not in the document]
firestack: > Currently, there's only one state "Detour Drawing". Is that state needed, or can Editing & Sharing Detour be top-level states in this machine?

I've been unsure how to make this decision. It _should_ be possible to model each step of our detours flow (then new "pick route pattern" work I'm doing, "editing" and "sharing") each as their own state machine by becoming an "actor" within a larger state machine.

I've been struggling to find resources for XState on composing state machines (though maybe looking into blogs about state charts would help me here) and have not figured out a easy way forward. I've tried experimenting a bit but have not found anything satisfying yet.	

I do think we could do this in the future, but for now I think having a top level state that we build off of gives us the best foundation for starting while we figure out state charts.

so tldr; yes but I don't think we have enough understanding and hands on experience with state charts yet to do so. but maybe that's an argument for trying it to find pitfalls. I'm expecting if we did move the state up, we'd have a potentially large diff and a bunch of strings to track down.


> If it is needed, does it still need to be set as the "initial" state if it's the only state?

An "initial" state is _always_ required and I believe xstate never makes an assumption, even if there's a single child.

assets/src/models/createDetourMachine.ts
```diff
@@ -1,23 +1,114 @@
...
+                  {
+                    guard: ({ context }) => context.waypoints.length === 1,
+                    actions: "detour.undo",
+                    target: "Pick Start Point",
+                  },
+                  { actions: "detour.undo", target: "Place Waypoint" },
+                ],
+              },
+            },
+            "Finished Drawing": {
```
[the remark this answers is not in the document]
firestack: Hmm, I don't think I understand what you mean by "Connect endpoint" here? The only actions left to do at this state is "finishing" or "undo"-ing, so I don't think it applies because there's nothing to "Connect"? Maybe I'm misunderstanding?

assets/src/models/createDetourMachine.ts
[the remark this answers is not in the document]
hannahpurcell: Beautiful answer, thanks for elaborating! I was mostly curious and not requesting a change, so I'm completely happy with the top-level state of the state machine encompassing the primary states "editing" / "share". In short, love it as-is and glad just to know the question was considered.

assets/src/models/createDetourMachine.ts
firestack: Does this work? https://github.com/mbta/skate/compare/ea2d1939077007fd6f1366a5e775afe13a3b1985..9f4979315bfdd52eeb1fb760a5f985596734b91e

assets/src/models/createDetourMachine.ts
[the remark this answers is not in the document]
firestack: cont. https://github.com/mbta/skate/pull/2607#discussion_r1611805683

assets/src/hooks/useDetour.ts
```diff
@@ -27,37 +28,74 @@ export enum DetourState {
   Finished,
 }
 
-export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
-  const [snapshot, send] = useMachine(createDetourMachine)
+const selectWaypoints = (snapshot: SnapshotFrom<typeof createDetourMachine>) =>
```
hannahpurcell, line 31: Since waypoints in one context != waypoints in another context (as in, the snapshot includes start and end points as waypoints) a comment here would be helpful as a waypoint ;)

assets/src/hooks/useDetour.ts
```diff
@@ -27,37 +28,74 @@ export enum DetourState {
...
+  const isInFinishedDetourState = isFinishedDrawing || isSharingDetour
+  return isInFinishedDetourState
+}
 
-  const [startPoint, setStartPoint] = useState<ShapePoint | null>(null)
-  const [endPoint, setEndPoint] = useState<ShapePoint | null>(null)
-  const [waypoints, setWaypoints] = useState<ShapePoint[]>([])
+const selectEndPoint = (snapshot: SnapshotFrom<typeof createDetourMachine>) =>
+  (selectFinishedState(snapshot) && snapshot.context.waypoints.at(-1)) ||
+  undefined
```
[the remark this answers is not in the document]
firestack: https://github.com/mbta/skate/compare/9f4979315bfdd52eeb1fb760a5f985596734b91e..68083f20876eb75e218941c6c1786838082f48d8

assets/src/hooks/useDetour.ts
firestack: I'm considering renaming this but I'm struggling to find names. I was thinking maybe "selectDetourWaypoints"/"selectDetourRouteWaypoints"/"selectRouteWaypoints" but I don't think it makes it any clearer?

Maybe `intermediateWaypoints`? Really not sure, I'll try to come up with a comment to describe this at the very least

firestack: Maybe this is more of a UI concern and `useDetour` shouldn't be computing this? (something to revisit later I suppose)

firestack: https://github.com/mbta/skate/compare/068d53704c5f20a0db8a02955c1161d454c7c612..def441d30e8274e09b9b1b6e6f86fc7e325c3502

assets/src/models/createDetourMachine.ts
hannahpurcell: totally!

assets/src/models/createDetourMachine.ts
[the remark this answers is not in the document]
hannahpurcell: Oooooo... I'm not sure what I was meaning here anymore! Agree that the current naming makes sense as-is. (I think I wrote the above when I was trying to understand the differences between some states and events?)

hannahpurcell (approved):
Ok! It was tricky to wrap my head around this new approach. I think I'm on top of things now, and I _think_ I'm still onboard, so win-win! Approving, and continuing to watch for Josh's feedback.

assets/src/hooks/useDetour.ts
```diff
@@ -25,37 +28,76 @@ export enum DetourState {
...
-  const [state, setState] = useState<DetourState>(DetourState.Edit)
+const selectFinishedState = (
+  snapshot: SnapshotFrom<typeof createDetourMachine>
+) => {
+  /*
+   * There's probably a better way to do this? Tags or maybe context?
+   * Tags seem more appropriate, but weird to manage Out-Of-Bounds state via tags.
+   * Maybe this entire API call could be moved to and managed by an actor
+   * combined with parallel child states within the state machine?
+   * -- https://stately.ai/docs/promise-actors */
```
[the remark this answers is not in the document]
firestack: I'm a little conflicted on this but what I'm thinking is if we know that we can make a API call as soon as we reach the "Finished Drawing" state, then we could use XState promise actors to do the API call we want. 

There's two downsides
- I can't find many/any docs about using promise actors within another machine, especially when you don't want to block the application on that API call finishing, this is similar to my [other comment about nested machines](https://github.com/mbta/skate/pull/2607#discussion_r1611797335)
- If we restore to a state that has already "past" the "Finished Drawing" state, then it would be a little more difficult to find the right place to trigger the promise actor to do the API request.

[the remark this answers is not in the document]
firestack: > Yeahhhhh it'd be nice to have a way to catch "any state that is past state X" in the flowchart instead of needing to check states individually. 🤔 Pretty simple here, but could get more complicated in the future?

Ideally https://github.com/mbta/skate/pull/2606 makes this better

assets/src/hooks/useDetour.ts
```diff
@@ -27,37 +28,76 @@ export enum DetourState {
...
-  const [startPoint, setStartPoint] = useState<ShapePoint | null>(null)
-  const [endPoint, setEndPoint] = useState<ShapePoint | null>(null)
-  const [waypoints, setWaypoints] = useState<ShapePoint[]>([])
+const selectEndPoint = (snapshot: SnapshotFrom<typeof createDetourMachine>) =>
+  (selectFinishedState(snapshot) && snapshot.context.waypoints.at(-1)) ||
+  undefined
+
+export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
+  const [snapshot, send, createDetourActor] = useMachine(createDetourMachine)
+  const startPoint = useSelector(createDetourActor, selectStartPoint)
```
joshlarson, line 67: Curiosity/Nit: How come you changed the way that waypoints/start point/end point are stored? 

I'm partly asking because I had been thinking of something like:
- From the `Pick Start Point` state, if the event `detour.edit.place-connection-point` happens, then it initializes the `startPoint` context variable and switches to the `Place Waypoint` state.
- From the `Place Waypoint` state, if the event `detour.edit.place-connection-point` happens, then it initializes the `endPoint` context variable and switches to the `Finished Detour` final state (and then shifts out of `Editing`.

That feels a bit more ... state machine-y to me, since the states then have tighter control over when the start point/end point/etc get set.

assets/src/hooks/useDetour.ts
```diff
@@ -27,37 +28,76 @@ export enum DetourState {
...
     apiCall: useCallback(async () => {
-      if (startPoint && endPoint) {
+      /* Until we have "typegen" in XState,
+       * we need to validate these exist for typescript
+       *
+       * > [Warning] XState Typegen does not fully support XState v5 yet. However,
+       * > strongly-typed machines can still be achieved without Typegen.
+       * > -- https://stately.ai/docs/migration#use-typestypegen-instead-of-tstypes
+       */
+      if (isInFinishedDetourState && startPoint && endPoint) {
```
joshlarson, line 87: Is it possible for `isInFinishedDetourState` to be `false` if `endPoint` isn't set?

assets/src/hooks/useDetour.ts
joshlarson: Thinking this through more....

We do need `startPoint`/`endPoint`/`waypoints` separately for rendering, and also we need `allPoints` (below) for directions. 

Given the choice between computing `allPoints` from `startPoint`/`waypoints`/`endPoint` versus computing `startPoint`/`waypoints`/`endPoint` from `allPoints`, I think I prefer concatenating over slicing.

But I wonder how evil it would be to just track `allPoints` separately from `sP`/`wp`/`eP`? I'm torn, because it would be nice to Just Have™ `allPoints` handed to us from the state machine, but also as soon as we start storing the same info twice, we run the risk of goofing and letting things get out of sync.

assets/src/hooks/useDetour.ts
joshlarson: As a "fun" "game", I removed the check for `isInFinishedDetourState &&` and all the tests still passed. I think that also means that the `isInFinishedDetourState` can be removed.

I _think_ if you change how `startPoint`/`endPoint`/`waypoints` are stored per my comment above, that could make `selectFinishedState` unnecessary.

assets/src/hooks/useDetour.ts
```diff
@@ -75,38 +115,32 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
...
-  }
+  const addConnectionPoint = (point: ShapePoint) =>
+    send({
+      type: "detour.edit.place-connection-point",
+      location: point,
+    })
 
-  const canUndo =
-    startPoint !== null && snapshot.matches({ "Detour Drawing": "Editing" })
+  const canUndo = snapshot.can({ type: "detour.edit.undo" })
```
joshlarson, line 136: I like this! Feels very state-machiney.

My instincts (admittedly new-ish instincts here) tell me that the main value of using a state machine comes when we otherwise would ask it about its data and variables and state, we instead ask it what it can do, and this feels like a perfect example of that!

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
joshlarson: I made a few comments below, but I wonder if we could get rid of this check entirely (and thus get rid of the pain point) by storing the data a little differently

https://github.com/mbta/skate/pull/2607/files#r1612405667
https://github.com/mbta/skate/pull/2607/files#r1612399064

I might be way off-base here, but I have this hunch that the pain that we're experiencing here is really a symptom of trying to use the state machine _slightly_ skew to a "state machine-y" style - that is, we're trying to do too much "tell me about who you are", and not enough "tell me about what you can do".

assets/src/models/createDetourMachine.ts
```diff
@@ -1,28 +1,134 @@
-import { setup } from "xstate"
+import { setup, assign } from "xstate"
+import { ShapePoint } from "../schedule"
 
 export const createDetourMachine = setup({
   types: {} as {
-    events: { type: "detour.edit.done" } | { type: "detour.edit.resume" }
+    context: {
```
joshlarson, line 6: I think I mentioned in another PR, but I really like how clean the diff is when we augment the state machine. It makes this tool a lot easier to think about!

assets/src/models/createDetourMachine.ts
```diff
@@ -1,28 +1,134 @@
...
+      waypoints: ShapePoint[]
+    }
+    events:
+      | { type: "detour.edit.clear-detour" }
+      | { type: "detour.edit.done" }
+      | { type: "detour.edit.place-connection-point"; location: ShapePoint }
```
joshlarson, line 12: We've discussed before that "Connection Point" actually means something different (the last un-missed stop before the detour begins and the first un-missed stop after the detour ends). Should we take this opportunity to come up with a different name?

```suggestion
      | { type: "detour.edit.place-point-on-original-route"; location: ShapePoint }
```
☝️ a possibility, although honestly... we can do better 😅

assets/src/hooks/useDetour.ts
firestack: Still thinking on this, but I think this counts as derived state, but that doesn't answer if the `startPoint` and `endPoint` should be stored vs `waypoints`, https://github.com/statecharts/statecharts/discussions/36

joshlarson: https://github.com/mbta/skate/pull/2629

hannahpurcell (approved):
This has been greatly simplified since I last reviewed! I found it easier to read 👍

assets/src/models/createDetourMachine.ts
firestack: how about `detour.edit.place-detour-terminus-point`?

firestack: I'm mainly thinking that we may have "ghost" connection points, that are on the original route, so I'm trying to get at the root user intention.

firestack: Though `terminus` is not great :/ looking for something to generalize the idea of "start" and "end"

joshlarson: Oh, huh.

Maybe I'm missing something, but isn't that a different question? Like, we're displaying the points in one spot, and computing them effectively from a different spot, but either way, the question is... what should they be called?

joshlarson (approved):
Looks great!

It _might_ be worth figuring out a better name per this comment --> https://github.com/mbta/skate/pull/2607#discussion_r1612424986 <--, although I wouldn't want to block on that, since I don't have any good ideas 🙃 

It _might_ also be worth waiting until https://github.com/mbta/skate/pull/2645 is merged, because I think they'll conflict, and https://github.com/mbta/skate/pull/2645 blocks our current sprint goal.

Either way - those are both up to you - this PR looks excellent as it is!

joshlarson (approved):

hannahpurcell (approved):

PR #2609, opened 2024-05-20
feat(storybook): Add storybook story for RoutePropertiesCard

joshlarson:


joshlarson:
@mbta/skate-developers at the risk of slightly bikeshedding... I started experimenting with atomic design, and put this in the `molecules` folder, but ... now that I think about it, it's more of an `organism`?

What do you all think?

hannahpurcell:
> @mbta/skate-developers at the risk of slightly bikeshedding... I started experimenting with atomic design, and put this in the `molecules` folder, but ... now that I think about it, it's more of an `organism`?

Whatever term you choose here can completely be reworked in the subsequent atomic design task. And perhaps some a definition guide will be useful when that work is done. Personally, I'd be drawn to terms that fit our context / needs like `atom`, `compound`, `layoutElement`, `page`. The atomic design terms like `organism` are great for the analogy / idea, but I wouldn't use it in the final product because they can be interpreted so differently based on context.

hannahpurcell (approved):

hannahpurcell (approved):
Oops, I should have nodded to my comment in my approval, but ultimately I think `molecule` is fine here, and I have some ideas for renaming when we do the rest of the organizational change to storybook

joshlarson:
> Whatever term you choose here can completely be reworked in the subsequent atomic design task. And perhaps some a definition guide will be useful when that work is done.

Yeah that's a good point. Maybe for the immediate moment, I'll put it at the top level, and then once we make a Real Decision™, we can move it around.


> Personally, I'd be drawn to terms that fit our context / needs like `atom`, `compound`, `layoutElement`, `page`. The atomic design terms like `organism` are great for the analogy / idea, but I wouldn't use it in the final product because they can be interpreted so differently based on context.

Huh, I hadn't even thought of straying away from the exact terms in the atomic design doc.... but yeah, we should definitely aim for things that meet our needs and aren't confusing!

PR #2611, opened 2024-05-20
feat: Allow RoutePropertiesCard to be open to "variants" or "stops" by default

joshlarson:
Asana Ticket: https://app.asana.com/0/0/1207363543546275/f

### DIFF
--- a/assets/src/components/mapPage/routePropertiesCard.tsx
+++ b/assets/src/components/mapPage/routePropertiesCard.tsx
@@ -224,20 +224,23 @@ const VariantPicker = ({
   }
 }
 
+export type RoutePropertiesCardOpened = "variants" | "stops" | null
+
 const RoutePropertiesCard = ({
   routePatterns,
   selectedRoutePatternId,
   selectRoutePattern,
   onClose,
+  defaultOpened = null,
 }: {
   routePatterns: ByRoutePatternId<RoutePattern>
   selectedRoutePatternId: RoutePatternId
   selectRoutePattern: (routePattern: RoutePattern) => void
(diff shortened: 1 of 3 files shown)


assets/stories/skate-components/molecules/routePropertiesCard.stories.tsx
```diff
@@ -95,3 +95,14 @@ export const WithInboundSelected: Story = {
...
+export const WithVariantsOpenedByDefault: Story = {
+  args: {
+    defaultOpened: "variants",
+  },
+}
+export const WithStopsOpenedByDefault: Story = {
+  args: {
+    defaultOpened: "stops",
+  },
+}
```
hannahpurcell, line 108: I'm on the fence about including these two stories, because these stories don't really surface a UI variation of the card, but rather serve as a kind of test that the parameter of "default openness" is getting correctly applied. If we ever had a regression in that "openness" variable, the below tests would catch it.

joshlarson: The reason I thought it was a good idea is that if anything ever changes about how the two "open" states look, Chromatic wouldn't catch it on the other stories where they default to closed.

I'm not attached to doing this - it's kind of small potatoes stakes - but that's what I was thinking when I added these.

hannahpurcell: Oh I seeeeeee, I see. Ok yeah, I was very easily persuaded. It makes sense to want some record of the look of those open states.

hannahpurcell (approved):
👍

PR #2612, opened 2024-05-20
feat: New route ladder header

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207260183650058


assets/src/helpers/bsIcons.tsx
```diff
@@ -145,19 +145,22 @@ export const ExclamationCircleFill = (props: SvgProps) => (
...
-    height="16"
-    fill="currentColor"
-    className="bi bi-exclamation-triangle-fill"
-    viewBox="0 0 16 16"
-    aria-hidden
-    {...props}
-  >
-    <path d="M8.982 1.566a1.13 1.13 0 0 0-1.96 0L.165 13.233c-.457.778.091 1.767.98 1.767h13.713c.889 0 1.438-.99.98-1.767zM8 5c.535 0 .954.462.9.995l-.35 3.507a.552.552 0 0 1-1.1 0L7.1 5.995A.905.905 0 0 1 8 5m.002 6a1 1 0 1 1 0 2 1 1 0 0 1 0-2" />
-  </svg>
+export const ExclamationTriangleFill = forwardRef(
```
joshlarson, line 148: Is it possible to do this by attaching the `forwardRef` to a parent element of the icon, rather than the bsIcon itself?

assets/src/components/routeLadder.tsx
```diff
@@ -66,14 +70,51 @@ export const Header = ({
...
+  routeName,
+  onClose,
+  hasAlert,
+}: {
+  routeName: string
+  onClose: () => void
+  hasAlert: boolean
+}) => {
+  return (
+    <div className="c-new-route-ladder__header">
```
firestack, line 92: hmm, Figma says this is supposed to be a bootstrap Card, is there a reason to prefer custom CSS for this instead of bootstrap classes?

assets/src/helpers/bsIcons.tsx
hannahpurcell: Oh yeah 💯 . I got stumped on this earlier and am still stumped. I just pushed up a commit with a few variations of the icon with `forwardRef` combos. I changed the `SvgProps` to allow refs to be passed, and set up an AlertThing component that forwards the ref, but even with that plumping, if I use either `ExclamationTriangleFillJustSvg` or the wrapper `ExclamationTriangleFillWithRef` in `<AlertThing />`, neither work as I would hope?? only `ExclamationTriangleFill`. I dunno what to try next

assets/src/components/routeLadder.tsx
firestack: What if you instead wrap the SVG in a `div` under the `Tippy`?

assets/src/helpers/bsIcons.tsx
firestack: What if you instead wrap the `<SVG/>` in a `<div/>` under the `<Tippy/`?

hannahpurcell: NUH UH. I thought if any nested component was under <Tippy/> then it would have problems 😭  and all this time, Tippy just didn't want an immediately nested component. Thank youuuuu 🥲

assets/src/components/routeLadder.tsx
hannahpurcell: Agh, there were a couple things I missed with this Figma handoff. I'm revisiting now. Thank you!

assets/src/helpers/bsIcons.tsx
firestack: I believe what is happening is that Tippy wants to register an `onClick` handler and isn't opinionated about what element is the container of the tooltip, so it uses `ref` to access the associated DOM element. Because the SVG's don't expose their ref's by default, Tippy wasn't able to register the click handler which I believe is an error?

So by giving it a direct reference to another element that does have a ref, you can avoid needing that on the SVGIcons.

This does bring up the question of if it's worth making refs work with our SVG's so we don't have to do this "wrapping" on the icon, just to make our icon components a little more complete?

firestack: Something for a future ticket I suppose

assets/src/components/routeLadder.tsx
```diff
@@ -66,14 +70,52 @@ export const Header = ({
...
+  hasAlert,
+}: {
+  routeName: string
+  onClose: () => void
+  hasAlert: boolean
+}) => {
+  return (
+    <Card className="c-new-route-ladder__header">
+      <Card.Body>
+        <div className="c-new-route-ladder__fake-dropdown" />
```
joshlarson, line 94: Non-blocking: Did we ever determine what the UX here should be if the user isn't allowed to add/edit detours? I assume that this "fake" dropdown shouldn't be present. 

This is a non-blocking comment because this isn't going live yet, but I'm _pretty_ sure that this is something we'll need to address before we do go live.

joshlarson (approved):
This looks great! Can you address https://github.com/mbta/skate/pull/2612/files#r1618952190 before merging?

assets/tests/components/routeLadder.test.tsx
```diff
@@ -187,6 +194,37 @@ describe("routeLadder", () => {
     expect(tree).toMatchSnapshot()
   })
 
+  test("renders a route ladder with the new header format", () => {
+    jest.mocked(getTestGroups).mockReturnValue(["route-ladder-header-update"])
```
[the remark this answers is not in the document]
hannahpurcell: Ah dang, I meant to bubble up [this change](https://github.com/mbta/skate/pull/2623#discussion_r1617228730)! Yes, will address

assets/src/components/routeLadder.tsx
joshlarson: No need! I don't think this needs to be addressed in THIS PR - I just wanted to make sure it wasn't fully dropped.

joshlarson (approved):
✅ Re-approving just to signal that I saw the update ✅

PR #2613, opened 2024-05-20
chore: New route ladder header test + stories

hannahpurcell:


joshlarson:
I know like 9 times out of 10, I'm all for splitting PR's up into smaller units, but I think this is that tenth time. I would find this (and its parent) easier to review if they were merged into one PR.

assets/stories/skate-components/newRouteLadderHeader.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import { NewHeader as RouteLadderHeader } from "../../src/components/routeLadder"
+
+const meta = {
+  component: RouteLadderHeader,
+  args: {
+    routeName: "39",
+    onClose: () => {},
```
firestack, line 9: Could you replace this with `fn()` from `@storybook/test`?
https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spy-function

hannahpurcell:
> I know like 9 times out of 10, I'm all for splitting PR's up into smaller units, but I think this is that tenth time. I would find this (and its parent) easier to review if they were merged into one PR.

Agh, I felt like all my PRs lately were too big, but that's fine. I can combine

assets/stories/skate-components/newRouteLadderHeader.stories.tsx
hannahpurcell: Agh ok! I hadn't started looking into the change, so you saved me before I had begun the wild goose chase

PR #2615, opened 2024-05-21
cleanup(ts/hooks/useDetour): replace `state` with snapshot from `createDetourMachine`

firestack:
https://github.com/mbta/skate/pull/2607/files#r1608932364

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -7,7 +7,7 @@ import React, {
 } from "react"
 import { DiversionPanel } from "./diversionPanel"
 import { DetourMap } from "./detourMap"
-import { DetourState, useDetour } from "../../hooks/useDetour"
+import { useDetour } from "../../hooks/useDetour"
 import { Alert, Button, CloseButton, Modal } from "react-bootstrap"
 import * as BsIcons from "../../helpers/bsIcons"
 import { OriginalRoute } from "../../models/detour"
@@ -31,7 +31,7 @@ export const DiversionPage = ({
   showConfirmCloseModal,
 }: DiversionPageProps) => {
   const {
-    state,
+    snapshot,
 
     addConnectionPoint,
     addWaypoint,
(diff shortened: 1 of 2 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -99,7 +99,7 @@ export const DiversionPage = ({
         </header>
 
         <div className="l-diversion-page__panel bg-light">
-          {state === DetourState.Edit && (
+          {snapshot.matches({ "Detour Drawing": "Editing" }) ? (
```
joshlarson, line 102: Hmmm - I don't love this, because it leaks the state machine guts out of `useDetour` and into `diversionPage`. I like the (imperfect but improving) idea that `useDetour`'s role is to tell `diversionPage` a bunch of facts about the detour (well, really the detour editing process) as it currently is, and I feel like calling `xstate` methods directly strays away from that.

I know [earlier I said](https://github.com/mbta/skate/pull/2607/files#r1608976742) that we should get rid of `DetourState`, but having _something_ at the level of `EditScreen`/`SharingScreen` (and eventually /`RouteSelectionScreen`) feels like an appropriate thing for `useDetour` to expose.

firestack: I somewhat disagree because I see the state machine replacing `useDetour`/`useDetour` being a light wrapper around the state machine. I prefer this, because to me the state machines states should ideally match what we're trying to do in the UI, so it should read well to use the state machines states directly in the UI, and IMO, exposing extra states from `useDetour` that the state machine is unaware of feels wrong in an anti-pattern type way (I could be wrong!), but I think that the state machine is enough of an `enum` that I'd argue it's typesafe enough to be part of the public api of `useDetour` and the UI

firestack:
Uhh, I dunno what you just did github, but I did not merge this

assets/src/components/detours/diversionPage.tsx
firestack: We should keep chatting about this, but the new context is https://github.com/mbta/skate/pull/2617

PR #2616, opened 2024-05-22
 feat(ts/hooks/useDetour): replace `state` variable with state machine 

firestack:
Splitting #2615 up into two PR's, while trying to preserve context.

https://github.com/mbta/skate/pull/2607#issuecomment-2123630961

Asana Ticket: https://app.asana.com/0/0/1207381767522633/f

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -6,6 +6,8 @@ import { OriginalRoute } from "../models/detour"
 import { useApiCall } from "./useApiCall"
 import { Ok, isErr, isOk } from "../util/result"
 import { useNearestIntersection } from "./useNearestIntersection"
+import { useMachine } from "@xstate/react"
+import { createDetourMachine } from "../models/createDetourMachine"
 
 const useDetourDirections = (shapePoints: ShapePoint[]) =>
   useApiCall({
@@ -26,7 +28,7 @@ export enum DetourState {
 }
 
 export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
-  const [state, setState] = useState<DetourState>(DetourState.Edit)
+  const [snapshot, send] = useMachine(createDetourMachine)
 
   const [startPoint, setStartPoint] = useState<ShapePoint | null>(null)
   const [endPoint, setEndPoint] = useState<ShapePoint | null>(null)
@@ -88,7 +90,8 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
     }
   }
 
-  const canUndo = startPoint !== null && state === DetourState.Edit
+  const canUndo =
+    startPoint !== null && snapshot.matches({ "Detour Drawing": "Editing" })
 
   const undo = () => {
     if (endPoint !== null) {
@@ -107,11 +110,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
   }
 
   const finishDetour = () => {
-    setState(DetourState.Finished)
+    send({ type: "detour.edit.done" })
   }
 
   const editDetour = () => {
-    setState(DetourState.Edit)
+    send({ type: "detour.edit.resume" })
   }
 
   const missedStops = finishedDetour?.missedStops || undefined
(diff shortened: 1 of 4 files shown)


assets/src/models/createDetourMachine.ts
hannahpurcell, line 1: Let's check to see if I'm interpreting this machine right!

We've got a machine identified as "Detours Machine". It currently only has 1 state: "Detour Drawing". Within that state, there are currently 2 states: "Editing" and "Share Detour". And "Editing" is made up of states "Idle" (default) and "Done". 

Thoughts:
- tiny wordsmithing, non-blocking: could "Detour Drawing" instead be "Detour Creation", perhaps? Since drawing only happens during editing
- I'm wrapping my head around why "Editing" needs to be broken into "Idle" and "Done". At first I thought the "Editing" state itself would have an "on" handler that would be like "on detour.share event, go to `Share Detour` state" instead of needing to funnel through an intermediary state of `type: "final"` that then triggers an `onDone` handler. So I guess my question is more to understand: why is that intermediary `Done` state needed?

firestack: > * So I guess my question is more to understand: why is that intermediary `Done` state needed?

I think this is actually a bad PR split, where these intermediate `Idle` and `Done` states are bleeding in from initial modeling in prep for the waypoints in #2607.

So you're correct that this could be simplified so that the `detour.share` event is just on the `Editing` state and transitions to `Share Detour` without any child states.

---

<details>
<summary>

Extra note: I wrote this out because I was considering this question in the context of #2607 and not this PR. This response does not apply correctly in the context of this PR
</summary>

> the `final` state `Done` allows the `Editing` state to not have to know about `Share Detour`, so if we decided later that we had a different place to send people to after editing, the logic inside of `Editing` doesn't need to know about that. 
> 
> If we did want to go directly from `Finished Drawing` to `Share Detour` we'd need to use ID paths to get to them,
> 
> so `Finished Drawing` is really `#Detours Machine.Detour Drawing.Editing.Finished Drawing`
> and `Share Detour` is really `#Detours Machine.Detour Drawing.Share Detour`
> 
> So when `Finished Drawing` is trying to transition to another state, the only non-id targets available are siblings and children of siblings, which is why `Finished Drawing` transitions to `Done` instead of `#Detours Machine.Detour Drawing.Share Detour`, mainly for cleanliness and encapsulation, as it seemed like keeping child state machines encapsulated is best practice when building state charts.

</details>

hannahpurcell (approved):
This makes sense to me from an iterative perspective. (So while the Idle/Done split doesn't seem to be needed here, I know it's developed further in the later PR, so I'm approving this one and looking more closely at the subsequent one. Function otherwise looks good.) Noted one proposed wording change, but non-blocking.

assets/src/models/createDetourMachine.ts
firestack: Cleaned up and simplified the state machine for this PR!
https://github.com/mbta/skate/compare/e8b1719ee288daacce6b63a98c0162bd33e9a2a3..c5999e059f2a5db282837cb620a18c2168166476

assets/src/hooks/useDetour.ts
```diff
@@ -126,7 +129,11 @@ export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
 
   return {
     /** The current state of the detour machine */
-    state,
+    state:
+      snapshot.matches({ "Detour Drawing": "Editing" }) &&
+      !snapshot.matches({ "Detour Drawing": "Share Detour" })
```
joshlarson, line 134: Nit: Is it possible for `snapshot.matches({ "Detour Drawing": "Share Detour" })` to be true if `snapshot.matches({ "Detour Drawing": "Editing" })` is true?

If not, then we don't need the `!snapshot.matches({ "Detour Drawing": "Share Detour" })` part of the condition.

joshlarson (approved):
Looks good to me!

I made a [comment](https://github.com/mbta/skate/pull/2616#discussion_r1611960942) that I'm interested in your thoughts on, but 👍 either way.

assets/src/hooks/useDetour.ts
firestack: It "MUST NOT" be true, in the words of [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) 😅 
When I was writing this I thought I came up with a pretty good way to do this but upon reflection and seeing it again, it's a bit silly.

PR #2617, opened 2024-05-23
cleanup(ts/hooks/useDetour): replace `state` with `snapshot` from `createDetourMachine`

firestack:
This used to be #2615 but github thought I closed it >.>

---

https://github.com/mbta/skate/pull/2607/files#r1608932364

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -7,7 +7,7 @@ import React, {
 } from "react"
 import { DiversionPanel } from "./diversionPanel"
 import { DetourMap } from "./detourMap"
-import { DetourState, useDetour } from "../../hooks/useDetour"
+import { useDetour } from "../../hooks/useDetour"
 import { Alert, Button, CloseButton, Modal } from "react-bootstrap"
 import * as BsIcons from "../../helpers/bsIcons"
 import { OriginalRoute } from "../../models/detour"
@@ -31,7 +31,7 @@ export const DiversionPage = ({
   showConfirmCloseModal,
 }: DiversionPageProps) => {
   const {
-    state,
+    snapshot,
 
     addConnectionPoint,
     addWaypoint,
@@ -99,7 +99,7 @@ export const DiversionPage = ({
         </header>
 
         <div className="l-diversion-page__panel bg-light">
-          {state === DetourState.Edit && (
+          {snapshot.matches({ "Detour Drawing": "Editing" }) ? (
             <DiversionPanel
               directions={extendedDirections}
               missedStops={missedStops}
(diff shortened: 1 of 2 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -99,7 +99,7 @@ export const DiversionPage = ({
         </header>
 
         <div className="l-diversion-page__panel bg-light">
-          {state === DetourState.Edit && (
+          {snapshot.matches({ "Detour Drawing": "Editing" }) ? (
```
joshlarson, line 102: Uprooting [a comment thread](﻿﻿﻿https://github.com/mbta/skate/pull/2615#discussion_r1609007990) from the original version of this PR and bringing it here:

Me:
> Hmmm - I don't love this, because it leaks the state machine guts out of useDetour and into diversionPage. I like the (imperfect but improving) idea that useDetour's role is to tell diversionPage a bunch of facts about the detour (well, really the detour editing process) as it currently is, and I feel like calling xstate methods directly strays away from that.
>
> I know [earlier I said](https://github.com/mbta/skate/pull/2607/files#r1608976742) that we should get rid of DetourState, but having something at the level of EditScreen/SharingScreen (and eventually /RouteSelectionScreen) feels like an appropriate thing for useDetour to expose.

@firestack 

> I somewhat disagree because I see the state machine replacing useDetour/useDetour being a light wrapper around the state machine. I prefer this, because to me the state machines states should ideally match what we're trying to do in the UI, so it should read well to use the state machines states directly in the UI, and IMO, exposing extra states from useDetour that the state machine is unaware of feels wrong in an anti-pattern type way (I could be wrong!), but I think that the state machine is enough of an enum that I'd argue it's typesafe enough to be part of the public api of useDetour and the UI

joshlarson: I think what's throwing me off is that I envisioned the responsibilities of the components as something like:
 - state machine does all the state management, like holding onto data and figuring out how to handle state transitions. Ya know - state machine stuff.
 - `useDetour` maps that onto ... sort of a _description_ of what's on the screen. Things like "the start point is _here_", and "the undo button should do _this thing_" (or "the undo button should be disabled right now")
 - `<DiversionPage />` and its friends `<DetourMap />` and the various `<*Panel />`s handle the layout, but don't do any interpretation beyond putting the things from `useDetour` in the right place on the screen.

So to me, the mapping of "this is the xstate state we're in" to "throw the whatever panel on the left" maps cleanly onto `useDetour`'s responsibilities. This code makes `<DiversionPage />` do that instead, which feels like it cuts through the boundaries ☝️ differently from everything else.

---

The original reason I had this thought was because I envisioned `useDetour` as a (thin) wrapper around the state machine, and this particular thing pokes through that wrapper. Everything ☝️ ☝️ ☝️  is, I think, why I kind of want to stand by my instinct here.

---

I would also _maybe_ be open to totally removing `useDetour` and putting `<DiversionPage />` in charge of the "xstate state" <-> "UI description" mapping instead.

I'm not _totally_ sold on that, because I think that there's value in keeping the layout details and the page description separate from each other (and I don't want `<DiversionPage />` to get too bloated), but there might also be a bit of a fear-of-change thing going on, so I'd be interested to see how it looks if we decided to try it.

firestack: > I would also maybe be open to totally removing useDetour and putting <DiversionPage /> in charge of the "xstate state" <-> "UI description" mapping instead.

...what if we just get rid of use detour? 😅 I'm in favor of this :) especially as I'm having to do a lot of the same calls in `useDetour` as the diversion page now.

firestack: I'll have something for us to chat about soon which I think might help here!

joshlarson: Can't wait!

firestack: https://mbta.slack.com/archives/C047ZT8ACP5/p1716994945665709?thread_ts=1716992038.426239&cid=C047ZT8ACP5

firestack: @joshlarson, given our conversation, do you think there's any more to do here?

joshlarson: Oh woops! Lemme give this another 👀

joshlarson (approved):

PR #2618, opened 2024-05-23
feat: Have RoutePropertiesCard on DetourRouteSelectionPanel open to "variants"

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207363543546281/f

### DIFF
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -69,6 +69,7 @@ export const DetourRouteSelectionPanel = ({
                 selectedRouteInfo
               )}
               selectRoutePattern={() => {}}
+              defaultOpened="variants"
             />
           ) : (
             <p className="fst-italic">


assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -69,6 +69,7 @@ export const DetourRouteSelectionPanel = ({
                 selectedRouteInfo
               )}
               selectRoutePattern={() => {}}
+              defaultOpened="variants"
```
firestack, line 72: btw, I don't know if you noticed this, but because this only "seeds" the state, and the RPC is programmed to close it's "opened details section" whenever something is selected, this only applies on first render.

firestack (approved):

assets/src/components/detours/detourRouteSelectionPanel.tsx
joshlarson: Hmmm - touché - I'm guessing (although I'll ask) that that's not Kathleen's intent

PR #2620, opened 2024-05-23
fix: Add selectRoutePattern to RoutePropertiesCard story

joshlarson:
Its absence was throwing errors in the console and hiding the auto-collapse behavior (which is being removed in https://github.com/mbta/skate/pull/2619).

### DIFF
--- a/assets/stories/skate-components/routePropertiesCard.stories.tsx
+++ b/assets/stories/skate-components/routePropertiesCard.stories.tsx
@@ -70,6 +70,7 @@ const meta = {
   args: {
     routePatterns: routePatterns,
     selectedRoutePatternId: routePattern0.id,
+    selectRoutePattern: () => {},
   },
   argTypes: {
     routePatterns: { table: { disable: true } },


assets/stories/skate-components/routePropertiesCard.stories.tsx
```diff
@@ -70,6 +70,7 @@ const meta = {
   args: {
     routePatterns: routePatterns,
     selectedRoutePatternId: routePattern0.id,
+    selectRoutePattern: () => {},
```
firestack, line 73: this bothers me because the [storybook docs](https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spy-function) say to use the `fn()` function, but I've learned that the `fn()` function is only available to storybook v8 users 🫠

firestack (approved):

assets/stories/skate-components/routePropertiesCard.stories.tsx
joshlarson: I really wish Storybook yelled louder if `args` didn't have the right type signature for `component` - this is a type error, but CI let it sail right on through.

firestack: is it a type error? You're allowed to ignore arguments if you want, and the return type is `void`.

joshlarson: Oh I meant that leaving `selectRoutePattern` out, which I did originally, is a type error. 

So this PR _fixes_ a type error. But a type error that typescript/storybook weren't able to detect :(

PR #2623, opened 2024-05-24
feat: New route ladder action dropdown

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207260183650049

Just adds the dropdown. Will need to be hooked up to the detour workflow once it's done.

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -53,12 +53,20 @@
     align-items: center;
   }
 
-  .c-new-route-ladder__fake-dropdown {
+  .c-route-ladder__dropdown-button {
     height: 36px;
     width: 36px;
-    border-radius: 4px;
-    background-color: $color-eggplant-500;
     margin: 6px 8px;
+  }
+
+  .c-route-ladder__dropdown .dropdown-item {
+    box-sizing: border-box;
+    display: inline-flex;
+    align-items: center;
+  }
+
+  .c-route-ladder__dropdown-item__icon {
+    margin-right: 6px;
   }
 
   .c-route-ladder__alert-icon {
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -59,16 +59,6 @@
     width: 36px;
   }
 
-  .c-route-ladder__dropdown .dropdown-item {
-    box-sizing: border-box;
-    display: inline-flex;
-    align-items: center;
-  }
-
-  .c-route-ladder__dropdown-item__icon {
-    margin-right: 6px;
-  }
-
   .c-route-ladder__alert-icon {
     display: grid;
     align-items: center;
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -49,14 +49,14 @@
 
   .card-body {
     display: inline-grid;
-    padding: 0;
+    grid-template-columns: 1fr 1fr auto 1fr 1fr;
+    padding: 6px 8px;
     align-items: center;
   }
 
   .c-route-ladder__dropdown-button {
     height: 36px;
     width: 36px;
-    margin: 6px 8px;
   }
 
   .c-route-ladder__dropdown .dropdown-item {
@@ -90,8 +90,7 @@
 
   .c-route-ladder__close-button-container {
     grid-column: 5/6;
-    margin-right: 8px;
-    margin-left: auto;
+    margin: 2px 0 2px auto;
   }
 }
 
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -55,8 +55,8 @@
   }
 
   .c-route-ladder__dropdown-button {
-    height: 36px;
-    width: 36px;
+    height: 2.25rem;
+    width: 2.25rem;
   }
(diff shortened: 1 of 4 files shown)


assets/src/components/routeLadder.tsx
```diff
@@ -91,7 +91,17 @@ export const NewHeader = ({
...
+        {inTestGroup(TestGroups.DetoursPilot) && (
+          <Dropdown>
+            <Dropdown.Toggle className="c-route-ladder__dropdown-button"></Dropdown.Toggle>
+            <Dropdown.Menu className="c-route-ladder__dropdown">
+              <Dropdown.Item>
+                <PlusSquare className="c-route-ladder__dropdown-item__icon" />
+                Add detour
+              </Dropdown.Item>
+            </Dropdown.Menu>
+          </Dropdown>
```
firestack, line 103: suggestion: Most of this styling can be done with bootstrap

```suggestion
          <Dropdown className="border-box inherit-box">
            <Dropdown.Toggle style={{aspectRatio: "1/1"}}/>
            <Dropdown.Menu>
              <Dropdown.Item className="icon-link">
                <PlusSquare />
                Add detour
              </Dropdown.Item>
            </Dropdown.Menu>
          </Dropdown>
```

Though you may want to keep `className="c-route-ladder__dropdown-button"` and put the `aspect-ratio` on that class.

```suggestion
          <Dropdown className="border-box inherit-box">
            <Dropdown.Toggle className="c-route-ladder__dropdown-button"/>
            <Dropdown.Menu>
              <Dropdown.Item className="icon-link">
                <PlusSquare />
                Add detour
              </Dropdown.Item>
            </Dropdown.Menu>
          </Dropdown>
```

assets/tests/components/routeLadder.test.tsx
```diff
@@ -225,6 +225,39 @@ describe("routeLadder", () => {
...
+        vehiclesAndGhosts={undefined}
+        selectedVehicleId={undefined}
+        deselectRoute={() => {}}
+        reverseLadder={() => {}}
+        toggleCrowding={() => {}}
+        ladderDirections={{}}
+        ladderCrowdingToggles={{}}
+        hasAlert={false}
+      />
+    ).container
```
firestack, line 256: suggestion: could you de-structure instead? (I see other tests did it this way, and I think they're wrong )
```suggestion
    const { container: tree } = render(
      <RouteLadder
        route={route}
        timepoints={timepoints}
        vehiclesAndGhosts={undefined}
        selectedVehicleId={undefined}
        deselectRoute={() => {}}
        reverseLadder={() => {}}
        toggleCrowding={() => {}}
        ladderDirections={{}}
        ladderCrowdingToggles={{}}
        hasAlert={false}
      />
    )
```

assets/tests/components/routeLadder.test.tsx
```diff
@@ -225,6 +225,39 @@ describe("routeLadder", () => {
     expect(tree).toMatchSnapshot()
   })
 
+  test("renders a route ladder with the new header and detour dropdown", () => {
+    jest
+      .mocked(getTestGroups)
+      .mockReturnValue(["route-ladder-header-update", "detours-pilot"])
```
firestack, line 231: Could you import the `TestGroups` enum and reference them via the symbols instead?
```suggestion
      .mockReturnValue([TestGroups.RouteLadderHeaderUpdate, TestGroups.DetoursPilot])
```

assets/css/_route_ladder.scss
```diff
@@ -49,16 +49,24 @@
...
     display: inline-grid;
-    padding: 0;
+    grid-template-columns: 1fr 1fr auto 1fr 1fr;
+    padding: 6px 8px;
     align-items: center;
   }
 
-  .c-new-route-ladder__fake-dropdown {
+  .c-route-ladder__dropdown-button {
     height: 36px;
```
firestack, line 58: note/slight issue(from previous PR?): Github won't let me comment on the unchanged lines, but I don't think `36px` is the proper unit to use here. If the user uses a different root font size than `16px`, this doesn't actually end up looking correct anymore, I think it'd be more appropriate to use `height: 2.25rem; width: 2.25rem;` here, and even better to use `aspect-ratio: 1/1;` as noted in my other comment.

firestack: Defer to this thread though https://github.com/mbta/skate/pull/2623#discussion_r1617224651

assets/src/components/routeLadder.tsx
hannahpurcell: Oh for sure, so first I want to caveat that `aspect-ratio` is not an attribute I tend to use, since usually I'm building something I do _not_ want to scale. When I tried `aspect-ratio` in this case, no effect; it only works if I set the height I want to use to `36px`. Then, since width is auto, the `aspect-ratio` is correctly applied and width expands to be `36px` as well. 

I'm not sure why the height is required here, because from the docs, I'd expect aspect-ratio to compromise between height and width if they are both set to auto. It seems to me that height is not auto, but inherited from somewhere. So in that case, if I need to set both height and aspect-ratio, then I might as well just set height and width.

firestack (approved):

PR #2630, opened 2024-05-29
fix: Reworked route ladder header layout

hannahpurcell:
After merging other PRs into the target branch and before merging the target branch into main, I realized a couple things were off. For one, the route pill needed to switch to the large format. For another, the alert icon didn't have the right margins both when visible / invisible.

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -49,9 +49,15 @@
 
   .card-body {
     display: inline-grid;
-    grid-template-columns: 1fr auto auto auto 1fr;
+    grid-template-columns: 1fr auto 1fr;
     padding: 6px 8px;
-    align-items: center;
+  }
+
+  .c-route-ladder__header__action-container {
+    display: inline-flex;
+    div {
+      padding-right: 0.5rem;
+    }
   }
 
   .c-route-ladder__dropdown-button {
@@ -60,10 +66,9 @@
   }
 
   .c-route-ladder__alert-icon {
-    display: grid;
+    display: flex;
     align-items: center;
     width: 1.6rem;
-    margin: 0 0.5rem;
 
     cursor: pointer;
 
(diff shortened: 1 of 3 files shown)


assets/css/_route_ladder.scss
```diff
@@ -49,7 +49,7 @@
 
   .card-body {
     display: inline-grid;
-    grid-template-columns: 1fr 1fr auto 1fr 1fr;
+    grid-template-columns: 1fr auto auto auto 1fr;
```
joshlarson, line 52: Screenshot 2024-05-29 at 2 38 32 PM

It looks like the route pill got knocked off-center?

joshlarson: My guess is that the change from `1fr` to `auto` isn't necessary, and that the change to `margin` is all that's needed, but I'm not going to pretend to know enough about `grid` to be certain of that.

hannahpurcell: Sadly, ended up needing a bigger change. Shifting to `auto` was needed to keep the centering nice with the dropdown. Since that was precarious, had to rework a bit

assets/css/_route_ladder.scss
```diff
@@ -49,21 +49,35 @@
...
+      padding-right: 0.5rem;
+    }
   }
 
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
     width: 2.25rem;
+    // TODO All of this is bad and should be reverted when we fix the left nav button colors
+    &.btn-primary {
+      --bs-btn-bg: #451b83;
```
joshlarson, line 68: Nit (non-blocking): I think these colors can be imported from the `tokens_2021` file. 

At the top of the file, you could do something like:
```css
@use "./color/tokens_2021" as colors;
```

And then `#451b83` becomes `colors.$eggplant-700` and `#260f54` becomes `colors.$eggplant-800`.

This is _very_ temporary, so not a huge deal either way.

joshlarson (approved):

assets/css/_route_ladder.scss
hannahpurcell: Yeah, since these will be deleted, I specifically didn't add an extra line of code for the import, so that the single chunk can be deleted.

joshlarson: Ah yeah that makes sense!

PR #2631, opened 2024-05-29
feat(ts/detours/detourRouteSelectionPanel): add events to panel

firestack:
Asana Ticket: https://app.asana.com/0/0/1207428503805261/f

### DIFF
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -27,9 +27,9 @@ interface DetourRouteSelectionPanelProps {
   allRoutes: Route[]
   selectedRouteInfo: SelectedRouteInfo
 
-  onConfirm?: () => void
-  onSelectRoute?: (routeId: Route | undefined) => void
-  onSelectRoutePattern?: (routePatternId: RoutePattern | undefined) => void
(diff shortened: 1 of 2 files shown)


assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -26,6 +26,10 @@ type SelectedRouteInfo =
 interface DetourRouteSelectionPanelProps {
   allRoutes: Route[]
   selectedRouteInfo: SelectedRouteInfo
+
+  onConfirm?: () => void
```
joshlarson, line 30: Silly question - do these need to be optional? Seems like in the real world, all of these functions would need to be there.

joshlarson: If they're required, we'll have to add simple noop functions to the storybook stories for this component, but that might still be a good idea.

firestack: I was going to merge without this and open another PR, but I didn't realize that this component wasn't used anywhere else, so I've added it in [`bd50e11` (#2631)](https://github.com/mbta/skate/pull/2631/commits/bd50e111d3eab5315b3000a6df2eae6a80a038c0)

PR #2633, opened 2024-05-30
feat(ts/detours): model pick route and pattern flow in state machine

firestack:
This models the main flow of the pick route panel, this intentionally does not wire up anything in the UI or change the interface of `useDetour` and `DiversionPage`, those will be addressed in followup PRs. These new states should also be inaccessible as they're not "called" from any of our existing flow.

---

Asana Ticket: https://app.asana.com/0/0/1207438967654848/f

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -4,15 +4,27 @@ import { Route, RoutePattern } from "../schedule"
 export const createDetourMachine = setup({
   types: {} as {
     context: {
-      route: Route
-      routePattern: RoutePattern
+      route?: Route
+      routePattern?: RoutePattern
     }
 
-    input: {
-      // Caller has target route pattern
-      route: Route
-      routePattern: RoutePattern
-    }
+    input:
+      | {
+          // Caller has target route pattern
+          route: Route
+          routePattern: RoutePattern
+        }
+      | {
+          // Caller has target route
+          route: Route
+          routePattern: undefined
+        }
+      | {
+          // Caller has no prior selection
+          route: undefined
+          routePattern: undefined
+        }
+
     events: { type: "detour.edit.done" } | { type: "detour.edit.resume" }
   },
 }).createMachine({
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -38,6 +38,21 @@ export const createDetourMachine = setup({
     "Detour Drawing": {
       initial: "Editing",
       states: {
+        "Pick Route Pattern": {
+          initial: "Pick Route ID",
+          states: {
+            "Pick Route ID": {
+            },
+            "Pick Route Pattern": {
+            },
+            Done: {
+              type: "final",
+            },
+          },
+          onDone: {
+            target: "Editing",
+          },
+        },
         Editing: {
           on: {
             "detour.edit.done": {
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -1,11 +1,25 @@
-import { setup } from "xstate"
+import { setup, ActorLogicFrom, InputFrom } from "xstate"
+import { Route, RoutePattern } from "../schedule"
 
 export const createDetourMachine = setup({
   types: {} as {
+    context: {
+      route: Route
+      routePattern: RoutePattern
+    }
+
+    input: {
+      // Caller has target route pattern
+      route: Route
+      routePattern: RoutePattern
+    }
     events: { type: "detour.edit.done" } | { type: "detour.edit.resume" }
   },
 }).createMachine({
   id: "Detours Machine",
+  context: ({ input }) => ({
+    ...input,
+  }),
 
   initial: "Detour Drawing",
   states: {
@@ -30,3 +44,11 @@ export const createDetourMachine = setup({
     },
   },
 })
+
+/**
+ * This refers to the type of `input` provided in
+ * {@linkcode createDetourMachine}'s {@linkcode setup} call
+ */
+export type CreateDetourMachineInput = InputFrom<
+  ActorLogicFrom<typeof createDetourMachine>
+>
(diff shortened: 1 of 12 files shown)


assets/src/models/createDetourMachine.ts
```diff
@@ -1,17 +1,142 @@
...
 
 export const createDetourMachine = setup({
   types: {} as {
-    events: { type: "detour.edit.done" } | { type: "detour.edit.resume" }
+    context: {
+      routeId?: RouteId
+      routePatternId?: RoutePatternId
+
+      routePatterns?: RoutePattern[]
+    }
```
firestack, line 12: note: I'm not sure it's worth making this type "complex" like the `input` type?

I _think_ that doing more complex types here would make `assign` and other interactions with `context` harder, but hopefully it doesn't, but I'd like to leave that for a later PR.

assets/src/models/createDetourMachine.ts
```diff
@@ -1,17 +1,142 @@
...
+
+    input:
+      | {
+          // Caller has target route pattern
```
firestack, line 16: note(minor): I'd have put these comments before the blocks, but then the formatter made it look weird

assets/src/models/createDetourMachine.ts
```diff
@@ -1,17 +1,142 @@
...
+    "fetch-route-patterns": fromPromise<
+      Awaited<ReturnType<typeof fetchRoutePatterns>>,
+      { routeId?: RouteId }
+    >(async ({ input: { routeId } }) => {
+      if (routeId) {
+        return fetchRoutePatterns(routeId)
+      } else {
+        // Hmm, this could leave us in the lurch if we don't handle it,
+        // I'd like to make this impossible but I can't get typescript to infer
+        // the context types but it's not letting me
```
firestack, line 52: note: I _can_ make the `input` to this actor typed so that you can't call it without a `routeId`, but xstate currently doesn't support type deduction via guards, which is the only way of doing conditional calls, so we end up needing to use the `!` operator, which our `eslint` config disallows.

assets/src/models/createDetourMachine.ts
```diff
@@ -1,17 +1,139 @@
...
   initial: "Detour Drawing",
   states: {
     "Detour Drawing": {
       initial: "Editing",
       states: {
+        "Pick Route Pattern": {
+          initial: "Pick Route ID",
+          on: {
+            "detour.route-pattern.select-route": {
+              target: ".Load Route Patterns",
```
hannahpurcell, line 81: Is this a typo? (or a syntax I don't know yet?)

firestack: Good question! This is valid syntax and not a typo.

See: 
> Targets the descendant `closed` state node
> -- https://stately.ai/docs/finite-states#targets

assets/src/models/createDetourMachine.ts
```diff
@@ -1,17 +1,139 @@
...
+                  target: "Finalize Route Pattern",
+                  actions: assign({
+                    routePatterns: ({ event }) => event.output,
+                    routePatternId: ({ event }) =>
+                      event.output.find((pattern) => pattern.directionId === 1)
+                        ?.id,
+                  }),
+                },
+              },
+            },
```
hannahpurcell, line 112: Checking my comprehension here: it seems that this state is a transitional state that doesn't wait for any inputs before moving along? Like, "Load Route Patterns" whenever the "select-route" event happens, and then it invokes the "fetch-route-patterns" logic. How does the onDone get triggered for this state? Does the state automatically go into "Finalize Route Pattern" when the fetch finishes? If that's the case, should "Load Route Patterns" be a state if it doesn't wait for an event / input?

firestack: Relevant docs: https://stately.ai/docs/invoke#ondone

---

I could theoretically see this state & actor being merged into `Pick Route ID`, where this actor is invoked every time that we enter that state, and we only transition out of that state once we have route patterns loaded.

Currently, I _think_ I prefer this as it's own state because I like that what the machine is doing is explicitly described, but I'm open to chatting about this more to figure out if that's desired!

I think if we did what you're saying, it would look like this?

<details>
<summary>Code</summary>

```ts
"Pick Route Pattern": {
  initial: "Pick Route ID",
  on: {
    "detour.route-pattern.select-route": {
      target: ".Pick Route ID",
      actions: assign({
        routeId: ({ event }) => event.routeId,
        routePatternId: undefined,
      }),
    },
  },
  states: {
    "Pick Route ID": {
      invoke: {
        src: "fetch-route-patterns",

        input: ({ context: { routeId } }) => ({ routeId: routeId }),
        onDone: {
          target: "Finalize Route Pattern",
          actions: assign({
            routePatterns: ({ event }) => event.output,
            routePatternId: ({ event }) =>
              event.output.find((pattern) => pattern.directionId === 1)
                ?.id,
          }),
        },
      },
    },
    "Finalize Route Pattern": {
      on: {
        "detour.route-pattern.done": {
          guard: ({ context }) => context.routePatternId !== undefined,
          target: "Done",
        },
        "detour.route-pattern.select-pattern": {
          actions: {
            type: "set.route-pattern",
            params: ({ event: { routePatternId } }) => ({
              routePatternId,
            }),
          },
        },
      },
    },
    Done: {
      type: "final",
    },
  },
  onDone: {
    target: "Editing",
  },
}
```

</details>

I'm still a little on the fence, but I somewhat like the above example code, so lets chat about it! Do you think that the implicit state in the above example reads better or anything else?

firestack: I'm warming up to this more and more FYI!
But I'm still feeling like there is _some_ reason to keep the explicit state that I can't put my finger on (and my intuition could be wrong here even if I do remember it)

firestack: One issue here, it makes the states read a little bit stranger when adding nested loading states, see
```ts
snapshot.matches({
  "Detour Drawing": {
    "Pick Route Pattern": { "Pick Route ID": "Loading" },
  },
})
```
vs
```ts
snapshot.matches({
  "Detour Drawing": {
    "Pick Route Pattern": { "Load Route Patterns": "Loading" },
  },
})
```

firestack: This makes it seem like the Route ID is loading, when it's the patterns that are loading.

PR #2638, opened 2024-05-30
feat: Make the part of the route shape before the start point unclickable

joshlarson:
As part of an effort to help dispatchers avoid errors, we want to prevent them from clicking on the portion of a route shape that would cause them to draw the detour backwards.

Screenshot 2024-06-03 at 3 52 49 PM

(In the above screenshot, the route flows from south to north, so the part of the route shape lying to the south of the start point is disabled.)

---

Asana Ticket: https://app.asana.com/0/1205385723132845/1207449517841759/f

Depends on:
- #2637
- #2641
- #2696

Blocks:
- #2639

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -51,13 +51,17 @@ import {
 } from "./models/locationSearchSuggestionData"
 import { LocationSearchSuggestion } from "./models/locationSearchSuggestion"
 import { DetourShapeData, detourShapeFromData } from "./models/detourShapeData"
-import { DetourShape, FinishedDetour } from "./models/detour"
+import { DetourShape, FinishedDetour, UnfinishedDetour } from "./models/detour"
 import {
   FinishedDetourData,
   finishedDetourFromData,
 } from "./models/finishedDetour"
 import { FetchResult, ok, fetchError } from "./util/fetchResult"
 import { Ok, Err, Result, map } from "./util/result"
+import {
+  UnfinishedDetourData,
+  unfinishedDetourFromData,
+} from "./models/unfinishedDetour"
 
 export interface RouteData {
   id: string
@@ -248,6 +252,21 @@ export const fetchDetourDirections = (
       coordinates,
     })
   ).then((v) => map(v, detourShapeFromData))
+
+export const fetchUnfinishedDetour = (
+  routePatternId: RoutePatternId,
+  connectionStart: ShapePoint
+): Promise<UnfinishedDetour | null> =>
+  checkedApiCall<UnfinishedDetourData, UnfinishedDetour | null>({
+    url: "/api/detours/unfinished_detour",
+    parser: unfinishedDetourFromData,
+    dataStruct: UnfinishedDetourData,
+    defaultResult: null,
+    fetchArgs: postJsonParameter({
+      route_pattern_id: routePatternId,
+      connection_start: connectionStart,
+    }),
+  })
 
 export const fetchFinishedDetour = (
   routePatternId: RoutePatternId,
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -18,7 +18,7 @@ import {
 } from "../../util/pointLiterals"
 import { MapTooltip } from "../map/tooltip"
 import { joinClasses } from "../../helpers/dom"
-import { RouteSegments } from "../../models/detour"
+import { RouteSegments, UnfinishedRouteSegments } from "../../models/detour"
 import { MapButton } from "../map/controls/mapButton"
 import { ArrowLeftSquare, XSquare } from "../../helpers/bsIcons"
 import ZoomLevelWrapper from "../ZoomLevelWrapper"
@@ -62,6 +62,11 @@ interface DetourMapProps {
    * Coordinates to display as the waypoints.
    */
   waypoints: ShapePoint[]
+
+  /**
+   * Two partial route-shape segments: before and after the start point
+   */
+  unfinishedRouteSegments?: UnfinishedRouteSegments
 
   /**
    * Three partial route-shape segments: before, during, and after the detour
@@ -116,6 +121,8 @@ export const DetourMap = ({
 
   onClickOriginalShape,
   onAddWaypoint,
+
+  unfinishedRouteSegments,
 
   routeSegments,
 
@@ -228,7 +235,15 @@ export const DetourMap = ({
         />
 
         {routeSegments ? (
-          <DivertedRouteShape segments={routeSegments}></DivertedRouteShape>
+          <DivertedRouteShape segments={routeSegments} />
+        ) : unfinishedRouteSegments ? (
+          <UnfinishedDiversionRouteShape
+            interactive={!streetViewEnabled}
+            segments={unfinishedRouteSegments}
+            onClick={(e) => {
+              onClickOriginalShape(latLngLiteralToShapePoint(e.latlng))
+            }}
+          />
         ) : (
           <OriginalRouteShape
             positions={originalShape.map(shapePointToLatLngLiteral)}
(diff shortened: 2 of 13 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -144,6 +146,7 @@ export const DiversionPage = ({
             startPoint={startPoint ?? undefined}
             endPoint={endPoint ?? undefined}
             waypoints={waypoints}
+            unfinishedRouteSegments={unfinishedRouteSegments}
```
joshlarson, line 149: This needs to be gated behind a test group, so that's why this PR is still in Draft. 

I think it's sort of ready for review though!

assets/src/components/detours/detourMap.tsx
```diff
@@ -378,6 +393,47 @@ const OriginalRouteShape = ({
...
+      positions={afterStartPoint.map(shapePointToLatLngLiteral)}
+      className="c-detour_map--original-route-shape-core"
+    />
+    {interactive && (
+      <Polyline
+        positions={afterStartPoint.map(shapePointToLatLngLiteral)}
+        weight={16}
+        className={joinClasses([
+          "c-detour_map--original-route-shape-unfinished-diversion",
+          "c-detour_map--original-route-shape__unfinished",
```
joshlarson, line 426: The combination of these two _totally_ different CSS naming conventions is gross, but I was thinking of waiting until a bit later to clean it up, because exactly _how_ to clean it up may depend on what happens when we remove the test group.

(It's possible that after doing some user-testing, Kathleen will decide that this wasn't the right way to signal the route-direction to dispatchers and we'll wind up backing out all of these changes. I think that's unlikely, but it is possible)

assets/tests/factories/finishedDetourFactory.ts
```diff
@@ -1,8 +1,23 @@
...
+import {
+  FinishedDetour,
+  RouteSegments,
+  UnfinishedDetour,
+  UnfinishedRouteSegments,
+} from "../../src/models/detour"
 import { shapePointFactory } from "./shapePointFactory"
 import stopFactory from "./stop"
 
+export const unfinishedRouteSegmentsFactory =
```
joshlarson, line 11: The name of this file is misleading now.

Should I:
a) Separate `unfinishedRouteSegmentsFactory` and `unfinishedDetourFactory` out to a different factory file (thus leaving this file unchanged); OR
b) Keep these as-is, but rename the file to `detourFactory.ts`?

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -870,8 +878,32 @@ describe("DiversionPage", () => {
...
     await waitFor(() => {
-      expect(originalRouteShape.get(container)).toBeVisible()
+      expect(screen.queryByTitle("Detour Start")).toBeVisible()
     })
+
+    expect(originalRouteShape.get(container)).toBeVisible()
+  })
+
+  test("replaces the original route shape with an unfinished segment after the start point is added if the user is in the right test group", async () => {
+    jest
```
joshlarson, line 888: This test (and the one above it) is (are?) a bit devious. 

Because route segments show up in response to an API call, the test above (the one that makes sure that `originalRouteShape` is still there when the test group isn't enabled) still passes, even if we don't gate it by test group, because the `fetch` call defaults to `neverPromise()`. 

But if I set the `fetch` call to default to having a realistic return value, then a whole bunch of the tests start complaining that there's a change not wrapped in `act(...)`, because the promise resolves after the test is finished executing.

I'm not sure how to get around this 🙃

assets/src/api.ts
```diff
@@ -249,6 +253,21 @@ export const fetchDetourDirections = (
     })
   ).then((v) => map(v, detourShapeFromData))
 
+export const fetchUnfinishedDetour = (
+  routePatternId: RoutePatternId,
+  connectionStart: ShapePoint
+): Promise<UnfinishedDetour | null> =>
+  checkedApiCall<UnfinishedDetourData, UnfinishedDetour | null>({
```
firestack, line 260: is there a reason to prefer `checkedApiCall` over maybe `apiCallResult`?

firestack: We could probably mark those other ones as deprecated (in another PR), thoughts?

joshlarson: Not _immediately_ opposed. Would be interested in thinking more about it later.

assets/src/components/detours/detourMap.tsx
```diff
@@ -378,6 +393,47 @@ const OriginalRouteShape = ({
...
+      interactive={false}
+      positions={beforeStartPoint.map(shapePointToLatLngLiteral)}
+      className="c-detour_map--original-route-shape-core"
+    />
+    <Polyline
+      weight={6}
+      interactive={false}
+      positions={afterStartPoint.map(shapePointToLatLngLiteral)}
+      className="c-detour_map--original-route-shape-core"
+    />
```
hannahpurcell, line 419: Could these be easily combined into 1 polyline by passing both segments into positions?

assets/src/components/detours/detourMap.tsx
hannahpurcell: Why is `c-detour_map--original-route-shape-unfinished-diversion` needed? It could be `c-detour_map--original-route-shape`, which seems to me a semantically accurate classname here.

assets/src/components/detours/detourMap.tsx
hannahpurcell: non-blocking question:  

Related to the above comment, I just looked at the `OriginalRouteShape` code for the first time, which, naturally, overlaps with this alternative `UnfinishedDiversionRouteShape`. It occurs to me, could they have been a single component? It could still be called `OriginalRouteShape`, as it still references the existing route pre-detour. It would need a new field for the positions of the interactive line, but that's it. Basically, once a detour is starting to be drawn, the interactive line is truncated, but the core line remains the same. 

Not yet asking for a change here (would want to weight the cost-benefit before genuinely suggesting a change)

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: > But if I set the fetch call to default to having a realistic return value, then a whole bunch of the tests start complaining

If you mock the `fetch` call right in the test test `keeps the original route shape after...`, I don't see why other tests would start complaining, so I think I'm missing something simple 🤔

assets/tests/factories/finishedDetourFactory.ts
hannahpurcell: I like `b)`!

assets/src/components/detours/detourMap.tsx
hannahpurcell: Ohhhhhhhhhh.... is it for the test??

hannahpurcell (commented):
A few questions, but would be ok approving after we discuss

assets/src/components/detours/detourMap.tsx
joshlarson: Hummmmmm - I'm wondering if it would make more sense to combine this PR with https://github.com/mbta/skate/pull/2639. A lot of the suggestions you're making.. I think they do make sense in the context of this work alone, but I think I'd have to undo them in order to to https://github.com/mbta/skate/pull/2639

joshlarson: But yeah... the "core" could be its own element, since it renders the same in all three of `OriginalRouteShape`, `UnfinishedDiversionRouteShape`, and `DivertedRouteShape`.

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: Oh - if I mock it only in that test, then that test fails because the `fetch` resolves after the test wraps up, and I don't know how to make the test wait for the `fetch` to resolve, since the `fetch` resolving doesn't cause any changes on the page....

At least I _think_ that's what's happening here?

assets/src/components/detours/detourMap.tsx
joshlarson: Yep. This is so that I can easily assert that the route shape that's displayed has changed.

joshlarson: Although I guess I could assert on the number of `core` shapes there are...

This might affect https://github.com/mbta/skate/pull/2641 also.

joshlarson: Oohhh actually, maybe [this would be a better way to make these assertions](https://github.com/mbta/skate/blob/e14d100c1db7df8bbf3c24ca4f8dbbe100d24d1e/assets/tests/components/detours/detourMap.test.tsx#L235-L273)?

firestack (commented):
awaiting changes and comments in response to @hannahpurcell's review. (commenting to get this out of my "review requests" page.)

assets/src/api.ts
joshlarson: My preference is to keep this and `fetchFinishedDetour` the same. Let's change them both over as a follow-up?

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: Turns out [this](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#an-alternative-waiting-for-the-mocked-promise) works! So we can wait for the promise returned by `fetchUnfinishedDetour` to resolve and _then_ assert that the route segments on the screen are correct, then that test tests what we want (and fails if I don't do the test group correctly)

assets/tests/factories/finishedDetourFactory.ts
joshlarson: [Will do this as a follow-up](https://app.asana.com/0/1205718271156548/1207845782956705/f)

hannahpurcell:
Noting here a conversation we had in slack: seems that there may have been a functionality regression after some of the forced-pushes 😥

hannahpurcell (approved):
Ok! Gave a briefer review this round because I gave my main review previously, and it functions beautifully. I think a future refactor of the OriginalShape / UnfinishedDiversionRouteShape overlap would make detourMap more readable, as well as the css. But at this point, I think it's best to get this feature out there, and we'll refactor _if / when_ that code gets touched again

PR #2640, opened 2024-05-30
fix: Missed stops bug when stop is close to start point

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207328357694128/f

Depends On:
- https://github.com/mbta/skate/pull/2664

### DIFF
--- a/lib/skate/detours/missed_stops.ex
+++ b/lib/skate/detours/missed_stops.ex
@@ -124,7 +124,7 @@ defmodule Skate.Detours.MissedStops do
     shape_after_anchor = [anchor_point] ++ points_after_anchor
 
     {segments_before_anchor, [incomplete_anchor_segment]} =
-      segment_shape_by_stops(shape_before_anchor, stops_before_anchor) |> Enum.split(-1)
+      shape_before_anchor |> segment_shape_by_stops(stops_before_anchor) |> Enum.split(-1)
 
     anchor_segment = %Skate.Detours.ShapeSegment{incomplete_anchor_segment | stop: anchor_stop}
 
@@ -141,35 +141,12 @@ defmodule Skate.Detours.MissedStops do
         ) :: non_neg_integer()
   defp get_index_by_min_dist(shape_segments, reference) do
     shape_segments
-    |> Enum.with_index(fn %Skate.Detours.ShapeSegment{points: points} = segment, idx ->
-      %{segment: segment, closest: closest_point(points, reference), index: idx}
-    end)
-    |> Enum.min_by(fn %{closest: %{dist: dist}} -> dist end)
-    |> Map.fetch!(:index)
-  end
+    |> Enum.with_index(fn %Skate.Detours.ShapeSegment{points: points}, index ->
+      {point, _} = NearestPoint.nearest_point_on_shape(points, reference)
 
-  @type point_dist_index :: %{
-          elem: Util.Location.From.t(),
-          dist: float(),
-          index: non_neg_integer()
-        }
-  @spec closest_point(
-          coordinates :: [Util.Location.From.t()],
-          reference :: Util.Location.From.t()
-        ) ::
-          point_dist_index()
-
-  defp closest_point(coordinates, reference) do
-    reference = Util.Location.as_location!(reference)
-
-    dist_from_reference = fn point ->
-      Location.distance(reference, Util.Location.as_location!(point))
-    end
-
-    coordinates
-    |> Enum.with_index(fn point, idx ->
-      %{elem: point, dist: dist_from_reference.(point), index: idx}
+      %{index: index, dist: Location.distance(reference, point)}
     end)
     |> Enum.min_by(& &1.dist)
+    |> Map.fetch!(:index)
   end
 end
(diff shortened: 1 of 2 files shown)


lib/skate/detours/missed_stops.ex
```diff
@@ -72,11 +73,11 @@ defmodule Skate.Detours.MissedStops do
           connection_end_segment: Skate.Detours.ShapeSegment.t() | nil
         }
   defp missed_segments({connection_start, connection_end}, segmented_shape) do
-    %{index: start_index} = get_index_by_min_dist(segmented_shape, connection_start)
+    start_index = get_index_by_min_dist(segmented_shape, connection_start)
```
firestack, line 76: nit(non-blocking): would've preferred either a specific commit that describes this change or a seperate PR, because IMO, changing the interface adds noise to the overall diff of this change.

lib/skate/detours/missed_stops.ex
```diff
@@ -92,99 +93,63 @@ defmodule Skate.Detours.MissedStops do
...
-    anchor_segment = %Skate.Detours.ShapeSegment{
-      (%Skate.Detours.ShapeSegment{stop: :none} = anchor_segment)
-      | stop: anchor
-    }
+    {segments_before_anchor, [incomplete_anchor_segment]} =
+      shape_before_anchor |> segment_shape_by_stops(stops_before_anchor) |> Enum.split(-1)
 
-    left_segments ++ [anchor_segment] ++ segment_shape_by_stops(right_shape, right_stops)
-  end
+    anchor_segment = %Skate.Detours.ShapeSegment{incomplete_anchor_segment | stop: anchor_stop}
```
firestack, line 130: question: I recall that the only reason we noticed that there was an issue originally was because we had the `%Skate.Detours.ShapeSegment{stop: :none} = anchor_segment` assertion, do you think it's worth dropping that assertion? 

I'd rather have it than not personally.

lib/skate/detours/missed_stops.ex
joshlarson: Good call!

Lemme see if I can get that out promptly.

lib/skate/detours/missed_stops.ex
joshlarson: Oooohhhoohooo THAT's why we had that assertion!

Yes - let's add that back in

test/skate/detours/missed_stops_test.exs
```diff
@@ -200,18 +201,267 @@ defmodule Skate.Detours.MissedStopsTest do
...
       assert %MissedStops.Result{missed_stops: missed_stops} =
-               Skate.Detours.MissedStops.missed_stops(param)
+               MissedStops.missed_stops(param)
 
       assert Enum.slice(stops, connection_start_index..(connection_end_index - 1)) ==
                missed_stops
     end
+
+    test "counts stops as missed if they're just inside the detour" do
+      #                               Both stops are missed
```
joshlarson, line 219: Normally I prefer Github's split-diff view, but it really messes with this otherwise truly beautiful ASCII-art 😂

lib/skate/detours/missed_stops.ex
```diff
@@ -92,99 +93,63 @@ defmodule Skate.Detours.MissedStops do
           shape :: [Util.Location.From.t()],
           stops :: [Util.Location.From.t()]
         ) :: [Skate.Detours.ShapeSegment.t()]
-  defp segment_shape_by_stops([] = _shape, [] = _stops),
+  def segment_shape_by_stops([] = _shape, [] = _stops),
```
firestack, line 96: thought(non-blocking/address in another PR): IIRC we originally considered this a implementation detail that shouldn't be tested, now that it's being tested, should this become a generic "segment shape by projected points" that's generic across `Util.Location.From` elements so we retain the same functionality?

I _think_ we could then use this for both route segments and missed stops?

firestack: Like, I think the important thing about this function is that it's "generic" across inputs as long as they implement `Util.Location.From`, and that this function allows extra data to be stuffed in with the algorithm's return value.

firestack (commented):
Having a little trouble groking the "works when the shape loops around and an intersection stop is closer to the wrong part of the route" and similar test, I think I just need to diagram it out in tldraw or something, otherwise looking good, I don't think I have any other comments to open.

lib/skate/detours/missed_stops.ex
joshlarson: https://github.com/mbta/skate/pull/2664

lib/skate/detours/missed_stops.ex
joshlarson: Ooooh I like that!

Definitely fodder for a follow-up refactor PR. Will queue that up on my own personal task-list.

lib/skate/detours/missed_stops.ex
joshlarson: [How's this?](https://github.com/mbta/skate/pull/2640/commits/6970a232b0c2a015f534641d36e0cb28cedcd18c)

joshlarson:
> Having a little trouble groking the "works when the shape loops around and an intersection stop is closer to the wrong part of the route" and similar test, I think I just need to diagram it out in tldraw or something, otherwise looking good, I don't think I have any other comments to open.

That makes sense! This is super confusing and it took me a whole bunch of conversations and several months to get it.

I'd love to not block merging this on that, but I'd be up for making an ASCII-art diagram for these test cases as well as a follow-up.

firestack (approved):

PR #2641, opened 2024-06-03
tests: Add tests that assert on visible route sections

joshlarson:
This is a prequel to some work that we're doing to disable clicking on the portion of a route shape that comes before the start point, in order to make it less likely that a dispatcher will draw a detour backwards. 

We need some tests to validate that we are doing ☝️ correctly, and that it's only enabled when the user is in a particular test group. This PR adds tests that validate the existing behavior.

---

Asana Ticket: https://app.asana.com/0/1205385723132845/1207475016947553/f

Blocks:
- #2638
- #2639

### DIFF
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -932,6 +932,50 @@ describe("DiversionPage", () => {
     })
   })
 
+  test("has (only) an original route shape at the beginning", async () => {
+    const { container } = render(<DiversionPage />)
+
+    await waitFor(() => {
+      expect(originalRouteShape.get(container)).toBeVisible()
+    })
+
+    expect(originalRouteShape.diverted.getAll(container)).toHaveLength(0)
+  })
+
+  test("keeps the original route shape after the start point is added", async () => {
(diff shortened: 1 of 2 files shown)


firestack (dismissed):

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -933,6 +933,52 @@ describe("DiversionPage", () => {
...
+
+    await waitFor(() => {
+      expect(originalRouteShape.get(container)).toBeVisible()
+    })
+  })
+
+  test("replaces the original route shape with a diverted segment after the end point is added", async () => {
+    jest.mocked(fetchFinishedDetour).mockResolvedValue(
+      finishedDetourFactory.build({
+        routeSegments: routeSegmentsFactory.build(),
```
firestack, line 961: question(non-blocking): is there a reason to make this explicit but without parameters?

firestack (approved):

assets/tests/components/detours/diversionPage.test.tsx
joshlarson: Lol nope

PR #2643, opened 2024-06-04
fix: Lil styling fix on the route ladder headers

hannahpurcell:
Before
Screenshot 2024-06-04 at 10 46 13 AM

After
Screenshot 2024-06-04 at 10 59 21 AM

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -55,7 +55,8 @@
 
   .c-route-ladder__header__action-container {
     display: inline-flex;
-    .dropdown, .c-route-ladder__alert-icon {
+    .dropdown,
+    .c-route-ladder__alert-icon {
       padding-right: 0.5rem;
     }
   }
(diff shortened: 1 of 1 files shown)


joshlarson:
I'm not sure I understand the change? The alert icon appears in the "after" picture, but not the "before". Is the change that the alert icon got covered up by the dropdown or something like that?

hannahpurcell:
@joshlarson Ah the contrast might not be showing up well enough for ya, but in both I'm hovering the dropdown content, and in the first the content doesn't extend to the right edge, and in the second it does. 

The culprit line was written that way so that it'd apply padding to both the dropdown toggle and the alert icon, so I wanted to make sure to get those buttons in frame for the fix, but the focus of the fix was the dropdown content.

joshlarson (approved):
OH HO HO HO It all makes sense now!

I thought that this was something that bootstrap `<Dropdown />` can do automagically, but yeah - it does look a lot better

(it really shows up when you click!)

firestack (changes requested):
I think there may be a better way, a moment

PR #2645, opened 2024-06-04
feat(ts/detours): add context and input to detour machine

firestack:
To be able to control the route and variant, the state machine needs to be the source of truth for the active route and pattern. This moves the values into the state machine and updates callers to send the route and pattern.

Asana Ticket: https://app.asana.com/0/0/1207481283409966/f

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -14,6 +14,25 @@ import { OriginalRoute } from "../../models/detour"
 import { joinClasses } from "../../helpers/dom"
 import { AsProp } from "react-bootstrap/esm/helpers"
 import { DetourFinishedPanel } from "./detourFinishedPanel"
+import { Route, RoutePattern } from "../../schedule"
+
+const displayFieldsFromRouteAndPattern = (
+  route: Route,
+  routePattern: RoutePattern
+) => {
+  const routeName = route.name
+
+  const routeDescription = routePattern.headsign ?? ""
+
+  const routeOrigin = routePattern?.name
+
+  const routeDirection =
+    routePattern && route.directionNames[routePattern.directionId]
+
+  const shape = routePattern.shape
+
+  return { routeName, routeDirection, routeOrigin, routeDescription, shape }
+}
 
 interface DiversionPageProps {
   originalRoute: OriginalRoute
@@ -66,11 +85,22 @@ export const DiversionPage = ({
     ? nearestIntersectionDirection.concat(directions)
     : undefined
 
+  const { route, routePattern } = snapshot.context
+  const {
+    routeName = undefined,
+    routeDirection = undefined,
+    routeOrigin = undefined,
+    routeDescription = undefined,
+    shape = undefined,
+  } = route && routePattern
+    ? displayFieldsFromRouteAndPattern(route, routePattern)
+    : {}
+
   useEffect(() => {
     setTextArea(
       [
-        `Detour ${originalRoute.routeName} ${originalRoute.routeDirection}`,
-        originalRoute.routeOrigin,
+        `Detour ${routeName} ${routeDirection}`,
+        routeOrigin,
         ,
         "Connection Points:",
         connectionPoints?.start?.name ?? "N/A",
@@ -84,7 +114,9 @@ export const DiversionPage = ({
       ].join("\n")
     )
   }, [
-    originalRoute,
+    routeName,
+    routeDirection,
+    routeOrigin,
     extendedDirections,
     missedStops,
     connectionPoints?.start?.name,
@@ -103,10 +135,10 @@ export const DiversionPage = ({
             <DiversionPanel
               directions={extendedDirections}
               missedStops={missedStops}
-              routeName={originalRoute.routeName}
-              routeDescription={originalRoute.routeDescription}
-              routeOrigin={originalRoute.routeOrigin}
-              routeDirection={originalRoute.routeDirection}
+              routeName={routeName}
+              routeDescription={routeDescription}
+              routeOrigin={routeOrigin}
+              routeDirection={routeDirection}
               detourFinished={finishDetour !== undefined}
               onFinishDetour={finishDetour}
             />
@@ -137,7 +169,7 @@ export const DiversionPage = ({
           )}
           {routingError?.type === "unknown" && <RoutingErrorAlert />}
           <DetourMap
-            originalShape={originalRoute.shape.points}
+            originalShape={shape?.points ?? []}
             center={originalRoute.center}
             zoom={originalRoute.zoom}
             detourShape={detourShape}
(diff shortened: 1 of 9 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -294,3 +300,18 @@ DiversionPagePanel.Body = DiversionPagePanelBody
 DiversionPage.Panel = DiversionPagePanel
 
 export const Panel = DiversionPagePanel
+
+function descriptors(route: Route, routePattern: RoutePattern) {
```
firestack, line 304: todo: needs a better name

assets/src/components/detours/diversionPage.tsx
```diff
@@ -137,7 +143,7 @@ export const DiversionPage = ({
           )}
           {routingError?.type === "unknown" && <RoutingErrorAlert />}
           <DetourMap
-            originalShape={originalRoute.shape.points}
+            originalShape={shape?.points ?? []}
```
firestack, line 146: note: is this a breaking change? Not sure what to do here since we can't derive shapes from the route pattern and expect that they're always there.

assets/src/components/detours/diversionPage.tsx
firestack: also probably shouldn't be `function`

assets/src/hooks/useDetour.ts
```diff
@@ -22,8 +22,12 @@ const useDetourDirections = (shapePoints: ShapePoint[]) =>
     }, [shapePoints]),
   })
 
-export const useDetour = ({ routePatternId, shape }: OriginalRoute) => {
-  const [snapshot, send] = useMachine(createDetourMachine)
+export const useDetour = (input: OriginalRoute) => {
```
firestack, line 25: note:suggestion: I'd almost like to remove `OriginalRoute` here altogether and use, haven't come up with any downsides yet.
```ts
InputFrom<
  ActorLogicFrom<typeof createDetourMachine>
>
```

assets/src/models/createDetourMachine.ts
```diff
@@ -1,11 +1,25 @@
...
+    input: {
+      // Caller has target route pattern
+      route: Route
+      routePattern: RoutePattern
+    }
     events: { type: "detour.edit.done" } | { type: "detour.edit.resume" }
   },
 }).createMachine({
   id: "Detours Machine",
+  context: ({ input }) => ({
```
firestack, line 20: note:nit: could be simplified

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -35,31 +33,23 @@ import {
...
+import { originalRouteFactory } from "../../factories/originalRouteFactory"
+import { DeepPartial } from "fishery"
 
 const DiversionPage = (
   props: Omit<
     Partial<ComponentProps<typeof DiversionPageDefault>>,
     "originalRoute"
   > & {
-    originalRoute?: Partial<
+    originalRoute?: DeepPartial<
```
firestack, line 44: note: this could be it's own PR if desired.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1039,13 +1029,16 @@ describe("DiversionPage", () => {
...
+          route: {
+            name: routeName,
+          },
+          routePattern: {
+            headsign: routeDescription,
+            name: routeOrigin,
+            shape: {
+              points: [connectionPoint],
+            },
+          },
```
firestack, line 1041: note: unsure about this due to how encapsulation is broken here.

joshlarson: IMO this is fine. The fact that `routedDescription`, for instance, comes from the `headsign` of the `routePattern` is a meaningful statement about how `<DiversionPage />` works, so it's not wrong to have the test be explicit about that here.

assets/src/hooks/useDetour.ts
firestack: This gives us the option to add more to `input` without having to change the `useDetour` interface to use it. going to add this to this PR.

assets/src/components/dummyDetourPage.tsx
```diff
@@ -18,16 +18,11 @@ export const DummyDetourPage = () => {
...
-      {routePattern && routePattern.shape && (
+      {route && routePattern && routePattern.shape && (
         <DiversionPage
           originalRoute={{
-            shape: routePattern.shape,
-            routeName: routePattern.routeId,
-            routeDescription: routePattern.headsign || "?",
-            routeOrigin: routePattern.name,
-            routeDirection:
-              route?.directionNames[routePattern.directionId] || "?",
```
joshlarson, line 29: Lol I forgot that I put that in there as `"?"` - I would have been interested to see the bug report when a dispatcher saw `Inbound`/`Outbound`/`?`

assets/src/components/detours/diversionPage.tsx
firestack: > Out of curiosity, how come you wanted this to be extracted as a function, rather than just inlined directly to `DiversionPage` itself?

It was just real messy and I was getting annoyed with how much noise it was adding

firestack: > And yeah, I _think_ our conventions are that this should be `const descriptors = (...) => {`?

Agreed, though I'm torn because the only reason this can be at the bottom of the file (which is where _I_ would prefer it) is due to hoisting. So converting this to `const` would mean it needs to live at the top of the file, before it's first invocation. So maybe we reconsider this practice at some point but I'll change it for now.

Given it'd have to be the first invocation, do you think in-lining it would be better?

joshlarson: Huhhhhhhh - I forgot about that. I think per what our team has been doing so far, it should be a `const` and live at the top of the file (I'm so used to that in our React code that it doesn't really bother me, but now that you mention it, it is weird that we like... sometimes have the "main component" at the top, and sometimes have it at the bottom). Happy to change our rules on that, but we should make an intentional choice to do that, and then follow the new rules.

No objection to in-lining, but no pressure to do so either.

firestack: @joshlarson what do you think of https://github.com/mbta/skate/compare/4829f3842bf9b7304a55362b17f8c5aa2aa74f80..e180162268a170207ddc2513135b1857d91ef998 ?

joshlarson: Probably fine, but since it's above anyway, does it need to be capitalized?

firestack: oh whoops, didn't mean to do that

firestack: https://github.com/mbta/skate/compare/b2625564eee19238849772cd11042fdccdff4e81..dca6719b349e19cc09ed41d051d104eed913fbbc

joshlarson (approved):

PR #2647, opened 2024-06-05
feat: Open the detour modal from a route ladder

joshlarson:
https://github.com/mbta/skate/assets/912020/89b2a491-7030-43dc-ae4e-7d331e8b0a1c

Depends on:
- https://github.com/mbta/skate/pull/2645
- https://github.com/mbta/skate/pull/2652

Asana Ticket: https://app.asana.com/0/1205385723132845/1207260183650060/f

### DIFF
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -17,8 +17,8 @@ interface SelectedRouteInfoWithRoutePatterns {
 
 interface SelectedRouteInfoWithoutRoutePatterns {
   selectedRoute: null | Route
-  routePatterns: undefined
-  selectedRoutePatternId: undefined
+  routePatterns?: undefined
+  selectedRoutePatternId?: undefined
 }
 
 type SelectedRouteInfo =
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -88,22 +88,26 @@ export const DetourRouteSelectionPanel = ({
 
           <section className="pb-3">
             <h2 className="c-diversion-panel__h2">Choose direction</h2>
-            {selectedRouteInfo.routePatterns ? (
+            {selectedRouteInfo.selectedRoute ? (
               <div className="position-relative">
-                <RoutePropertiesCard
-                  routePatterns={selectedRouteInfo.routePatterns}
-                  selectedRoutePatternId={selectedRoutePatternFromInfo(
-                    selectedRouteInfo
-                  )}
-                  selectRoutePattern={onSelectRoutePattern}
-                  defaultOpened="variants"
-                />
-                <div
-                  hidden={!isLoadingRoutePatterns}
-                  className="position-absolute inset-0 bg-light opacity-75 d-flex justify-content-center align-items-center"
-                >
-                  <Spinner />
-                </div>
+                {selectedRouteInfo.routePatterns && (
+                  <>
+                    <RoutePropertiesCard
+                      routePatterns={selectedRouteInfo.routePatterns}
+                      selectedRoutePatternId={selectedRoutePatternFromInfo(
+                        selectedRouteInfo
+                      )}
+                      selectRoutePattern={onSelectRoutePattern}
+                      defaultOpened="variants"
+                    />
+                    <div
+                      hidden={!isLoadingRoutePatterns}
+                      className="position-absolute inset-0 bg-light opacity-75 d-flex justify-content-center align-items-center"
+                    >
+                      <Spinner />
+                    </div>
+                  </>
+                )}
               </div>
             ) : (
               <p className="fst-italic">
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -9,19 +9,21 @@ import {
 } from "../../schedule"
 import RoutePropertiesCard from "../mapPage/routePropertiesCard"
 
-interface SelectedRouteInfoWithRoute {
+interface SelectedRouteInfoWithRoutePatterns {
   selectedRoute: Route
   routePatterns: ByRoutePatternId<RoutePattern>
   selectedRoutePatternId: RoutePatternId | null
 }
 
-interface SelectedRouteInfoWithoutRoute {
-  selectedRoute: null
+interface SelectedRouteInfoWithoutRoutePatterns {
+  selectedRoute: null | Route
+  routePatterns: undefined
+  selectedRoutePatternId: undefined
 }
 
 type SelectedRouteInfo =
-  | SelectedRouteInfoWithRoute
-  | SelectedRouteInfoWithoutRoute
+  | SelectedRouteInfoWithRoutePatterns
+  | SelectedRouteInfoWithoutRoutePatterns
 
 interface DetourRouteSelectionPanelProps {
   allRoutes: Route[]
@@ -35,7 +37,7 @@ interface DetourRouteSelectionPanelProps {
 }
 
 const selectedRoutePatternFromInfo = (
-  selectedRouteInfo: SelectedRouteInfoWithRoute
+  selectedRouteInfo: SelectedRouteInfoWithRoutePatterns
 ): RoutePatternId =>
   selectedRouteInfo.selectedRoutePatternId ||
   Object.values(selectedRouteInfo.routePatterns).find(
@@ -86,7 +88,7 @@ export const DetourRouteSelectionPanel = ({
 
           <section className="pb-3">
             <h2 className="c-diversion-panel__h2">Choose direction</h2>
-            {selectedRouteInfo.selectedRoute ? (
+            {selectedRouteInfo.routePatterns ? (
               <div className="position-relative">
                 <RoutePropertiesCard
                   routePatterns={selectedRouteInfo.routePatterns}
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -157,7 +157,11 @@ export const DiversionPage = ({
                       selectedRoute: route,
                       selectedRoutePatternId: routePattern?.id ?? null,
                     }
-                  : { selectedRoute: null }
+                  : {
+                      routePatterns: undefined,
+                      selectedRoute: route ?? null,
+                      selectedRoutePatternId: undefined,
+                    }
               }
               onSelectRoute={(route) => {
                 if (route) {
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -158,9 +158,7 @@ export const DiversionPage = ({
                       selectedRoutePatternId: routePattern?.id ?? null,
                     }
                   : {
-                      routePatterns: undefined,
                       selectedRoute: route ?? null,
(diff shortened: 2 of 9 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -157,7 +157,11 @@ export const DiversionPage = ({
                       selectedRoute: route,
                       selectedRoutePatternId: routePattern?.id ?? null,
                     }
-                  : { selectedRoute: null }
+                  : {
+                      routePatterns: undefined,
+                      selectedRoute: route || null,
```
firestack, line 162: suggestion: this will convert `""` to `null`, instead, if we use the `??` operator, it'll only convert `undefined` and `null` to `null`.
```suggestion
                      selectedRoute: route ?? null,
```

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -9,19 +9,21 @@ import {
...
+interface SelectedRouteInfoWithRoutePatterns {
   selectedRoute: Route
   routePatterns: ByRoutePatternId<RoutePattern>
   selectedRoutePatternId: RoutePatternId | null
 }
 
-interface SelectedRouteInfoWithoutRoute {
-  selectedRoute: null
+interface SelectedRouteInfoWithoutRoutePatterns {
+  selectedRoute: null | Route
```
firestack, line 19: question(non-blocking): Would this be a chance to convert this also to `undefined`?

assets/src/components/ladderPage.tsx
```diff
@@ -143,6 +145,9 @@ const LadderPage = (): ReactElement<HTMLDivElement> => {
 
   const mobileMenuClass = mobileMenuIsOpen ? "blurred-mobile" : ""
 
+  const [showDetourModal, setShowDetourModal] = useState(false)
+  const [routeForDetour, setRouteForDetour] = useState<Route | null>(null)
```
firestack, line 149: question: should it be possible to have the route closed while keeping the last route saved?

That doesn't seem like intended behavior to me, so I think we could derive `showDetourModal`'s state from `routeForDetour !== null` instead of keeping track of two states which are implicitly related.

assets/src/components/routeLadder.tsx
```diff
@@ -241,6 +248,12 @@ const RouteLadder = ({
             inTestGroup(TestGroups.DetoursPilot) &&
             inTestGroup(TestGroups.DetourRouteSelection)
           }
+          onClickAddDetour={
+            onAddDetour &&
+            (() => {
+              onAddDetour(route)
+            })
```
firestack, line 255: suggestion(non-blocking): since we're not changing the display based on the presensce of this variable, we could simplify this callback, but of course is just taste and is optional.

But maybe I'm missing something about the function being passed in conditionally?
```suggestion
            (() => {
              onAddDetour?.(route)
            })
```

assets/tests/components/ladderPage.test.tsx
```diff
@@ -525,6 +532,52 @@ describe("LadderPage", () => {
...
+    const mockState = {
+      ...initialState,
+      routeTabs: [
+        routeTabFactory.build({
+          selectedRouteIds: ["1"],
+          ordering: 0,
+          isCurrentTab: true,
+        }),
+      ],
+    }
```
firestack, line 554: suggeston: it's inconsitent in this file, but we should be using the factory `stateFactory` to generate the global state.

Additionally, but not required, you shouldn't need to specify ordering because the `routeTabFactory` handles that AFAIK.

```suggestion
    const mockState = stateFactory.build({
      routeTabs: [
        routeTabFactory.build({
          selectedRouteIds: ["1"],
          isCurrentTab: true,
        }),
      ],
    })
```

(non-blocking, just a nit, don't feel like you have to) I'd personally replace the `"1"` route id with `routeFactory.build().id` but I figure you'd prefer the explicitness 😜

assets/tests/components/ladderPage.test.tsx
```diff
@@ -525,6 +532,52 @@ describe("LadderPage", () => {
...
+    const { container } = render(
+      <StateDispatchProvider state={mockState} dispatch={mockDispatch}>
```
firestack, line 556: suggestion: I don't think we should be accessing a "global" variable in tests, can we replace this?

```suggestion
      <StateDispatchProvider state={mockState} dispatch={jest.fn()}>
```

assets/tests/components/ladderPage.test.tsx
```diff
@@ -525,6 +532,52 @@ describe("LadderPage", () => {
...
+
+    await userEvent.click(
+      container.querySelector(".c-route-ladder__dropdown-button") as Element
+    )
+
+    expect(screen.getByRole("button", { name: "Add detour" })).toBeVisible()
+
+    await userEvent.click(screen.getByRole("button", { name: "Add detour" }))
+
+    expect(screen.getByText("Create Detour")).toBeVisible()
```
firestack, line 579: question(non-blocking): I'd prefer the `getByRole("heading", { name: "Create Detour" })` call here, is there a reason you wouldn't or would prefer `getByText` here?

firestack (changes requested):
A few comments we should chat about but it's looking good!

assets/tests/components/ladderPage.test.tsx
joshlarson: No - I was actually hoping to figure out the Right™ way to do this. I copied this from [the analogous mapPage.test](https://github.com/mbta/skate/blob/25c2f662b818f072b82afd7f1911795c675cc4df/assets/tests/components/mapPage.test.tsx#L1852).

assets/tests/components/ladderPage.test.tsx
```diff
@@ -525,6 +532,52 @@ describe("LadderPage", () => {
...
+
+    expect(container.querySelector(".c-new-route-ladder__header")).toBeVisible()
+
+    expect(
+      container.querySelector(".c-route-ladder__dropdown-button")
+    ).toBeVisible()
+
+    await userEvent.click(
+      container.querySelector(".c-route-ladder__dropdown-button") as Element
+    )
```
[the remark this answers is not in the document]
joshlarson: Oooh no - I just didn't know about the `!` operator. Although I have seen it used - I didn't really _get_ it.

assets/tests/components/ladderPage.test.tsx
joshlarson: LOL this comment and https://github.com/mbta/skate/pull/2652/files#r1631385371 are beautiful mirror images of each other

assets/src/components/detours/detourRouteSelectionPanel.tsx
joshlarson: I'm happy to do that, but I have a vague memory of:
- `undefined` is an implicit missing, like data might just not be initialized yet, or might just be an enhancement that doesn't really _mean_ anything
- `null` is a semantic "missing", in the sense that the user has selected a "missing" option, or cleared an input.

So my sense was that `null` was actually correct here.

Like I said though - happy to correct this if I misunderstood.

assets/src/components/ladderPage.tsx
joshlarson: I was thinking the same thing. I took the general concept from `<MapPage />`, which does the same thing, but it does ... seem odd to me.

joshlarson: I figured we did it in `<MapPage />` for a reason, but now that I think about it, it might have been more than I forgot to remove the unnecessary boolean in `<MapPage />` also 🤔

assets/src/components/detours/detourRouteSelectionPanel.tsx
firestack: I'm good with that rationale 👍🏻 thanks for sharing your thought process!

assets/src/components/ladderPage.tsx
joshlarson: [Fixed](https://github.com/mbta/skate/pull/2647/commits/dcdf89dfff324438f08a006e4c1bdb5584623896)

joshlarson: Ooooh I know the reason....

It's because of the exit animation. If we derive this state from `routeForDetour !== null`, then the whole detour modal disappears from the DOM immediately, while having `showDetourModal` as a separate prop allows it to have an exit animation.

assets/tests/components/ladderPage.test.tsx
joshlarson: This is fixed now, but I squash-rebased and lost the specific `fixup!` commit that did this.

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -86,7 +88,7 @@ export const DetourRouteSelectionPanel = ({
 
           <section className="pb-3">
             <h2 className="c-diversion-panel__h2">Choose direction</h2>
-            {selectedRouteInfo.selectedRoute ? (
+            {selectedRouteInfo.routePatterns ? (
```
firestack, line 91: nit(non-blocking): if you wrapped the RPC in this check, and left the `selectedRoute` check, then we'd properly display nothing when a route _is_ selected and then only show the RPC after it's loaded. This then makes the change to add a loading spinner while `routePatterns` is loading easier, but it's small enough that that this is fine to leave as is for now.

assets/tests/components/ladderPage.test.tsx
```diff
@@ -525,6 +532,50 @@ describe("LadderPage", () => {
...
+
+    expect(container.querySelector(".c-new-route-ladder__header")).toBeVisible()
+
+    expect(
+      container.querySelector(".c-route-ladder__dropdown-button")
+    ).toBeVisible()
+
+    await userEvent.click(
+      container.querySelector(".c-route-ladder__dropdown-button")!
+    )
```
firestack, line 571: nit(non-blocking): I don't like these selectors, so I opened #2657 to make it easier to test this button. Doesn't need to be addressed in this PR, but a followup PR to #2657 that adjust these tests would be nice.

assets/tests/components/ladderPage.test.tsx
firestack: I see it! [`48c6fba1c3f543a154189e273dc683f6c0507986..e47aa0764f9a0a259897026543c9194eec20744a`#diff-41b776228e](https://github.com/mbta/skate/compare/48c6fba1c3f543a154189e273dc683f6c0507986..e47aa0764f9a0a259897026543c9194eec20744a#diff-41b776228e88467a5a40b857932c4e403705be68b76270e8b50f440ec6abcb3cR577)

firestack (approved):
I left a few more nits, that I'm interested in chatting about, but they don't need to block merging necessarily

assets/src/components/detours/detourRouteSelectionPanel.tsx
joshlarson: Is [this](https://github.com/mbta/skate/pull/2647/commits/f5b33ec89b27e206ab880710157fb1cd37dd09d7) what you had in mind?

assets/tests/components/ladderPage.test.tsx
joshlarson: Oh yeah - that's a good call.

assets/src/components/detours/detourRouteSelectionPanel.tsx
firestack: lgtm

PR #2648, opened 2024-06-06
fix(ts/components/rpc): make it possible to find RPC routes via `getByRole`

firestack:
When writing tests for trying to assert how many route patterns are on screen, it wasn't actually possible to identify the group which contained all the radio buttons.

This enables test assertions such as the following
```ts
expect(
  await within(
    screen.getByRole("group", { name: "Variants" })
  ).findAllByRole("radio")
).toHaveLength(2)
```


hannahpurcell (approved):
👍

PR #2649, opened 2024-06-06
fix(ts/components/detourSelect): Allow selecting form select via role

firestack:
This enables test code such as 
```ts
await userEvent.selectOptions(
  screen.getByRole("combobox", { name: "Choose route" }),
  route2.name
)
```

### DIFF
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -56,8 +56,11 @@ export const DetourRouteSelectionPanel = ({
     <Panel.Body className="d-flex flex-column">
       <Panel.Body.ScrollArea className="d-flex flex-column">
         <section className="pb-3">
-          <h2 className="c-diversion-panel__h2">Choose route</h2>
+          <h2 className="c-diversion-panel__h2" id="choose-route-select">
+            Choose route
+          </h2>
           <FormSelect
+            aria-labelledby="choose-route-select"
             value={selectedRouteInfo.selectedRoute?.id}
             onChange={(changeEvent) => {
               onSelectRoute?.(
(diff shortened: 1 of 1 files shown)


hannahpurcell (approved):
✨

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -56,8 +56,11 @@ export const DetourRouteSelectionPanel = ({
     <Panel.Body className="d-flex flex-column">
       <Panel.Body.ScrollArea className="d-flex flex-column">
         <section className="pb-3">
-          <h2 className="c-diversion-panel__h2">Choose route</h2>
+          <h2 className="c-diversion-panel__h2" id="choose-route-select">
```
joshlarson, line 59: Are we worried at all about using `id` here, on the off-chance that there's more than one detour modal in the DOM?

firestack: Just a little, I could throw in a `useId` call, I just liked keeping the arrow function as a pure return 😭

joshlarson: It's ... proooooooobably worth doing that? I guess I don't fully know the implications of what happens if there are duplicate `id`'s, but it feels like the sort of area where a bug will creep in like four years from now and future-us will never find it 🙃 

RIP pure arrow function :(

joshlarson: (I disabled auto-merge so that this PR doesn't get surprise-merged while you/we are still thinking about this)

firestack: [There's a conflict which prevents this from being merged](https://github.com/mbta/skate/pull/2649/commits/3229af55870fd6cbfb9757f8468ce20e74bd2e30)

PR #2650, opened 2024-06-06
fix(ts/components/rpc): make radio buttons react controlled

firestack:
tests are succeeding where they should be failing because the HTML element is setting itself to `checked` even if the parent component did not instruct so.


hannahpurcell (approved):
✨

PR #2651, opened 2024-06-06
fix(ts/components/detourRouteSelection): control form value from react

firestack:
Without this, we fail to control this combo box when we have an initial selected value.

```
  ● DiversionPage › 'Route Selection Panel' › when a route pattern is already selected, opens to selected pattern

    expect(element).toHaveValue(route59)

    Expected the element to have value:
      route59
    Received:
      Select a route

      1444 |       expect(
      1445 |         screen.getByRole("combobox", { name: "Choose route" })
    > 1446 |       ).toHaveValue(selectedRoutePattern.routeId)
           |         ^
      1447 |       expect(
      1448 |         screen.getByRole("radio", {
      1449 |           name: RPCRadioButtonName(selectedRoutePattern),

      at tests/components/detours/diversionPage.test.tsx:1446:9
      at step (tests/components/detours/diversionPage.test.tsx:44:23)
      at Object.next (tests/components/detours/diversionPage.test.tsx:25:53)
      at fulfilled (tests/components/detours/diversionPage.test.tsx:16:58)
```

PR #2652, opened 2024-06-06
feat: add route selection panel

firestack:
Depends On:
- #2645
- #2648 
- #2649 
- #2650 
- #2651 

Asana Ticket: https://app.asana.com/0/1203014709808707/1207372095414476/f

### DIFF
--- a/assets/css/bootstrap/_map_overrides.scss
+++ b/assets/css/bootstrap/_map_overrides.scss
@@ -11,3 +11,14 @@ $theme-colors: (
   "dark": new_tokens.$gray-900,
   "missed-stop": semantic.$missed-stop,
 );
+
+$utilities: (
+  // Shorthand for specifying utility classes `top-0 start-0 bottom-0 end-0`
+  "inset":
+    (
+      property: inset,
+      values: (
+        0: 0,
+      ),
+    )
+);
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -1,5 +1,5 @@
 import React, { useId } from "react"
-import { Button, FormSelect } from "react-bootstrap"
+import { Button, FormSelect, Spinner } from "react-bootstrap"
 import { Panel } from "./diversionPage"
 import {
   ByRoutePatternId,
@@ -27,6 +27,8 @@ interface DetourRouteSelectionPanelProps {
   allRoutes: Route[]
   selectedRouteInfo: SelectedRouteInfo
 
+  isLoadingRoutePatterns: boolean
+
   onConfirm: () => void
   onSelectRoute: (route: Route | undefined) => void
   onSelectRoutePattern: (routePattern: RoutePattern | undefined) => void
@@ -47,6 +49,7 @@ export const DetourRouteSelectionPanel = ({
   onConfirm,
   onSelectRoute,
   onSelectRoutePattern,
+  isLoadingRoutePatterns,
 }: DetourRouteSelectionPanelProps) => {
   const selectId = "choose-route-select" + useId()
   return (
@@ -84,14 +87,22 @@ export const DetourRouteSelectionPanel = ({
           <section className="pb-3">
             <h2 className="c-diversion-panel__h2">Choose direction</h2>
             {selectedRouteInfo.selectedRoute ? (
-              <RoutePropertiesCard
-                routePatterns={selectedRouteInfo.routePatterns}
-                selectedRoutePatternId={selectedRoutePatternFromInfo(
-                  selectedRouteInfo
-                )}
-                selectRoutePattern={onSelectRoutePattern}
-                defaultOpened="variants"
-              />
+              <div className="position-relative">
+                <RoutePropertiesCard
+                  routePatterns={selectedRouteInfo.routePatterns}
+                  selectedRoutePatternId={selectedRoutePatternFromInfo(
+                    selectedRouteInfo
+                  )}
+                  selectRoutePattern={onSelectRoutePattern}
+                  defaultOpened="variants"
+                />
+                <div
+                  hidden={!isLoadingRoutePatterns}
+                  className="position-absolute inset-0 bg-light opacity-75 d-flex justify-content-center align-items-center"
+                >
+                  <Spinner />
+                </div>
+              </div>
             ) : (
               <p className="fst-italic">
                 Select a route in order to choose a direction.
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -2,6 +2,7 @@ import React, {
   ComponentProps,
   ComponentPropsWithoutRef,
   PropsWithChildren,
+  useContext,
   useEffect,
   useState,
 } from "react"
@@ -14,7 +15,10 @@ import { OriginalRoute } from "../../models/detour"
 import { joinClasses } from "../../helpers/dom"
 import { AsProp } from "react-bootstrap/esm/helpers"
 import { DetourFinishedPanel } from "./detourFinishedPanel"
+import { DetourRouteSelectionPanel } from "./detourRouteSelectionPanel"
 import { Route, RoutePattern } from "../../schedule"
+import RoutesContext from "../../contexts/routesContext"
+import inTestGroup, { TestGroups } from "../../userInTestGroup"
 
 const displayFieldsFromRouteAndPattern = (
   route: Route,
@@ -51,6 +55,7 @@ export const DiversionPage = ({
 }: DiversionPageProps) => {
   const {
     snapshot,
+    send,
 
     addConnectionPoint,
     addWaypoint,
@@ -85,7 +90,11 @@ export const DiversionPage = ({
     ? nearestIntersectionDirection.concat(directions)
     : undefined
 
-  const { route, routePattern } = snapshot.context
+  const { route, routePattern, routePatterns } = snapshot.context
+  const routePatternsById = Object.fromEntries(
+    routePatterns?.map((rp) => [rp.id, rp]) ?? []
+  )
+
   const {
     routeName = undefined,
     routeDirection = undefined,
@@ -123,6 +132,8 @@ export const DiversionPage = ({
     connectionPoints?.end?.name,
   ])
 
+  const routes = useContext(RoutesContext)
+
   return (
     <>
       <article className="l-diversion-page h-100 border-box inherit-box">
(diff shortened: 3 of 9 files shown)


assets/css/bootstrap/_map_overrides.scss
```diff
@@ -11,3 +11,14 @@ $theme-colors: (
   "dark": new_tokens.$gray-900,
   "missed-stop": semantic.$missed-stop,
 );
+
+$utilities: (
+  // Shorthand for specifying utility classes `top-0 start-0 bottom-0 end-0`
```
joshlarson, line 16: I'm curious about this comment - are `top-0`, etc examples of possible utility classes?

joshlarson: I ask because unless I'm missing something special about the `$utilities` syntax, we're only actually short-handing `inset`?

assets/src/models/createDetourMachine.ts
```diff
@@ -1,19 +1,68 @@
...
+      | {
+          // Caller has target route
+          route: Route
+          routePattern: undefined
+        }
+      | {
+          // Caller has no prior selection
+          route: undefined
+          routePattern: undefined
+        }
```
joshlarson, line 29:
```suggestion
```

I don't think this is a use case that we currently want to support. The two entrypoints are that we already have a route pattern (from the search map), or that we have a route but no route pattern ( from the route ladders).

assets/src/models/createDetourMachine.ts
```diff
@@ -24,10 +73,120 @@ export const createDetourMachine = setup({
...
+              guard: ({ context }) =>
+                context.routePattern !== undefined &&
+                context.route !== undefined,
+              target: "Editing",
+            },
+            { target: "Pick Route Pattern" },
+          ],
+        },
+        "Pick Route Pattern": {
+          initial: "Pick Route ID",
```
joshlarson, line 90:
```suggestion
          initial: "Load Route Patterns",
```

Related to https://github.com/mbta/skate/pull/2652/files#r1630163656, I think we'll land in this state when the input has a route, but no route pattern, and when that happens, we want to be loading the route patterns.

(I had to make that change [here](https://github.com/mbta/skate/pull/2647/files#diff-2697d2574fc8e2dec397eba594a8b6fa60a8f51bf3ea7fa188e87d697a87bb50R90) in order to get the detour modal to load properly, but I think it makes more sense to make that change in this PR.

assets/src/models/createDetourMachine.ts
firestack: This is not something we _currently_ have implemented, but it would not be difficult to update the `<DummyDetourPage>` to use this behavior.

I think we _do_ want this behavior, as per this [figma slide: Detour Opened from Detours Page](https://www.figma.com/design/XGmmwVROXdOvnWtJeUFC50/Skate-%7C-Unplanned-Detours-%7C-Hand-Off?node-id=2855-6968&t=XF9NKdbU2WB3iojO-1), but you could argue it shouldn't be implemented yet because that's not what we're doing yet. 

IMO, I'd leave it so that it's clear in the future what the machine _can_ handle, but we could drop it.

assets/src/models/createDetourMachine.ts
joshlarson: UPDATE: ☝️ was based on an old version of this branch, and is no longer true. 

In fact, making my change would break this because the `Load Route Patterns` state doesn't exist anymore!

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -84,14 +87,22 @@ export const DetourRouteSelectionPanel = ({
...
+                    selectedRouteInfo
+                  )}
+                  selectRoutePattern={onSelectRoutePattern}
+                  defaultOpened="variants"
+                />
+                <div
+                  hidden={!isLoadingRoutePatterns}
+                  className="position-absolute inset-0 bg-light opacity-75 d-flex justify-content-center align-items-center"
+                >
+                  <Spinner />
```
joshlarson, line 103: Not exactly a nit, but a UX nit. This works GREAT if we're transitioning from a state with route patterns to another state with route patterns,

https://github.com/mbta/skate/assets/912020/107a02b5-5383-4f1f-8e95-ed35f22bbc73

...but is a much less smooth UX when you're starting with no route patterns

https://github.com/mbta/skate/assets/912020/958155c5-8d5f-4d36-8a0d-822b05c2d905

This can be a follow-up ticket, but I just wanted to flag it.

assets/src/models/createDetourMachine.ts
```diff
@@ -24,10 +73,120 @@ export const createDetourMachine = setup({
...
+              }),
+            },
+            /** This seems like it's needed, since it's possible to select
+             * ["Select a route"](https://github.com/mbta/skate/blame/e14d100c1db7df8bbf3c24ca4f8dbbe100d24d1e/assets/src/components/detours/detourRouteSelectionPanel.tsx#L68)
+             * But the `select` is "controlled"(in react terms),
+             * then we need a way to represent that state?
+             * Otherwise there's no way to see the "select a route" help message
+             *
+             * There is not yet a test for this because this is not a project requirement (yet. tbd).
+             * This also depends on the combobox implementation decisions.
```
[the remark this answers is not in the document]
firestack: > The current `FormSelect` implementation has an entry with a `null` route ID also (I think), which should also trigger this event.

Yes, and just for clarity, I've explicitly handled that state instead of ignoring it https://github.com/mbta/skate/pull/2652/files#diff-a5c2ebb8d72dc32209d3598dadecd6793515e43155078116e8176cc292c32a0aR162-R173


---

I'll go ahead and edit this comment to remove the project requirement bit and probably the test bit?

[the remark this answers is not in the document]
firestack: I'm deciding to remove the comment altogether, but if you disagree, lets chat about what to put instead.

assets/src/components/detours/diversionPage.tsx
```diff
@@ -131,7 +142,46 @@ export const DiversionPage = ({
         </header>
 
         <div className="l-diversion-page__panel bg-light">
-          {snapshot.matches({ "Detour Drawing": "Editing" }) ? (
+          {snapshot.matches({ "Detour Drawing": "Pick Route Pattern" }) ? (
```
joshlarson, line 145: Non-blocking: Would it make sense to lift the `Pick Route Pattern` and `Share Detour` machines to top level? I would expect the top-level states to basically line up with the panels on the left, and it surprised me a bit to see `Detour Drawing` as part of the condition that governs the panel that shows up _before_ you start drawing.

joshlarson: Non-blocking for a few reasons, one of which is that I don't know enough about state machines / xstate to know whether this is even a good idea.

But also because I think it would be good for me or Hannah to try our hands at doing a state machine refactor like this.

And most importantly because I don't want to block merging of this (and the work that depends on this) on another style change.

firestack: > But also because I think it would be good for me or Hannah to try our hands at doing a state machine refactor like this.

I agree with this! Maybe open a ticket and put it on the [Skate Tech Cleanup board](https://app.asana.com/0/1152340551558956/1204993290118919)?



> Non-blocking: Would it make sense to lift the `Pick Route Pattern` and `Share Detour` machines to top level? I would expect the top-level states to basically line up with the panels on the left, and it surprised me a bit to see `Detour Drawing` as part of the condition that governs the panel that shows up _before_ you start drawing.

I'm not really sure on this, I initially nested them in case we needed say, parallel states running at all times, or other modes we haven't thought of. Since there's no real easy LSP way of refactoring the `matches` calls, I wouldn't opt to do this _myself_ but I don't mind entertaining the idea and trying it if one of you would like to make that refactor.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1222,7 +1233,7 @@ describe("DiversionPage", () => {
      * components, and we want to make sure that we don't accidentally
      * render it when we didn't intend to.
      */
-    test("the button doesn't render yet in context", async () => {
+    test("when not in test group, is not present", async () => {
```
joshlarson, line 1236: I think the comment above this can go away now.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1232,5 +1243,311 @@ describe("DiversionPage", () => {
...
+          screen.getByRole("radio", {
+            name: RPCRadioButtonName(selectedRoutePattern),
+          })
+        ).toBeChecked()
+      })
+    })
   })
 })
+
+function RPCRadioButtonName(rp2: RoutePattern) {
```
joshlarson, line 1549: Nit: Should this be `const rpcRadioButtonName = (rp2: RoutePattern) => {...}` instead (and probably live at the top of the file 😩)?

Other nit: The variable name should perhaps just be `rp`?

---

Also...

I also would sooner see the tests be a bit more explicit about their data, so something like

```typescript
rp = routePatternFactory.build({
  name: "Harvard via Allston",
  description: "from Nubian Station",
})

/* ... */

screen.getByRole("radio", {
  name: "Harvard via Allston from Nubian Station",
})
```

That way the test is super explicit about what it's testing (and it's clearer what type of thing we expect `name` and `description` to be). As it currently stands, in order to figure out what the assertion actually is, we have to trace through `RPCRadioButtonName`, and if `patternDisplayName` is broken then the assertion will assert the broken/wrong thing.

This can be a follow-up, and doesn't need to block merging this PR.

joshlarson (changes requested):
Arg - I didn't mean to make my comment a review comment. 

The first part (the two nits) I think are worth changing before merging. The other part is good to be a follow-up. 

I'm close to done, but not quite done, reviewing this!

assets/tests/components/detours/diversionPage.test.tsx
firestack: Does it make it any better that `patternDisplayName` is tested? https://github.com/mbta/skate/blob/main/assets/tests/components/mapPage/routePropertiesCard.test.tsx#L58-L104

assets/tests/components/detours/diversionPage.test.tsx
firestack: https://github.com/mbta/skate/compare/67e118cb595b85d64dd29f50cd987ca1a0776045..a0147d026bc04ce11131c677e0e07c2f18d9bc0b

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1232,5 +1238,371 @@ describe("DiversionPage", () => {
...
+      )
+
+      expect(
+        await screen.getByRole("heading", { name: "Choose route" })
+      ).toBeVisible()
+    })
+  })
+
+  describe("'Route Selection Panel'", () => {
+    beforeEach(() => {
```
joshlarson, line 1284: **Non-blocking**

In general, I don't love having a `beforeEach` in the middle of the test file, because it makes it a lot harder to tell what context each test is working with. For any of the tests within this `describe`, they've got an extra test group enabled, but it's hard to tell when looking at a given test which `describe` it's in.

A possible solution would be to enable this test group at the top for _all_ tests in this file, and then disable it only for the one test that asserts that the button isn't there (I think there's only one test that cares).

Non-blocking because this will be short-lived anyway, but food for thought.

assets/tests/components/detours/diversionPage.test.tsx
firestack: Part of it https://github.com/mbta/skate/compare/356f1eb5c81370545b67bc2e28304b7595914805..aeb951f1a4020482c2a6040753fe5bfc440d5de2

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -56,16 +64,25 @@ const DiversionPage = (
   )
 }
 
+function rpcRadioButtonName(rp: RoutePattern) {
+  const { name, description } = patternDisplayName(rp)
+  const radioName = name + " " + description
+  return radioName
```
joshlarson, line 70:
```suggestion
  return name + " " + description
```

Nit

assets/src/models/createDetourMachine.ts
joshlarson: Eh that's fair. Since we're pretty certain we'll want to support it, I'm okay leaving it in.

joshlarson (approved):
Looks great!

I raised a few things as follow-ups, but those (discussions and/or fixes) can happen later!

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1232,5 +1243,358 @@ describe("DiversionPage", () => {
...
+      await waitFor(() => {
+        expect(
+          screen.getByRole("radio", {
+            name: rpcRadioButtonName(selectedRoutePattern),
+          })
+        ).toBeChecked()
+      })
+    })
+
+    test("when route is changed from a route, to empty, can still select another route", async () => {
```
joshlarson, line 1502:
```suggestion
    test("when route is changed from a route to empty, can still select another route", async () => {
```

Tiny grammar nit.

joshlarson (approved):

PR #2657, opened 2024-06-10
feat(ts/components/routeLadder): add accessible name to route ladder toggle

firestack:
Tests are a bit annoying to comprehend when trying to get this button solely via class names, this instead makes it possible to do `screen.getByRole("button", { name: "Route Options" })`

---

While reviewing the [tests](https://github.com/mbta/skate/blob/e47aa0764f9a0a259897026543c9194eec20744a/assets/tests/components/ladderPage.test.tsx#L565-L571) in #2647, I found it a little annoying to translate the class names into the actual button that was being referenced. For both accessibility and avoiding class names in our tests, this makes it possible to find the button solely on it's textual meaning.

### DIFF
--- a/assets/tests/components/routeLadder.test.tsx
+++ b/assets/tests/components/routeLadder.test.tsx
@@ -7,7 +7,7 @@ import {
   afterEach,
 } from "@jest/globals"
 import React from "react"
-import { render } from "@testing-library/react"
+import { render, screen } from "@testing-library/react"
 import RouteLadder from "../../src/components/routeLadder"
 import { StateDispatchProvider } from "../../src/contexts/stateDispatchContext"
 import { LadderCrowdingToggles } from "../../src/models/ladderCrowdingToggle"
@@ -261,6 +261,8 @@ describe("routeLadder", () => {
         hasAlert={false}
       />
     )
+
+    expect(screen.getByRole("button", { name: "Route Options" })).toBeVisible()
 
     expect(tree).toMatchSnapshot()
   })
--- a/assets/tests/components/routeLadder.test.tsx
+++ b/assets/tests/components/routeLadder.test.tsx
@@ -262,7 +262,9 @@ describe("routeLadder", () => {
       />
     )
 
-    expect(screen.getByRole("button", { name: "Route Options" })).toBeVisible()
+    expect(
+      screen.getByRole("button", { name: "28 Route Options" })
+    ).toBeVisible()
 
     expect(tree).toMatchSnapshot()
   })
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -64,6 +64,11 @@
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
     width: 2.25rem;
+
+    &.dropdown-toggle::after {
+      // Force bootstrap carat to center even with hidden text
+      margin-left: 0;
+    }
   }
 
   .c-route-ladder__alert-icon {
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useId } from "react"
 import {
   AlertIcon,
   CrowdingIcon,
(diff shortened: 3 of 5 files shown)


joshlarson (changes requested):
Screenshot 2024-06-10 at 3 39 56 PM

On Firefox at least, this looks like it's nudged the arrow off-center - I suspect by adding that `<span />`. You can see this both on the route ladders page and in storybook.

firestack:
> Screenshot 2024-06-10 at 3 39 56 PM
> 
> On Firefox at least, this looks like it's nudged the arrow off-center - I suspect by adding that `<span />`. You can see this both on the route ladders page and in storybook.

Turns out, I think we're using "dropdown" wrong here, if you look at the [bootstrap dropdown docs](https://getbootstrap.com/docs/5.3/components/dropdowns/#split-button) you can see that they also did this same pattern, the difference is that by using `dropdown-split-toggle` it removes the `margin-left` that the dropdown arrow has by default. I'm not sure why we haven't encountered this issue yet though and why a dropdown without text seems to _look_ correct, even with the margin (IMO likely due to margin collapse).

So I _think_ the correct fix would be to add `dropdown-split-toggle` but that also seems strange when we're not using a split toggle (also it didn't work for some reason, seemingly due to specificity). So more to look into, but I found the root cause at least.

firestack:
> Screenshot 2024-06-10 at 3 39 56 PM
> 
> On Firefox at least, this looks like it's nudged the arrow off-center - I suspect by adding that `<span />`. You can see this both on the route ladders page and in storybook.

I couldn't get the bootstrap provided classes to have enough specificity to remove the margin added to the carat on non-empty toggles, so I [overrode `margin-left` explicitly](https://github.com/mbta/skate/compare/7c69255191a84de912d5c3372a2266dbbb220c41..c687ca482b25e0951aa8463fe19d407573f7b86a#diff-44e788e6ce2b1f6fa671a698b44daf3a71e6b6ef344a1ea95c615f4e2d143112R68-R70)

assets/tests/components/routeLadder.test.tsx
```diff
@@ -262,6 +262,10 @@ describe("routeLadder", () => {
       />
     )
 
+    expect(
+      screen.getByRole("button", { name: "28 Route Options" })
```
joshlarson, line 266: Curiosity: Where does the `28` here come from? I see how `Route Options` gets into the button's name, and I theoretically see why `28` is appropriate, but I can't figure out how they get stitched together.

firestack: Hmm, would this be a better assertion?
```suggestion
      screen.getByRole("button", { name: "${route.name} Route Options" })
```

joshlarson: Ooooh does `aria-labelledby=[routePillId, routeOptionsToggleId]` translate the `[routePillId, routeOptionsToggleId]` into their `name`'s `["28", "Route Options"]` and then concatenate?

joshlarson (approved):
🔽

assets/tests/components/routeLadder.test.tsx
joshlarson: > Hmm, would this be a better assertion?

Heh no I like the assertion as it is - I just didn't understand the code mechanism that made the test pass 😅

firestack: > Ooooh does `aria-labelledby=[routePillId, routeOptionsToggleId]` translate the `[routePillId, routeOptionsToggleId]` into their `name`'s `["28", "Route Options"]` and then concatenate?

That's essentially correct AFAIK, sorry about the comment before, I was misremembering things

firestack: > Ooooh does `aria-labelledby=[routePillId, routeOptionsToggleId]` translate the `[routePillId, routeOptionsToggleId]` into their `name`'s `["28", "Route Options"]` and then concatenate?

Just to be clear, we create a [space-separated id reference list](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby), and then screen readers and accessibility apis are responsible for doing the concatenation, this doesn't introduce any clones of the data from the DOM, and instead acts like pointers

PR #2658, opened 2024-06-10
fix(ts/models/createDetourMachine): clear detour when selecting route pattern

firestack:
A fix to #2652 that #2607 enables.

Split into another PR because discussing the test may block #2607 

---

Depends On:
- #2607

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -219,6 +219,7 @@ export const createDetourMachine = setup({
           on: {
             "detour.route-pattern.open": {
               target: "Pick Route Pattern",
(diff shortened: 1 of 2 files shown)


assets/src/models/createDetourMachine.ts
```diff
@@ -219,6 +219,7 @@ export const createDetourMachine = setup({
           on: {
             "detour.route-pattern.open": {
               target: "Pick Route Pattern",
+              actions: "detour.clear",
```
joshlarson, line 222: Delighted that the fix for this was a single line 😂

joshlarson (approved):

PR #2659, opened 2024-06-12
feat: Made searchForm into generic combobox with two variations (for search & select)

hannahpurcell:
There is 1 failing test, but I can't test it without the vpn! Will do tomorrow at the office

### DIFF
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -382,7 +382,7 @@ export const Combobox = ({
             />
           ) : (
             <GrouplessAutocompleteFromArray
-              controlName=""
+              controlName="Search Suggestions"
               fallbackOption={autocompleteOption(
                 "No matching routes",
                 undefined
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -89,13 +89,13 @@ type SearchComboboxProps = ComboboxProps & {
   comboboxType: "map_search"
   dispatch: Dispatch
   query: SearchQuery
-  options: null
+  options?: undefined
 }
 
 type SelectComboboxProps = ComboboxProps & {
   comboboxType: "select"
-  dispatch: null
-  query: null
+  dispatch?: undefined
+  query?: undefined
   options: AutocompleteOptionData[]
 }
 
@@ -437,7 +437,6 @@ const SearchFormFromStateDispatchContext = ({
       comboboxType="map_search"
       dispatch={dispatch}
       query={query}
-      options={null}
     />
   )
 }
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -229,7 +229,7 @@ export const MapSearchCombobox = ({
 }
 
 /**
- * Form which exposes all configurable state and callbacks via {@link ComboboxProps props}.
+ * Search form which exposes all configurable state and callbacks via {@link ComboboxProps props}.
  */
 export const Combobox = ({
   inputText,
@@ -358,13 +358,9 @@ export const Combobox = ({
           {comboboxType == "select" && (
             <button
               // TODO: rename this class (follow-up PR)
-              // TODO: for some reason, this button is still submitting???
+              type="button"
               className="c-search-form__submit"
-              onClick={() =>
-                setAutocompleteEnabled(
-                  (autocompleteEnabled) => !autocompleteEnabled
-                )
-              }
+              onClick={() => setAutocompleteEnabled(true)}
             >
               <ChevronDown />
             </button>
--- a/assets/src/components/searchForm.tsx
+++ b/assets/src/components/searchForm.tsx
@@ -1,4 +1,5 @@
 import React, {
+  MutableRefObject,
   SyntheticEvent,
   useContext,
   useId,
@@ -11,6 +12,7 @@ import { SearchIcon } from "../helpers/icon"
 import {
   SearchProperties,
   SearchPropertyQuery,
+  SearchQuery,
   isValidSearchText,
   searchPropertyDisplayConfig,
 } from "../models/searchQuery"
@@ -24,14 +26,18 @@ import {
 
 import { CircleXIcon } from "./circleXIcon"
 import {
+  AutocompleteOptionData,
   GroupedAutocompleteControls,
-  GroupedAutocompleteFromSearchTextEventProps,
   GroupedAutocompleteFromSearchTextResults,
+  GrouplessAutocompleteFromArray,
   autocompleteOption,
 } from "./groupedAutocomplete"
 import { SocketContext } from "../contexts/socketContext"
 import useSocket from "../hooks/useSocket"
 import { FilterAccordion } from "./filterAccordion"
+import { Ghost, Vehicle } from "../realtime"
+import { ChevronDown } from "../helpers/bsIcons"
+import { Dispatch } from "../state"
 
 // #region Search Filters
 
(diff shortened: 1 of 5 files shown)


assets/src/components/searchForm.tsx
```diff
@@ -146,25 +160,93 @@ const Filters = ({
...
+    )
+  }
+
+  const onSelectedLocationText = (text: string) => {
+    dispatch(setSearchText(text))
+    dispatch(submitSearch())
+  }
+
+  return (
+    <SocketContext.Provider value={useSocket()}>
```
firestack, line 204: question: what's going on here? initially, this seems unnecessary?

firestack (commented):
I have a few thoughts on this but haven't been able to read through and grok all the code yet, sorry!

I'm running out of time today, so I'll try and look at this later tonight or tomorrow morning at the very latest

assets/tests/components/groupedAutocomplete.test.tsx
```diff
@@ -893,4 +894,107 @@ describe("<GroupedAutocompleteFromSearchTextResults/>", () => {
...
+        onSelectedLocationText={onSelectLocationText}
+      />
+    )
+
+    await userEvent.click(option(locationSuggestion.text).get())
+
+    expect(onSelectLocationText).toHaveBeenCalledWith(locationSuggestion.text)
+
+    expect(listbox().get()).not.toBeVisible()
+  })
```
firestack, line 999: note: I believe this test is failing because this test is testing something that's not the responsibility of the autocomplete component itself. The autocomplete is not responsible for handling it's visibility, which is why this test lived in `searchForm` and not in `groupedAutocomplete`. So I believe tests will pass if you do this/leave this test on the search form instead.

assets/src/components/searchForm.tsx
```diff
@@ -257,64 +343,75 @@ export const SearchForm = ({
...
+              // TODO(design): add error states instead of using `disabled`
+              disabled={!isValidSearchText(inputText)}
+            >
+              <SearchIcon />
+            </button>
+          )}
+          {comboboxType == "select" && (
+            <button
+              // TODO: rename this class (follow-up PR)
+              // TODO: for some reason, this button is still submitting???
```
firestack, line 361: > // TODO: for some reason, this button is still submitting???

you likely need `type="button"`

assets/src/components/searchForm.tsx
firestack: omg I forgot about this, my bad, I didn't see that this existed originally in the place it was extracted from. When I built this originally, I hated having to add this hack to make the sockets work 😭 Which I guess is why I questioned why it was here 😅

assets/tests/components/groupedAutocomplete.test.tsx
hannahpurcell: Ahhhhaaaaaa. I moved the dropdown openness check to the searchForm, and kept the `onSelectLocationText` check in groupedAutocomplete

assets/src/components/searchForm.tsx
hannahpurcell: I didn't expect this! Nailed it, though.

assets/src/components/searchForm.tsx
```diff
@@ -257,64 +343,71 @@ export const SearchForm = ({
...
+            >
+              <SearchIcon />
+            </button>
+          )}
+          {comboboxType == "select" && (
+            <button
+              // TODO: rename this class (follow-up PR)
+              type="button"
+              className="c-search-form__submit"
+              onClick={() => setAutocompleteEnabled(true)}
```
firestack, line 363: question: what happens here if the user has the option typed out exactly, and then clicks this button or tabs onto it to submit it?

Maybe you can use `onClick={onSubmit}`?

assets/src/components/searchForm.tsx
```diff
@@ -146,25 +160,93 @@ const Filters = ({
...
-}: SearchFormProps) => {
+  dispatch,
+  query,
+  options,
+}: SearchComboboxProps | SelectComboboxProps) => {
   const formSearchInput = useRef<HTMLInputElement | null>(null)
   const [autocompleteEnabled, setAutocompleteEnabled] = useState(true)
 
   const onSubmit = (e: SyntheticEvent) => {
+    e.preventDefault()
```
firestack, line 249: question: do any tests fail if this is removed, or is any strange behavior introduced?

I'm curious because I wonder if this was something added when initially working on the TODO from this resolved comment https://github.com/mbta/skate/pull/2659/files/671ac7f7775f73b7f349d4679412442b7d279b9a#r1638262192

assets/src/components/searchForm.tsx
```diff
@@ -257,64 +343,71 @@ export const SearchForm = ({
...
+              formSearchInput={formSearchInput}
+              onSubmit={onSubmit}
+              query={query}
+              setAutocompleteEnabled={setAutocompleteEnabled}
+            />
+          ) : (
+            <GrouplessAutocompleteFromArray
+              controlName=""
+              fallbackOption={autocompleteOption(
+                "No matching routes",
```
firestack, line 387: nit(non-blocking): we chatted about this in a huddle, but it _may_ be beneficial if we rename this from `Combobox` to something about how this is a Map search OR route select _at this moment_. 

I'll post the combobox refactor ticket we talked about on this comment thread for posterity once it's created.

assets/tests/components/groupedAutocomplete.test.tsx
```diff
@@ -893,4 +894,105 @@ describe("<GroupedAutocompleteFromSearchTextResults/>", () => {
...
+          operator: [],
+          run: [],
+        }
+      }
+      return {
+        operator: [],
+        run: [],
+        vehicle: [],
+      }
+    }) as typeof useAutocompleteResults)
```
firestack, line 919: suggestion: since we're using `jest.mocked`, I think the `as` is unneeded?

```suggestion
    }))
```

assets/src/components/searchForm.tsx
hannahpurcell: Ahh! Yes, that was needed, as well as passing controllerRef to the GrouplessAutocomplete! Totally missed out on checking that piece of functionality.

assets/src/components/searchForm.tsx
hannahpurcell: Good memory, totally right

assets/src/components/searchForm.tsx
```diff
@@ -78,17 +83,22 @@ type SearchFormProps = SearchFormEventProps &
...
+  comboboxType: "map_search"
+  dispatch: Dispatch
+  query: SearchQuery
+  options: null
+}
+
+type SelectComboboxProps = ComboboxProps & {
+  comboboxType: "select"
+  dispatch: null
+  query: null
```
firestack, line 98: nit/suggestion(non-blocking): I _think_ if you make these optional, then they don't need to be specified when the `comboboxType` is `select`
Which would also mean if we're able to change the interface at some point, we wouldn't need to track these props down at that point

same for `options` on `SearchComboboxProps`

```suggestion
  dispatch?: undefined
  query?: undefined
```

firestack (approved):
Feeling good that all feedback is either addressed or being addressed, left one more nit if you want to do it, otherwise I don't think I need to rereview 👍🏻

assets/src/components/searchForm.tsx
hannahpurcell: Oh my goodness, there ya go! I tried a couple typing adjustments but not this. Thank you 🙇

PR #2632, opened 2024-05-30
feat: Basic validation on "select route" dropdown

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207402376744565

PR #2661, opened 2024-06-12
feat: ⚙️🚧 Clicking "start" with no route selected shows an error

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1207402376744565/f 

Superceeds #2632

### DIFF
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -66,24 +66,29 @@ export const DetourRouteSelectionPanel = ({
             <h2 className="c-diversion-panel__h2" id={selectId}>
               Choose route
             </h2>
-            <Form.Select
-              aria-labelledby={selectId}
-              value={selectedRouteInfo.selectedRoute?.id}
-              onChange={(changeEvent) => {
-                onSelectRoute?.(
-                  allRoutes.find(
-                    (route) => route.id === changeEvent.target.value
+            <Form>
+              <Form.Select
+                aria-labelledby={selectId}
+                value={selectedRouteInfo.selectedRoute?.id}
+                onChange={(changeEvent) => {
+                  onSelectRoute?.(
+                    allRoutes.find(
+                      (route) => route.id === changeEvent.target.value
+                    )
                   )
-                )
-              }}
-            >
-              <option>Select a route</option>
-              {allRoutes.map((route) => (
-                <option key={route.id} value={route.id}>
-                  {route.name}
-                </option>
-              ))}
-            </Form.Select>
+                }}
+              >
+                <option>Select a route</option>
+                {allRoutes.map((route) => (
+                  <option key={route.id} value={route.id}>
+                    {route.name}
+                  </option>
+                ))}
+              </Form.Select>
+              <Form.Control.Feedback type="invalid">
+                Select a route to continue.
+              </Form.Control.Feedback>
+            </Form>
           </section>
 
           <section className="pb-3">
--- a/assets/src/components/detours/detourRouteSelectionPanel.tsx
+++ b/assets/src/components/detours/detourRouteSelectionPanel.tsx
@@ -68,28 +68,30 @@ export const DetourRouteSelectionPanel = ({
               Choose route
             </h2>
             <Form>
-              <Form.Select
-                required
-                aria-labelledby={selectId}
-                value={selectedRouteInfo.selectedRoute?.id}
-                onChange={(changeEvent) => {
-                  onSelectRoute?.(
-                    allRoutes.find(
-                      (route) => route.id === changeEvent.target.value
+              <Form.Group>
+                <Form.Select
+                  required
+                  aria-labelledby={selectId}
+                  value={selectedRouteInfo.selectedRoute?.id}
+                  onChange={(changeEvent) => {
+                    onSelectRoute?.(
+                      allRoutes.find(
+                        (route) => route.id === changeEvent.target.value
+                      )
                     )
-                  )
-                }}
-              >
-                <option value="">Select a route</option>
-                {allRoutes.map((route) => (
-                  <option key={route.id} value={route.id}>
-                    {route.name}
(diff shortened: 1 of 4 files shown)


assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -63,27 +67,43 @@ export const DetourRouteSelectionPanel = ({
...
-              aria-labelledby={selectId}
-              value={selectedRouteInfo.selectedRoute?.id}
-              onChange={(changeEvent) => {
-                onSelectRoute?.(
-                  allRoutes.find(
-                    (route) => route.id === changeEvent.target.value
-                  )
-                )
-              }}
+            <Form
```
firestack, line 70: note: Since we're not actually doing "form" things, and just using it for it's css classes, I think we maybe should use the `as` prop to change this from a `form` to something else, I guess a `div`?

I'd keep it a form if we wrapped all of the contents in the form (RPC and the "submit" button), but those elements aren't really setup to be part of a form either IMO.
```suggestion
            <Form
              as="div"
```

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -52,8 +53,11 @@ export const DetourRouteSelectionPanel = ({
   onSelectRoute,
   onSelectRoutePattern,
   isLoadingRoutePatterns,
+  isRouteInvalid = false,
```
firestack, line 56: note: this seemed like the easiest route, but we _could_ pass in the `snapshot` and `send` values instead if we _wanted_.

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1668,5 +1668,32 @@ describe("DiversionPage", () => {
...
+        screen.getByRole("combobox", { name: "Choose route" })
+      ).not.toHaveAccessibleErrorMessage()
+
+      await userEvent.click(
+        screen.getByRole("button", { name: "Start drawing detour" })
+      )
+
+      expect(
+        screen.getByRole("combobox", { name: "Choose route" })
+      ).toHaveAccessibleErrorMessage("Select a route to continue.")
```
firestack, line 1696: note: I would've liked to use `toBeVisible` in these tests, as that's more accurate to what we expect, but because we don't have CSS in our tests, we don't have the ability to check that because bootstrap styles the visibility entirely with the presence of a class on the `<Form>` element, we could _manually_ add the `hidden` prop on the error container though, which would be a stopgap until we could get our tests to check CSS as well

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1668,5 +1668,32 @@ describe("DiversionPage", () => {
...
+              routePattern: undefined,
+              route: undefined,
+            }}
+          />
+        </RoutesProvider>
+      )
+
+      expect(
+        screen.getByRole("combobox", { name: "Choose route" })
+      ).not.toHaveAccessibleErrorMessage()
```
firestack, line 1688: note: this is here because I originally had a test which didn't assert this, and silently succeeded because it was always "visible". I then asserted `not.toBeVisible()` on the error content, which actually failed the test because we don't have access to the CSS in jsdom at this point. So this assertion makes sure it's known that there isn't an error _until_ the button is clicked.

Hmm, should this be a comment in the code?

assets/src/components/detours/detourRouteSelectionPanel.tsx
```diff
@@ -63,27 +67,43 @@ export const DetourRouteSelectionPanel = ({
...
+                    )
+                  }}
+                >
+                  <option value="">Select a route</option>
+                  {allRoutes.map((route) => (
+                    <option key={route.id} value={route.id}>
+                      {route.name}
+                    </option>
+                  ))}
+                </Form.Select>
```
hannahpurcell, line 101: Hmmm.. a lot of plumbing was set up here that will get wiped by the combobox work, though I didn't try to set up this error state in the combobox yet. Not sure what to do with that sequencing yet

assets/src/models/createDetourMachine.ts
```diff
@@ -188,6 +188,12 @@ export const createDetourMachine = setup({
                   },
                 },
                 Loading: {},
+                "No Route Error": {},
```
hannahpurcell, line 191: Can we use punctuation in state names? This almost reads like there _isn't_ a route error, but it's like "No-Route Error", or "Error: No Route" would also work

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: It wouldn't hurt to keep that comment around!

assets/src/components/detours/detourRouteSelectionPanel.tsx
hannahpurcell: After finishing review, I see that the plumbing wasn't high effort or anything, and all of it was a logical change given the combobox wasn't finished. I'll have to add more work to the combobox to ensure the same validation styling is applied

hannahpurcell (approved):
Great! None of my comments are blocking. Even if the error wording changes, I don't need to re-review. I trust your judgment there.

assets/src/models/createDetourMachine.ts
firestack: Great callout! Done! https://github.com/mbta/skate/compare/d35f2b2cd0934ed079f0bce75ff9b3d7d4b04ec2..ee0a7305856cca85e6b6ec0161ce34cf18b1a981

assets/tests/components/detours/diversionPage.test.tsx
firestack: Thanks! Done! https://github.com/mbta/skate/compare/d35f2b2cd0934ed079f0bce75ff9b3d7d4b04ec2..ee0a7305856cca85e6b6ec0161ce34cf18b1a981

PR #2662, opened 2024-06-14
fix: Detour map not interactive until route pattern selected

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207555257756259/f

### DIFF
--- a/assets/tests/components/detours/diversionPage.test.tsx
+++ b/assets/tests/components/detours/diversionPage.test.tsx
@@ -1713,5 +1713,34 @@ describe("DiversionPage", () => {
         screen.getByRole("combobox", { name: "Choose route" })
       ).toHaveAccessibleErrorMessage("Select a route to continue.")
     })
+
+    test("while on this panel, route is not interactive", async () => {
+      const route = routeFactory.build()
+      const routePattern = routePatternFactory.build({
+        routeId: route.id,
+      })
+
+      jest.mocked(fetchRoutePatterns).mockResolvedValue([routePattern])
+
+      const { container } = render(
+        <RoutesProvider routes={[route]}>
+          <DiversionPage originalRoute={{ route, routePattern }} />
+        </RoutesProvider>
+      )
+
+      await userEvent.click(
+        screen.getByRole("button", { name: "Change route or direction" })
+      )
+
+      // Non-Interactive route shape
+      expect(
+        container.querySelectorAll(".c-detour_map--original-route-shape-core")
+      ).toHaveLength(1)
+
+      // Interactive route shape
+      expect(
+        container.querySelectorAll(".c-detour_map--original-route-shape")
+      ).toHaveLength(0)
+    })
   })
 })
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -97,6 +97,11 @@ interface DetourMapProps {
    */
   center?: LatLngLiteral
   zoom?: number
+
+  /*
+   * Map is in editing state, and is therefore interactive
+   */
+  editing: boolean
 }
 
 export const DetourMap = ({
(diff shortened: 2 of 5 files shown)


firestack:
Looks good here! One question, do you think it'd be worth trying to add tests for this? I didn't see any similar tests for the share panel, so I'm not sure we have anything existing to build off of?

joshlarson (dismissed):
LGTM!

joshlarson:
@hannahpurcell I think you could assert on the lack of a `.c-detour_map--original-route-shape` selector (because that's the interactive bit - the core, non-interactive one is `.c-detour_map--original-route-shape-core`) after the `Change route or direction` button has been pressed.

hannahpurcell:
Kayla's right in that we really didn't have enough existing scaffolding to easily add a test for this. I've spent an hour on it, tried making a state machine snapshot factory that was less than successful, and can only think to mock the state machine, which we should certainly do, but is also certainly overkill for the small change happening here. Thoughts?

firestack:
> Kayla's right in that we really didn't have enough existing scaffolding to easily add a test for this. I've spent an hour on it, tried making a state machine snapshot factory that was less than successful, and can only think to mock the state machine, which we should certainly do, but is also certainly overkill for the small change happening here. Thoughts?

You may be able to take the `"when a route pattern is already selected, opens to selected pattern"` test and turn it into a test for this

```tsx
    test("<rename this>", async () => {
      const route = routeFactory.build()
      const routePattern = routePatternFactory.build({
        routeId: route.id,
      })

      jest.mocked(fetchRoutePatterns).mockResolvedValue([routePattern])

      const { container } = render(
        <RoutesProvider routes={[route]}>
          <DiversionPage
            originalRoute={{ route, routePattern }}
          />
        </RoutesProvider>
      )

      await userEvent.click(
        screen.getByRole("button", { name: "Change route or direction" })
      )

      expect(
        container.querySelectorAll(".c-detour_map--original-route-shape-core")
      ).toHaveLength(<N>)

      expect(
      	container.querySelectorAll(".c-detour_map--original-route-shape")
      ).toHaveLength(<M>)
    })
```

The query selectors are a bit hard to read though, so it may be worth doing the abstraction where we can make the selectors within the `expect` a bit more readable to understand _which_ one is _actually_ not interactive. I think we have a ticket for this?

At the very least, should probably add comments for which one is interactive and which one isn't

hannahpurcell:
> You may be able to take the `"when a route pattern is already selected, opens to selected pattern"` test and turn it into a test for this

I'm OOO today but think this shouldn't be blocked from going out today by a test (given Kathleen's prioritization of the issue). If making the above change would be super smooth (taking <30min of effort / review), would you be interested / able to do that? Otherwise, do you feel ok with shipping, and I add the test on Thursday?

firestack:
> > You may be able to take the `"when a route pattern is already selected, opens to selected pattern"` test and turn it into a test for this
> 
> I'm OOO today but think this shouldn't be blocked from going out today by a test (given Kathleen's prioritization of the issue). If making the above change would be super smooth (taking <30min of effort / review), would you be interested / able to do that? Otherwise, do you feel ok with shipping, and I add the test on Thursday?

Yeah sure I can add the test and then assign Josh to review!

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1713,5 +1713,34 @@ describe("DiversionPage", () => {
         screen.getByRole("combobox", { name: "Choose route" })
       ).toHaveAccessibleErrorMessage("Select a route to continue.")
     })
+
+    test("while on this panel, route is not interactive", async () => {
```
joshlarson, line 1717: Awesome sauce!

This test is a perfect example of a test that will be a lot easier to write and work with once we build some better scaffolding for detour-related tests.

Thanks for adding it!

joshlarson (approved):

PR #2663, opened 2024-06-17
 test(ts/components/diversionPage): add helpers for route interativity 

firestack:
Followup for https://github.com/mbta/skate/pull/2662#issuecomment-2173266737

> The query selectors are a bit hard to read though, so it may be worth doing the abstraction where we can make the selectors within the expect a bit more readable to understand _[which]_ one is _actually_ not interactive. I think we have a ticket for this?


firestack:
Oh I can't mark this as auto-merge because it's merging into #2663 🤣

joshlarson (approved):

PR #2665, opened 2024-06-20
feat: New TripModification module

joshlarson:
Still TODO before merging this PR:
- [x] Typespecs
- [x] ~Jason~
- [x] `last_modified_time`
- [x] Rename `for` to `new`
- [x] Move `TripModification` module to the right place
- [x] `service_dates`
- [x] Fix `StopSelector` type
- [x] Fix `last_modified_time` type

Follow-up:
- Stop sequences instead of Stop ID's
- Start times (unless that would be easy)

Asana Ticket: https://app.asana.com/0/1205718271156548/1207383192300466/f


lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+          }
+
+    defstruct [:start_stop_selector, :end_stop_selector]
+  end
+
+  @type t :: %__MODULE__{
+          selected_trips: [SelectedTrip.t()],
+          modifications: [Modification.t()]
+        }
+  defstruct [:selected_trips, :modifications]
```
joshlarson, line 57: @mbta/skate-developers this will eventually need to be JSON-encoded, but I'm not familiar enough with `Jason` to know whether that's necessary right now.

...or ever. I found a few other examples of things that get JSON-encoded, where it looks like Jason isn't involved at all, and now I'm wondering if we'll even need to use it here, or if the data structure being right is enough.

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+        route_pattern: %RoutePattern{representative_trip_id: trip_id},
+        missed_stops: missed_stops
+      }) do
+    %__MODULE__{
+      selected_trips: [
+        %SelectedTrip{trip_ids: [trip_id]}
+      ],
+      modifications: [
+        %Modification{
+          start_stop_selector: missed_stops |> List.first() |> Map.get(:id),
```
joshlarson, line 100: In the medium-to-long term, we'll need to use stop sequences here instead of stop ID's, but that'll involve some changes to the missed stops algorithm.

I'm thinking we should probably do that as a near-term follow-up, because otherwise this will sneak back up on us. But I was thinking that this PR didn't need to be blocked on that! 😅

firestack: > but that'll involve some changes to the missed stops algorithm.
Can you elaborate? My understanding was that it's not missed stops that needs to change, but the underlying `Stop` struct needs to have `stop_sequence`

joshlarson: Oh, that would also work.

I had been thinking that the missed stops algorithm should include the index into the array, so instead of returning:
```elixir
[
  %Stop{...},
  %Stop{...}
]
```

It would return
```elixir
[
  {4, %Stop{...}},
  {5, %Stop{...}}
]
```

or something like that.

But I like the idea of putting `stop_sequence` directly into the data structure better - I just didn't realize that was an option!

What do you think... follow-up, or do it as part of this PR?

firestack: > I had been thinking that the missed stops algorithm should include the index into the array

Just noting this: This also won't work AFAIK, because we need to source `stop_sequence` from GTFS data
> `stop_times.txt`
> `stop_sequence` 	`Non-negative integer` 	`Required`
>
> Order of stops, location groups, or GeoJSON locations for a particular trip. The values must increase along the trip but do not need to be consecutive.
> _Example: The first location on the trip could have a stop_sequence=1, the second location on the trip could have a stop_sequence=23, the third location could have a stop_sequence=40, and so on._
>
> -- https://gtfs.org/schedule/reference/#stop_timestxt

firestack: > But I like the idea of putting `stop_sequence` directly into the data structure better - I just didn't realize that was an option!
> 
> What do you think... follow-up, or do it as part of this PR?

Definitely follow-up, because this involves parsing the `stop_sequence` data from GTFS and storing it, then changing the way we source stop data to include this when we get it from a trip. Then we need to add the field to `Stop` and make sure it doesn't affect current `stops.txt` parsing code, but is filled when we get stops associated with a trip. (that's just off the top of my head, I may have repeated myself or not said everything, but I figure we'll get there okay)

firestack: (Just a little side note, because I think it's neat but I don't think I mentioned or wrote down this, but the reason that I originally "architechted" MissedStops to sort objects which implement `Util.Location.From` instead of doing something "easier" or "simplier" at first, is because of this situation with `stop_sequence`!)

lib/skate/detours/trip_modification.ex
firestack: I _think_ this'll be okay!

lib/skate/detours/trip_modification.ex
firestack: > What do you think... follow-up, or do it as part of this PR?

A random thought though, to be in compliance with GTFS-TM spec, maybe we should be asserting that the associated trip does not contain duplicate `stop_id`'s?

That'd be great if it could be in this PR, but the original spec of this module is to be a "simple transformer" and ideally doesn't need to access any outside data. So either the caller would need to provide a Route Pattern and it's Shape + Stops for this module to assert on, or the caller would need to assert, which then doesn't centralize the assertion.

Anyway, not blocking, "just food for thought".

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+
+    defstruct [:start_stop_selector, :end_stop_selector]
+  end
+
+  @type t :: %__MODULE__{
+          selected_trips: [SelectedTrip.t()],
+          modifications: [Modification.t()]
+        }
+  defstruct [:selected_trips, :modifications]
+
```
firestack, line 58: question: is there a reason why the required field [`service_dates`](https://gtfs.org/realtime/reference/#message-tripmodifications) is not in this struct currently?

I'm okay with transit data modifying it down the line, but it seems like we should be providing _something_ here since it's required.

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+    More info: https://gtfs.org/realtime/reference/#message-selectedtrips
+    """
+
+    @type t :: %__MODULE__{
+            trip_ids: [String.t()]
+          }
+
+    defstruct [:trip_ids]
+  end
```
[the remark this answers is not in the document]
firestack: I'm open to doing this as a follow-up if you'd _prefer_.

firestack:
> @joshlarson have you thought anything about emitting [`Shape` Messages](https://gtfs.org/realtime/reference/#message-shape) as well?

At the very least, do we have the emitting shapes work tracked somewhere?

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+    @moduledoc """
+    Struct representing the `modifications` part of `Skate.Detours.TripModification`.
+
+    More info: https://gtfs.org/realtime/reference/#message-modification
+    """
+
+    @type t :: %__MODULE__{
+            start_stop_selector: String.t(),
+            end_stop_selector: String.t()
+          }
```
firestack, line 48: issue: It seems like this is the wrong type?

From what I see in the [`StopSelector` message docs](https://gtfs.org/realtime/reference/#message-stopselector), `StopSelector` is a object and not a string.

Is there a reason to prefer a string here?

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+    defstruct [:start_stop_selector, :end_stop_selector]
+  end
+
+  @type t :: %__MODULE__{
+          selected_trips: [SelectedTrip.t()],
+          modifications: [Modification.t()]
+        }
+  defstruct [:selected_trips, :modifications]
+
+  @spec for(input :: Input.t()) :: __MODULE__.t()
```
firestack, line 59: nit(non-blocking): I think `new` would be a more appropriate name?

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,106 @@
...
+    More info: https://gtfs.org/realtime/reference/#message-modification
+    """
+
+    @type t :: %__MODULE__{
+            start_stop_selector: String.t(),
+            end_stop_selector: String.t()
+          }
+
+    defstruct [:start_stop_selector, :end_stop_selector]
+  end
```
firestack, line 51: suggestion(non-blocking): if it's easy, I'd also love to get `last_modified_time` on this as well.

firestack (changes requested):
Left some comments!

lib/skate/detours/trip_modification.ex
joshlarson: I like that idea!

I _think_ that would be an attribute of the route pattern, so if I can easily get that into the route pattern here, I'll add it to this PR.

That does remind me that `{:ok, result}` is the Elixir Way™, so it's worth switching to that!

joshlarson:
@firestack I agree with all of your comments, and I think at least some of them should be addressed before this work can be considered done. But also, getting this merged even in its current incomplete state will unblock [⚙️🤝GTFS-TM:Create and Send Trip Modification when User "Finishes" Detour](https://app.asana.com/0/1205385723132845/1207383192300475/f), so I'm wondering if it's worth it to merge this as an incomplete implementation of the feature and create follow-up tickets and/or subtasks for anything that moves this past midday tomorrow?

Also... the ticket for this work has no A/C's and is pointed as a 1, so I think maybe even the comically incomplete implementation is enough for now?

lib/skate/detours/trip_modification.ex
```diff
@@ -0,0 +1,108 @@
+defmodule Skate.Detours.TripModification do
```
firestack, line 1: I'm in the middle of writing a different comment about naming with relation to my ticket [⚙️🤝GTFS-TM:A module that takes modification events to send via MQTT](https://app.asana.com/0/1203014709808707/1207383192300471/f), but I realized while writing that that because this implements a GTFS spec, maybe this file should live at `Realtime.TripModification` since this is a GTFS-RT struct and not a Skate specific struct?

That also makes it worth thinking about if we should flatten all the "messages" into Modules within `Realtime`?

firestack: fwiw, I still think having a [`new/1`](https://github.com/mbta/skate/pull/2665/files#r1648067277) function is worth it on this module, to make it easy to construct these, but I think it should live in `Realtime`

lib/skate/detours/trip_modification.ex
firestack: Okay I know I said non-blocking but I think I'm gonna block on having a _conversation_ about this 😅

firestack (commented):
> @firestack I agree with all of your comments, and I think at least some of them should be addressed before this work can be considered done. But also, getting this merged even in its current incomplete state will unblock [⚙️🤝GTFS-TM:Create and Send Trip Modification when User "Finishes" Detour](https://app.asana.com/0/1205385723132845/1207383192300475/f), so I'm wondering if it's worth it to merge this as an incomplete implementation of the feature and create follow-up tickets and/or subtasks for anything that moves this past midday tomorrow?
> 
> Also... the ticket for this work has no A/C's and is pointed as a 1, so I think maybe even the comically incomplete implementation is enough for now?

I think I'm okay with this, I'd _like_ to see these two things addressed/talked about at least and then we can merge and fix the rest in followup 👍🏻 
- [Converse on renaming `for/1` to `new/1`](https://github.com/mbta/skate/pull/2665/files#r1648067277)
- [Move file into `Realtime`](https://github.com/mbta/skate/pull/2665/files/fcc012ce63d7ed1e4bb27e99e9a3cae08b10cf35#diff-aa1fce2b5b6d69e85a0b6d988ddca08c58df5da982c819fdb9602370d7b90768)

joshlarson:
@firestack Yeah I agree - both of those are simple changes, and the follow-up work would be blocked on those changes anyway! (Or would have conflicts)

Will address before re-requesting review.

lib/skate/detours/trip_modification.ex
joshlarson: [Done](https://github.com/mbta/skate/pull/2665/commits/28e0ef48eff87a249a4a6aa315cbba3abdc3ba39)

lib/skate/detours/trip_modification.ex
joshlarson: [Here we go](https://github.com/mbta/skate/pull/2665/commits/a35857cb1975ada9a670bf22956e8ef5890cb9db)

I thought about just using `DateTime.utc_now`, but decided that having the time as an input was a better choice (especially because the `last_modified_time` might not actually be the time we create the `TripModification` 😅)

lib/skate/detours/trip_modification.ex
joshlarson: I'm unresolving this because this thread included the `Realtime` recommendation as well, which I haven't done yet!

Stay tuned...

joshlarson: [Realtime move!](https://github.com/mbta/skate/pull/2665/commits/df4871b05a174ee14241834ece7752874e93512d)

(Kind of fitting that I kept you updated about this change in real time, huh!)

firestack (approved):
Assuming we get to the rest of the feedback in followup PR's, I'm good merging this 👍🏻

lib/skate/detours/trip_modification.ex
joshlarson: Oh - I want to fix this before merging.

lib/skate/detours/trip_modification.ex
joshlarson: @firestack what would put in as a simple default? Just like... today's date to start?

firestack: Today's date seems like a good default to me.

joshlarson: [Added `service_dates`](https://github.com/mbta/skate/pull/2665/commits/0f750f523c6719bfd87faa5ededb365630c9250e)

lib/skate/detours/trip_modification.ex
joshlarson: [Fixed](https://github.com/mbta/skate/pull/2665/commits/5de88c009b263655c61a1bad2dbd7fca1d01d2f0)

lib/skate/detours/trip_modification.ex
joshlarson: I made it an input 😇

lib/skate/detours/trip_modification.ex
joshlarson: Long-term fix: [⚙️GTFS-TM: Send stop sequence instead of stop ID as the stop selectors](https://app.asana.com/0/1206070633318963/1207634607164315/f)

Short-term fix (will do v. soon): [⚙️GTFS-TM: (Stopgap) return an error if a route-pattern has any duplicate stops](https://app.asana.com/0/1206070633318963/1207635746210198/f)

lib/skate/detours/trip_modification.ex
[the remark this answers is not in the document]
joshlarson: [⚙️GTFS-TM: Add shape_id (with "dummy" value) to TripModification](https://app.asana.com/0/1206070633318963/1207635746210190/f) - I'll pick this up immediately, but I want to get this PR out of the way!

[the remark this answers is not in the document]
joshlarson: Real long-term fix: [⚙️GTFS-TM: Publish the detour shape and attach it to the TripModification message](https://app.asana.com/0/1206070633318963/1207634607164317/f)

PR #2668, opened 2024-06-21
fix: Updated nav buttons to use bootstrap primary token 2024 $eggplant-500

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1207446071151352

I originally thought I'd want to update the nav buttons to bootstrap, but that seemed a big scope bloat when only the colors needed shifting for now. So instead the scope bloat that I accepted was also changing the drawer-tab on the map page to be the lighter purple, since that was a simple fix (and it was obviously mismatched to the updated navbar.

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -64,15 +64,6 @@
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
     width: 2.25rem;
-    // TODO All of this is bad and should be reverted when we fix the left nav button colors
-    &.btn-primary {
-      --bs-btn-bg: #451b83;
-      --bs-btn-border-color: #451b83;
-      --bs-btn-hover-bg: #260f54;
-      --bs-btn-hover-border-color: #260f54;
-      --bs-btn-active-bg: #260f54;
-      --bs-btn-active-border-color: #260f54;
-    }
   }
 
   .c-route-ladder__alert-icon {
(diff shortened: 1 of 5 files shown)


assets/css/_route_ladder.scss
```diff
@@ -64,15 +64,6 @@
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
     width: 2.25rem;
-    // TODO All of this is bad and should be reverted when we fix the left nav button colors
```
joshlarson, line 67: Yay!

assets/css/_ui_kit.scss
```diff
@@ -278,6 +279,27 @@ $color-tooltip-background: #333333;
...
 }
 
+@mixin button-primary-new {
+  color: $white;
+  padding: 8px;
+  background-color: new_tokens.$eggplant-500;
+  border-radius: 4px;
+
+  &:hover {
+    background-color: #5d3092;
```
joshlarson, line 289: I'm pretty sure this hex code is a value that bootstrap came up with based on `$eggplant-500`. I did briefly have this dream of just getting rid of this and just using the bootstrap logic, but that messed up the c-drawer-tab on the route ladders page.

I think there's a Real Solution™ hiding out somewhere, but I'm hoping this is an acceptable incremental change 😅

firestack (approved):

PR #2672, opened 2024-06-24
test(factories): Use sequences in GTFS Stop factory

joshlarson:


test/realtime/trip_modification_test.exs
```diff
@@ -18,13 +18,7 @@ defmodule Realtime.TripModificationTest do
     shape_with_stops =
       build(:shape_with_stops,
         id: "id-of-the-shape",
-        stops: [
```
joshlarson, line 21: I'm on the fence about whether I should punt these changes to a separate PR, but since these changes are _why_ we did the `sequence` thing in the first place, it felt at least somewhat reasonable to ship them together (and none of this is a large enough change that it's confusing).

firestack: I'm so glad to see this!

firestack (approved):
🎉

PR #2673, opened 2024-06-24
refactor: Rename "finish" variables and functions to "review"

joshlarson:
Asana Ticket: https://app.asana.com/0/1205526445275136/1207636402402295/f

Follow-up To:
- https://github.com/mbta/skate/pull/2671

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -77,7 +77,7 @@ export const DiversionPage = ({
     canUndo,
     undo,
     clear,
-    finishDetour,
+    reviewDetour,
     editDetour,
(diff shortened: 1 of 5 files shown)


firestack (approved):
I'm a little iffy on this, as I'd love to find a name that doesn't need to change with the copy, but while our functions match the copy, this is at least an improvement.

joshlarson:
> I'd love to find a name that doesn't need to change with the copy

Yeah I'm kind of torn, but since there's such a close mapping between states and copy right now (because in a sense, the whole point of the states is to determine what shows up on the screen), I wasn't sure what else to do.

I also really didn't like the term `Finish` there, since it has always kind of felt misleading.

PR #2675, opened 2024-06-28
refactor: Move fetchFinishedDetour into the state machine

joshlarson:
Asana Ticket: https://app.asana.com/0/0/1207659713660474/f

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -318,6 +318,8 @@ export const createDetourMachine = setup({
                     finishedDetour: ({ event }) => event.output,
                   }),
                 },
+
+                onError: {},
               },
               on: {
                 "detour.edit.undo": {
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -1,7 +1,8 @@
 import { setup, assign, fromPromise, ActorLogicFrom, InputFrom } from "xstate"
-import { ShapePoint } from "../schedule"
+import { RoutePatternId, ShapePoint } from "../schedule"
 import { Route, RouteId, RoutePattern } from "../schedule"
-import { fetchRoutePatterns } from "../api"
+import { fetchFinishedDetour, fetchRoutePatterns } from "../api"
+import { FinishedDetour } from "./detour"
 
 export const createDetourMachine = setup({
   types: {} as {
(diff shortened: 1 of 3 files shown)


assets/src/models/createDetourMachine.ts
```diff
@@ -282,6 +303,22 @@ export const createDetourMachine = setup({
...
+                  routePatternId: routePattern?.id,
+                  startPoint,
+                  endPoint,
+                }),
+
+                onDone: {
+                  actions: assign({
+                    finishedDetour: ({ event }) => event.output,
+                  }),
+                },
```
firestack, line 320: issue: btw, if you don't stub out the `onError` handler, the state machine will lockup if the API call fails in _some_ way. So similar to the `fetch-route-patterns` actor, you should add an empty handler.
https://github.com/mbta/skate/blob/jdl/refactor/finished-detour-into-state-machine/assets/src/models/createDetourMachine.ts#L180-L181

firestack (commented):
aside from the one issue, LGTM!

assets/src/models/createDetourMachine.ts
firestack: Hmm, actually I'm not to sure on this anymore, I had _thought_ that this was the case, but lets check this assumption while pairing this afternoon.

assets/src/models/createDetourMachine.ts
```diff
@@ -93,11 +111,13 @@ export const createDetourMachine = setup({
     }),
     "detour.remove-end-point": assign({
       endPoint: undefined,
+      finishedDetour: undefined,
```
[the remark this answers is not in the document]
firestack: Chatted about this, we'll see what things are like in the future, definitely an interesting topic!

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -908,6 +938,29 @@ describe("DiversionPage", () => {
     ).toBeVisible()
   })
 
+  test("When 'Review Detour' button is clicked, shows 'Share Detour Details' screen but there was an error", async () => {
```
firestack, line 941: I think we should make the test name a bit clearer.
```suggestion
  test("When API call errors and 'Review Detour' button is clicked, shows 'Share Detour Details' screen", async () => {
```

firestack (approved):
One more note about the test name but I'm going to approve and let you decide.

PR #2677, opened 2024-07-01
fix: Simplify `useReactDivIcon`

firestack:
This changes `useReactDivIcon` to create the portal element and the `divIcon` at the same time so that we don't need two renders for this component to complete initialization. This prevents some issues in tests where "`An update to ReactMarker inside a test was not wrapped in act(...).`" occurs.

### DIFF
--- a/assets/src/components/map/utilities/reactDivIcon.tsx
+++ b/assets/src/components/map/utilities/reactDivIcon.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useState } from "react"
+import { useMemo } from "react"
 
-import L, { DivIcon, DivIconOptions as LeafletDivIconOptions } from "leaflet"
+import L, { DivIconOptions as LeafletDivIconOptions } from "leaflet"
 
 // Prevent user from setting parameters we intend to provide
 export type DivIconOptions = Omit<LeafletDivIconOptions, "html">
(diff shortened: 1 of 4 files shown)


assets/src/components/map/utilities/reactDivIcon.tsx
```diff
@@ -23,26 +23,14 @@ const defaultOptions = {}
 export function useReactDivIcon(options?: DivIconOptions) {
   const opts = options || defaultOptions
   // Persistent element for react portal to use between `divIcon` recreations
-  const [iconContainer, setIconContainer] = useState<HTMLDivElement | null>(
-    null
+  const iconContainer = useMemo(() => createPortalElement(), [])
```
joshlarson, line 26: Would this work?
```suggestion
  const iconContainer = useMemo(createPortalElement, [])
```

I was also wondering if we could just get rid of the `useMemo()` call and make it
```ts
const iconContainer = createPortalElement()
```

but I'm thinking that it's important that `createPortalElement()` is only called once?

firestack: It's _very_ important that `createPortalElement()` is only called once because that's the element we hand to leaflet to manipulate, we could change it to `usePortalElement`, but it'd still be _some_ react state somewhere.

firestack: okay I'm not actually _certain_ that we couldn't do `const iconContainer = createPortalElement()` but I personally think we should keep it as is.

firestack: It's a good question though! Any invariants that need to be upheld by only calling this once _should_ be documented.

firestack: I get a few test failures when I outright remove it, not sure about if those could be fixed or not but I'm not going to spend more time on it.

PR #2678, opened 2024-07-01
refactor: Move fetchDetourDirections calls into state machine

joshlarson:
Asana Ticket: https://app.asana.com/0/1205526445275136/1207707249234239/f

Depends On:
- https://github.com/mbta/skate/pull/2675

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -312,6 +312,8 @@ export const createDetourMachine = setup({
                     detourShape: ({ event }) => event.output,
                   }),
                 },
+
+                onError: {},
               },
               on: {
                 "detour.edit.place-waypoint": {
@@ -394,6 +396,8 @@ export const createDetourMachine = setup({
                       },
                     }),
                   },
+
+                  onError: {},
                 },
               ],
 
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -1,8 +1,14 @@
 import { setup, assign, fromPromise, ActorLogicFrom, InputFrom } from "xstate"
 import { RoutePatternId, ShapePoint } from "../schedule"
 import { Route, RouteId, RoutePattern } from "../schedule"
-import { fetchFinishedDetour, fetchRoutePatterns } from "../api"
-import { FinishedDetour } from "./detour"
+import { Ok, Result, isOk } from "../util/result"
+import {
+  FetchDetourDirectionsError,
+  fetchDetourDirections,
+  fetchFinishedDetour,
+  fetchRoutePatterns,
+} from "../api"
+import { DetourShape, FinishedDetour } from "./detour"
 
 export const createDetourMachine = setup({
   types: {} as {
@@ -15,6 +21,12 @@ export const createDetourMachine = setup({
       waypoints: ShapePoint[]
       startPoint: ShapePoint | undefined
       endPoint: ShapePoint | undefined
+
+      detourDirections:
+        | Result<DetourShape, FetchDetourDirectionsError>
+        | undefined
+
+      showRegularRouteText: boolean
 
       finishedDetour: FinishedDetour | undefined | null
     }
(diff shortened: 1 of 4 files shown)


assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -544,6 +544,56 @@ describe("DiversionPage", () => {
...
+      fireEvent.click(screen.getByRole("button", { name: "Clear" }))
+    })
+    await waitFor(() => {
+      expect(
+        getByText(
+          "Click a point on the regular route to start drawing your detour. As you continue to select points on the map, turn-by-turn directions will appear in this panel."
+        )
+      ).toBeVisible()
+
+      expect(queryByText("Turn left on Main Street")).not.toBeInTheDocument()
```
firestack, line 593: thought(**non-blocking**): I wonder if we should start using [`waitForElementToBeRemoved`](https://testing-library.com/docs/dom-testing-library/api-async/#waitforelementtoberemoved) to check the removal of an element 🤔

assets/src/models/createDetourMachine.ts
```diff
@@ -297,30 +343,64 @@ export const createDetourMachine = setup({
...
+
+                        return map(detourShape, (shape) => {
+                          return {
+                            ...shape,
+                            directions: shape.directions?.concat({
+                              instruction: "Regular Route",
+                            }),
+                          }
+                        })
+                      },
```
firestack, line 396: nit/suggestion(**non-blocking**): Just wondering, do you prefer this syntax over something like this?
```suggestion
                      detourShape: ({ event: { output: detourShape } }) =>
                        map(detourShape, (shape) => ({
                          ...shape,
                          directions: shape.directions?.concat({
                            instruction: "Regular Route",
                          }),
                        })),
```

assets/src/models/createDetourMachine.ts
```diff
@@ -68,6 +76,23 @@ export const createDetourMachine = setup({
...
+    >(async ({ input: { points } }) => {
+      if (points) {
+        if (points.length < 2) {
+          return Ok({ coordinates: [], directions: undefined })
+        } else {
+          return fetchDetourDirections(points)
+        }
+      } else {
+        throw "Missing detour direction inputs"
+      }
```
firestack, line 93: nit/suggestion(**_non-blocking_**): thoughts? 😅 I kinda wanna do this to `fetch-route-patterns` now too(later).

I don't wanna blow up your PR, so I've formatted this and tested this so that you could use `commit-suggestion`, but don't feel like you have to.

```suggestion
      if (!points) {
        throw "Missing detour direction inputs"
      }
      if (points.length < 2) {
        return Ok({ coordinates: [], directions: undefined })
      }
      return fetchDetourDirections(points)
 
```

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -544,6 +544,56 @@ describe("DiversionPage", () => {
...
+    })
```
firestack, line 594: suggestion(non-blocking): what do you think about flattening the test a bit?

```suggestion
    jest.mocked(fetchDetourDirections).mockResolvedValue(
      Ok(
        detourShapeFactory.build({
          directions: [
            { instruction: "Turn left on Main Street" },
            { instruction: "Turn right on High Street" },
            { instruction: "Turn sharp right on Broadway" },
          ],
        })
      )
    )

    const { container } = render(<DiversionPage />)
    const { queryByText, findByText } = within(
      screen
        .getByRole("heading", { name: "Detour Directions" })
        .closest("section")!
    )

    act(() => {
      fireEvent.click(originalRouteShape.get(container))
    })
    act(() => {
      fireEvent.click(container.querySelector(".c-vehicle-map")!)
    })

    const directionText = await findByText("Turn left on Main Street")
    expect(directionText).toBeVisible()
    expect(
      queryByText(
        "Click a point on the regular route to start drawing your detour. As you continue to select points on the map, turn-by-turn directions will appear in this panel."
      )
    ).not.toBeInTheDocument()

    act(() => {
      fireEvent.click(screen.getByRole("button", { name: "Clear" }))
    })

    const startingHelpText = await findByText(
      "Click a point on the regular route to start drawing your detour. As you continue to select points on the map, turn-by-turn directions will appear in this panel."
    )
    expect(startingHelpText).toBeVisible()
    expect(directionText).not.toBeInTheDocument()

```

assets/tests/components/detours/diversionPage.test.tsx
firestack: Nevermind! I tried something and didn't actually use it https://github.com/mbta/skate/pull/2678/files#r1661568857

firestack (approved):
Looks good! Left some non-blocking comments that'd be nice to chat about but this works for me.

PR #2679, opened 2024-07-02
feat: create module to publish trip modifications to MQTT

firestack:
This is the initial work to create a module to send MQTT events. This does not retain events yet and also does not have tests, but I'm intending on adding those.

---

Asana Ticket: https://app.asana.com/0/1148853526253420/1207383192300471/f
Partially Implements ADRs:
- https://github.com/mbta/skate/pull/2685
- https://github.com/mbta/skate/pull/2686


lib/skate/mqtt_connection.ex
```diff
@@ -0,0 +1,39 @@
...
+    EmqttFailover.Connection.publish(connection, message)
+  end
+
+  @spec publish_in_topic_prefix(server(), EmqttFailover.Message.t()) :: :ok | {:error, term()}
+  def publish_in_topic_prefix(connection, %EmqttFailover.Message{topic: topic} = message) do
+    EmqttFailover.Connection.publish(connection, %{
+      message
+      | topic: (topic_prefix() || "") <> topic
+    })
+  end
```
firestack, line 34: note: Because topic prefixes are allow-listed based on the user in our devops configuration, I figured we should make it so that we automatically are sending messages to the correct topic prefix. I'm not sure this is actually correct though so maybe this should be dropped

lib/skate/detours/trip_modification_publisher.ex
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  @impl GenServer
+  def handle_call(
+        {:new_modification, %{is_draft?: is_draft?, modification: modification}},
+        _from,
+        %__MODULE__{connection: connection} = state
+      )
+      when not is_nil(connection) do
+    # id = "fake-id"
```
joshlarson, line 83:
```suggestion
```

🧟 🧟 🧟

lib/skate/detours/trip_modification_publisher.ex
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def publish_modification(
+        %Realtime.TripModification{} = modification,
+        opts
+      ) do
+    is_draft? = Keyword.get(opts, :is_draft?, false)
+    server = Keyword.get(opts, :server, @default_name)
+
+    # check that server exists?
```
joshlarson, line 26: Does this comment mean that you were considering adding in a check here, but haven't yet?

lib/realtime/trip_modification.ex
```diff
@@ -29,6 +29,7 @@ defmodule Realtime.TripModification do
       :service_date,
       :last_modified_time
     ]
+    @derive Jason.Encoder
```
joshlarson, line 32: I just removed a bunch of `@derive Jason.Encoder` annotations without breaking anything. Are these necessary?

joshlarson (commented):
How much lift would it be to add documentation for how to get this running locally if you wanted to?

I like that it's not required, and that the absense of a local MQTT won't make Skate fall over, but if I ever want to run this on my local machine, I'd love for that to be relatively easy.

lib/realtime/trip_modification.ex
firestack: this one was unneeded, but the rest are [`a05e410f4938f20dbe67da1dccfb2a8a4cb7364b..bf11cdfe783cf811525acebb49d8a7934716336b`#diff-025e8c9325](https://github.com/mbta/skate/compare/a05e410f4938f20dbe67da1dccfb2a8a4cb7364b..bf11cdfe783cf811525acebb49d8a7934716336b#diff-025e8c9325d66be4be78b1284705c85fd7ba87a26ebaed06a3c38b14cd167c36L32)



```
     ** (EXIT from #PID<0.608.0>) an exception was raised:
         ** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for %Realtime.TripModification.SelectedTrip{trip_ids: ["39-0-0-1"], shape_id: "id-of-the-shape"} of type Realtime.TripModification.SelectedTrip (a struct), Jason.Encoder protocol must always be explicitly implemented.

     If you own the struct, you can derive the implementation specifying which fields should be encoded to JSON:

         @derive {Jason.Encoder, only: [....]}
         defstruct ...

     It is also possible to encode all fields, although this should be used carefully to avoid accidentally leaking private information when new fields are added:

         @derive Jason.Encoder
         defstruct ...

     Finally, if you don't own the struct you want to encode to JSON, you may use Protocol.derive/3 placed outside of any module:

         Protocol.derive(Jason.Encoder, NameOfTheStruct, only: [...])
         Protocol.derive(Jason.Encoder, NameOfTheStruct)

             (jason 1.4.1) lib/jason.ex:164: Jason.encode!/2
             (skate 0.1.0) lib/skate/detours/trip_modification_publisher.ex:90: Skate.Detours.TripModificationPublisher.handle_call/3
             (stdlib 5.1.1) gen_server.erl:1113: :gen_server.try_handle_call/4
             (stdlib 5.1.1) gen_server.erl:1142: :gen_server.handle_msg/6
             (stdlib 5.1.1) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

joshlarson: Yup - I just played around with it myself and saw the same error. Sorry for the 🎢!

lib/skate/detours/trip_modification_publisher.ex
firestack: I left that for myself to consider later but never did 😅 
So I'm real unsure what to do here. 

If this module is non-critical, then we could theoretically use `GenServer.cast`, because we don't need any response, maybe the server is there, maybe it's not, does the caller want to know that?
That's what I was originally thinking when I did this.

Then, I was having issues with tests and not sure that the call was ever completing successfully, due to the "fire and forget" nature of `cast`.
So I changed it to `call` and then added the `id` return value to the return.

I never did go back and reevaluate fully if `call` _is_ what we want, and I'm not _certain_ it is. If Mqtt is optional, then a caller shouldn't care if it succeeds or not, and we shouldn't crash on any pattern matching of `{:ok, id} = publish_modification([...])`, at the same time, maybe callers _do_ want to know that, and we can selectively ignore the return value, but then that's more places that could crash.

If we use `call`, then we don't need to check the server is alive, but if we use `cast`, does the caller want to know if we know it didn't succeed because the process doesn't exist?(I'm not how to check for existence of a process either, but I'd _imagine_ it's easy)

joshlarson: I think we care whether a message was sent, since if we fail to deliver a message, we'll need to re-send it. Otherwise some detour info won't make it to Transit Data, which wouldn't be great. So I think that `call` is the right 😎 _call_ 😎  here.

We should address a failed MQTT connection at some point, but a) I don't think we need to block this work on that, and b) I don't think we're in a good state to do that yet (I'm thinking that the _right_ handling is some kind of DB-backed retry, and we're not really set up for doing something in the DB yet).

For now, probably crashing via a bad pattern match is fine? Or the production code could just ignore the return value?

firestack: > Or the production code could just ignore the return value?

Right, what I'm trying to get at is we'll need to _remember_ to do so, and to fail to remember to do something is only human :)

firestack:
> How much lift would it be to add documentation for how to get this running locally if you wanted to?
> 
> I like that it's not required, and that the absense of a local MQTT won't make Skate fall over, but if I ever want to run this on my local machine, I'd love for that to be relatively easy.

Added to [`84a1b43` (#2679)](https://github.com/mbta/skate/pull/2679/commits/84a1b432f2e5e11aae8f76ac2ef6ef0fdf55fa5a), let me know what you think

lib/skate/detours/trip_modification_publisher.ex
joshlarson: True, but we have tools for that :D

My instinct is to not block merging this, and instead create a follow-up ticket that blocks the ["🤝 Detours that Skate produces are complete and accurate" milestone](https://app.asana.com/0/1206070633318963/1207635746210201/f).

joshlarson (approved):
Happy to talk more about how we want to deal with https://github.com/mbta/skate/pull/2679#discussion_r1667093857, but so long as we make follow-up tickets for anything that's missing, I'm good to roll forward with this as-is.

lib/skate/detours/trip_modification_publisher.ex
```diff
@@ -0,0 +1,147 @@
+defmodule Skate.Detours.TripModificationPublisher do
```
joshlarson, line 1: Curiosity/Non-blocking: When we add `Shape` publishing as well, do you envision us adding a whole new GenServer like `ShapePublisher`, or would it make more sense to genericize this as `GtfsPublisher`, or `MqttPublisher` or something like that?

test/skate/detours/trip_modification_publisher_test.exs
```diff
@@ -0,0 +1,78 @@
...
+    expected_topic =
+      id |> TripModificationPublisher.trip_modification_topic() |> MqttConnection.prefix_topic()
+
+    assert_receive {:message, ^reader_pid,
+                    %EmqttFailover.Message{payload: incoming_message, topic: ^expected_topic}}
+
+    message =
+      message
+      |> Jason.encode!()
+      |> Jason.decode!()
```
joshlarson, line 73: Curiosity: Is this encode/decode cycle necessary?

joshlarson (approved):

lib/skate/detours/trip_modification_publisher.ex
firestack: I don't envision the name changing for that reason, IMO, publishing a shape is also part of Trip Modifications, so I'd imagine it being another parameter to `publish` and a second call to `MqttConnection.publish` within the `handle_call` function.

test/skate/detours/trip_modification_publisher_test.exs
firestack: > Curiosity: Is this encode/decode cycle necessary?

Partly; I otherwise didn't know how to convert the struct to string keys with json encoded values (which I _think_ is mainly the string keys part? But I wasn't certain)

https://elixirforum.com/t/how-to-convert-map-keys-from-atoms-to-string-in-elixir/30870/6

lib/skate/detours/trip_modification_publisher.ex
joshlarson: I'm okay with that (especially for now), but what if something about a detour changes other than its `Shape`. For instance, if it goes longer than we originally thought, and so we add more `selected_trips`? Then we probably wouldn't want to re-publish the `Shape`, but just an updated `TripModification` that references the original `Shape`.

That can be a future problem though.

PR #2680, opened 2024-07-02
refactor: Consolidate /detours/finished_detour and /detours/directions endpoints

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207657169700544/f

Depends on:
- https://github.com/mbta/skate/pull/2678
- https://github.com/mbta/skate/pull/2681

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -103,16 +103,15 @@ export const createDetourMachine = setup({
       async ({
         input: { routePatternId, startPoint, waypoints, endPoint },
       }) => {
-        if (routePatternId && startPoint && endPoint) {
-          return fetchFinishedDetour(
-            routePatternId,
-            startPoint,
-            waypoints,
-            endPoint
-          )
-        } else {
+        if (!(routePatternId && startPoint && endPoint)) {
           throw "Missing finished detour inputs"
         }
+        return fetchFinishedDetour(
+          routePatternId,
+          startPoint,
+          waypoints,
+          endPoint
+        )
       }
     ),
   },
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -1,7 +1,7 @@
 import { setup, assign, fromPromise, ActorLogicFrom, InputFrom } from "xstate"
 import { RoutePatternId, ShapePoint } from "../schedule"
 import { Route, RouteId, RoutePattern } from "../schedule"
-import { Ok, Result, map } from "../util/result"
+import { Ok, Result } from "../util/result"
 import {
   FetchDetourDirectionsError,
   fetchDetourDirections,
@@ -96,15 +96,25 @@ export const createDetourMachine = setup({
       {
         routePatternId?: RoutePatternId
         startPoint?: ShapePoint
+        waypoints: ShapePoint[]
         endPoint?: ShapePoint
       }
-    >(async ({ input: { routePatternId, startPoint, endPoint } }) => {
-      if (routePatternId && startPoint && endPoint) {
-        return fetchFinishedDetour(routePatternId, startPoint, endPoint)
-      } else {
-        throw "Missing finished detour inputs"
+    >(
+      async ({
+        input: { routePatternId, startPoint, waypoints, endPoint },
+      }) => {
+        if (routePatternId && startPoint && endPoint) {
+          return fetchFinishedDetour(
+            routePatternId,
+            startPoint,
+            waypoints,
+            endPoint
+          )
+        } else {
+          throw "Missing finished detour inputs"
+        }
       }
-    }),
+    ),
   },
   actions: {
     "set.route-pattern": assign({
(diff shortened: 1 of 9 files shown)


assets/src/models/createDetourMachine.ts
```diff
@@ -348,51 +358,35 @@ export const createDetourMachine = setup({
...
-                  src: "fetch-detour-directions",
-                  input: ({
-                    context: { startPoint, waypoints, endPoint },
-                  }) => ({
-                    points: (startPoint ? [startPoint] : [])
-                      .concat(waypoints || [])
-                      .concat(endPoint ? [endPoint] : []),
-                  }),
+              invoke: {
+                src: "fetch-finished-detour",
```
joshlarson, line 362: NGL - it was pretty neat that I was able to do what I was trying to do by just going here and removing one of the `invoke`s.

assets/src/models/finishedDetour.ts
```diff
@@ -16,6 +17,7 @@ export const FinishedDetourData = type({
     detour: array(Coordinate),
     after_detour: array(Coordinate),
   }),
+  detour_shape: DetourShapeData,
```
joshlarson, line 20: I feel like this interface and this type have gotten kind of _messy_, but I almost would rather clean that up when we make a real `/detours` API.

assets/src/models/createDetourMachine.ts
```diff
@@ -348,51 +358,35 @@ export const createDetourMachine = setup({
...
+                          {
                             instruction: "Regular Route",
-                          }),
-                        })),
-                    }),
-                  },
-
-                  onError: {},
+                          }
+                        ),
```
firestack, line 383: note: I do wonder if we should be surfacing the potential errors from directions, so I've asked in Slack https://mbta.slack.com/archives/C024N3SKX33/p1720010049121199

joshlarson: This PR does mean that there's one interaction that used to cause a visible error banner that would now not do that. I don't love the idea of introducing a bug, but in the grand scheme of things, this feels relatively low-impact.

Here's why I thought this was okay:
- There are two user interactions that can create this error banner: placing a waypoint and placing the end point.
- If you'd get the error from placing a waypoint, you'll still get that error.
- So the only actual change is if an error happens that would only manifest in the directions between the last waypoint and the end point, at which point you get the bad UX of no feedback and no route shape.
- I'm guessing ☝️ is super unlikely, if not impossible, which is why I don't feel strongly that we need to fix this right this instant.

The only reason that I'm sort of okay with moving forward with introducing the bug is that it does unblock the MQTT shape work (you can't build a `Shape` without info from the consolidated `/finished_detour` endpoint), and I don't know what the Right™ thing to do here is.

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -64,12 +69,16 @@ defmodule SkateWeb.DetoursController do
           connection_end_location
         )
 
+      {:ok, ors_result} =
+        OpenRouteServiceAPI.directions([connection_start] ++ waypoints ++ [connection_end])
+
```
firestack, line 74: issue: this assertion now means that any successful computation of missed stops and the route segments will be overshadowed by ORS errors, which I don't think is a great UX

(I put a waypoint in the ocean in this example)
image

I don't think we should require here that directions are successful to be able to compute the other things.

assets/src/models/createDetourMachine.ts
```diff
@@ -96,15 +96,25 @@ export const createDetourMachine = setup({
...
+        if (routePatternId && startPoint && endPoint) {
+          return fetchFinishedDetour(
+            routePatternId,
+            startPoint,
+            waypoints,
+            endPoint
+          )
+        } else {
+          throw "Missing finished detour inputs"
+        }
```
firestack, line 115: suggestion(non-blocking): It'd be nice to flatten this while were here
```suggestion
        if (!(routePatternId && startPoint && endPoint)) {
          throw "Missing finished detour inputs"
        }
        return fetchFinishedDetour(
          routePatternId,
          startPoint,
          waypoints,
          endPoint
        )
```

firestack (changes requested):
A few more changes but looking good 👍🏻

lib/skate_web/controllers/detours_controller.ex
joshlarson: I'm thinking of this as a follow-up, but would it be possible for the state machine to disallow placing more waypoints/endpoint/etc when there's a routing error? Because if there's a routing error for waypoints `1` thru `n`, then we already know that there'll be a routing error for waypoints `1` thru `n+1`.

(We probably only want to have this behavior on routing error, not on other errors)

That feels like something the state machine would be able to do pretty easily.

firestack: > but would it be possible for the state machine to disallow placing more waypoints/endpoint/etc when there's a routing error?

Sounds easily doable to me! Happy to find some time to pair on it if you'd like.

joshlarson: Do you want ☝️ to be a part of this PR? Or should we pair on it as a follow-up?

firestack: Lets do it as a follow-up 👍🏻

joshlarson: https://app.asana.com/0/1148853526253420/1207750816978131/f

joshlarson: Take 2: https://app.asana.com/0/1148853526253420/1207768837488463/f

firestack (approved):

PR #2685, opened 2024-07-04
doc: create ADR 0003 for Skate MQTT

firestack:
We've somewhat already made this decision, but in trying to record this, I've created this ADR to record our decision to use MQTT in Skate for Trip Modifications communications.

---

Rendered: https://github.com/mbta/skate/blob/kf/doc/mqtt-adr/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
Asana Ticket: https://app.asana.com/0/1205385723132845/1207716595296862/f

### DIFF
--- a/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
+++ b/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
@@ -1,0 +1,83 @@
+# 3. Use MQTT to publish Trip Modifications
+
+Date: 2024-07-04
+
+## Status
+
+Accepted
+
+## Context
+
+We need to send [GTFS Trip Modifications](https://gtfs.org/realtime/feed-entities/trip-modifications/)
+to Transit Data.
+
+## Decision
+
+To minimize the amount of resources used and latency between Skate publishing or
+un-publishing a Trip Modification and Transit Data becoming aware, we are going
+to use MQTT to form an event stream between Skate and Transit Data.
+
+Since this is an output of Skate, we will not make this service required to do
+local development.
+
+### Alternatives Considered
+
+#### Polling based alternatives
+1. HTTP API
+
+    We considered delivering this information by way of an API that
+    Transit Data could poll. That would have the benefit of allowing
+    us to "render straight from the database" and avoid any sort of
+    expiration/synchronization logic with an external service.
+    Additionally, it is already an established practice within Transit
+    Data to poll an API (Ex: `Concentrate` polls `RTR`).
+
+    This might be an appropriate choice if we expected there to be
+    more "writes" than "reads" -- that is, if we expected officials to
+    be modifying detours more frequently than Transit Data would be
+    querying them. But we actually expect many more "reads" than
+    "writes" - we'll need to access this data any time a rider looks
+    up a bus route, or any time an operator loads turn-by-turn
+    directions, or possibly once per second, while writes will only
+    happen a few times a week, when an official is creating or
+    modifying a detour. Creating an API would create performance
+    concerns that messaging wouldn't.
+
+    Additionally, Skate's current access control mechanism is
+    concerned entirely with _users_ and not providing pragmatic
+    access. Using an external service allows us to push authentication
+    for this access to other services and IaC instead of requiring
+    that Skate also concerns itself with these issues.
+
+2. S3
+
+    [Paul Swartz suggested "[...] if the data isn't updating frequently, S3 is 
+    still a good option."](https://www.notion.so/Detours-Transit-Data-Skate-8f39df3d5601437b81a657efdad270c4?d=fabf76c1c0b24acf9f3bb93f7c019dec&pvs=4#b0b68315daf2449ea39582cf2463841b)  
+    We considered S3 and decided that it wasn't a good match due to having the
+    same complexity of introducing a new service as MQTT and also requiring
+    logic to manage old trip modifications, without the benefit of lower latency
+    from an event based method.[^ps-1]
+
+#### Event based alternatives
+3. AWS Kinesis
+
+    We briefly considered Kinesis because we had thought that MQTT might not
+    have a method for catching new subscribers up to speed on any existing Trip
+    Modifications, which is a requirement because "Concentrate doesn't currently
+    have a mechanism for maintaining state". Upon learning that MQTT has
+    `retained` messages, we opted for MQTT due to Kinesis being proprietary.
+
+
+## Consequences
+
+Since we'll be adding a new technology and service to Skate, there will be more
+surface area to test.
+
+By using a message queue, we risk sending events out of order and
+having inaccurate ordering of data, we'll need to be careful to address these
+issues as they arise during implementation.
+
+[^ps-1]:
+    In this comment [Paul Swartz mentioned that using MQTT but sending all
+    messages using `retain` will effectively be the same as uploading to S3, but
+    with lower latency to clients](https://app.asana.com/0/0/1207124404105396/1207386033651835/f).


documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
```diff
@@ -0,0 +1,26 @@
...
+un-publishing a Trip Modification, we are going to use MQTT to form a 
+Event Stream between Skate and Transit Data.
+
+Since this is an output of Skate, we will not make this service required to do
+local development.
+
+## Consequences
+
+We'll be adding a new technology and service to Skate, this is more surface area
+to test.
```
joshlarson, line 26: Are there any additional risks that we need to consider?

Here are a few possible ones:
1. We need to be super careful in order to ensure that Transit Data gets up-to-date detour info. If the wrong message winds up being retained, then Transit Data won't have any way to get the actual latest message.
   - If we run purely on fire-and-forget, then if a message doesn't make it to the MQTT broker for whatever reason, then it never will. 
   - If we do retry sending messages on failure, then we could wind up sending messages out of order.
1. We may need to keep stale topics around for a long time so that TD knows that a detour has been deactivated.

documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
```diff
@@ -0,0 +1,26 @@
...
+## Context
+
+We need to send [GTFS Trip Modifications](https://gtfs.org/realtime/feed-entities/trip-modifications/)
+to Transit Data.
+
+## Decision
+
+To minimize the amount of resources used and latency between Skate publishing or
+un-publishing a Trip Modification, we are going to use MQTT to form a 
+Event Stream between Skate and Transit Data.
```
joshlarson, line 18:
```suggestion
event stream between Skate and Transit Data.
```

Should "event stream" be capitalized?

documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
```diff
@@ -0,0 +1,26 @@
...
+
+Since this is an output of Skate, we will not make this service required to do
+local development.
+
+## Consequences
+
+We'll be adding a new technology and service to Skate, this is more surface area
```
joshlarson, line 25:
```suggestion
We'll be adding a new technology and service to Skate, so there will be more surface area
```

OR

```suggestion
Since we'll be adding a new technology and service to Skate, there will be more surface area
```

Tiny grammar nit.

joshlarson (changes requested):
Made a few tiny grammar nitpicks.

I'm wondering if it would be worth documenting the alternatives as well (specifically I'm thinking of an HTTP-based API). I read through [Michael Nygard's description of ADR's](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions) and there isn't anything explicit about alternatives, but my intuition about this is that a large part of the benefit of an ADR is so that it would answer the question of "wait, why didn't we do this other thing?".

I don't want it to look like we just didn't think of it, or didn't give it serious consideration 😅

documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
firestack: > If the wrong message winds up being retained

I don't envision this as being an issue, can you elaborate on why it would be?

Also, I don't believe this risk would belong in this ADR either.

firestack: > We may need to keep stale topics around for a long time so that TD knows that a detour has been deactivated.

I think this is also not a concern for this ADR. But I also don't think we need to keep stale topics around, we can discuss further in #2686 if you'd like.

firestack:
> I'm wondering if it would be worth documenting the alternatives as well (specifically I'm thinking of an HTTP-based API). I read through [Michael Nygard's description of ADR's](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions) and there isn't anything explicit about alternatives, but my intuition about this is that a large part of the benefit of an ADR is so that it would answer the question of "wait, why didn't we do this other thing?".
> 
> I don't want it to look like we just didn't think of it, or didn't give it serious consideration 😅

@joshlarson PTALA https://github.com/mbta/skate/compare/c5392140c2176ba262a0db04ed3f30fa696c9484..aa8ed2017614c56c50142707acedbd282798d798

documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
joshlarson: > I don't envision this as being an issue, can you elaborate on why it would be?

If we're not careful (and implement retry in the wrong way), we could wind up with a sequence of events like:
- Create Detour A (communicated with `message 1`)
- Sending `message 1` fails - queue a retry
- Update Detour A in some way (communicated with `message 2`)
- Sending `message 2` succeeds
- Sending `message 1` retry succeeds

Now the latest message about Detour A is `message 1`, which doesn't have whatever the updated info from `message 2` is.

Making sure that the right message is retained is a solvable problem, but any time messages are a thing, there is that bit of risk that messages will get delivered out of order.

firestack: > If we're not careful (and implement retry in the wrong way),

What do you think about this?
```diff
diff --git a/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md b/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
index f72ba5caf2...8cb8b8c8bc 100644
--- a/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
+++ b/documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
@@ -57,6 +57,10 @@
 Since we'll be adding a new technology and service to Skate, there will be more
 surface area to test.
 
+By using a message queue, we risk sending events out of order and
+having inaccurate ordering of data, we'll need to be careful to address these
+issues as they arise during implementation.
+
 [^ps-1]: In this comment [Paul Swartz mentioned that using MQTT but sending all
   messages using `retain` will effectively be the same as uploading to S3, but
   with lower latency to clients](https://app.asana.com/0/0/1207124404105396/1207386033651835/f).
```

documentation/adr/0003-use-mqtt-to-publish-trip-modifications.md
```diff
@@ -0,0 +1,62 @@
...
+  benefit of this is we can "render straight from the database" and avoid any
+  sort of expiration/synchronization logic with an external service.
+  Additionally, it is already an established practice within Transit Data to
+  poll an API (Ex: `Concentrate` polls `RTR`).  
+  The downside being that we are then requiring extra resources to continuously
+  poll, at 1s intervals, to check for any changes. Additionally, Skate's current
+  access control mechanism is concerned entirely with _users_ and not providing
+  pragmatic access. Using an external service allows us to push authentication
+  for this access to other services and IaC instead of requiring that Skate also
+  concerns itself with these issues.
```
joshlarson, line 37: Another worthwhile point is that an API would be a good choice if "reads" were infrequent while "writes" were frequent, but the reality is the other way around - "reads" happen once a second, or any time a user wants to look something up, while write will only happen a few times a week

Some possible wordsmithing (no need to take this suggestion exactly):

```suggestion
  - We considered delivering this information by way of an API that
    Transit Data could poll. That would have the benefit of allowing
    us to "render straight from the database" and avoid any sort of
    expiration/synchronization logic with an external service.
    Additionally, it is already an established practice within Transit
    Data to poll an API (Ex: `Concentrate` polls `RTR`).

    This might be an appropriate choice if we expected there to be
    more "writes" than "reads" -- that is, if we expected officials to
    be modifying detours more frequently than Transit Data would be
    querying them. But we actually expect many more "reads" than
    "writes" - we'll need to access this data any time a rider looks
    up a bus route, or any time an operator loads turn-by-turn
    directions, or possibly once per second, while writes will only
    happen a few times a week, when an official is creating or
    modifying a detour. Creating an API would create performance
    concerns that messaging wouldn't.

    Additionally, Skate's current access control mechanism is
    concerned entirely with _users_ and not providing pragmatic
    access. Using an external service allows us to push authentication
    for this access to other services and IaC instead of requiring
    that Skate also concerns itself with these issues.
```

Rendered:
>  - We considered delivering this information by way of an API that
    Transit Data could poll. That would have the benefit of allowing
    us to "render straight from the database" and avoid any sort of
    expiration/synchronization logic with an external service.
    Additionally, it is already an established practice within Transit
    Data to poll an API (Ex: `Concentrate` polls `RTR`).
>
>    This might be an appropriate choice if we expected there to be
    more "writes" than "reads" -- that is, if we expected officials to
    be modifying detours more frequently than Transit Data would be
    querying them. But we actually expect many more "reads" than
    "writes" - we'll need to access this data any time a rider looks
    up a bus route, or any time an operator loads turn-by-turn
    directions, or possibly once per second, while writes will only
    happen a few times a week, when an official is creating or
    modifying a detour. Creating an API would create performance
    concerns that messaging wouldn't.
>
>    Additionally, Skate's current access control mechanism is
    concerned entirely with _users_ and not providing pragmatic
    access. Using an external service allows us to push authentication
    for this access to other services and IaC instead of requiring
    that Skate also concerns itself with these issues.

hannahpurcell (approved):
Glad to have the "considered alternatives" included. I think the edit you note [here](https://github.com/mbta/skate/pull/2685#discussion_r1670755368) would be useful. (I'm unsure if there are any other "risks" that need capturing?) 

One tweak: the footnote is used in two places, but I think it only applies in 1 place (not under the Kinesis bullet point)

Whoops (approved):

joshlarson (approved):

PR #2686, opened 2024-07-04
 doc: create ADR 0004 for Skate MQTT Hierarchy 

firestack:
This proposes the initial schema and topic hierarchy.

---

Rendered: https://github.com/mbta/skate/blob/kf/doc/mqtt-hierarchy-adr/documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
Asana Ticket: https://app.asana.com/0/1205385723132845/1207716595296863/f


documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,134 @@
...
+
+#### Active Trip Modifications State
+Messages pushed to these topics will have the
+[`retain` property](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/)
+set so that new connections get active Trip Modifications and Shapes published
+to them on connect.
+
+Deactivating a Trip Modification will be done by
+["deleting"](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/#heading-how-to-delete-retained-messages-in-mqtt)
+the retained messages on each topic within the `<prefix>/trip_modifications/#{id}/`
```
paulswartz, line 82: question: does this send a message to current subscribers telling them that the message has been deleted?

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,134 @@
...
+[MQTT "topic level"](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/)
+`trip_modifications`, which will be under our "topic prefix"
+configured in the DevOps repo for each environment.
+
+For Example: (spaces used for alignment, and are not part of the topic)
+```
+skate/prod     /trip_modifications/
+skate/dev      /trip_modifications/
+skate/dev-blue /trip_modifications/
+skate/dev-green/trip_modifications/
```
joshlarson, line 37:
```suggestion
     skate/prod/trip_modifications/
      skate/dev/trip_modifications/
 skate/dev-blue/trip_modifications/
skate/dev-green/trip_modifications/
```

The spaces confused me despite the annotation 😢.

This is not a huge deal at all, but if we want the `/trip_modifications/` to all be aligned, maybe right-aligning the whole thing?

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+
+For each new and unique trip modification, we'll generate a unique ID
+(likely `UUID` using `Ecto.UUID.generate()`)
```
joshlarson, line 42:
```suggestion
```

Nit: I don't think we need to (or want to) specify how we'll generate the ID's. So long as they're unique, I think we're good.

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
joshlarson: The mechanism for deleting a retained message is to publish a 0-byte message to the topic, which overwrites the previous retained message. If a subscriber is currently online when that happens, I'd be pretty surprised if they didn't get alerted to the "new 0-byte message".

I'm less certain of what happens if no subscribers are online when we publish the 0-byte message. My guess is that the topic would still be there, but with no retained message (which Transit Data could interpret as "any previously-existing detour with this ID has been deactivated"). But it's possible that the topic itself gets deleted at that point, which means that if Transit Data is having an outage at that point, they wouldn't get that update.

If we decide to go a different route, and signal detour deactivation in some way that leaves a retained message on its MQTT topic, are we worried about cluttering up our MQTT space? Either cognitive load associated with a whole bunch of topics lying around representing ancient detours, or actual space constraints?

And/or is there an easy way to tell MQTT itself to delete topics that haven't had any activity for, say, a week, even if they have a retained message?

joshlarson (changes requested):
Similar to https://github.com/mbta/skate/pull/2685#pullrequestreview-2161228424, I think it'd be worth discussing some alternatives, and why we're choosing this approach.

Alternatives that we should at least mention:
- A single topic with no retained messages where we put explicit "diff" messages (Detour A exists now, Detour B now applies to trip xyz, Detour C is deactivated, etc)
- Topics split out by detour ID with retained messages, but instead of deactivating with a 0-byte message, deactivating with an explicit "deactivated" message
- No deactivating at all - we just stop adding trip ID's to a particular detour (or remove them if they've already been added) once a detour doesn't apply to future or current trips anymore.

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: > question: does this send a message to current subscribers telling them that the message has been deleted?

Great question, this is actually something I wanted to write a test for, thanks for bringing this up! I'll test it in #2679 



> If a subscriber is currently online when that happens, I'd be pretty surprised if they didn't get alerted to the "new 0-byte message".

This is my assumption so far, but will test it 👍🏻.

firestack: > > If a subscriber is currently online when that happens, I'd be pretty surprised if they didn't get alerted to the "new 0-byte message".
> 
> This is my assumption so far, but will test it 👍🏻.

TL;DR: When a "zero-byte payload" is published, subscribers get a zero byte message.

I tested locally, with Skate code and running `mosquitto_sub` locally.
The MQTT Spec says "publish a retained message with a zero-byte payload" to delete a retained message on a topic. 

To publish a retained "zero-byte payload" with the `EmqttFailover` library, you must send a message that fits the following pattern:
``` elixir
%EmqttFailover.Message{payload: "", retain?: true} = sent_message
```

Subscribers will receive a similar message when this is published, but the retain flag will not be true from what I can tell
``` elixir
%EmqttFailover.Message{payload: "", retain?: true} = received_message
```

When new subscriptions are created though, a message is __not__ sent containing this zero byte payload. So this means the topic is effectively "deleted" for new subscribers, as they'll never know it existed until a new message is published to that topic.

firestack: > And/or is there an easy way to tell MQTT itself to delete topics that haven't had any activity for, say, a week, even if they have a retained message?

[As of MQTT 5, the spec says that messages can have a Message Expiry Interval](https://www.hivemq.com/blog/mqtt5-essentials-part4-session-and-message-expiry/#heading-message-expiry-interval-in-mqtt-5). This doesn't _seem_ to be exposed in `EmqttFailover`, I don't know if it's exposed in the underlying library either.

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: What do you think about doing something like this?
````diff
diff --git a/documentation/adr/0004-trip-modifications-mqtt-topic-structure.md b/documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
index a512c6f91e...f249ca111e 100644
--- a/documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
+++ b/documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
@@ -29,13 +29,21 @@
 `trip_modifications`, which will be under our "topic prefix"
 configured in the DevOps repo for each environment.
 
-For Example: (spaces used for alignment, and are not part of the topic)
+For Example:
+
+Our `<prefix>`s would be derived from the environment and are described via IaC
+in the devops repo
 ```
-skate/prod     /trip_modifications/
-skate/dev      /trip_modifications/
-skate/dev-blue /trip_modifications/
-skate/dev-green/trip_modifications/
+skate/prod     
+skate/dev      
+skate/dev-blue 
+skate/dev-green
 ```
+And the final combined topic hierarchy would be formed by prepending _one_ of 
+the `<prefix>`s
+```
+<prefix>/trip_modifications/
+```
 
 
 For each new and unique trip modification, we'll generate a unique ID and use
````

Comment (joshlarson) [reply documentation/adr/0004-trip-modifications-mqtt-topic-structure.md:37]: Love it

I would add something like:

```diff
+And the final combined topic hierarchy would be formed by prepending _one_ of 
+the `<prefix>`s
+```
+<prefix>/trip_modifications/
+```
+For instance,
+```
+skate/dev-blue/trip_modifications/
+```
```

firestack: https://github.com/mbta/skate/compare/3d9b684ffe2f00c5c9ee208312fe0b769699ed9e..124dfdc2beb3cc2b22d6279b2fdefc5676cd89d1

firestack:
> Similar to [#2685 (review)](https://github.com/mbta/skate/pull/2685#pullrequestreview-2161228424), I think it'd be worth discussing some alternatives, and why we're choosing this approach.
> 
> Alternatives that we should at least mention:
> 
>     * A single topic with no retained messages where we put explicit "diff" messages (Detour A exists now, Detour B now applies to trip xyz, Detour C is deactivated, etc)
> 
>     * Topics split out by detour ID with retained messages, but instead of deactivating with a 0-byte message, deactivating with an explicit "deactivated" message
> 
>     * No deactivating at all - we just stop adding trip ID's to a particular detour (or remove them if they've already been added) once a detour doesn't apply to future or current trips anymore.

I copied your thoughts and added some responses, let me know what you think? 
https://github.com/mbta/skate/compare/124dfdc2beb3cc2b22d6279b2fdefc5676cd89d1..00ca2f322b14d8813ad2450c3f7215cdc48ba1ef

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+[`retain` property](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/)
+set so that new connections get active Trip Modifications and Shapes published
+to them on connect.
+
+Deactivating a Trip Modification will be done by
+[publishing a zero-byte payload](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/#heading-how-to-delete-retained-messages-in-mqtt)
+each topic within the `<prefix>/trip_modifications/#{id}/`
+topic hierarchy, which will clear any retained messages and ensure that new
+subscribers are not notified of these now deleted messages and existing
+are made aware of the removal subscribers.
```
hannahpurcell, line 96: perhaps this? (I didn't understand the sentence and am taking a stab at copyediting)

```suggestion
subscribers are not notified of these now deleted messages and existing
subscribers are made aware of the removal.
```

firestack: oops, thanks!

hannahpurcell:
Ok great! I'm giving my casual approval, as I see you have multiple other reviewers who have spent more time with this doc and I would like to see their approval. But this ADR covers what I need to see

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+  "data": {
+    # <GTFS-RT Trip Modification Message as JSON>
+  }
+  "meta": {
+    # this field will be removed once we give users the ability to activate and
+    # deactivate detours themselves.
+    #
+    # Currently, this indicates that this message should not be published to
+    # applications outside of the MBTA, and is for internal testing and
+    # development only.
```
joshlarson, line 115:
```suggestion
    // this field will be removed once we give users the ability to activate and
    // deactivate detours themselves.
    //
    // Currently, this indicates that this message should not be published to
    // applications outside of the MBTA, and is for internal testing and
    // development only.
```

This is rendering funny, and I think it's because `#` isn't the right comment indicator for JSON.


```json5
{
  # This is not a comment
  "elixir comment": "#",

  // This is a comment
  "json comment": "//",
}
```

firestack: Ah yep, great catch!! I knew this looked a bit weird, I've gotten to used to not having proper highlighting 😅 

[Thanks a bunch!](https://github.com/mbta/skate/compare/8d768330b70c9ff37649a6a1c99793fc6a59d1ad..dc0d1bb510f12ff35ae2a1ec3940cf1b36c269a5)

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+3.  No deactivating at all - we just stop adding trip ID's to a particular detour
+    (or remove them if they've already been added) once a detour doesn't apply to
+    future or current trips anymore.
+
+    Because we need to keep Trip Modifications up to date with the trips that
+    the modification applies to, we'll be publishing frequent updates depending
+    on how many trips out we're configured to say will be affected. We
+    theoretically could publish a Trip Modification with an empty
+    `selected_trips` field, but this would go against the
+    [GTFS spec of Required: Many, Cardinality: One](https://gtfs.org/realtime/reference/#message-selectedtrips)
```
joshlarson, line 178: This might be a misunderstanding on my part, but I thought that a trip ID uniquely specified a trip. Like the run of the 39 departing Back Bay today at 4:35 is a specific trip ID, and no other trip has the same ID. 

If I'm right, then is there anything wrong with having a `TripModification` with a `selected_trip_ids` that includes a trip that happened in the past?

Not saying that we need to go that route by any stretch, but this alternative would mean that "`TripModification` where all of the `selected_trip_ids` are trips that already happened" === "Deactivated Detour".

firestack: Trip Id's are not unique across time.
So if the service ID for a trip _doesn't_ change, then that trip will be marked as detoured in perpetuity.

https://gtfs.org/schedule/reference/#tripstxt

> ### Trips
> [...]
>
> You may also notice there is no date field. That is because in GTFS, a `trip` is not an individual trip as a rider would perceive it (i.e. “the Inbound 89 trip that occurs at 8:39am on October 27th, 2020”), but a trip that occurs once on every date defined for a given service. 
>
> – https://github.com/firestack/gtfs_onboarding.nix/blob/82c011138de121e8ee1c520cc56f39bca880b5ba/GTFS_Onboarding.sqlite3.ipynb#Trips

firestack:
> * Transit Data comes back online
> 
> * No new message (since what we did was a deletion, not a "delete message")
> 
> * Everyone downstream of Transit Data thinks the detour is still active

As I understand it, Concentrate is "stateless", it'll re-read open topics when it comes online, and publish a new GTFS-RT file without the trip modification in it.

firestack:
> MQTT could delete the topic. If that happens, I'm not sure what we do, TBH.

Topics don't really "exist", you only know about them when a message is sent on it.

> MQTT could deliver the zero-byte message on startup, which means this isn't even a concern (I'd be surprised if that's how this works, but I like surprises 🙂)

[A zero-byte message "deletes the topic", by deleting the last retained message, and therefore new subscribers after the deletion are not even aware the message or topic ever existed](https://github.com/mbta/skate/pull/2686#discussion_r1668898384).

> MQTT could keep the topic around with no retained message (I think this is the most likely outcome)

I don't understand? But see above, I think this is an incorrect assumption?

> [..] when new clients come online, it treats the topic as empty. Transit Data could iterate through all of the topics, and treat any that have no retained message as "deactivated"

This is what's expected, except it's that when it first connects it'll receive all retained messages for topics it subscribes too. (which **_should_** include the wildcard to subscribe to all `id`s)

> and do its Transit Data thing to deactivate them downstream. 

There's not much to do aside from publish a new GTFS-RT file with the current data, because it's stateless, there's nothing to "remove", only to "omit" (and omission is a side-effect of the data disappearing from the feed via the zero-byte message)

> This seems fine, but does run the risk of Transit Data having to iterate over an increasingly large collection of topics on startup.

We should be deleting any topics that aren't live anymore, if the collection is growing, either all those detours should be live, or there's a bug IMO.

joshlarson:
> As I understand it, Concentrate is ephemeral and does not have a DB, it'll re-read open topics when it comes online, and publish a new GTFS-RT file without the removed trip modification in it.

This is helpful context, thank you!

I guess I wasn't (maybe am still not?) totally clear on who TD's downstream consumers are, and what data they need - for instance, I thought that TD was going to publish something directly to Swiftly, which _might_ mean that they need to publish a deactivation to Swiftly as well?

Anyway, assuming that Walton/TransitData approves, then I think this looks great!

firestack:
> for instance, I thought that TD was going to publish something directly to Swiftly, which _might_ mean that they need to publish a deactivation to Swiftly as well?

My understanding is that Swiftly consumes our GTFS-RT feed, so they'd just implicitly get these once we started publishing them to the feed, and they'd continue to make predictions for us based on this info. I'd defer to @Whoops to make sure, but I'm fairly certain we won't be tying into Swiftly via any specific API's at this point.

Whoops:
> > for instance, I thought that TD was going to publish something directly to Swiftly, which _might_ mean that they need to publish a deactivation to Swiftly as well?
> 
> My understanding is that Swiftly consumes our GTFS-RT feed, so they'd just implicitly get these once we started publishing them to the feed, and they'd continue to make predictions for us based on this info. I'd defer to @Whoops to make sure, but I'm fairly certain we won't be tying into Swiftly via any specific API's at this point.

@fsaid90 This is the question I mentioned in our meeting. Can you put what you said in writing? :-)

fsaid90:
> > > for instance, I thought that TD was going to publish something directly to Swiftly, which _might_ mean that they need to publish a deactivation to Swiftly as well?
> > 
> > 
> > My understanding is that Swiftly consumes our GTFS-RT feed, so they'd just implicitly get these once we started publishing them to the feed, and they'd continue to make predictions for us based on this info. I'd defer to @Whoops to make sure, but I'm fairly certain we won't be tying into Swiftly via any specific API's at this point.
> 
> @fsaid90 This is the question I mentioned in our meeting. Can you put what you said in writing? :-)

Yep!

Right now the flow is:

_Busloc -> Swiftly -> Consumers_

Busloc assembles GTFS-RT VehiclePositions + TripUpdates _sans predictions_ and delivers them to Swiftly. Swiftly send adjusted feeds with predictions. Concentrate consumes + applies block waivers to cancel trips then publishes to consumers.

We'll have a new flow as well:

_Skate -> Somewhere? S3? Concentrate? -> Swiftly -> Concentrate -> consumers_

Skate would send TripModifications to us, we'd have to forward them on to Swiftly (to adjust predictions), then we'd receive updated GTFS-RT from Swiftly that factors in the TripModifications, and we'd publish that GTFS-RT (alongside our TripModifications). I don't expect that Swiftly will _adjust_ TripModifications, they just need to know about them to update predictions accordingly (i.e. we can publish the TripModifications as is from Skate).

@Whoops you mentioned this might be a circular dependency, but I'm not sure that it is! Let me know what you think.

firestack:
> _Skate -> Somewhere? S3? Concentrate? -> Swiftly -> Concentrate -> consumers_



> @Whoops you mentioned this might be a circular dependency, but I'm not sure that it is! Let me know what you think.

Seems a little circular to me 😅 

```mermaid
graph TD
	Skate -- MQTT --> Concentrate -- without predictions --> Swiftly -- with predictions --> Concentrate --> Consumers
```

fsaid90:
> > _Skate -> Somewhere? S3? Concentrate? -> Swiftly -> Concentrate -> consumers_
> 
> > @Whoops you mentioned this might be a circular dependency, but I'm not sure that it is! Let me know what you think.
> 
> Seems a little circular to me 😅
> 
> ```mermaid
> graph TD
> 	Skate
> 	-- MQTT --> Concentrate
> 	-- without predictions --> Swiftly
> 	-- with predictions --> Concentrate
> 	--> Consumers
> ```
> But also, Skate _could_ become a consumer at some point because we'd potentially want the Alert information that Alerts would add eventually.
> 
> ```mermaid
> graph TD
> 	Skate
> 	-- MQTT --> Concentrate
> 	-- without predictions --> Swiftly
> 	-- with predictions --> Concentrate
> 	--> Consumers -.->  Skate
> ```

Haha definitely looks circular 😅 

I'm really thinking about it as two separate flows:

GTFS-RT TripUpdates + VehiclePositions flow continues to be the same: _Busloc -> S3 -> Swiftly -> Concentrate -> public_

TripModifications: _Skate -> Concentrate -> Public_ AND _Skate -> Concentrate -> S3 -> Swiftly_ <- it ends there since we're not waiting on the resource to come back to us?

_But obviously I'm not a developer so I'll defer to y'all 😃_

Whoops:
I think the big question is what exactly is Switly expecting to consume? Right now Concentrate takes in a bunch of RT (including Swiftly) feeds, and soon these Skate messages and outputs them into an RT feed. If Swiftly is going to then consume that feed then we have a situation where Swiftly is both consuming that RT feed to generate it's output, and Switfly's output goes into that RT feed.

I'm curious what you mean by "without predictions" in the diagram above. The only Concentrate output that doesn't have predictions in my mind is VehiclePositions.

fsaid90:
@Whoops 

> I think the big question is what exactly is Switly expecting to consume? 

- VehiclePositions from Busloc (via S3)
- TripUpdates from Busloc (via S3)
- TripModifications from _somewhere_ 

>  If Swiftly is going to then consume that feed then we have a situation where Swiftly is both consuming that RT feed to generate it's output, and Swiftly's output goes into that RT feed.

Here's where I might be missing something - in my thinking:

- Swiftly is taking in Busloc VehiclePositions + TripUpdates output as per the above
- Swiftly now also takes in TripModifications _(to make proper modifications for adjusted trips)_
- Swiftly sends us back _only_ TripUpdates as they currently do
- Concentrate publishes:
  - VehiclePositions from Busloc
  - TripModifications from Skate _(Swiftly won't be modifying TripModifications - I don't see a reason for them to do so unless you do?)_
  - TripUpdates from Swiftly _(factoring in TripModifications if they exist)_

---

@firestack 

> Where/How does Swiftly get our VehiclePosistions and TripUpdates without predictions currently?

Swiftly retrieves VehiclePositions and TripUpdates that Busloc produces and dumps into S3 - basically polling

joshlarson:
>  I don't expect that Swiftly will adjust TripModifications, they just need to know about them to update predictions accordingly (i.e. we can publish the TripModifications as is from Skate).

I'm wondering about the `propagated_modification_delay` field in the [Modification message](https://gtfs.org/realtime/reference/#message-modification).

Our current code doesn't do anything to compute that field, and I'd been thinking about that field as effectively a prediction, which puts it into Swiftly's wheelhouse. So I had envisioned a TripModification going through a pipeline like

- Skate generates `TripModification` message without `propagated_modification_delay` and sends it to TransitData
- TransitData sends that message to Swiftly
- Swiftly updates predictions for the selected trips and lets TransitData know in some way (as well as updating predictions that Skate receives)
- TransitData publishes `TripModification` messages with `propagated_modification_delay` populated to the actual GTFS-RT feed

How do you all envision `propagated_modification_delay` being set? Do we even need that field (I think yes, or else rider-facing info will have inaccurate predictions)? Should Skate figure it out and attach it to the message, rather than relying on Swiftly?

firestack:
> How do you all envision `propagated_modification_delay` being set? Do we even need that field (I think yes, or else rider-facing info will have inaccurate predictions)? Should Skate figure it out and attach it to the message, rather than relying on Swiftly?

I think that because Swiftly is responsible for providing the TripUpdates data, that their predictions will just include the info computed from the Trip Modification data we give them. We should assert that this assumption is correct, but this seems most correct to me. So I don't think we _need_ to provide the `propagated_modification_delay`, because IMO that's only needed if we didn't have live updating TripUpdates but instead the client needs to propagate this delay through a static schedule.

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
paulswartz: ActiveMQ 5.17 only supports MQTT 3.

firestack: I don't think we need Message Expiry anyway, but good to note thank you!

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+```
+skate/dev-blue/trip_modifications/
+```
+
+
+For each new and unique trip modification, we'll generate a unique ID and use
+that as our Trip Modification identifier, named `id`.
+
+> [!NOTE]  
+> The GTFS-RT Trip Modifications spec does not currently have a ID for each modification,
```
paulswartz, line 57: suggestion: the `FeedEntity` (of which `TripModification` is a child) does have an ID: https://github.com/google/transit/blob/master/gtfs-realtime/proto/gtfs-realtime.proto#L86-L95

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+Any associated [`Shape`](https://gtfs.org/realtime/reference/#message-shape)
+Messages will be published to the topic level `shape` in the hierarchy.
+```
+<prefix>/trip_modifications/#{id}/shape
+```
+
+On `Shape` Messages, the
+[`Shape.shape_id`](https://gtfs.org/realtime/reference/#message-shape)
+field will also have a unique ID generated by Skate, which will be
+referenced by the corresponding Trip Modification's
```
paulswartz, line 81: question: if the shape also has an ID, why not:
- `<prefix>/trip_modifications/#{modification id}`
- `<prefix>/shapes/#{shape id}`

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+[`SelectedTrips.shape_id`](https://gtfs.org/realtime/reference/#message-selectedtrips)
+field.
+
+#### Active Trip Modifications State
+Messages pushed to these topics will have the
+[`retain` property](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/)
```
paulswartz, line 87: suggestion: better references (not to a proprietary site) might be to the MQTT protocol: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718038

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
```diff
@@ -0,0 +1,179 @@
...
+set so that new connections get active Trip Modifications and Shapes published
+to them on connect.
+
+Deactivating a Trip Modification will be done by
+[publishing a zero-byte payload](https://www.hivemq.com/blog/mqtt-essentials-part-8-retained-messages/#heading-how-to-delete-retained-messages-in-mqtt)
```
paulswartz, line 92: suggestion: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718038

Also, this clarifies how servers should handle the zero-byte message:
> A PUBLISH Packet with a RETAIN flag set to 1 and a payload containing zero bytes will be processed as normal by the Server and sent to Clients with a subscription matching the topic name. Additionally any existing retained message with the same topic name MUST be removed and any future subscribers for the topic will not receive a retained message [MQTT-3.3.1-10]. “As normal” means that the RETAIN flag is not set in the message received by existing Clients. A zero byte retained message MUST NOT be stored as a retained message on the Server [MQTT-3.3.1-11].

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: Given this [feedback](https://github.com/mbta/skate/pull/2686#discussion_r1678341041), this does give me pause!

cc: @mbta/skate-developers

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: It's interesting that the field is not present on the actual documentation definition https://gtfs.org/realtime/reference/#message-feedentity

paulswartz: It's there for me?
Screenshot 2024-07-16 at 1 09 09 PM

firestack: Oh I mean there's no `trip_modification` field.
image

firestack: Because it is there on a different page https://gtfs.org/realtime/feed-entities/#trip-modifications

(A side note: This kinda points to why I'd like to attempt to propose to define the GTFS spec in a machine readable format so docs can be autogenerated as much as possible)

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: I'm a little stuck here.

One issue with separating them out, is that you have to predict what id's you want to subscribe to if you want to subscribe to only a single modification and all of it's dependencies.

I know that this is attempting to predict the future, and we don't have a use case for subscribing to individual topics, but following the ["Best Practices" from HiveMQ](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/#heading-embrace-extensibility) I wonder if having a shared ID for identifying all related resources to a trip modification is more extensible?

OTOH, IMO it's _possible_ (but unlikely, given ordering requirements if for example it applied to an inbound and outbound trip) that we'd have a single shape referenced by multiple trip modifications, in which the proposed structure falls apart.

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: https://github.com/mbta/skate/compare/785ac3ebd742a6a64fe2b032145c36d268b88cd8..34831342d74980d12fe8dc47778013db2fc45c12

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: https://github.com/mbta/skate/compare/785ac3ebd742a6a64fe2b032145c36d268b88cd8..34831342d74980d12fe8dc47778013db2fc45c12

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
hannahpurcell: I cannot find the relationship between shapes and trip modifications to decide whether the risk you note in the last paragraph ("... in which the proposed structure falls apart") is likely to be a problem. `TripModifications` contains a list of `SelectedTrips` which have `Shape`s of the existing route/trip, but the `Modifications` field doesn't have any nested references to `Shape`s, it seems. If that's the case, then the logic in paragraph 2 especially makes sense to me

firestack: _I think_ that if two routes had different lists of missed stops but the same detour shape, then we'd need to publish two trip modifications. I don't think we're at the point where we'd do that sort of optimization, but the current structure would not cope with that _as well_.
It's still doable and not that much of an issue if the subscriber subscribes to the wildcards.
```
<prefix>/trip_modifications/+/trip_modification
<prefix>/trip_modifications/+/shape
```

documentation/adr/0004-trip-modifications-mqtt-topic-structure.md
firestack: I've updated this to use a `FeedEntity`. This now means that Skate's implementation of this and this ADR are not in sync anymore.

joshlarson: Do we (should we?) have a ticket to bring the implementation in line with the ADR?

Or do we want to wait until we confirm that the ADR is fully what we want it to be?

firestack: https://app.asana.com/0/1148853526253420/1207900514697511/f

joshlarson (approved):

hannahpurcell (approved):

PR #2687, opened 2024-07-10
feat: send trip modification when finished detour api is called

firestack:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207383192300475/f

Depends on:
- https://github.com/mbta/skate/pull/2693
- https://github.com/mbta/skate/pull/2694

### DIFF
--- a/test/skate_web/controllers/detours_controller_test.exs
+++ b/test/skate_web/controllers/detours_controller_test.exs
@@ -1,4 +1,6 @@
 defmodule SkateWeb.DetoursControllerTest do
+  alias Realtime.Shape
+  alias Realtime.TripModification
   alias Skate.OpenRouteServiceAPI.DirectionsRequest
   use SkateWeb.ConnCase
 
@@ -13,10 +15,15 @@ defmodule SkateWeb.DetoursControllerTest do
       {:ok, build(:ors_directions_json)}
     end)
 
+    stub(Skate.Detours.MockTripModificationPublisher, :publish_modification, fn _, _, _ -> nil end)
+
+    reassign_env(:skate_web, :missed_stops_fn, fn _ -> build(:missed_stops_result) end)
     reassign_env(:skate_web, :shape_with_stops_fn, fn _ -> build(:shape_with_stops) end)
 
     :ok
   end
+
+  setup :verify_on_exit!
 
   describe "unfinished_detour/2" do
     @tag :authenticated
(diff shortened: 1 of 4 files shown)


firestack (commented):
I can't approve my own PR but this looks good!

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -407,6 +414,76 @@ defmodule SkateWeb.DetoursControllerTest do
...
+
+          assert %Shape{
+                   shape_id: ^shape_id
+                 } = shape
+
+          assert opts == [is_draft?: true]
+        end
+      )
+
+      post(conn, ~p"/api/detours/finished_detour",
```
hannahpurcell, line 467: There's no verification after this post that the response indicated success, and I gather that's unneeded?

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -407,6 +414,76 @@ defmodule SkateWeb.DetoursControllerTest do
...
+      reassign_env(:skate_web, :missed_stops_fn, fn _ ->
+        build(:missed_stops_result,
+          missed_stops: [missed_stop]
+        )
+      end)
+
+      missed_stop_id = missed_stop.id
+      representative_trip_id = route_pattern.representative_trip_id
+
+      expect(
```
hannahpurcell, line 438: Oh wow, the syntax of this expect is more complex than what I've seen before!

hannahpurcell (approved):
Great! This is my first time looking at the trip_modification work _at all_. (Was away during the bulk of that work!) Neither of my notes should block merging, just conversation points

test/skate_web/controllers/detours_controller_test.exs
joshlarson: There are other tests that validate that the response was successful. This one is solely concerned with the `TripModification` and `Shape` messages.

firestack: > What is the purpose of the post, then?

The `post` is what triggers the call to the mocked module which asserts that the function is called with the right parameters. Without this post, `mox` helpfully fails the test[^mox-failure] because `publish_modification` is never called, therefore the assertions in the defined in the `expect` are never run.

[^mox-failure]: 
	```
	  1) test finished_detour/2 sends data to TripModificationPublisher (SkateWeb.DetoursControllerTest)
	     test/skate_web/controllers/detours_controller_test.exs:418
	     ** (Mox.VerificationError) error while verifying mocks for #PID<0.551.0>:
	
	       * expected Skate.Detours.MockTripModificationPublisher.publish_modification/3 to be invoked once but it was invoked 0 times
	```

PR #2688, opened 2024-07-11
feat: Add `Shape` as an argument for `TripModificationPublisher.publish_modification/3`

joshlarson:
Asana Ticket: https://app.asana.com/0/1205718271156548/1207790645006778/f

### DIFF
--- a/lib/skate/detours/trip_modification_publisher.ex
+++ b/lib/skate/detours/trip_modification_publisher.ex
@@ -75,6 +75,29 @@ defmodule Skate.Detours.TripModificationPublisher do
     )
   end
 
+  @doc """
+  Publishes a `Realtime.Shape` to the configured `MQTT` server.
+
+  MQTT is optional for Skate, and callers should remember to handle both the
+  `:ok` and `:error` return values
+  """
+  def publish_shape(
+        %Realtime.Shape{} = shape,
+        opts \\ []
+      ) do
+    server = Keyword.get(opts, :server, @default_name)
+
+    GenServer.call(
+      server,
+      {
+        :new_shape,
+        %{
+          shape: shape
+        }
+      }
+    )
+  end
+
   @type t :: %{
           connection: Skate.MqttConnection.on_start() | nil,
           on_connect_subscribers: [pid()]
@@ -142,6 +165,34 @@ defmodule Skate.Detours.TripModificationPublisher do
     }
   end
 
+  @impl GenServer
+  def handle_call(
+        {:new_shape, %{shape: shape}},
+        _from,
+        %__MODULE__{connection: connection} = state
+      )
+      when not is_nil(connection) do
+    id = Ecto.UUID.generate()
+
+    res =
+      Skate.MqttConnection.publish(connection, %EmqttFailover.Message{
+        topic: shape_topic(id),
+        payload:
+          Jason.encode!(%{
+            data: shape
+          }),
+        # Send at least once
+        qos: 1
+      })
+
+    {
+      :reply,
+      {res, id},
+      state
+    }
+  end
+
   def trip_modification_topic(id), do: "#{trip_modifications_topic(id)}/trip_modification"
+  def shape_topic(id), do: "#{trip_modifications_topic(id)}/shape"
   defp trip_modifications_topic(id), do: "trip_modifications/#{id}"
 end
--- a/lib/skate/detours/trip_modification_publisher.ex
+++ b/lib/skate/detours/trip_modification_publisher.ex
@@ -58,6 +58,7 @@ defmodule Skate.Detours.TripModificationPublisher do
   """
   def publish_modification(
         %Realtime.TripModification{} = modification,
+        %Realtime.Shape{} = shape,
         opts \\ []
       ) do
     is_draft? = Keyword.get(opts, :is_draft?, false)
@@ -69,7 +70,8 @@ defmodule Skate.Detours.TripModificationPublisher do
         :new_modification,
         %{
           is_draft?: is_draft?,
-          modification: modification
+          modification: modification,
+          shape: shape
         }
       }
     )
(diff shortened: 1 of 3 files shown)


lib/skate/detours/trip_modification_publisher.ex
```diff
@@ -75,6 +75,29 @@ defmodule Skate.Detours.TripModificationPublisher do
...
+      server,
+      {
+        :new_shape,
+        %{
+          shape: shape
+        }
+      }
+    )
+  end
+
```
firestack, line 100: issue: Given [`shape_id` is a required part of selected trips](https://gtfs.org/realtime/reference/#message-selectedtrips), I don't like publishing these separately and in two calls, it's fine if that's internal to this module, but IMO, I don't think callers should have the option. I get that this would potentially result in publishing the shape multiple times with the same data, but I don't think it matters, and I'd rather solve that at a later point when/if it becomes an issue.

I would rather see `%Realtime.Shape{}` become an argument to `publish_modification`.

Also, right now, you'll have a different ID for the Shape and for the Trip Modification, which also seems incorrect.

joshlarson: > Also, right now, you'll have a different ID for the Shape and for the Trip Modification, which also seems incorrect.

Heh I think https://github.com/mbta/skate/pull/2689 resolves ☝️  (unless I'm misunderstanding your comment here)

---

> I don't like publishing these separately and in two calls

Yeah that's fair. I was of mixed minds about how this should be done, so I appreciate the nudge in that direction! Will update

firestack: > > Also, right now, you'll have a different ID for the Shape and for the Trip Modification, which also seems incorrect.
> 
> Heh I think #2689 resolves ☝️ (unless I'm misunderstanding your comment here)

I'm thinking of the [generated ID in the `handle_call`](https://github.com/mbta/skate/pull/2688/files/055bf3ac369ecba1fe07fe83ac92bafd58701ac7#diff-9573b7338e9a4517a639fc5d4c6aab49d95b8334e56717f41aaddbbdd9729786R175)

joshlarson: Oohhhh.

Question - in the medium-long-term, are those ID's supposed to correspond to, say, the trip modification ID, rather than being generated inside of the MQTT publisher?

firestack: > are those ID's supposed to correspond to, say, the trip modification ID

"Trip Modification ID" is not a concept in the current Trip Modifications spec, so we're in charge of deciding what the ID of a trip modification is for internal communications and synchronization and it will not make it into GTFS-RT.

If eventually there is a spec which add's an ID, I could see the topic corresponding to the ID within the message.

As for whether the Trip Modification Publisher _should_ be responsible for generating ID's, I'm not sure. I like that it can generate it for us, but that poses the need for a new function[^1], that basically does the same thing as `publish_modification` but with an ID instead. 

[^1]: `publish_update`?

joshlarson: I posed my question poorly, but that makes sense. Basically, we'll definitely need a version of this that takes an ID at some point, but we don't necessarily need that right now.

joshlarson:
Now that I've changed the signature of `TripModificationPublisher.publish_modification/2` to `TripModificationPublisher.publish_modification/3`, this needs to wait until https://github.com/mbta/skate/pull/2687 is merged, and make the change in `DetoursController`. That's why I moved this back into Draft.

firestack:
> But if this gets merged then #2687 will need to hand a `Shape` to `TripModificationPublisher`!

*nod*, happy to do it, or have you push it as a commit on top.

test/skate/detours/trip_modification_publisher_test.exs
```diff
@@ -29,15 +30,15 @@ defmodule Skate.Detours.TripModificationPublisherTest do
 
   @tag "Test.Integration": :mqtt
   test "can publish as draft via metadata" do
-    {:ok, reader_pid} = MqttConnection.start_link(["trip_modifications/+/trip_modification"])
+    {:ok, reader_pid} = MqttConnection.start_link(["trip_modifications/+/+"])
```
firestack, line 33: **discussion**/suggestion(non-blocking&low-priority): this is good, but do you think it'd be useful to be explicit about our subscriptions?

I'm on the fence about this but thinking of [MQTT [...]Wildcard subscriptions best practices](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/), 
> ### Avoid Subscribing to Wildcards (`#`)
> Sometimes, it is necessary to subscribe to all messages that are transferred over the broker. For example, to persist all messages into a database. Do not subscribe to all messages on a broker by using an MQTT client and subscribing to a multi-level wildcard. Frequently, the subscribing client is not able to process the load of messages that results from this method (especially if you have a massive throughput)
>
> –– https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/

The heading has the `#` wildcard and not the `+` wildcard, and this seems to be talking about multi level wildcards, but we now have two wildcards, and I _wonder_ if we should consider being more explicit both for readability and as a ~"contract"?

Anyway, none of this is important enough to block over merging.

---

```suggestion
    {:ok, reader_pid} =
      MqttConnection.start_link([
        "trip_modifications/+/trip_modification",
        "trip_modifications/+/shape"
      ])
```

lib/skate/detours/trip_modification_publisher.ex
```diff
@@ -114,26 +116,41 @@ defmodule Skate.Detours.TripModificationPublisher do
...
+                     is_draft?: is_draft?
+                   }
+                 }),
+               # Send at least once
+               qos: 1
+             }) do
+        :ok
+      else
+        res -> res
+      end
```
firestack, line 153: **praise**: this is so clean 🎉 🍰

test/skate/detours/trip_modification_publisher_test.exs
joshlarson: [Makes sense! I didn't notice that it was an array](https://github.com/mbta/skate/pull/2688/commits/88d2873ea5905d595fa486291fcfd3b4a597f8de)

firestack (approved):

PR #2696, opened 2024-07-16
fix(tests): Don't change the input on every useDetour test rerender

joshlarson:
While debugging test issues with [this PR][1], I noticed that useDetour.test.ts was re-invoking originalRouteFactory.build() every time something would cause the hook to rerender, which can cause any hooks that depend on the input to rerender as well (which is not what would happen in real life, since the input isn't randomly changing).

This change updates `useDetourWithFakeRoutePattern` to take in the `originalRoute` as an argument so that the input is only computed once.

[1]: https://github.com/mbta/skate/pull/2638

Blocks:
- https://github.com/mbta/skate/pull/2638

### DIFF
--- a/assets/tests/hooks/useDetour.test.ts
+++ b/assets/tests/hooks/useDetour.test.ts
@@ -14,7 +14,6 @@ import { finishedDetourFactory } from "../factories/finishedDetourFactory"
 import { routeSegmentsFactory } from "../factories/finishedDetourFactory"
 import { originalRouteFactory } from "../factories/originalRouteFactory"
 import { Err, Ok } from "../../src/util/result"
-import { OriginalRoute } from "../../src/models/detour"
 
 jest.mock("../../src/api")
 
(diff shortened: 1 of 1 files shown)


assets/tests/hooks/useDetour.test.ts
```diff
@@ -48,7 +45,8 @@ describe("useDetour", () => {
       return Promise.resolve(Ok(detourShape))
     })
 
-    const { result } = renderHook(useDetourWithFakeRoutePattern)
+    const originalRoute = originalRouteFactory.build()
+    const { result } = renderHook(() => useDetour(originalRoute))
```
firestack, line 49: I think you're looking for this syntax?

```suggestion
    const { result } = renderHook(useDetour, {
      initialProps: originalRouteFactory.build(),
    })
```

joshlarson: Omg!

joshlarson: Addressed [here](https://github.com/mbta/skate/pull/2696/commits/6cc24892ad37299888cb0c28a64b9db5cc7a820b), but I think [this (next commit)](https://github.com/mbta/skate/pull/2696/commits/7fdb8c9931158b351b62daa225379e0db0afcb67) might be even better.

firestack (approved):

PR #2697, opened 2024-07-16
cleanup: Remove detour-route-selection and route-ladder-header-update test groups

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207483512351237/f


assets/src/components/routeLadder.tsx
```diff
@@ -48,39 +42,6 @@ interface Props {
...
-        <OldCloseButton closeButtonType="l_darker" onClick={onClose} />
-      </div>
-
-      <div className="c-route-ladder__route-name">{routeName}</div>
-    </div>
-  )
-}
-
-// TODO: delete old header after roll-out
 export const NewHeader = ({
```
hannahpurcell, line 45: What do you think about renaming `NewHeader` to `Header` in this PR? I think it should be renamed with this work, but understand if you'd want it in a new PR

assets/src/components/routeLadder.tsx
```diff
@@ -130,8 +91,11 @@ export const NewHeader = ({
               trigger="click"
               onShow={() => tagManagerEvent("alert_tooltip_clicked")}
             >
-              <div className="c-route-ladder__alert-icon">
-                <ExclamationTriangleFill aria-label="Route Alert" />
+              <div
+                className="c-route-ladder__alert-icon"
+                aria-label="Route Alert"
```
hannahpurcell, line 96: Could you teach me why the `aria-label` needed moving?

assets/src/components/routeLadder.tsx
hannahpurcell: Oooh additionally, beyond just changing the component name, the "new" label on the classnames will need changing. In hindsight, this should have been done in my work! I should have used the term "oldHeader" instead of going with "newHeader"

joshlarson: I created a [follow-up subtask](https://app.asana.com/0/1205385723132845/1207827639549180/f) for this in order to keep the diff for this smaller and easier to review. I could lump them together if you'd prefer.

assets/src/components/routeLadder.tsx
joshlarson: [This test](https://github.com/mbta/skate/blob/c64f1f1477f2b2b1466bdceec1b1c256c985a19e/assets/tests/components/routeLadder.test.tsx#L656) uses a [selector](https://github.com/mbta/skate/blob/c64f1f1477f2b2b1466bdceec1b1c256c985a19e/assets/tests/testHelpers/selectors/components/routeLadder.ts#L3) that tries to find the route alert icon using `byRole("generic", { name: "Route alert" })`. For reasons I don't fully understand, having the `aria-label` on the bootstrap icon doesn't actually put the aria-label onto the rendered component, but putting it on the wrapping `div` does.

I think this means that the new route ladder header route alert icon isn't doing accessibility quite right.

(FWIW - on `main` right now, the test is testing the old route ladder header, because the test group defaults to not being enabled, so the new route ladder header wasn't being tested for accessibility or an alert icon)

firestack: But also, as I understand it, `aria-label` should be used as a last resort. This wasn't known when this was originally implemented, but a better way to do this would be to add the label text as a DOM node and linking it to the icon with `aria-labelledby` and have the text be visually hidden

firestack: We could attempt to make things smarter, where if `aria-label` or `aria-labelledby` is provided, we could not set the [default `aria-hidden` value](https://github.com/mbta/skate/blob/c64f1f1477f2b2b1466bdceec1b1c256c985a19e/assets/src/helpers/bsIcons.tsx#L226), but I'm not sure it's worth it.

joshlarson: > but a better way to do this would be to add the label text as a visually hidden DOM node and linking it to the icon with `aria-labelledby`

Want to either:
- pair on this?
- ...address as a follow-up?

firestack: > Want to either:
> 
>     * pair on this?
> 
>     * ...address as a follow-up?

I'm good either way! Since it was old code, I'm also fine with how it is for now.

joshlarson: I tried `aria-hidden={false}` and it still said `Unable to find an accessible element with the role "generic" and name "Route Alert"`

I also tried `aria-labelledby`, but that gave this error: 
```
Expected element to have accessible description:
  /Active Detour/i
Received:
```

PR (into this branch for now) below
- https://github.com/mbta/skate/pull/2703

hannahpurcell (approved):
Great, thanks for writing that follow-up sub-task for renaming the header. As for the `aria-label` question, I feel the same as Kayla, that it can be left as-is for now, and addressed in the future if/when needed. But it's above and beyond what you need to cover here!

PR #2699, opened 2024-07-18
feat: persist last detour state to localStorage

firestack:
This persists the state machine to `localStorage` and retrieves the last stored snapshot when the dummy detour page is opened.

> [!note]
> This is reviewable by commit to make it easier to understand all the changes.

----

Asana Ticket: https://app.asana.com/0/1205385723132845/1207529620198256/f

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -78,7 +78,9 @@ export const DiversionPage = ({
     clear,
     reviewDetour,
     editDetour,
-  } = useDetour(originalRoute)
+  } = useDetour({
+    input: originalRoute
+  })
 
   const [textArea, setTextArea] = useState("")
 
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -18,6 +18,8 @@ import { DetourFinishedPanel } from "./detourFinishedPanel"
 import { DetourRouteSelectionPanel } from "./detourRouteSelectionPanel"
 import { Route, RoutePattern } from "../../schedule"
 import RoutesContext from "../../contexts/routesContext"
+import inTestGroup, { TestGroups } from "../../userInTestGroup"
+import { Snapshot } from "xstate"
 
 const displayFieldsFromRouteAndPattern = (
   route: Route,
@@ -37,20 +39,32 @@ const displayFieldsFromRouteAndPattern = (
   return { routeName, routeDirection, routeOrigin, routeDescription, shape }
 }
 
-interface DiversionPageProps {
-  originalRoute: OriginalRoute
+interface DiversionPageFunctions {
   onClose?: () => void
   onConfirmClose?: () => void
   onCancelClose?: () => void
   showConfirmCloseModal: boolean
 }
 
+interface DiversionPageFromInput extends DiversionPageFunctions {
+  originalRoute: OriginalRoute
+}
+
+interface DiversionPageFromSnapshot extends DiversionPageFunctions {
+  /** A _validated_ snapshot from which to initialize {@linkcode createDetourMachine} with */
+  snapshot: Snapshot<unknown>
+}
+
+export type DiversionPageProps =
+  | DiversionPageFromInput
+  | DiversionPageFromSnapshot
+
 export const DiversionPage = ({
-  originalRoute,
   onClose,
   onConfirmClose,
   onCancelClose,
   showConfirmCloseModal,
+  ...useDetourProps
 }: DiversionPageProps) => {
   const {
     snapshot,
@@ -78,9 +92,11 @@ export const DiversionPage = ({
     clear,
     reviewDetour,
     editDetour,
-  } = useDetour({
-    input: originalRoute
-  })
+  } = useDetour(
+    "snapshot" in useDetourProps
+      ? useDetourProps
+      : { input: useDetourProps.originalRoute }
+  )
 
   const [textArea, setTextArea] = useState("")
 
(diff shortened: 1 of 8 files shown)


assets/src/components/dummyDetourPage.tsx
```diff
@@ -16,19 +45,60 @@ export const DummyDetourPage = () => {
...
+            // clear snapshot from storage and reinitialize state machine
+            localStorage.setItem("snapshot", "")
+            reload()
+          }}
         />
+      ) : (
+        <>
+          {
+            /* temp: since this is an internal page, show snapshot error as text
+             */
```
hannahpurcell, line 84: I liked seeing the error on the dummy detour page! Relatedly, was this only applied to the dummyDetourPage as a POC? I envisioned this change [persist last detour state to localstorage] being applied to DetourModal, given that this dummy page is going away.

assets/src/components/dummyDetourPage.tsx
```diff
@@ -16,19 +45,60 @@ export const DummyDetourPage = () => {
...
+                route,
+                routePattern,
+              }}
+              showConfirmCloseModal={false}
+              onClose={() => {}}
+            />
+          )}
+        </>
       )}
+      {}
```
hannahpurcell, line 101: 👀 What's this lil bracket doing on its lonesome?

assets/src/hooks/useDetour.ts
```diff
@@ -6,11 +6,33 @@ import {
...
+
+export type UseDetourInput =
+  | {
+      /** Initial arguments for {@linkcode createDetourMachine} */
+      input: CreateDetourMachineInput
+    }
+  | {
+      /** A _validated_ snapshot from which to initialize {@linkcode createDetourMachine} with */
+      snapshot: Snapshot<unknown>
+    }
```
hannahpurcell, line 20: Oooooo nifty, so useMachine takes either initialization approach

assets/src/components/dummyDetourPage.tsx
```diff
@@ -1,8 +1,37 @@
...
+import { createDetourMachine } from "../models/createDetourMachine"
+import { AnyActorLogic, Snapshot, createActor } from "xstate"
+import { reload } from "../models/browser"
+import { isOk, Err, Ok, Result } from "../util/result"
+
+/**
+ * Checks if the provided object, {@linkcode snapshot}, is a "valid" snapshot
+ * for the provided state machine {@linkcode actorLogic}.
+ */
+const isValidSnapshot = (
```
hannahpurcell, line 15: I'm wondering if this should be a util function in the `useDetour` hook file?

hannahpurcell (commented):
Ooooo nice! The last detour state is nicely persisted: start/end points, waypoints, missed stops, route id. When retrieved from localStorage, the directions and shape don't appear because FetchDetourDirections would need to be run again. That seems like something that should be done here, but what do you think?

Screenshot 2024-07-25 at 2 53 13 PM

assets/src/components/dummyDetourPage.tsx
firestack: No idea! Thanks for catching!

assets/src/components/dummyDetourPage.tsx
firestack: I wasn't sure where to put it because I made it a bit more generic than "just" the `createDetourMachine` (see the `actorLogic` arg)

Given that, do you have any other thoughts on _where_ it should go? I'm open to it going in `useDetour` for now anyway

assets/src/components/dummyDetourPage.tsx
firestack: Good point! I've created a [commit](https://github.com/mbta/skate/pull/2699/commits/606ce272d77a07bc649ab57604cb10b162dc4265) that also exposes the snapshot property on `DetourModal`, IMO, ideally the modal aspect of the detour component would not be it's own component, so that we didn't need to plumb this through, but that's for another day.

firestack:
> Ooooo nice! The last detour state is nicely persisted: start/end points, waypoints, missed stops, route id. When retrieved from localStorage, the directions and shape don't appear because FetchDetourDirections would need to be run again. That seems like something that should be done here, but what do you think?

For posterity: Not sure what happened here, but thinking it may have been an old commit, as this doesn't happen on Dev Green, so it seems like this is resolved.

There _is_ an issue with the ordering of directions, and I've narrowed that down to the way the directions are being keyed and messing with React's caching. I'll open another PR for that.

assets/src/components/detours/diversionPage.tsx
```diff
@@ -38,20 +39,32 @@ const displayFieldsFromRouteAndPattern = (
...
+  | DiversionPageFromInput
+  | DiversionPageFromSnapshot
+
 export const DiversionPage = ({
-  originalRoute,
   onClose,
   onConfirmClose,
   onCancelClose,
   showConfirmCloseModal,
+  ...useDetourInput
```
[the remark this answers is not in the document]
firestack: Done! let me know what you think!

assets/src/components/dummyDetourPage.tsx
```diff
@@ -16,19 +45,60 @@ export const DummyDetourPage = () => {
...
   const route = useRoute(routePattern?.routeId)
 
+  // The state machine won't "reinitialize", if it's inputs change, so only
+  // compute the snapshot from `localStorage` once.
+  const snapshot = useMemo(() => {
+    const snapshot_string = localStorage.getItem("snapshot")
+
+    if (!snapshot_string) {
+      return Err<"No Snapshot Present">("No Snapshot Present")
+    }
```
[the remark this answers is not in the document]
firestack: This doesn't _need_ to be an "error", it was just a bit easier to do it this way.

assets/src/components/dummyDetourPage.tsx
firestack: I decided to move it into it's own file https://github.com/mbta/skate/compare/af7ed658055f2747717e946df684ca5b6f88343b..5c74ac9ed78cf2434377aebbf020ced67c8b92eb

assets/src/components/detours/diversionPage.tsx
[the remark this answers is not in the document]
hannahpurcell: Yeah! Looks good 👍

assets/src/components/dummyDetourPage.tsx
hannahpurcell: Oh yes, that's great. Love that it's generic and reusable. Thanks!

hannahpurcell (approved):
Great! I'm still having a little wonkiness locally, but things are looking great in dev-green. I say let's roll and just double-check things look great in dev once merged

PR #2705, opened 2024-07-24
feat: Created db table for detours

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1207527734305934

Db files created!


priv/repo/migrations/20240723145806_create_detours.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Skate.Repo.Migrations.CreateDetours do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
```
firestack, line 5: question: are these only needed due to creating the index in the same migration as the table?

_If so_, I think we should split these up into two migrations, so that we're not disabling transactions for the table creation.

firestack (commented):
Everything looks good to me so far! But I also kinda helped write this, I wonder if we need @joshlarson to look?

I wanna chat about the migration file but otherwise I'm ready to approve.

priv/repo/migrations/20240723145806_create_detours.exs
hannahpurcell: Oh I seeeeeeeee. No, these are not only needed due to creating the index in the same migration as the table: disabling transactions was [recommended](https://mbtace.sentry.io/issues/5644599926/?project=5303878&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0) for creating indexes concurrently at all.

But! Your comment points out the best answer to [this question](https://mbta.slack.com/archives/C047ZT8ACP5/p1721835486727689?thread_ts=1721834111.986619&cid=C047ZT8ACP5), and that is we should definitely separate these creations (so that we don't disable transactions for the table, as you note)

hannahpurcell: They're split out now ✅

hannahpurcell:
> Everything looks good to me so far! But I also kinda helped write this, I wonder if we need @joshlarson to look?

Yeah I originally assigned Josh, but also a lot of items in this sprint are dependent upon the db foundation, so I'm considering maybe proceeding as-is? I don't have strong feelings about requiring an extra set of eyes when two people pair on a task, given there is an element of intra-review during the pairing

firestack (approved):

PR #2707, opened 2024-07-25
fix: Basic check missed stops exists

hannahpurcell:
Had a new BadMapError when drawing a detour the wrong way. We should not try to send a published detour in that case.

https://mbtace.sentry.io/issues/5644599926/?project=5303878&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0

### DIFF
--- a/lib/skate_web/controllers/detours_controller.ex
+++ b/lib/skate_web/controllers/detours_controller.ex
@@ -81,6 +81,7 @@ defmodule SkateWeb.DetoursController do
                route_segments: route_segments,
                detour_shape: ors_result
              }),
+           # credo:disable-for-next-line
            # TODO: Revisit the idea of `TripModification`s without missed stops
            false <- missed_stops == [],
            {:ok, modification} <-
--- a/lib/skate_web/controllers/detours_controller.ex
+++ b/lib/skate_web/controllers/detours_controller.ex
@@ -81,6 +81,7 @@ defmodule SkateWeb.DetoursController do
                route_segments: route_segments,
                detour_shape: ors_result
              }),
+           false <- length(missed_stops) == 0,
            {:ok, modification} <-
              TripModification.new(%TripModification.Input{
                route_pattern: route_pattern,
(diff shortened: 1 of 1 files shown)


lib/skate_web/controllers/detours_controller.ex
```diff
@@ -81,6 +81,7 @@ defmodule SkateWeb.DetoursController do
                route_segments: route_segments,
                detour_shape: ors_result
              }),
+           false <- missed_stops == [],
```
firestack, line 84: Huh, I'm actually somewhat surprised that a `TripModification` can't have 0 missed stops, because IMO it'd still be useful to at least update the shape, even if there are no missed stops. I think for now this is a good choice though.

What do you think about a todo for this or something? maybe something like this?
```suggestion
           # TODO: Revisit the idea of `TripModification`s without missed stops
           false <- missed_stops == [],
```

firestack (approved):

lib/skate_web/controllers/detours_controller.ex
hannahpurcell: Ohhhhh hmmmm. I didn't consider the need to still preserve the trip mod for turn-by-turn reasons, even if no stops are affected. But that makes a lot of sense. I think I'll even make a ticket for it, just so we have some extra tracking for this weirdness that will need product thinking

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -81,6 +81,8 @@ defmodule SkateWeb.DetoursController do
                route_segments: route_segments,
                detour_shape: ors_result
              }),
+           # TODO: Revisit the idea of `TripModification`s without missed stops
+           false <- missed_stops == [],
```
hannahpurcell, line 85:
```suggestion
           # credo:disable-for-next-line
           # TODO: Revisit the idea of `TripModification`s without missed stops
           false <- missed_stops == [],
```

PR #2713, opened 2024-07-30
feat: Add `Activate` button gated behind detours-list test group

joshlarson:
Screenshot 2024-07-30 at 2 53 56 PM

Asana Ticket: https://app.asana.com/0/1205526445275136/1207935854187629/f
Subtask Of: https://app.asana.com/0/1148853526253420/1205836718156544/f

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -1,3 +1,5 @@
+@use "./color/definitions" as semantic;
+
 .l-diversion-page {
   display: grid;
   // Default layout is linear
@@ -106,4 +108,8 @@
 
 .c-diversion-panel__help-text {
   line-height: 1.5;
+}
+
+.c-diversion-panel__activate-button {
+  background-color: semantic.$missed-stop;
 }
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -1,5 +1,3 @@
-@use "./color/definitions" as semantic;
-
 .l-diversion-page {
   display: grid;
   // Default layout is linear
@@ -108,8 +106,4 @@
 
 .c-diversion-panel__help-text {
   line-height: 1.5;
-}
-
-.c-diversion-panel__activate-button {
-  background-color: semantic.$missed-stop;
 }
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -233,6 +233,11 @@ export const DiversionPage = ({
               onNavigateBack={editDetour}
               detourText={textArea}
               onChangeDetourText={setTextArea}
+              onActivateDetour={
+                inTestGroup(TestGroups.DetoursList)
+                  ? () => alert("Activate!")
+                  : undefined
+              }
             />
           ) : null}
         </div>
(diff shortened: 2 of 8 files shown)


assets/css/_diversion_page.scss
```diff
@@ -107,3 +109,7 @@
 .c-diversion-panel__help-text {
   line-height: 1.5;
 }
+
+.c-diversion-panel__activate-button {
+  background-color: semantic.$missed-stop;
+}
```
firestack, line 115: Do we need bespoke css for this? I'd prefer to use a `variant` on the button itself https://react-bootstrap.github.io/docs/components/buttons

joshlarson: Ahh - I didn't realize I could do [that](https://github.com/mbta/skate/pull/2713/commits/a9c43b6ba83f5d10c7e69f2268f8565334c65619)

(It gives it nicer hover behavior too!)

assets/tests/components/detours/diversionPageActivate.test.tsx
```diff
@@ -0,0 +1,86 @@
+import React from "react"
```
joshlarson, line 1: This is a thing I've been thinking about doing for a while.

`diversionPage.test.tsx` is getting very large, which makes it kind of unwieldy. There's no reason we need to have all of the detour tests live in the same file. I was thinking that we could start splitting out test cases for different angles of the detours feature into different test files.

For instance, all of the test cases that make assertions about which route-shape pieces are visible could be in `diversionPageRouteShapes.test.tsx`, and the test cases that assert when the undo and clear buttons are disabled or enabled might live in `diversionPageUndoClearButtons.test.tsx`.

Exactly how to divvy the use cases into files is a matter of taste, and I don't have strong feelings about it. But I figured that test cases related to saving/activating might be happiest away from everything else.

assets/src/components/detours/diversionPage.tsx
```diff
@@ -233,6 +233,11 @@ export const DiversionPage = ({
               onNavigateBack={editDetour}
               detourText={textArea}
               onChangeDetourText={setTextArea}
+              onActivateDetour={
+                inTestGroup(TestGroups.DetoursList)
+                  ? () => alert("Activate!")
```
firestack, line 238: nit: I can't quite articulate why, so I guess take this as non-blocking, but I'd prefer to not use `alert` if we don't really need it.

hannahpurcell: Does knowing that this is only a temporary thing change your feeling?

firestack: Slightly, I'm fine with it being non-blocking, but I'm inclined to enable the https://eslint.org/docs/latest/rules/no-alert rule

joshlarson: Oh I'm happy to remove it - I used `alert` because it was an easy way to locally test it, but I'm not attached.

joshlarson: [It's gone, btw](https://github.com/mbta/skate/pull/2713/commits/cec1abd6fbdbb76e57c60e4d4b11ed0fbe431cc1)

firestack (approved):

PR #2716, opened 2024-07-31
feat: Suppress "Add detour" dropdown on mobile

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207448960080664/f

### DIFF
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -27,7 +27,6 @@ import { ExclamationTriangleFill, PlusSquare } from "../helpers/bsIcons"
 import { RoutePill } from "./routePill"
 import { Card, CloseButton, Dropdown } from "react-bootstrap"
 import { joinTruthy } from "../helpers/dom"
-import useScreenSize from "../hooks/useScreenSize"
 
 interface Props {
   route: Route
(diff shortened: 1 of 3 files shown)


assets/src/components/routeLadder.tsx
```diff
@@ -213,6 +214,11 @@ const RouteLadder = ({
 
   const displayCrowding = someVehicleHasCrowding(vehiclesAndGhosts, route.id)
 
+  const screenSize = useScreenSize()
```
firestack, line 217: can we do this within CSS instead?

joshlarson: We could - I think that would make it a bit more difficult to test, but perhaps that's okay?

firestack: Yeah I'm not really sure we'd be able to test it until we have CSS in Jest tests, but that's fine with me. We could assert a certain class is on it, I think we just need to change it to `d-none d-sm-block`, or something like that https://getbootstrap.com/docs/5.3/utilities/display/ (see the `Hidden only on xs` text)

joshlarson: @firestack I updated to match ☝️, but I'm not sure that I love the lack of testability. I'll let it sit with me over the evening.

joshlarson: @firestack I've let it sit with me, and while I don't _love_ relying on snapshot tests, they will solve this issue.

Ready for re-review!

firestack (approved):

PR #2718, opened 2024-07-31
tests: Update selectors for "Add detour" dropdown in tests

joshlarson:
Context: https://github.com/mbta/skate/pull/2647#discussion_r1633194691. At the time, https://github.com/mbta/skate/pull/2657 hadn't been merged yet, so this couldn't be done, but that's no longer the case.

Asana Ticket: https://app.asana.com/0/1205718271156548/1207533417539273/f

### DIFF
--- a/assets/tests/components/ladderPage.test.tsx
+++ b/assets/tests/components/ladderPage.test.tsx
@@ -557,11 +557,11 @@ describe("LadderPage", () => {
     expect(container.querySelector(".c-route-ladder__header")).toBeVisible()
 
     expect(
-      container.querySelector(".c-route-ladder__dropdown-button")
+      screen.getByRole("button", { name: "1 Route Options" })
     ).toBeVisible()
 
     await userEvent.click(
-      container.querySelector(".c-route-ladder__dropdown-button")!
+      screen.getByRole("button", { name: "1 Route Options" })
     )
 
     expect(screen.getByRole("button", { name: "Add detour" })).toBeVisible()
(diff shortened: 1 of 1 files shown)


assets/tests/components/ladderPage.test.tsx
```diff
@@ -557,11 +557,11 @@ describe("LadderPage", () => {
     expect(container.querySelector(".c-route-ladder__header")).toBeVisible()
 
     expect(
-      container.querySelector(".c-route-ladder__dropdown-button")
+      screen.getByRole("button", { name: "1 Route Options" })
```
firestack, line 560: note/nit(non-blocking): btw, I think this `expect().toBeVisible()` is redundant? I'd want to check, but I'm fairly certain you can't use `screen.getBy` to get hidden elements without passing the `{ hidden: true }` option, and it'll stop the test if it can't find it. So both asserts that it's visible and clicks it.
```ts
    await userEvent.click(
      screen.getByRole("button", { name: "1 Route Options" })
    )
```

firestack (approved):
looks good, left one note[^1] that I don't think will change anything, but in case you do want to, I disabled automerge. Feel free to ignore.

[^1]: https://github.com/mbta/skate/pull/2718#discussion_r1700119216

assets/tests/components/ladderPage.test.tsx
joshlarson: Yeah I was thinking the same thing, but at the time I didn't want to unilaterally get rid of it.

Now it won't be unilateral!

PR #2714, opened 2024-07-30
feat: Enter the "Activated" state when "Activate Detour" button is clicked

joshlarson:
https://github.com/user-attachments/assets/748b3abf-29cf-4927-b3b1-6680e4598b39

Asana Ticket: https://app.asana.com/0/1205718271156548/1207935854187630/f
Subtask Of: https://app.asana.com/0/1148853526253420/1205836718156544/f

Depends on:
- https://github.com/mbta/skate/pull/2713

PR #2721, opened 2024-08-01
feat: Add ability to deactivate an active detour

joshlarson:
https://github.com/user-attachments/assets/de51f86a-4cdc-461a-8ce6-341bc9e19d12

Asana Ticket: https://app.asana.com/0/1205385723132845/1206554755512191/f

Depends on:
- https://github.com/mbta/skate/pull/2713
- https://github.com/mbta/skate/pull/2714
- https://github.com/mbta/skate/pull/2715

### DIFF
--- a/assets/src/components/detours/activeDetourPanel.tsx
+++ b/assets/src/components/detours/activeDetourPanel.tsx
@@ -1,12 +1,27 @@
 import React from "react"
 import { Panel } from "./diversionPage"
+import { Button } from "react-bootstrap"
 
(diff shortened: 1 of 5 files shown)


joshlarson:
I'm open to splitting this up into smaller PR's, like I did for activating. But also this doesn't seem overly large, and since it traces a similar path as activating, combining it into one PR made sense to me.

Lemme know what you all think!

hannahpurcell (approved):
Looks great! It does chart the same path as the activate work, so indeed was a smooth review 👍

PR #2723, opened 2024-08-01
fix(warnings): Fix warnings associated with Elixir 1.17 upgrade

joshlarson:
Asana Ticket: https://app.asana.com/0/1152340551558956/1207955284475165/f


mix.exs
```diff
@@ -87,7 +87,7 @@ defmodule Skate.MixProject do
       {:server_sent_event_stage, "~> 1.1.0"},
       {:sobelow, "~> 0.13", only: [:dev, :test], runtime: false},
       {:ssl_verify_fun, "~> 1.1"},
-      {:stream_data, "~> 0.6.0", only: :test},
+      {:stream_data, "~> 1.1.1", only: :test},
```
joshlarson, line 90: The ancient version of `stream_data` we were using was calling a deprecated version of `register_test`

(`register_test/4` is deprecated - `register_test/6` is the new hotness)

hannahpurcell (approved):
Solid, thanks!!

PR #2724, opened 2024-08-02
feat: Basic react-bootstrap table for detour list

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1207890779629110

We decide to go ahead and merge this table-in-progress, even though the styling is not where we want it to be, yet. Let's make it hi-fi and add it to storybook once those designs are available, and after we've turned on the "reboot" bootstrap styles


assets/src/components/detourListPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <Table hover className="c-detour-table">
+        <thead>
+          <tr>
+            <th>Route</th>
+            {["desktop", "tablet"].includes(screenSize) && (
+              <>
+                <th>Starting Intersection</th>
+                <th>On detour since</th>
+              </>
+            )}
```
firestack, line 89: I'm fairly certain we can do this entirely within CSS, is there a better reason to do it like this?

assets/src/components/detourListPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+      activeSince: 1721361948,
+    },
+  ]
+
+  const epochNowInSeconds = useCurrentTimeSeconds()
+  const screenSize = useScreenSize()
+
+  return (
+    <>
+      <Table hover className="c-detour-table">
```
joshlarson, line 80: How would you feel about extracting this to a separate component, like `<DetoursTable data={...}/>` or something like that?

I see two benefits from that:
1. Storybook
2. As we iterate on styles and bootstraps and mobile layouts etc, this page doesn't have to change at all - only the component (whose changes would then show up in Storybook)

joshlarson (changes requested):
I'm interested in at least discussing this, but for now I feel pretty strongly about it https://github.com/mbta/skate/pull/2724/files#r1702166477

assets/src/components/detourListPage.tsx
firestack: Fwiw, I think it's worth waiting on making this it's own component.

assets/src/components/detourListPage.tsx
hannahpurcell: Ah yes yes! I noticed this in your review of [Josh's PR](https://github.com/mbta/skate/pull/2716). 

The problem is that our [Skate-set breakpoints](https://github.com/mbta/skate/blob/a9a862e1a7dc4e51f1d4861978dbd8d22eab148f/assets/css/_skate_ui.scss#L1) are different from the [bootstrap breakpoints](https://getbootstrap.com/docs/5.0/layout/breakpoints/). This might need to be a deeper conversation with design to determine whether our chosen breakpoints are preferred to Bootstrap's, and whether we should override the Bootstrap defaults. For now, it seems Skate primarily uses our own custom breakpoints instead of the Bootstrap ones, so I'd lean towards consistency for now?

firestack: you can still do this within CSS with our own breakpoints, which addresses the consistency issue.

assets/src/components/detourListPage.tsx
hannahpurcell: I did like moving to its own component, I think that's fine and went ahead and did that. (Kayla, feel free to still weigh in on why it's worth waiting.) But didn't add to Storybook yet, because I see that as "visual change control" and think that can/should come in once the table is properly formatted. Thoughts?

firestack: > Kayla, feel free to still weigh in on why it's worth waiting

I think we tend to "over-componentize" and it prevents us from using composition to our advantage and forces us into managing props and prop drilling more often. I think react makes it seem like you're supposed to make a component out of everything but I think acting too soon leaves you with bad abstractions and bad data structures.

That said, now that it's done, I wouldn't revert.

hannahpurcell: Hmm I can see that. In this case, this table will be repeated 3 times on the page with slight variation, so I saw it as an especially apt candidate for a component. (Initially, I wasn't sure if I should do it now or later once we populate the tables, but figured I might as well after Josh mentioned it, too)

joshlarson: I agree with everything both of you said, and still think that this in particular deserves to be a component.

Relatedly...
> acting too soon leaves you with bad abstractions and bad data structures.

I think we have a few bad abstractions and bad data structures lying around (at least in our detours code), and I definitely support refactoring them to be less bad.

joshlarson: As far as Storybook is concerned, I think it's still valuable to have a source of truth of *What Is*, and I actually think that it's even more valuable right now because of the contrast between *What Is* and *What Should Be*.

Plus, (I dunno maybe this is just me) it's extremely satisfying to see the Chromatic diff when we improve the design of a component.

Happy to let Storybook be a follow-up though. No need for it to be in this PR.

assets/src/components/detourListPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+      name: "Rindge Ave",
+      intersection: "Pearl Street & Clearwell Ave",
+      activeSince: 1721361948,
+    },
+  ]
+
+  return (
+    <>
+      <DetoursTable className="active" data={fakeData} />
+      <DetoursTable className="past" data={fakeData} />
```
joshlarson, line 82:
```suggestion
      <DetoursTable detourStatus="active" data={fakeData} />
      <DetoursTable detourStatus="past" data={fakeData} />
```

I like this abstraction, but calling it `className` confused me, since `className` typically translates directly to CSS classes, rather than being processed (by in this case, being appended to a thing).

Calling it `detourStatus` (or something else - I'm not attached to that name in particular) would make it clearer what's actually happening here.

hannahpurcell: I think `className` is more accurate, since it serves no purpose beyond being part of the table's className. (And FWIW I have seen it done this way, though that definitely doesn't mean it's the proper way!)

joshlarson: At the implementation level, it is accurate that the only effect (for now) is to update the CSS, which means that `classNameSuffix` would be an accurate name.

But from the caller's perspective, it still doesn't feel like the thing I care about. When I'm constructor a `<DetoursTable />`, I don't really care about the CSS classes applied to it; I care about whether the table is full of active detours, or past detours (or eventually draft detours), so I'd rather just explicitly specify that.

That's part of my reasoning. The other part is that when I saw `className=`, I looked in the CSS for classes called `active` or `past` and couldn't find them.

hannahpurcell: I'll remove the prop for now! Finessing what we care about with these props feels too early for this PR, so I'll leave it for the table implementor to do.

assets/css/_nav.scss
```diff
@@ -43,4 +43,5 @@
 .l-nav__app-content {
   grid-area: content;
   position: relative;
+  overflow: scroll;
```
firestack, line 46: suggestion/question: could this be put on Detours List Page instead? [`overflow-auto`](https://getbootstrap.com/docs/5.3/utilities/overflow/) should scroll only if it overflows

assets/css/detours/_detour_table.scss
```diff
@@ -0,0 +1,21 @@
...
+
+.c-detours-table__route-info-text {
+  display: flex;
+  flex-direction: column;
+  margin-left: 10px;
+}
+
+.c-detours-table__route-info-text--bold {
+  font-weight: var(--font-weight-bold);
+}
```
firestack, line 21: nit(non-blocking): could we "bootstrap-ify" this and use  `fw-bold` instead?

assets/src/components/detourListPage.tsx
```diff
@@ -0,0 +1,84 @@
+import React from "react"
+import { DetoursTable } from "./detoursTable"
+
+export interface Detour {
+  number: number
```
firestack, line 5: I think this needs to be a string? https://www.mbta.com/schedules/627/line

firestack (changes requested):
One issue, but otherwise 👍🏻

assets/src/components/detourListPage.tsx
hannahpurcell: Oh definitely, that was a careless mistake, especially given the routePill takes the id in the format of a string! And the SLs! :shakes-head:

assets/css/_nav.scss
hannahpurcell: I need to get more comfortable with these utilities! Thanks!

firestack (approved):
I am wondering if there's any tests we should/could add, but otherwise this looks good

hannahpurcell:
> I am wondering if there's any tests we should/could add, but otherwise this looks good

Yep, you're right that there should be _some_ test here. I hesitated on merging to think about it a bit, and so I just stubbed out the existing detourListPage as a jest snapshot test. Then at least the test will break when we do the hi-fi work, and the test suite can be built out further depending on how we adjust the table at that point.

PR #2729, opened 2024-08-05
feat: Save detour creation snapshot to db

hannahpurcell:
Still need to write tests and format, but wanted to gauge this approach asap

- [x] Tests!


assets/src/hooks/useDetour.ts
```diff
@@ -28,8 +28,12 @@ export const useDetour = (input: UseDetourInput) => {
   // Record snapshots when changed
   useEffect(() => {
     const snapshotSubscription = actorRef.subscribe(() => {
-      const serializedSnapshot = JSON.stringify(actorRef.getPersistedSnapshot())
+      const persistedSnapshot = actorRef.getPersistedSnapshot()
+      const serializedSnapshot = JSON.stringify(persistedSnapshot)
       localStorage.setItem("snapshot", serializedSnapshot)
+      if (persistedSnapshot.context.uuid) {
+        putDetourUpdate(persistedSnapshot.context)
```
firestack, line 35: issue: I think the entire `persistedSnapshot` object needs to be saved? not just the `context`

lib/skate/detours/db/detour.ex
```diff
@@ -7,6 +7,7 @@ defmodule Skate.Detours.Db.Detour do
   alias Skate.Settings.Db.User
 
   schema "detours" do
+    field :uuid, :string
```
firestack, line 10: question: would it be possible to use the table `id` instead of creating a new column? I think we should be able to configure the ID to be a UUID, and I think we should have the database create the ID and not the frontend.

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -4,10 +4,25 @@ defmodule SkateWeb.DetoursController do
...
   use SkateWeb, :controller
 
+  alias Skate.Detours.Detours
   alias Skate.Detours.MissedStops
   alias Skate.Detours.RouteSegments
+  alias SkateWeb.AuthManager
   alias Util.Location
 
+  @spec update_snapshot(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def update_snapshot(conn, %{"snapshot" => snapshot} = _params) do
```
firestack, line 14: nit: I'd name this `upsert_snapshot` since it inserts or creates.

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -4,10 +4,25 @@ defmodule SkateWeb.DetoursController do
...
+    uuid = Map.get(snapshot, "uuid")
+
+    %{id: user_id} = AuthManager.Plug.current_resource(conn)
+
+    Detours.update_or_create_detour_for_user(user_id, uuid, %{
+      state: snapshot
+    })
+
+    send_resp(conn, 200, "")
```
firestack, line 23: thought: along the lines of my [comment](https://github.com/mbta/skate/pull/2729/files#r1704411158), I would suggest that if `uuid` is not in the `context`, then we insert and return the UUID back to the frontend for the state machine to save. What do you think

lib/skate/detours/detours.ex
```diff
@@ -55,6 +56,31 @@ defmodule Skate.Detours.Detours do
...
+
+  @doc """
+  Update or create a detour given a user id & detour id.
+  """
+  def update_or_create_detour_for_user(user_id, uuid, attrs \\ %{}) do
+    case get_detour_by_uuid!(uuid) do
+      # If there isn't a detour with this uuid, create one.
+      nil -> create_detour_for_user(user_id, uuid, attrs)
+      existing_detour -> update_detour(existing_detour, attrs)
+    end
```
firestack, line 81: can we use [`upserts`](https://hexdocs.pm/ecto/constraints-and-upserts.html#upserts) instead of creating this manually?

lib/skate_web/controllers/detours_controller.ex
firestack, line 23:
```suggestion
    send_resp(conn, :ok, "")
```

firestack: I think we should do something like this though
```suggestion
    if is_nil(uuid) do
      send_resp(conn, :created, new_uuid)
    else
      send_resp(conn, :ok, "")
    end
```

firestack: And we'd get `new_uuid` from the database upsert

assets/src/hooks/useDetour.ts
```diff
@@ -28,8 +28,12 @@ export const useDetour = (input: UseDetourInput) => {
   // Record snapshots when changed
   useEffect(() => {
     const snapshotSubscription = actorRef.subscribe(() => {
-      const serializedSnapshot = JSON.stringify(actorRef.getPersistedSnapshot())
+      const persistedSnapshot = actorRef.getPersistedSnapshot()
+      const serializedSnapshot = JSON.stringify(persistedSnapshot)
       localStorage.setItem("snapshot", serializedSnapshot)
+      if (persistedSnapshot.context.uuid) {
```
firestack, line 34: suggestion: I think what this conditional is getting at is that we don't want to save the detour unless we're in certain states. This code is currently using `uuid` as a sentinel value/proxy for "should we save the detour?"

What if instead, we use a tag, maybe called `no-save`, and do `snapshot.hasTag("no-save")` as the guard for if we call `putDetourUpdate`? 

I'll admit I played around with that idea when I built the localStorage stuff, so I'd be happy to pair on it if that'd be helpful.

assets/src/hooks/useDetour.ts
```diff
@@ -28,8 +28,12 @@ export const useDetour = (input: UseDetourInput) => {
...
+        putDetourUpdate(persistedSnapshot.context)
+      }
```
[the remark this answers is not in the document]
hannahpurcell: I might as well do it here 👍

lib/skate/detours/db/detour.ex
hannahpurcell: Oh nvm! I see your subsequent comment explains this exact process. Yeah, makes sense to me

firestack: that sounds correct.

lib/skate/detours/detours.ex
hannahpurcell: Hmmm I'm not sure. After removing the uuid from the db and using the `id` column as the unique identifier, the uuid coming from the frontend is either null (indicating the detour hasn't yet been added to the db) or it's a reference to the id of an existing detour. 

If the uuid was always non-null from the frontend, I see I could use upsert like in [Skate.Settings.User line 51](https://github.com/mbta/skate/blob/hp/test-branch-saving-to-db/lib/skate/settings/user.ex#L51).

lib/skate_web/controllers/detours_controller.ex
hannahpurcell: Did a variation of this: always return the UUID, but only sometimes save it in the state machine. In hindsight, unsure if I took the right approach?

hannahpurcell: Switched to using `json(conn)`

assets/src/hooks/useDetour.ts
hannahpurcell: Ohhh I forgot to apply this. Given my other updates, would you still recommend using a tag?

assets/src/api.ts
```diff
@@ -441,6 +441,22 @@ export const putRouteTabs = (routeTabs: RouteTab[]): Promise<Response> =>
...
+    defaultResult: null,
+    fetchArgs: {
+      method: "PUT",
+      headers: {
+        "Content-Type": "application/json",
+        "x-csrf-token": getCsrfToken(),
+      },
+      body: JSON.stringify({ snapshot: snapshot }),
+    },
+  })
```
firestack, line 458: suggestion: Can we use `apiCallResult` instead? This would make the error state a lot more explicit.

I can post a code example later if needed.

assets/src/hooks/useDetour.ts
```diff
@@ -27,15 +27,21 @@ export const useDetour = (input: UseDetourInput) => {
...
-    const snapshotSubscription = actorRef.subscribe(() => {
-      const serializedSnapshot = JSON.stringify(actorRef.getPersistedSnapshot())
+    const snapshotSubscription = actorRef.subscribe((snap) => {
+      const persistedSnapshot = actorRef.getPersistedSnapshot()
+      const serializedSnapshot = JSON.stringify(persistedSnapshot)
       localStorage.setItem("snapshot", serializedSnapshot)
+      putDetourUpdate(persistedSnapshot).then((uuid: string | null) => {
+        if (uuid && snap.matches({ UUID: "Unset" })) {
+          send({ type: "detour.set.uuid", uuid })
+        }
```
firestack, line 37: I don't think you need this. _maybe_ the UUID,
```suggestion
        if (uuid) {
          send({ type: "detour.set.uuid", uuid })
        }
```

but if we should _always_ get a UUID, then it should always be possible to send it right?
```suggestion
    send({ type: "detour.set.uuid", uuid })
```

[(I just realized) Maybe we should change out the type, because I think `null` is representing an error state here?](https://github.com/mbta/skate/pull/2729/files#r1711449321)

```suggestion
        if (isOk(uuid)) {
          send({ type: "detour.set.uuid", uuid: uuid.ok })
        }
```

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -4,10 +4,28 @@ defmodule SkateWeb.DetoursController do
...
+    uuid = Map.get(context, "uuid")
+
+    %{id: user_id} = AuthManager.Plug.current_resource(conn)
+
+    {:ok, returned_detour} =
+      Detours.update_or_create_detour_for_user(user_id, uuid, %{
+        state: snapshot
+      })
+
+    returned_uuid = Map.get(returned_detour, :id)
```
firestack, line 24: suggestion: would this work?
```suggestion
    {:ok, %Skate.Detours.Db.Detour{id: returned_uuid} = returned_detour} =
      Detours.update_or_create_detour_for_user(user_id, uuid, %{
        state: snapshot
      })
```

assets/src/api.ts
firestack:
```suggestion
  apiCallResult(
    `/api/detours/update_snapshot`,
    object({ data: string() }),
    any(),
    {
      method: "PUT",
      headers: {
        "Content-Type": "application/json",
        "x-csrf-token": getCsrfToken(),
      },
      body: JSON.stringify({ snapshot: snapshot }),
    }
  )
```

lib/skate/detours/detours.ex
```diff
@@ -55,6 +56,39 @@ defmodule Skate.Detours.Detours do
...
+      nil ->
+        create_detour_for_user(user_id, attrs)
+
+      _ ->
+        Repo.insert(
+          Detour.changeset(%Detour{author: user, id: uuid}, attrs),
+          returning: true,
+          conflict_target: [:id],
+          on_conflict: {:replace, [:state, :updated_at]}
+        )
```
firestack, line 88: issue: is `author` respected here? It seems like a user could update a detour that doesn't belong to them by injecting whatever ID they want.

I'm looking at providing a suggestion for all of this though, so you don't need to jump on this issue immediately, just wanted to bring it to your attention, maybe we write tests for this.

assets/src/models/createDetourMachine.ts
```diff
@@ -59,6 +60,7 @@ export const createDetourMachine = setup({
       | { type: "detour.edit.place-waypoint-on-route"; location: ShapePoint }
       | { type: "detour.edit.place-waypoint"; location: ShapePoint }
       | { type: "detour.edit.undo" }
+      | { type: "detour.set.uuid"; uuid: string }
```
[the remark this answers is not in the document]
hannahpurcell: Oh yep, totally!! 🧠

assets/src/api.ts
hannahpurcell: Gotcha. (I was following in the predominant pattern of the rest of the api functions.)

So, I see `apiCallResult` is a recent add, but I'm not sure what new-and-different thing it does as compared to the other call methods? I'm thinking `apiCall`, `checkedApiCall`, and `apiCallResult` also need comments, especially if we're trying to shift towards a newer function over time

lib/skate/detours/detours.ex
firestack: I tested with this and found that a user can update a detour that does not belong to them

```elixir
    test "update_or_create_detour_for_user/3 should not update a detour made by another user" do
      detour = detour_fixture()
      other_user = insert(:user)

      {:ok, new_detour} = Detours.update_or_create_detour_for_user(
          other_user.id,
          detour.id,
          %{state: %{a: 1}}
        )

      assert detour.id == new_detour.id
      assert detour.author_id == new_detour.author_id
      assert detour.state == new_detour.state
    end
```

assets/src/hooks/useDetour.ts
hannahpurcell: I think this leads to a loop? The snapshotSubscription will trigger whenever anything with the snapshot changes, then will try to update the detour, and the detour update will try to update the snapshot, triggering the subscription again. My goal with the `snap.matches` was to only permit the initial set ID change

lib/skate/detours/detours.ex
firestack: Hmm yeah that's fair! but it does bring up the issue of permissions, roles and access control, because as is, this still allows someone who doesn't have the `detours-pilot` role to update a detour, if they knew the endpoint. I don't think this matters for our threat model, but maybe we need to figure out our threat model.

assets/src/hooks/useDetour.ts
hannahpurcell: We discussed on a slack huddle and decided yeah, the `matches` is annoyingly needed. Refactored anyway because of different issues

lib/skate_web/controllers/detours_controller.ex
hannahpurcell: You bet!

assets/src/hooks/useDetour.ts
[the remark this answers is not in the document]
hannahpurcell: Actually! I'd rather not do that transition, because I'd also need to make sure that we read from the database instead of localStorage, and that just feels like more than this PR should do.

assets/src/hooks/useDetour.ts
hannahpurcell: Gosh did I try to make this work. But since we moved the `putDetourUpdate` to the state machine, there seems to be NO way of incorporating `hasTag` with a guard, and I'm so so frustrated about it! My "solution" ends up being pretty janky to get the same effect 😒

firestack: dang that sucks, there's even [unofficial suggestions that there _should_ be a `hasTag`](https://github.com/statelyai/xstate/discussions/4347#discussioncomment-7404932) guard 😭

hannahpurcell: I knooooowwwww I saw that toooo ugh

assets/src/hooks/useDetour.ts
```diff
@@ -27,9 +27,29 @@ export const useDetour = (input: UseDetourInput) => {
...
+      const persistedSnapshot = actorRef.getPersistedSnapshot()
+      const serializedSnapshot = JSON.stringify(persistedSnapshot)
       localStorage.setItem("snapshot", serializedSnapshot)
+
+      // check for no-save tag before
+      if (snap.hasTag("no-save")) {
+        return
+      }
+
+      actorRef.getSnapshot().can({ type: "detour.save.begin-save" }) &&
```
joshlarson, line 40: What happens if `.send` is called even if `.can` returns false? I assumed it wouldn't do anything, but I may have made that up.

hannahpurcell: 🤔 The `can` clause serves as a guard here, so the `send` cannot execute if the `can` returns false, right? Now I'm second guessing everything

assets/tests/components/detours/diversionPage.saveDetour.test.tsx
```diff
@@ -0,0 +1,140 @@
+import { jest, describe, test, expect, beforeEach } from "@jest/globals"
```
joshlarson, line 1: Nice!

assets/src/hooks/useDetour.ts
joshlarson: Well, I tried removing the `.can` check and it caused the tests to hang, so I think my guess about how `.send` works was wrong 😅

firestack: if `.send` is called, even though it _can't_ be handled more than once(with our current machine), it seems to cause the `subscribe` callback to be called again with a new snapshot, which gets into a loop where it's constantly triggering `send`->`subscribe`, unfortunately you'd need to guard the `send` some other way because of that, but it is a little surprising that `.send` doesn't do a `.can` internally and prevent `subscribe` from being invoked (though maybe it makes sense because `subscribe` does not describe if you're subscribing to anything _other_ than just snapshot updates and what causes a snapshot update does not seem to be explicitly defined either.)

assets/src/models/createDetourMachine.ts
```diff
@@ -61,7 +62,19 @@ export const createDetourMachine = setup({
...
+      | { type: "detour.save.set-uuid"; uuid: number },
+
+    // We're making an assumption that we'll never want to save detour edits to the database when in particular stages
+    // of detour drafting:
+    // -- when starting a detour, before any user input
+    // -- when the route id / route pattern is getting selected
+    // -- right after the route pattern is finalized, before any waypoints are added
+    // That leads to the following interface: if the user begins drafting a detour, adds waypoints, and then changes the route,
+    // the database will reflect the old route and old waypoints up until the point where a new waypoint is added.
+    // If that UX assumption isn't the right one, we can iterate in the future!
```
firestack, line 76: **praise**: Love this comment!

assets/tests/components/detours/diversionPage.saveDetour.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    await userEvent.click(
+      screen.getByRole("button", { name: "Start drawing detour" })
+    )
+
+    await waitFor(() => {
+      expect(putDetourUpdate).toHaveBeenCalledTimes(0)
+    })
+  })
+
+  // We made an assumption that we'll never want to save detour edits in response to changing
```
firestack, line 111:
```suggestion
  // We made an assumption that we'll never want to save detour edits in response to changing route/variant
```

assets/tests/components/detours/diversionPage.saveDetour.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    act(() => {
+      fireEvent.click(originalRouteShape.get(container))
+    })
+
+    await waitFor(() => {
+      expect(putDetourUpdate).toHaveBeenCalledTimes(3)
+    })
+  })
+
+  // We made an assumption that we'll never want to save detour edits in response to changing
```
firestack, line 79:
```suggestion
  // We made an assumption that we'll never want to save detour edits in response to changing route/variant
```

assets/tests/api.test.ts
```diff
@@ -1122,3 +1125,33 @@ describe("putRouteTabs", () => {
...
+  const actor = createActor(testToggleMachine).start()
+  const persistedSnapshot = actor.getPersistedSnapshot()
+
+  test("uses PUT and CSRF token", () => {
+    mockFetch(200, { data: 1 })
+    putDetourUpdate(persistedSnapshot)
+    expect(window.fetch).toHaveBeenCalledTimes(1)
+    const args = jest.mocked(window.fetch).mock.calls[0][1]
+    expect(args!.method).toEqual("PUT")
+    expect(args!.headers).toHaveProperty("x-csrf-token")
```
firestack, line 1155: question/suggestion(non-blocking): could this be written using jest asymmetric matchers?

This may be the wrong code, but something like this?
```ts
expect(jest.mocked(window.fetch))
	.toHaveBeenCalledWith([
		expect.anything(),
		expect.objectContaining({
			method: "PUT",
			headers: expect.objectContaining({
				"x-csrf-token": expect.any(String) // or expect.anything(), I dunno the right syntax for specifying the type
			})
		})
	])
```

firestack: https://jestjs.io/docs/expect#asymmetric-matchers

firestack (commented):
Looks good! I'll leave the review up to @joshlarson and come and look at this later once I'm back online and approve if it's still up

assets/src/hooks/useDetour.ts
hannahpurcell: Ohhh oh oh, I understand your initial question now, after Kayla's reply. Yeah, we did that exact thing and tried just `send`ing, but ran into issues

firestack (approved):
I'd have to leave test structure/organization and methodology review to @joshlarson but LGTM!

assets/src/api.ts
firestack: Documentation! https://github.com/mbta/skate/pull/2747

hannahpurcell:
Merging now, but would take later feedback for iteration / improvement!

PR #2730, opened 2024-08-05
fix: Run database migrations before starting anything else

joshlarson:
This allows us to put migrations and code that relies on those migrations into the same commits.

If the code relies on changes made in a migration, and the migration runs after the code starts, then there's a potential race condition where that code runs before the migrations have run, which would cause an error. This eliminates that possibility.

### DIFF
--- a/lib/skate/application.ex
+++ b/lib/skate/application.ex
@@ -37,9 +37,9 @@ defmodule Skate.Application do
           []
         end ++
         [
+          Skate.Migrate,
           {Phoenix.PubSub, name: Skate.PubSub},
           SkateWeb.Endpoint,
-          Skate.Migrate,
           {Oban, Application.fetch_env!(:skate, Oban)},
           {Skate.Detours.TripModificationPublisher,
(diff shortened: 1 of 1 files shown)


lib/skate/application.ex
```diff
@@ -37,9 +37,9 @@ defmodule Skate.Application do
           []
         end ++
         [
+          Skate.Migrate,
```
firestack, line 40: I can see that _maybe_ this needs to be done before the Endpoint is started, but does it need to come before the PubSub?

Also, _when_ would code in the Endpoint conflict with the Migrations? I assume this would never happen in a deployment due to our health checks?

joshlarson: Basically I was thinking that code in the Endpoint would refer to some column in the database that's introduced in a migration. That'll cause an error in the 

Do the healthchecks fail if there are any un-run migrations?

My reason for this was remembering that a while ago, when I added a migration, I was asked to pull the migration out to a pre-PR that got deployed before the main PR was merged. Maybe that wasn't necessary though?

PR #2725, opened 2024-08-02
cleanup: Remove `keycloak-sso` test group and Cognito

joshlarson:
Asana Ticket: https://app.asana.com/0/0/1207261350055998/f

PR #2732, opened 2024-08-05
cleanup: Remove Cognito-related environment variables

joshlarson:
Depends on:
- https://github.com/mbta/skate/pull/2725

Asana Ticket: https://app.asana.com/0/1205385723132845/1207967466069593/f


config/config.exs
```diff
@@ -142,9 +142,8 @@ config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
-# "code" is the secret value returned by AWS to /auth/cognito/callback
 log_filter_params =
-  ~w(password code token guardian_default_claims guardian_default_resource guardian_default_token)
+  ~w(password token guardian_default_claims guardian_default_resource guardian_default_token)
```
joshlarson, line 146: I'm unsure about this - perhaps we should keep `code` filtered?

firestack: I _think_ this is okay? I can't see a variable in the keycloak response that's simlar to the code variable in the cognito response. It doesn't really hurt to keep though it seems?

I'm good either way I think

firestack (approved):

PR #2734, opened 2024-08-06
cleanup: move $metadata to front of $level

firestack:
Asana Ticket: https://app.asana.com/0/0/1207978932098578/f

Depends On:
- #2733

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -139,7 +139,7 @@ config :skate, Oban,
 
 # Configures Elixir's Logger
 config :logger, :console,
-  format: "$time $metadata[$level] $message\n",
+  format: "$time [$level] $metadata$message\n",
   metadata: [:mfa, :request_id]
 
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -139,7 +139,7 @@ config :skate, Oban,
 
 # Configures Elixir's Logger
 config :logger, :console,
-  format: "$time $metadata[$level] $message\n",
+  format: "$time [$level] $metadata$message\n",
```
joshlarson, line 142: Curiosity: How come we don't need a space between `$metadata` and `$message`?

firestack: The `$metadata` variable has a space appended to the end of it apparently. Which is also why a space wasn't needed between `$metadata[$level]`

> `$metadata` - user controlled data presented in `"key=val key2=val2 "` format
> – https://hexdocs.pm/logger/1.17.2/Logger.Formatter.html#module-formatting

PR #2735, opened 2024-08-06
feat: Add validation to ensure no whitespace in test group names

joshlarson:
This is the first part of a multi-part (I think two-part, but who knows) effort to change test-group URL's to be `/test_groups/<name>` instead of `/test_groups/<database_id>`, so for instance, `https://skate.mbta.com/test_groups/detours-pilot` instead of `https://skate.mbta.com/test_groups/10`.

This will only work (nicely) if test group names aren't allowed to have spaces, so I figured it could be a good idea to enforce that now.

No Asana Ticket.


joshlarson:
Decided this isn't necessary for now.

PR #2736, opened 2024-08-07
fix: Allow Hastus import to work when a block has as-directeds and regular trips

joshlarson:
Asana Ticket: https://app.asana.com/0/1205718271156548/1208002136498399/f

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -839,7 +839,10 @@ const isBreak = (activity: Piece | Break): activity is Break =>
   Object.prototype.hasOwnProperty.call(activity, "breakType")
 
 const isTrip = (trip: Trip | AsDirected): trip is Trip =>
-  Object.prototype.hasOwnProperty.call(trip, "id")
+  !Object.prototype.hasOwnProperty.call(trip, "kind")
+
+const isAsDirected = (trip: Trip | AsDirected): trip is AsDirected =>
+  Object.prototype.hasOwnProperty.call(trip, "kind")
 
 const isDeadhead = (trip: Trip | AsDirected): boolean =>
   isTrip(trip) && trip.routeId == null
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -544,27 +544,25 @@ const TripSchedule = ({
           activeStatus={layoverActiveStatus}
         />
       ) : null}
-      {isTrip(trip) ? (
-        isDeadhead(trip) ? (
-          <DeadheadTrip
-            trip={trip}
-            status={sequenceToNonRevenueStatus(sequence)}
-            timeBasedStyle={deadheadTimeBasedStyle}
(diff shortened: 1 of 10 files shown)


joshlarson:
The [error](https://github.com/mbta/skate/pull/2736#issuecomment-2276106543) mentioned above is resolved by marking `%AsDirected{}`'s as [not being revenue trips](https://github.com/mbta/skate/pull/2736/files#diff-6efd3069f4e6ec418138f852efb794bdceb49ef5b27e06785e9f0221c443e706R180) for the purpose of finding the next revenue trip for a block.

Before merging this, we should confirm that we're comfortable with changing the behavior of `Trip.revenue_trip?/1`.

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -839,7 +837,10 @@ const isBreak = (activity: Piece | Break): activity is Break =>
   Object.prototype.hasOwnProperty.call(activity, "breakType")
 
 const isTrip = (trip: Trip | AsDirected): trip is Trip =>
-  Object.prototype.hasOwnProperty.call(trip, "id")
+  !Object.prototype.hasOwnProperty.call(trip, "kind")
```
firestack, line 840:
```suggestion
  !(trip && "kind" in trip)
```

assets/src/models/minischeduleData.ts
```diff
@@ -122,10 +123,12 @@ const tripFromData = (tripData: TripData): Trip => ({
...
   kind: asDirectedData.kind,
   startTime: asDirectedData.start_time,
   endTime: asDirectedData.end_time,
 })
 
-const isTripData = (data: TripData | AsDirectedData): data is TripData =>
-  Object.prototype.hasOwnProperty.call(data, "id")
+const isAsDirectedData = (
+  data: TripData | AsDirectedData
+): data is AsDirectedData => Object.prototype.hasOwnProperty.call(data, "kind")
```
firestack, line 134:
```suggestion
): data is AsDirectedData => !Object.hasOwn(trip, "kind")
```

firestack (approved):
@hannahpurcell I tried to pull this apart to no avail, so I went ahead and described each of the commits but I think we can merge this as is and I've found we can't really avoid the :id value or the fix to `AsDirectedRow`

test/schedule/block_test.exs
```diff
@@ -80,6 +80,31 @@ defmodule Schedule.BlockTest do
            pieces: [@piece]
          )
 
+  @as_directed %AsDirected{
+    id: "ad1",
+    kind: :wap,
```
hannahpurcell, line 85: 😂 um

```suggestion
    kind: :wad,
```
Interesting that this didn't lead to failed tests?

hannahpurcell: After looking at it `next_revenue_trip`, I'm not concerned there isn't a failed test, as that function wasn't responsible for checking it. But why didn't elixir's typing catch this??

firestack: oh wow thanks for catching this @hannahpurcell !

PR #2741, opened 2024-08-16
fix: Include Sentry LoggerBackend

hannahpurcell:
Not only log to the console (Splunk) but to Sentry

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -137,6 +137,10 @@ config :skate, Oban,
     }
   ]
 
+# Include 2 logger backends
+  config :logger,
+  backends: [:console, Sentry.LoggerBackend]
+
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time [$level] $metadata$message\n",


config/config.exs
```diff
@@ -137,6 +137,10 @@ config :skate, Oban,
     }
   ]
 
+# Include 2 logger backends
+config :logger,
+  backends: [:console, Sentry.LoggerBackend]
```
firestack, line 142: question: Hmm, we don't have Sentry in local dev or test, should this be in `prod.exs` instead?

hannahpurcell: Yeah, I waffled on this, because prod has the dsn check and would prevent anything in local dev / test from getting sent to Sentry anyway. But I don't mind moving it either

firestack: Oh that's a great point actually, thanks! I now realize that if it's in the base config, that means you can test sentry locally by providing the variable, and have it disabled by default, like it would be in a deployed environment. I do like that.

So now I don't really have a strong opinion on where it should go 🫤 I'd be happy with dropping that commit if you wanted.

firestack (approved):
I'm good either way on my comment https://github.com/mbta/skate/pull/2741#discussion_r1720375442 (though do take a sec to read/consider)

hannahpurcell:
> I'm good either way on my comment [#2741 (comment)](https://github.com/mbta/skate/pull/2741#discussion_r1720375442) (though do take a sec to read/consider)

Thanks for the thoughtful re-consideration! Testing Sentry locally could be an interesting option 👀  I'll drop the last commit, then

PR #2744, opened 2024-08-19
fix: Make as-directed parsing not-crash and default to regular trips

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1208002136498399/f

### DIFF
--- a/lib/schedule/hastus/activity.ex
+++ b/lib/schedule/hastus/activity.ex
@@ -152,10 +152,11 @@ defmodule Schedule.Hastus.Activity do
     end
 
     as_directeds_and_schedule_trips =
-      if operator_is_as_directed?(activity) do
-        [as_directed_from_trips(trips_in_piece)]
+      with true <- operator_is_as_directed?(activity),
+           {:ok, as_directed} <- as_directed_from_trips(trips_in_piece) do
+        [as_directed]
       else
-        Enum.map(trips_in_piece, &Map.fetch!(schedule_trips_by_id, &1.trip_id))
+        _ -> Enum.map(trips_in_piece, &Map.fetch!(schedule_trips_by_id, &1.trip_id))
       end
 
     %Piece{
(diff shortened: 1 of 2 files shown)


test/schedule/data_test.exs
```diff
@@ -1762,6 +1762,250 @@ defmodule Schedule.DataTest do
              } = Data.run_from_hastus(run_key, activities, trips, %{}, %{}, %{})
     end
 
+    test "allows as directed trips and regular trips in the same piece named like a regular block" do
```
joshlarson, line 1765: This test passed without the changes here (because the block name doesn't have `wad`/`rad` in it).

lib/schedule/hastus/activity.ex
```diff
@@ -175,27 +176,32 @@ defmodule Schedule.Hastus.Activity do
...
-      end
-
-    %AsDirected{
-      kind: kind,
-      start_time: as_directed_trip.start_time,
-      end_time: as_directed_trip.end_time,
-      start_place: as_directed_trip.start_place,
-      end_place: as_directed_trip.end_place
-    }
+    case trips_in_piece do
```
joshlarson, line 181: The main change here was to have this function return `{:ok, as_directed}` or `:error` instead of `as_directed` so that the calling function can handle both cases gracefully.

firestack (approved):
A thought: I'm not sure how easy this would be, but I wonder if testing this by mocking the date for elixir would be possible? 

I tried looking into it and didn't find a good way to do so, but I would like the peace of mind that our current rating doesn't break at 1200 after this change (I'm _fairly_ certain the [Sentry Error: `KeyError: block not found in map`](https://mbtace.sentry.io/issues/5731019670/?alert_rule_id=15157549&alert_type=issue&notification_uuid=fa73826f-75fd-45d0-92ab-4b8cae16e04d&project=5303878) is reproducible locally?).

But I think this addresses it well enough that we'll be okay without it

PR #2745, opened 2024-08-21
refactor: Move `fetchNearestIntersection` to the state machine

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207539825544288/f

### DIFF
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -6,6 +6,7 @@ import {
   FetchDetourDirectionsError,
   fetchDetourDirections,
   fetchFinishedDetour,
+  fetchNearestIntersection,
   fetchRoutePatterns,
 } from "../api"
 import { DetourShape, FinishedDetour } from "./detour"
@@ -21,6 +22,8 @@ export const createDetourMachine = setup({
       waypoints: ShapePoint[]
       startPoint: ShapePoint | undefined
       endPoint: ShapePoint | undefined
+
+      nearestIntersection: string | null
 
       detourShape: Result<DetourShape, FetchDetourDirectionsError> | undefined
 
@@ -76,6 +79,18 @@ export const createDetourMachine = setup({
         // the context types but it's not letting me
         throw "No Route ID"
       }
+    }),
+
+    "fetch-nearest-intersection": fromPromise<
+      Awaited<ReturnType<typeof fetchNearestIntersection>>,
+      {
+        startPoint?: ShapePoint
+      }
+    >(async ({ input: { startPoint } }) => {
+      if (!startPoint) {
+        throw "Missing nearest intersection inputs"
+      }
+      return fetchNearestIntersection(startPoint.lat, startPoint.lon)
     }),
 
     "fetch-detour-directions": fromPromise<
@@ -163,6 +178,7 @@ export const createDetourMachine = setup({
     waypoints: [],
     startPoint: undefined,
     endPoint: undefined,
+    nearestIntersection: null,
     finishedDetour: undefined,
     detourShape: undefined,
   }),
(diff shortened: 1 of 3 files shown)


assets/src/models/createDetourMachine.ts
```diff
@@ -308,22 +324,38 @@ export const createDetourMachine = setup({
...
-                input: ({ context: { startPoint, waypoints } }) => ({
-                  points: (startPoint ? [startPoint] : []).concat(
-                    waypoints || []
-                  ),
-                }),
-
-                onDone: {
-                  actions: assign({
-                    detourShape: ({ event }) => event.output,
+              invoke: [
```
joshlarson, line 327: Discussion about this:

This change will cause `fetch-nearest-intersection` to be invoked any time a waypoint is placed. This is a change from the hook implementation, where it was only called when the start point was placed (or changed).

I don't think it's a huge problem for right now, which is why I'm PR'ing this as-is, but I do wonder if the Right Solution is to add another state, so that the drawing-detour states would be something like:
- `No Start Point`
- `Start Point Only` <-- calls `fetchNearestIntersection` and eventually `fetchUnfinishedDetour`
- `Some Waypoints` <-- calls `fetchDetourDirections`
- `End Point Placed` <-- calls `fetchFinishedDetour`

@mbta/skate-developers thoughts?

joshlarson: Hmmm, this also seems to leave open a race condition, where if the end point is placed before the `fetch-nearest-intersection` actor finished its work, then the `fetch-nearest-intersection` actor's work is ignored, and we end up with `From null`. Is there a way to force an actor to keep plugging on even if the state machine has moved to a different state?

Screenshot 2024-08-21 at 3 40 45 PM

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -1288,6 +1288,10 @@ describe("DiversionPage", () => {
       fireEvent.click(originalRouteShape.get(container))
     })
 
+    await act(async () => {
```
joshlarson, line 1291: This is why I noticed [this](https://github.com/mbta/skate/pull/2745/files#r1725668252), because if this `await act` is moved down below the second `fireEvent.click`, then the test fails.

assets/src/models/createDetourMachine.ts
firestack: You want to [spawn an actor](https://stately.ai/docs/spawn)
A while ago, I tried doing this refactor once and ran into this, and was struggling to grasp the concepts and how to use `spawn` for this, maybe you'll have better luck, or we can try pairing on it.

firestack: If we want to avoid spawning, we'll probably need to do the intermediate state thing,
which could be done within the `place start point` state and use `onDone` to transition out of `place start point` and into `place waypoint`

Spawning has the benefit of not blocking the user on the intersection fetch, but then means we have to figure out how to store it and figure out how that gets serialized and persisted, so that we're not storing an actor state we don't mean to.

joshlarson: Okay here's a silly question - because this tool is primarily (exclusively?) used by dispatchers, where the internet connection is reasonably reliable, can we roll forward with this as-is, and migrate towards `spawn` later?

I think `spawn` is the right call in the long/medium term, but the `spawn` contract is so different from `invoke` that actually making that change is more complicated than I was expecting.

I think though that merging this as-is should unblock the things that are blocked, and changing to `spawn` shouldn't change anything (aside from fixing the issues mentioned ☝️☝️)

firestack: I think that's fine, would you document your reasoning inline? that might help future "us"?

My day is up, but I'll revisit this first thing tomorrow

joshlarson: Okay! I'll be around a little bit tomorrow, so let's see how we're feeling about this tomorrow.

firestack (approved):

PR #2747, opened 2024-08-23
doc: document api calls and deprecate old api calls

firestack:
We've had many API helper functions in api.ts, and it's been up to the programmer to understand when and why to use something, so now I've documented them and (hopefully) deprecated the old ones.

Context: https://github.com/mbta/skate/pull/2729#discussion_r1713933871


hannahpurcell (approved):
I LOVE this, thank you so much for adding!!

PR #2748, opened 2024-08-27
refactor: Move fake detour list data to backend and query it with API call

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1207933819206606/f

PR #2757, opened 2024-08-27
feat: API / controller work to populate detours table

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1207890779629113

Relied also on @joshlarson's work drafted [here](https://github.com/mbta/skate/pull/2748/files#diff-1e6f1a30aedafd3da50391010b59ab7b3acb63621d711125ffe161c4ef03f15d)

Needs:
- [x] Tests
- [x] Fixing the frontend timestamp

Polishing the tables frontend can be handled in [this ticket](https://app.asana.com/0/1205385723132845/1207890780042656), or in other appropriate tickets.


lib/skate/detours/detours.ex
```diff
@@ -5,54 +5,120 @@ defmodule Skate.Detours.Detours do
 
   import Ecto.Query, warn: false
   alias Skate.Repo
-
-  alias Skate.Detours.Db.Detour
+  alias Schedule.Gtfs.Direction
+  alias Skate.Detours.Db.Detour, as: DbDetour
```
joshlarson, line 9: Is there a struct called `Detour` that will run into a naming collision with this? Or is this mainly for clarity? [The other example](https://github.com/mbta/skate/blob/445884d91d766641b165a0a02af345f13696929a/lib/skate/settings/test_group.ex#L4) where I've seen this kind of this is in `TestGroup`, but the parent module and the DB module were both named `TestGroup`, so it was necessary.

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -24,6 +24,24 @@ defmodule SkateWeb.DetoursController do
...
   end
 
+  @spec get_detours(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def get_detours(conn, _params) do
+    %{id: user_id} = AuthManager.Plug.current_resource(conn)
+
+    detours =
+      user_id
+      |> Detours.list_detours()
+      |> Enum.group_by(fn detour -> detour.status end)
```
joshlarson, line 34: I think I'd prefer this to live in `&Detours.list_detours/1`. Moving it there would leave the controller to focus entirely on HTTP-layer stuff, rather than having to worry about figuring out which ones are `:draft`, etc.

Plus, right now, the `draft`/`active`/`past` trifurcation happens in the `Detours` module and then also a little bit here, and it would be nice to unify them.

I'll do some thinking about whether this is a blocking comment or a non-blocking nit.

hannahpurcell: 👍 That works for me. I reframed the function in Detours as `grouped_detours`

lib/skate/detours/detours.ex
hannahpurcell: Totally right! At one point, I thought an error I was seeing was a typing issue, so I was pulling in a Detour model that looked a little different. Later became unnecessary, so rolled this change back

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,99 @@ defmodule Skate.Detours.Detours do
...
+    date =
+      db_detour.updated_at
+      |> DateTime.from_naive!("Etc/UTC")
+      |> DateTime.to_unix()
+
+    %__MODULE__{
+      route: route_name,
+      direction: direction,
+      name: headsign,
+      intersection: "TBD",
```
joshlarson, line 82: This info is available now, I think.

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,99 @@ defmodule Skate.Detours.Detours do
...
+  @type detour_type :: :active | :draft | :past
+
+  @spec categorize_detour(map(), integer()) :: detour_type
+  defp categorize_detour(%{state: %{"value" => %{"Detour Drawing" => "Active"}}}, _user_id),
+    do: :active
+
+  defp categorize_detour(%{state: %{"value" => %{"Detour Drawing" => "Past"}}}, _user_id),
+    do: :past
+
+  defp categorize_detour(%{author_id: author_id}, user_id) when author_id == user_id, do: :draft
```
joshlarson, line 98: I do like this usage of Elixir pattern matching - makes it very easy to separate out that drafts are intended to be author-specific, while the other two aren't.

lib/schedule/gtfs/direction.ex
```diff
@@ -46,4 +46,16 @@ defmodule Schedule.Gtfs.Direction do
...
   def id_from_string("0"), do: 0
   def id_from_string("1"), do: 1
+
+  @doc """
+  Returns a inbound / outbound terms from a Direction ID
+
+  iex> Schedule.Gtfs.Direction.id_from_string(0)
+  "Outbound"
+  iex> Schedule.Gtfs.Direction.id_from_string(1)
+  "Inbound"
```
joshlarson, line 56: You might want to add a brief `moduledoc` to the top of this file, since right now it's tagged as `@moduledoc false`, thus not generating these docs.

Also, in order to get these examples to render nicely, you need to indent them 4 spaces.

```suggestion
      iex> Schedule.Gtfs.Direction.id_from_string(0)
      "Outbound"
      iex> Schedule.Gtfs.Direction.id_from_string(1)
      "Inbound"
```

# Before

Screenshot 2024-08-28 at 12 10 00 PM

# After

Screenshot 2024-08-28 at 12 09 27 PM


You can see the docs rendered by running `mix docs` and then bringing up `/path/to/skate/doc/api-reference.html` in the browser, or for this particular doc, `/path/to/skate/doc/Schedule.Gtfs.Direction.html`.

lib/schedule/gtfs/direction.ex
```diff
@@ -46,4 +46,16 @@ defmodule Schedule.Gtfs.Direction do
...
+  """
+  @spec id_to_string(id()) :: String.t()
+  def id_to_string(0), do: "Outbound"
+  def id_to_string(1), do: "Inbound"
```
firestack, line 60: Gtfs does not specify if 0 means outbound and 1 means inbound, and we have a `direction_names` map on our route objects, are we certain that this will _always_ be the case?

firestack: Did some digging, it seems like at the MBTA we should be using values from [`directions.txt`](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#directionstxt)?

hannahpurcell: Whooaaaa I did NOT realize that GTFS doesn't specify this?! I received that by word of mouth over time, which means a lot of folks might be making the same assumption? 😳  Ok, I'll use `direction_names`. Thanks for the pointer!!

assets/src/models/detour.ts
```diff
@@ -1,6 +1,25 @@
...
 import { ShapePoint, Stop } from "../schedule"
 import { CreateDetourMachineInput } from "./createDetourMachine"
+import { array, Infer, nullable, number, string, type } from "superstruct"
+
+export const SimpleDetour = type({
+  route: string(),
+  direction: string(),
+  name: string(),
+  intersection: string(),
+  updated_at: number(),
```
joshlarson, line 11: I'm a little torn about this field. It really means different things in different contexts, so it wounds me a bit to have it just called `updated_at` (for active this is "active since", for instance)

For draft detours, `updated_at` means "last edited"
For active detours, `updated_at` means "active since"
For past detours, `updated_at` might mean "last used" (I asked a clarifying question about that - this might be more complicated)

The fact that the same field can be used for all three things feels like an implementation detail, and it feels to me like it's an implementation detail that really belongs in the backend.

But I also see merit in not over-complicating this at this stage.

lib/schedule/gtfs/direction.ex
hannahpurcell: I reverted the changes to this file anyway, given [this rec](https://github.com/mbta/skate/pull/2757#discussion_r1734968669)

assets/src/models/detour.ts
hannahpurcell: Hmm... At first, I felt the variations are presentational. `updated_at` should be taken at face-value on the backend and presented appropriately on the frontend.

On the other hand, I see how that meaning could get lost between backend => frontend, and so defining the meaning more explicitly could be valuable. Especially depending on how we implement being able to edit active detours down the road. Given we don't know yet how that will be presented on the frontend / what data model we'll need for detours at that point, I might be leaning towards the current simplicity?

assets/src/hooks/useAllDetours.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A hook to fetch all detours, organized by "Active", "Drafts", "Past"
+ * @returns a list of {@link SimpleDetours}, or null if loading.
+ */
+export const useAllDetours = (): SimpleDetourData | null => {
+  const [detours, setDetours] = useState<SimpleDetourData | null>(null)
+
+  useEffect(() => {
+    fetchDetours().then((detours) => {
+      return isOk(detours) && setDetours(detours.ok)
```
firestack, line 15:
```suggestion
      isOk(detours) && setDetours(detours.ok)
```

assets/tests/components/detours/detoursPage.addDetour.test.tsx
```diff
@@ -13,6 +13,10 @@ import getTestGroups from "../../../src/userTestGroups"
 
 jest.mock("../../../src/userTestGroups")
 
+jest.mock("../../../src/hooks/useAllDetours", () => ({
+  useAllDetours: jest.fn(() => {}),
+}))
```
firestack, line 18: note: we should mock the fetch, not the hook, this makes sure we test anything related to promises resolving

We should also be settings mocks in `beforeEach` and not in 
```suggestion
import { fetchDetours } from "../../../src/api"

import { neverPromise } from "../../testHelpers/mockHelpers"

jest.mock("../../../src/api")

beforeEach(() => {
    jest.mocked(fetchDetours).mockReturnValue(neverPromise())
})
```

Because this is prevalent in Skate's existing test code, I decided to write up my thoughts on this in our notion, feel free to add or modify or ask any questions, https://www.notion.so/mbta-downtown-crossing/Code-and-Coding-Norms-7f7a302480dd4c5e8c8e7274d5a7ca49?pvs=4#0d1e961a6b8946c69a9f2424b6cfb09b

assets/tests/components/detourListPage.test.tsx
```diff
@@ -2,11 +2,52 @@ import { describe, test, expect, jest } from "@jest/globals"
...
 import renderer from "react-test-renderer"
 import { DetourListPage } from "../../src/components/detourListPage"
+import { useAllDetours } from "../../src/hooks/useAllDetours"
 
-jest.useFakeTimers().setSystemTime(new Date("2024-08-16T13:00:00"))
+jest.useFakeTimers().setSystemTime(new Date("2024-08-29T20:00:00"))
+
+jest.mock("../../src/hooks/useAllDetours", () => ({
+  useAllDetours: jest.fn(() => {}),
+}))
```
firestack, line 11: https://github.com/mbta/skate/pull/2757#discussion_r1738629215

assets/src/hooks/useAllDetours.ts
```diff
@@ -0,0 +1,20 @@
...
+  const [detours, setDetours] = useState<SimpleDetourData | null>(null)
+
+  useEffect(() => {
+    fetchDetours().then((detours) => {
+      isOk(detours) && setDetours(detours.ok)
+    })
+  }, [])
+
+  return detours
+}
```
firestack, line 20: note: Oh I realize you should be able to implment this using `useApiCall` instead of by hand

firestack: This changes the semantics a _little bit_, so you could keep that by processing the result of `useApiCall` like `usePatternsByIdForRoute` or `useNearestIntersectionFetchResult`

assets/tests/components/detourListPage.test.tsx
hannahpurcell: Ok so I had no trouble applying this comment to the `addDetour` test suite, but did run into issues here, and I can't gather why. The below change leads to an empty snapshot.

```
beforeEach(() => {
  jest.mocked(fetchDetours).mockReturnValue(neverPromise())
})

describe("DetourListPage", () => {
  test("renders detour list page", () => {
    jest.mocked(fetchDetours).mockResolvedValue(Ok({ ...
```

assets/src/hooks/useAllDetours.ts
hannahpurcell: Ok a couple attempts I've tried:

in `detourListPage`, I've replaced the reference to useAllDetours with:
```
  const { result: result } = useApiCall({
    apiCall: useCallback(async () => 
      fetchDetours(), [])
  })

  const detours = isOk(result) && result.ok
```
but result is not a permissible format for isOk, which expects a `Result`. 

Also tried just adjusting the hook, in a similar pattern to the other hooks you noted, but then no detours ever populated the screen.

```
export const useAllDetours = (): any =>
  useApiCall({
    apiCall: useCallback(async () => 
      fetchDetours(), [])
  })
```

assets/tests/components/detourListPage.test.tsx
firestack: I _think_ part of the issue _may_ be using `renderer.create` and the other is that now that we're mocking a _promise_, we have to wait for the table to show up, so you'll need `await screen.findBy<something>()` before the snapshot

hannahpurcell: Marvelous, the combined changes covered it. Thank you 🙏

assets/src/models/detour.ts
```diff
@@ -1,6 +1,25 @@
...
+})
+
+export type SimpleDetour = Infer<typeof SimpleDetour>
+
+export const SimpleDetourData = type({
```
joshlarson, line 16: Nit (non-blocking): How would you feel about renaming this to `SimpleDetourCollection` or `GroupedSimpleDetours` or something like that?

Reasoning: It's not clear from the name what the difference is between `SimpleDetour` and `SimpleDetourData`, or that `SimpleDetourData` is actually a collection of detours.

assets/tests/components/detourListPage.test.tsx
```diff
@@ -1,13 +1,63 @@
...
+          },
+          {
+            route: "2",
+            direction: "Outbound",
+            name: "Headsign B",
+            intersection: "Street C & Avenue D",
+            updated_at: 1724856392,
+          },
+        ],
+        draft: null,
```
joshlarson, line 37: Question: Will `draft` ever be `null`? Or is it more that it could be empty?

joshlarson: Wait no, looking at the backend logic, I get why this would be `null` rather than `[]`.

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,114 @@ defmodule Skate.Detours.Detours do
...
+      active: Map.get(detours, :active),
+      draft: Map.get(detours, :draft),
+      past: Map.get(detours, :past)
+    }
+  end
+
+  @spec db_detour_to_detour(Detour.t(), integer()) :: t()
+  defp db_detour_to_detour(
+         # If the schema isn't like this, the state machine shape has changed.
+         # How do we want to handle that possibility?
```
joshlarson, line 74: We probably want to replace this question with an answer, right?

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -58,6 +58,57 @@ defmodule SkateWeb.DetoursControllerTest do
...
+                     "author_id" => some_number,
+                     "updated_at" => other_number
+                   }
+                 ],
+                 "past" => nil
+               }
+             } = json_response(conn, 200)
+
+      assert is_integer(some_number)
+      assert is_integer(other_number)
```
joshlarson, line 108:
```suggestion
                     "author_id" => author_id,
                     "updated_at" => updated_at
                   }
                 ],
                 "past" => nil
               }
             } = json_response(conn, 200)

      assert is_integer(author_id)
      assert is_integer(updated_at)
```

Can we assert anything more specific about `author_id`? Like that the author who created the detour is the same as the author whose ID shows up here?

assets/src/components/detoursTable.tsx
```diff
@@ -37,7 +37,7 @@ export const DetoursTable = ({ data }: DetoursTableProps) => {
               {detour.intersection}
             </td>
             <td className="align-middle u-hide-for-mobile">
-              {timeAgoLabel(epochNowInSeconds, detour.activeSince)}
+              {timeAgoLabel(epochNowInSeconds, detour.updated_at)}
```
joshlarson, line 40: I think that `updated_at` being called here shows that there's a missing translation layer somewhere around `api.ts` and `models/`, since we typically want JS fields to use `camelCase` instead of `snake_case`.

I'm happy to pair on adding in that layer if you'd like (probably tomorrow because my work environment is not conducive to pairing today).

test/skate_web/controllers/detours_controller_test.exs
hannahpurcell: Makes sense! 👍

assets/src/components/detoursTable.tsx
hannahpurcell: 👍 ✅ done

lib/skate/detours/detours.ex
hannahpurcell, line 74:
```suggestion
```

assets/src/models/detour.ts
```diff
@@ -1,6 +1,65 @@
...
+
+export type SimpleDetourFromApi = Infer<typeof SimpleDetourFromApi>
+
+const simpleDetourFromData = (
+  detourData: SimpleDetourFromApi
+): SimpleDetour => ({
+  route: detourData.route,
+  direction: detourData.direction,
+  name: detourData.name,
+  intersection: detourData.name,
```
joshlarson, line 32:
```suggestion
  intersection: detourData.intersection,
```

joshlarson: Fixes [this](https://github.com/mbta/skate/pull/2757#issuecomment-2332166168), I think.

assets/src/components/detourListPage.tsx
```diff
@@ -93,7 +27,14 @@ export const DetourListPage = () => {
           Add detour
         </Button>
       )}
-      <DetoursTable data={fakeData} />
+      {detours && (
+        <>
+          {detours.active && <DetoursTable data={detours.active} />}
+          {detours.draft && <DetoursTable data={detours.draft} />}
+          {detours.past && <DetoursTable data={detours.past} />}
```
joshlarson, line 34:
```suggestion
          <h2>Active detours</h2>
          {detours.active && <DetoursTable data={detours.active} />}
          <h2>Draft detours</h2>
          {detours.draft && <DetoursTable data={detours.draft} />}
          <h2>Closed detours</h2>
          {detours.past && <DetoursTable data={detours.past} />}
```

Non-blocking, but I had to add this in on my local machine in order to be able to understand what I was seeing, and I figured it might be helpful.

No big deal though - something like this will happen sooner or later as we move closer and closer to the hi-fi designs.

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,121 @@ defmodule Skate.Detours.Detours do
 
   import Ecto.Query, warn: false
   alias Skate.Repo
-
   alias Skate.Detours.Db.Detour
   alias Skate.Settings.User
 
+  @type t :: %__MODULE__{
```
joshlarson, line 11: Non-blocking, more of a thought than a specific suggestion: This doesn't feel like the right place for this struct to be defined. Specifically I mean top-level in the main `Detours` module. 

I _think_ I'd sooner see this struct defined either within a sub-module here, or in a separate file.

Happy for that to be a follow-up PR though.

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,121 @@ defmodule Skate.Detours.Detours do
...
+  """
+  def grouped_detours(user_id) do
+    detours =
+      Detour
+      |> Repo.all()
+      |> Enum.map(fn detour ->
+        try do
+          db_detour_to_detour(detour, user_id)
+        rescue
+          e in FunctionClauseError ->
```
joshlarson, line 65: Nit (non-blocking): What if instead of rescuing an exception here, we added another pattern-match-y definition of `db_detour_to_detour/2` like so:

```elixir
db_detour_to_detour(invalid_detour) do
  Sentry.capture_message("Invalid detour", data: %{detour: invalid_detour})

  nil
end
```

(I made up the `data:` key being the way to communicate extra data to Sentry - we should double-check that if we go this route).

Happy to do this as a follow-up PR instead (if we even want to do this).

assets/src/models/detour.ts
hannahpurcell: Ahhhh! I think this was a regression in my model update, thanks for catching

assets/src/components/detourListPage.tsx
hannahpurcell: Right, I figured this will be done in your hi-fi work. Fine to add it here, too

lib/skate/detours/detours.ex
hannahpurcell: Why switch to capture_message?

lib/skate/detours/detours.ex
hannahpurcell: Ok, so a `detour` sub-module?

assets/src/components/detourListPage.tsx
hannahpurcell: Nevermind, rolling back because it wasn't worth the resulting CI issues and was out-of-scope

lib/skate/detours/detours.ex
joshlarson: The approach I was suggesting doesn't have an exception to capture. Maybe `capture_exception` would still work though?

hannahpurcell: Ohhhhh I see. What's the improvement by switching to this approach?

joshlarson: I think it's more idiomatic Elixir to rely on pattern-matching, rather than an exception.

Also it's more efficient (I don't think we're worried about performance in this case, but exceptions are slow).

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -58,6 +58,59 @@ defmodule SkateWeb.DetoursControllerTest do
...
+               "data" => %{
+                 "active" => nil,
+                 "draft" => [
+                   %{
+                     "direction" => "Outbound",
+                     "intersection" => "Street A & Avenue B",
+                     "name" => "Headsign",
+                     "route" => "23",
+                     "status" => "draft",
+                     "author_id" => author_id_on_get,
```
joshlarson, line 101:
```suggestion
                     "author_id" => ^author_id_on_put,
```

☝️ non-blocking, but then you don't need to have two author_id variables, and could probably even just rename it to `author_id`.

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -58,6 +58,59 @@ defmodule SkateWeb.DetoursControllerTest do
     end
   end
 
+  describe "get_detours/2" do
+    @tag :authenticated
+    test "fetches detours from database and groups by active, past, draft", %{conn: conn} do
```
joshlarson, line 63: I think it would be worthwhile to add test cases that validate:
1. that a user cannot retrieve a different user's drafts, and
2. that the endpoint will return the properly-formatted detours, even if there are some detours that are not properly formatted.

joshlarson:
This looks good! The only thing keeping me from approving at this point is that I'm interested in your thoughts on https://github.com/mbta/skate/pull/2757#discussion_r1745961216. I made a number of other comments as well, but those can all either be ignored or treated as follow-ups if you'd prefer.

test/skate_web/controllers/detours_controller_test.exs
hannahpurcell: 1. User cannot retrieve a different user's drafts: isn't that covered by [this comment](https://github.com/mbta/skate/pull/2757#discussion_r1742458418) which you re-reviewed given [this comment](https://github.com/mbta/skate/pull/2757#discussion_r1745956857)?
2. This is new and can do.

lib/skate/detours/detours.ex
hannahpurcell: Leaving as-is.

test/skate_web/controllers/detours_controller_test.exs
joshlarson: 1. That asserts that the author ID comes back correctly, but there's no test asserting that if I create a draft and then you load your drafts, that my draft won't show up in that list.
2. 🎉

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,121 @@ defmodule Skate.Detours.Detours do
...
   def list_detours do
     Repo.all(Detour)
   end
 
+  @doc """
+  Returns the detours grouped by active, draft, and past.
+
+  ## Examples
+
+      iex> grouped_detours()
```
joshlarson, line 50:
```suggestion
      iex> grouped_detours(my_user_id)
```

lib/skate/detours/detours.ex
hannahpurcell: Leaving as-is. I cannot easily find documentation about the accepted fields for `captureMessage`

lib/skate/detours/detours.ex
```diff
@@ -5,23 +5,129 @@ defmodule Skate.Detours.Detours do
...
+      direction: direction,
+      name: headsign,
+      intersection: nearest_intersection,
+      updated_at: date,
+      author_id: db_detour.author_id,
+      status: categorize_detour(db_detour, user_id)
+    }
+  end
+
+  defp db_detour_to_detour(invalid_detour, _) do
```
joshlarson, line 113: Question: Does this work, and mean that the rescue that I alluded to [here](https://github.com/mbta/skate/pull/2757/files#r1745907474) can go away? Or was this an experiment gone wrong that should be removed? Or other?

(Non-blocking)

joshlarson (approved):

PR #2762, opened 2024-08-28
feat: Update borders on the draft detours table

joshlarson:
I also snuck a few other changes in, like fixing the font and background colors, since it looked annoyingly bad without those changes.

# Before

Screenshot 2024-08-28 at 5 52 11 PM

# After

Screenshot 2024-08-28 at 5 52 52 PM

---

Asana Ticket: https://app.asana.com/0/1205385723132845/1208155044119310/f

PR #2763, opened 2024-08-28
feat: Space out components on detour list page

joshlarson:
# Before

Screenshot 2024-08-28 at 5 52 52 PM

# After

Screenshot 2024-08-28 at 5 54 58 PM

---

Asana Ticket: https://app.asana.com/0/1205385723132845/1208155044119311/f

Depends on:
- https://github.com/mbta/skate/pull/2762
  - Mostly because there's a git conflict if you try to merge this first, but also because this looks pretty bad without the updated table styling.

### DIFF
--- a/assets/css/_detour_list_page.scss
+++ b/assets/css/_detour_list_page.scss
@@ -1,0 +1,5 @@
+.c-detour-list-page {
+  .c-detour-list-page__button {
+    margin-bottom: 2.25rem;
+  }
+}
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -39,6 +39,7 @@ $vpp-location-padding: 1rem;
 @import "crowding";
 @import "cutout_overlay";
 @import "data_status_banner";
+@import "detour_list_page";
 @import "detours/detour_map";
 @import "detours/detour_modal";
 @import "detours/detour_table";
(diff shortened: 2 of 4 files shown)


assets/css/_detour_list_page.scss
```diff
@@ -0,0 +1,8 @@
+.c-detour-list-page {
+  background-color: white;
+}
+
+.c-detour-list-page__button {
+  padding: 0.5rem 1rem 0.5rem 1rem;
+  margin: 24px 24px 36px 24px;
+}
```
firestack: I love css, but is there a reason to do this and add to the bundle size vs using bootstrap utilities?

joshlarson: Couple thoughts:

1. I forgot to push a fix I had where I swapped all of these `px`'s for `rem`'s, so this should actually be `margin-bottom: 2.25rem`.
2. I can replace almost all of these with bootstrap utilities, except for `margin-bottom: 2.25rem`. There's no bootstrap utility for `2.25 * $spacer`.
3. BUT... I think it still looks pretty good with `mb-4`, which is `1.5 * $spacer`, so maybe we can ditch the CSS entirely for now.

Screenshot 2024-09-03 at 12 42 51 PM

joshlarson: I changed my mind and added back in a teeny bit of CSS that enables the bigger margin. I replaced everything else with utilities though.

PR #2764, opened 2024-09-03
storybook: Add story for DetoursTable

joshlarson:
Asana Ticket: https://app.asana.com/0/home/1205718271156544/1205857827225843

---
  - https://app.asana.com/0/0/1205857827225843

### DIFF
--- a/assets/stories/skate-components/detours/detoursTable.stories.tsx
+++ b/assets/stories/skate-components/detours/detoursTable.stories.tsx
@@ -1,0 +1,58 @@
+import type { Meta, StoryObj } from "@storybook/react"
+
+import React from "react"
+import { DetoursTable } from "../../../src/components/detoursTable"
+
+const meta = {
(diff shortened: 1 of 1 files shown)


joshlarson:
Actually this will generate an infinite diff in Chromatic, since the timestamps will change to "800 hours ago" and then "824 hours ago", etc. 

We should refactor the detours table to not do that calculation before merging this.

assets/stories/skate-components/detours/detoursTable.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const WithData: Story = {}
+
+export const WithoutData: Story = {
+  args: {
+    data: [],
```
joshlarson, line 56: A lack of data will be signified by `data=null`, not `data=[]`.

PR #2765, opened 2024-09-03
refactor: rename `realtime.d.ts` => `realtime.ts`

firestack:
I'd like to add an `enum` to `realtime.d.ts` but turns out you can't really add and `enum` to a "definitions" file


assets/src/realtime.ts
```diff
@@ -7,7 +7,7 @@ import {
   TimepointId,
   TripId,
   ViaVariant,
-} from "./schedule.d"
+} from "./schedule"
```
joshlarson, line 10: Curiosity: Is there another `schedule.d.ts` ==> `schedule.ts` change associated with this, or can you import things from `*.d.ts` without the `.d`?

Looks like the build is passing, so it looks like this didn't break anything O_o

joshlarson (approved):

assets/src/realtime.ts
firestack: AFAIK you can leave off all the extensions when importing, didn't notice this change though sorry!

I'd be open to reverting that if you feel strongly enough about it.

joshlarson: No feelings whatsoever - I was just surprised that it worked.

PR #2768, opened 2024-09-03
feat: Adds content to active detour panel

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208153509160250

My last commit is optional! If this rearrangement is not valuable, I'm happy to revert. I just wanted to make a few pieces of the detour panels reusable.

Another idea for another PR: putting all detour panels in a subfolder together

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -43,7 +43,7 @@ const displayFieldsFromRouteAndPattern = (
 
 interface DiversionPageFunctions {
   onClose?: () => void
-  onConfirmClose?: () => void
+  onConfirmClose: () => void
   onCancelClose?: () => void
   showConfirmCloseModal: boolean
 }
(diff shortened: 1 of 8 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -245,6 +245,17 @@ export const DiversionPage = ({
...
+              connectionPoints={[
+                connectionPoints?.start?.name ?? "N/A",
+                connectionPoints?.end?.name ?? "N/A",
+              ]}
+              missedStops={missedStops}
+              routeName={routeName ?? "??"}
+              routeDescription={routeDescription ?? "??"}
+              routeOrigin={routeOrigin ?? "??"}
+              routeDirection={routeDirection ?? "??"}
+              onNavigateBack={() => onConfirmClose}
```
joshlarson, line 258:
```suggestion
              onNavigateBack={onClose}
```

In order to make the back button behave the same as the close button.

joshlarson (changes requested):
I'm a big fan of extracting the reusable components, and I like the way this shook out.

Found a bug with the `<- Back` button, and want to take another look in a little bit, but overall I like it!

assets/src/components/detours/diversionPage.tsx
hannahpurcell: We don't want it to behave the same as the close button. For an active detour, we don't need the confirm close modal to appear

joshlarson: Oh yeah I guess the confirm dialog isn't really necessary from the active state.

In that case:
```suggestion
              onNavigateBack={onConfirmClose}
```

joshlarson (approved):

PR #2770, opened 2024-09-04
feat: Styling for components in detours table cells

joshlarson:
# Before

Screenshot 2024-09-04 at 2 26 13 PM

# After

Screenshot 2024-09-04 at 2 25 16 PM

---

Asana Ticket: https://app.asana.com/0/1205718271156548/1208155044119313/f

Depends on:
- https://github.com/mbta/skate/pull/2762

---

I considered extracting this to a separate component, but didn't for now because it's not exactly the same as other instances of similar-looking things. I'm open to doing that, either as a follow-up, or as a pre-work to this, if you all think it's a good idea.

### DIFF
--- a/assets/src/components/detoursTable.tsx
+++ b/assets/src/components/detoursTable.tsx
@@ -16,27 +16,29 @@ export const DetoursTable = ({ data }: DetoursTableProps) => {
     <Table hover className="c-detours-table">
       <thead>
         <tr>
-          <th>Route</th>
-          <th className="u-hide-for-mobile">Starting Intersection</th>
-          <th className="u-hide-for-mobile">On detour since</th>
+          <th className="px-3 py-4">Route and direction</th>
+          <th className="px-3 py-4 u-hide-for-mobile">Starting Intersection</th>
+          <th className="px-3 py-4 u-hide-for-mobile">On detour since</th>
         </tr>
       </thead>
       <tbody>
         {data.map((detour, index) => (
           <tr key={index}>
-            <td className="align-middle">
-              <div className="c-detours-table__route-info-cell">
+            <td className="align-middle p-3">
+              <div className="d-flex">
                 <RoutePill routeName={detour.route} />
-                <div className="c-detours-table__route-info-text">
-                  <span>{detour.direction}</span>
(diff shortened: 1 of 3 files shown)


assets/src/components/detoursTable.tsx
```diff
@@ -16,27 +16,29 @@ export const DetoursTable = ({ data }: DetoursTableProps) => {
     <Table hover className="c-detours-table">
       <thead>
         <tr>
-          <th>Route</th>
-          <th className="u-hide-for-mobile">Starting Intersection</th>
-          <th className="u-hide-for-mobile">On detour since</th>
+          <th className="px-3 py-4">Route and direction</th>
```
joshlarson, line 19: I'd be interested to pick apart my instincts/choices here. My general strategy was to do everything that I could easily do using bootstrap utilities, and then put the rest into CSS.

I don't LOVE this because it means that our visual/styling logic is kind of spread out between CSS and HTML, but it does reduce CSS bundle size and lines up more closely with how we've been doing things so far.

Plus, when they work, the bootstrap utilities are quite nice!

assets/src/components/detoursTable.tsx
```diff
@@ -16,27 +16,29 @@ export const DetoursTable = ({ data }: DetoursTableProps) => {
...
-            <td className="align-middle">
-              <div className="c-detours-table__route-info-cell">
+            <td className="align-middle p-3">
+              <div className="d-flex">
                 <RoutePill routeName={detour.route} />
-                <div className="c-detours-table__route-info-text">
-                  <span>{detour.direction}</span>
-                  <span className="fw-bold">{detour.name}</span>
+                <div className="c-detours-table__route-info-text d-inline-block">
+                  <div className="pb-1 fs-4 fw-bold">{detour.name}</div>
```
hannahpurcell, line 31: `fw-bold` is a touch too bold. Design: font-weight 600, while fw-bold is 700

joshlarson: [Oops](https://github.com/mbta/skate/pull/2770/commits/b86438962d9fcc4a8d6f0f757148b916431c7357)

hannahpurcell (approved):

PR #2771, opened 2024-09-04
feat: Add new nav icon for the detours page

joshlarson:
# Before

Screenshot 2024-09-04 at 1 26 02 PM
Screenshot 2024-09-04 at 1 26 08 PM

# After
Screenshot 2024-09-04 at 1 26 26 PM
Screenshot 2024-09-04 at 1 26 31 PM

---

Asana Ticket: https://app.asana.com/0/1205718271156548/1208226797938050/f


assets/src/navLinkData.ts
```diff
@@ -1,19 +1,17 @@
...
 
 type HTMLElementProps = React.PropsWithoutRef<React.HTMLAttributes<HTMLElement>>
 
 export interface LinkData {
   title: string
   path: string
-  navIcon: React.JSXElementConstructor<HTMLElementProps>
+  navIcon:
+    | React.JSXElementConstructor<HTMLElementProps>
+    | ((props: NavIconProps) => React.JSX.Element)
```
joshlarson, line 14: This feels unideal to me, but I'm not sure how to convince typescript that this is okay without doing this.

firestack: IMO, the best way to handle this would be to change navIcon to a `ReactNode` and stop having the nav supply classes to the content it's trying to render, but the way our CSS is built is probably less friendly to that idea.

assets/src/helpers/navIcons.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ComponentPropsWithoutRef } from "react"
+
+export type NavIconProps = ComponentPropsWithoutRef<"span">
+
+export const DetourNavIcon = (props: NavIconProps) => (
+  <span {...props}>
+    <svg
```
firestack, line 7: suggestion: I think we should make this function like the icons in our bsIcons, so we'd ideally have the same props and we'd be putting `{...props}` on the `<svg/>` and get rid of the `<span>` inside here. If anything relies on the `<span/>` being there, I'd instead wrap this icon in some intermediate component.

If it follows our `bsicons` conventions we could potentially reuse some of that testing code as well that verifies the correct properties are present.

joshlarson: I had to torture `navLinkData.ts` (now `navLinkData.tsx`) a little bit in order to get this to work.

[Thoughts?](https://github.com/mbta/skate/pull/2771/commits/3056e3e2435b323dfd8bd46bc4ca37b3bbd74862?diff=split&w=1)

firestack: Works for me!

firestack (approved):

PR #2773, opened 2024-09-05
/refactor: move notification types into `content` field

firestack:
Asana Ticket: https://app.asana.com/0/0/1208212248665131/f

When looking to implement a new notification type, I ran into an issue where it's hard to deduce on the frontend which fields should be present because the same struct was being shared for multiple notification types. This finishes up the work of having multiple notification types by splitting it up for the frontend to be able to match on.


lib/notifications/notification.ex
```diff
@@ -131,48 +123,31 @@ defmodule Notifications.Notification do
...
-      reason: reason,
-      route_ids: @chelsea_st_bridge_route_ids,
-      run_ids: [],
-      trip_ids: [],
-      start_time: db_record.created_at,
-      end_time: end_time,
-      state: :unread
-    }
-
-    if source == :new_record, do: log_creation(notification)
```
joshlarson, line 174: **Praise**: It took me a little while to wrap my mind around why we needed the whole `source` thing. Removing that (and moving the log line to the right spot) makes this a lot easier to understand!

lib/notifications/notification_server.ex
```diff
@@ -181,9 +181,36 @@ defmodule Notifications.NotificationServer do
     %{status: bridge_status, lowering_time: lowering_time}
   end
 
-  defp broadcast(notification, registry_key) do
+  @chelsea_bridge_route_ids [
```
joshlarson, line 184: How did this work before? As far as I can tell, everything worked without these values here before - why add them?

Question-behind-the-question: Is there a way to avoid magic values in the code itself?

Non-blocking.

firestack: This was "magick-ed" here before https://github.com/mbta/skate/blob/e51917e03ea495402166a0df9a5dd45665fae82c/lib/notifications/notification.ex#L166

I'd be open to looking at a post-refactor which cleans this up and consolidates a bit more, as I'm noticing this data is now duplicated between linking the notifications to affected users and broadcasting to users.

joshlarson: Thanks for the explanation - it didn't show up when I searched earlier because I searched for `"743"` with the quotes.

assets/src/realtime.ts
```diff
@@ -53,26 +53,59 @@ export interface Ghost {
 
 export type NotificationId = string
 
-export interface Notification {
-  id: NotificationId
+// Matches the Module in which this notification was created by
```
hannahpurcell, line 56: Interesting! How does this work? This seems a little magical to me, how the Elixir module is referenced by this TS file. This change is the foundation of the refactor (I see NotificationType is referenced throughout the repo), so having some breadcrumbs for future readers could be helpful

assets/src/realtime.ts
```diff
@@ -53,26 +53,59 @@ export interface Ghost {
...
+  $type: NotificationType.BridgeMovement
+  status: "raised"
+  loweringTime: Date
+}
+
+export type BridgeNotification =
+  | BridgeLoweredNotification
+  | BridgeRaisedNotification
+
+export type BlockWaiverNotification = {
```
hannahpurcell, line 77: `createdAt` is part of the base type `NotificationContentTypes`, but it is also included in the `BlockWaiverNotification` type. If I'm understanding this right, I think the `createdAt` can be removed from this type spec?

assets/src/models/notificationData.ts
```diff
@@ -22,37 +37,88 @@ export const NotificationData = type({
...
-})
+): Notification => {
+  let content: NotificationContentTypes
+
+  switch (notificationData.content.$type) {
+    case NotificationType.BlockWaiver: {
+      content = {
+        $type: NotificationType.BlockWaiver,
+        reason: notificationData.content.reason,
+        createdAt: notificationData.content.created_at,
```
hannahpurcell, line 87: Maybe my prior comment about `created_at` is incorrect because I see here there is a difference between `notificationData.created_at` and `notificationData.content.created_at`, but I'm not sure why a BlockWaiver would have an additional `created_at` value from the db in its content

lib/notifications/notification.ex
```diff
@@ -111,13 +102,14 @@ defmodule Notifications.Notification do
           )
       end
 
-    struct!(__MODULE__, Map.merge(notification_values, %{id: db_record.id}))
+    from_db_notification(db_record)
   end
 
+  @spec get_or_create_from_bridge_movement(any()) :: t()
```
hannahpurcell, line 108: I think the `any()` can be replaced with a true type, given the output type of `get_db_values_from_bridge_movement`

assets/src/components/notificationCard.tsx
```diff
@@ -45,40 +56,55 @@ export const NotificationCard = ({
...
-            value:
-              notification.operatorName !== null &&
-              notification.operatorId !== null
-                ? `${notification.operatorName} #${notification.operatorId}`
-                : null,
-            sensitive: true,
-          },
-        ]}
-      />
+      {isBlockWaiverNotification(notification) && (
```
hannahpurcell, line 59: Why is this new conditional needed when it wasn't before? Does this mean that this section was previously rendered for BridgeMovements, but it shouldn't have been?

firestack: These properties _could_ render a Bridge Movement notification, if it had runIds and/or an operator name, but in practice, we prevented those states on the backend.

On the backend, a bridge notification would always an empty run_ids key https://github.com/mbta/skate/blob/main/lib/notifications/notification.ex#L162-L172 and operator name and ID is left out entirely. So it would be impossible for the backend to produce a bridge notification which triggered this UI logic.

You will find that due to this lack of type safety, our tests used to expect that a bridge notification would render with run ids, and would look for it when trying to click elements 
https://github.com/mbta/skate/blob/main/assets/tests/components/notificationCard.test.tsx#L250
[which had to be changed](https://github.com/mbta/skate/pull/2773/commits/78c9cf9b27b521e4aacf316eca944222219ec164#diff-54dea08de39a5c7398212550c06dcad3fd2c2f6af9f39e1c0718d0cb869c0038L250-R349)

lib/notifications/notification.ex
firestack: Oh whoops I meant for this refactor not to introduce anything, VSCode must've added this `@spec`, I'd like to remove it for this PR.

assets/src/realtime.ts
firestack: These just happen to be the module name because that's what I set the `default:` to. https://github.com/mbta/skate/pull/2773/files#diff-226c071046feb083bddbfbfd9ff74180d55bb17a2a0b27a1e1b423b59ddde1f9R40

We could use a different value but I didn't think it was worth agonizing over when I was in the middle of this, and I like that this is "self documenting" in that it points you to which module it's from.

I'll think about how to document this 👍🏻 

I did try and think of an alternative to adding a `$type` virtual key to the ecto schema, but I didn't think it was worth creating a new struct type just for that, and I didn't want to write a custom `Jason.Encode` implementation _just_ to add a type discriminator for typescript. This way, ecto just puts the key on the struct for us, but maybe this isn't a good pattern?

assets/src/realtime.ts
firestack: I'm not quite sure why they're duplicated, but I was trying to not introduce or remove any behavior in this PR.

You'll find that this column exists on both the notifications table and the block waiver table, and they seem to mean _something_ different between them? But I think that meaning is hidden deeper in the block waiver code.

https://github.com/mbta/skate/blob/310c7270adce01d8492e3baabf7c5d2601b57e83/lib/notifications/db/notification.ex#L13
https://github.com/mbta/skate/blob/310c7270adce01d8492e3baabf7c5d2601b57e83/lib/notifications/db/block_waiver.ex#L9

assets/src/models/notificationData.ts
firestack: Yeah me either :/

assets/src/realtime.ts
firestack: I added two commits which change the key on the Elixir side from `$type` to `__struct__`, which I help is a bit more explanatory in that it doesn't require knowing that the struct would contain a `$type` key with the `__MODULE__` but would instead just reference the actual Elixir struct we expect to receive. Do you think this makes this any better?

See commits beginning with `review!`, e.g.,

[review! fix: convert $type in notifications to __struct__](https://github.com/mbta/skate/pull/2773/commits/f3fedf8e1c0c504a9a28fe146bcd49a724b76dd5)

lib/notifications/notification.ex
firestack: fixed: https://github.com/mbta/skate/compare/8c751cb8ce6ca2a6b0f937346261f509cf9c259b..4cabc12ae223821d4394dd2a4932756eb179eb81

assets/src/realtime.ts
hannahpurcell: I like it! Thank you

assets/src/realtime.ts
hannahpurcell: Ohhhhh I see. How odd that it exists in both places, but I completely agree with your approach to avoid adding/subtracting existing behavior

lib/notifications/notification.ex
hannahpurcell: ok! that works

hannahpurcell (approved):
Great!! Thanks for covering all my questions. This was quite the undertaking, so bravo

PR #2774, opened 2024-09-05
feat: Update detour list page and table styles for mobile

joshlarson:
# Desktop

Screenshot 2024-09-05 at 4 54 05 PM

# Tablet Portrait

Screenshot 2024-09-05 at 4 50 21 PM

# Mobile

Screenshot 2024-09-05 at 4 50 35 PM

---

Asana Ticket: https://app.asana.com/0/1205385723132845/1208226797938041/f

Depends on:
- https://github.com/mbta/skate/pull/2778

### DIFF
--- a/assets/css/detours/_detour_table.scss
+++ b/assets/css/detours/_detour_table.scss
@@ -30,7 +30,9 @@ $table-border-radius: 0.5rem;
   tr {
     th:first-child,
     td:first-child {
-      border-left: $table-border;
+      @include media-breakpoint-up(md) {
+        border-left: $table-border;
+      }
     }
 
     th:last-child,
--- a/assets/src/components/detourListPage.tsx
+++ b/assets/src/components/detourListPage.tsx
@@ -83,10 +83,10 @@ export const DetourListPage = () => {
   const [showDetourModal, setShowDetourModal] = useState(false)
 
   return (
-    <div className="c-detour-list-page h-100 overflow-y-auto p-4 bg-white">
+    <div className="c-detour-list-page h-100 overflow-y-auto p-0 p-md-4 bg-white">
       {userInTestGroup(TestGroups.DetoursPilot) && (
         <Button
-          className="c-detour-list-page__button icon-link fw-light px-3 py-2"
+          className="c-detour-list-page__button icon-link fw-light px-3 py-2 u-hide-for-mobile"
           onClick={() => setShowDetourModal(true)}
         >
           <PlusSquare />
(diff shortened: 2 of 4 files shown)


assets/css/utilities/_hideable.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 .u-hide-for-mobile {
-  @media screen and (max-width: map-get($breakpoints, "max-mobile-landscape-tablet-portrait-width")) {
+  @include media-breakpoint-down(md) {
```
joshlarson, line 14: @mbta/skate-developers want to surface this change. It's not a huge change, but this changes the "hide for mobile" breakpoint from 800px to 768px. 

I did this because `media-breakpoint-...` uses bootstrap's breakpoints, and I wanted to bias towards bootstrap's breakpoints rather than our own `_skate_ui` ones, especially since we want to stop "caring about what's in _skate_ui", per this [Slack Convo](https://mbta.slack.com/archives/C047ZT8ACP5/p1725383093148209?thread_ts=1725382780.112419&cid=C047ZT8ACP5).

If you'd all prefer, I can split it out into a separate PR. Or ditch this and find a different way to keep our breakpoints consistent!

firestack: Btw I found the variable we need to update if we do want to make bootstrap use our breakpoints as a temporary measure https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L483-L492

joshlarson: @firestack Amazing, thank you!

What do you think of [this](https://github.com/mbta/skate/pull/2774/files#diff-7781a17abc4cbb8e10f3131a4da4fcc8dc20efd270178f5b756e6be32ae32cafR20-R31)?

And at this point, should the `_variable_overrides` change and this one be split to a separate PR, do you think?

firestack: > And at this point, should the `_variable_overrides` change and this one be split to a separate PR, do you think?

I'm good either way 👍🏻 


> What do you think of [this](https://github.com/mbta/skate/pull/2774/files#diff-7781a17abc4cbb8e10f3131a4da4fcc8dc20efd270178f5b756e6be32ae32cafR20-R31)?

Looks good, I don't _think_ you need `!default`? Did you try it without that? AFAIK bootstrap has that in their config so that we can override it easily.

Maybe it also should document _why_ these values are set like this, because right now we're doing this to be compatible with the old breakpoints "system", but I don't know what design will want to do with breakpoints in the future.

joshlarson: [Added a comment](https://github.com/mbta/skate/pull/2774/files#diff-7781a17abc4cbb8e10f3131a4da4fcc8dc20efd270178f5b756e6be32ae32cafR20-R24)

I think I do want to split this out into two PR's.

firestack (approved):

PR #2777, opened 2024-09-09
feat: Display empty detour table icon when the table is empty

joshlarson:
# Before
## Desktop
Screenshot 2024-09-09 at 11 15 20 AM

## Mobile
Screenshot 2024-09-09 at 11 15 14 AM

# After
## Desktop
Screenshot 2024-09-09 at 11 19 48 AM

## Mobile
Screenshot 2024-09-09 at 11 19 54 AM

---

Asana Ticket: https://app.asana.com/0/1205526445275136/1208226799631632/f

Depends on:
- https://github.com/mbta/skate/pull/2778
- https://github.com/mbta/skate/pull/2774


joshlarson:
I originally did this work as two separate chunks - creating the icon (and the storybook story), and then adding it to the table, but then combined the work into this one PR.

If you'd prefer that I split this back into two PR's (see the two commits on this branch), lemme know - I'm happy to do it.

assets/src/components/detourListPage.tsx
```diff
@@ -93,7 +93,10 @@ export const DetourListPage = () => {
           <span className="c-detour-list-page__button-text">Add detour</span>
         </Button>
       )}
+      <h2>Active</h2>
       <DetoursTable data={fakeData} />
+      <h2>Draft</h2>
```
joshlarson, line 98: This change is here in order to make it easier to actually view the changes while reviewing, but I'm planning on removing it before merging (for fear that it'll create unnecessary conflicts with [this work](https://github.com/mbta/skate/pull/2757/files#diff-9c377c74d3f4b789891f6671a6ff311ea53c18bb05476f0f9ea9ddeffb99f71e)

assets/stories/skate-components/icons/emptyDetourTableIcon.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import type { Meta, StoryObj } from "@storybook/react"
+import { EmptyDetourTableIcon } from "../../../src/components/detoursTable"
+
+const meta = {
+  component: EmptyDetourTableIcon,
+  parameters: {
+    layout: "centered",
+    stretch: false,
```
firestack, line 8: question: `stretch: false` is the default behavior, does this need to be specified?

joshlarson: I copied that from the [detour end icon story](https://github.com/mbta/skate/blob/cb2763e6ed74ba194033bc6487e2b68e3b3a7e9f/assets/stories/skate-components/detours/markers/startIcon.stories.tsx#L9), which I think I copied from the [map location search marker story](https://github.com/mbta/skate/blob/cb2763e6ed74ba194033bc6487e2b68e3b3a7e9f/assets/stories/skate-components/map/markers/locationSearchMarker.stories.tsx).

But it does look like removing that property doesn't change anything. Perhaps I'll remove it from some of the other stories as well (as a follow-up).

joshlarson: Oh and I'm the one who added it to the map location search marker story too, so I have no idea what's going on lol.

assets/css/detours/_detour_table.scss
```diff
@@ -61,6 +61,12 @@ $table-border-radius: 0.5rem;
     td:last-child {
       border-bottom-right-radius: $table-border-radius;
     }
+
+    @include media-breakpoint-down(md) {
```
joshlarson, line 65: This is an unrelated visual bug fix, but I noticed it while working on this PR and didn't realize until just now that I had accidentally lumped it into this PR.

I can split it out to a separate PR if you'd all prefer, but I didn't necessarily think it was worth the effort.

assets/src/components/detoursTable.tsx
```diff
@@ -1,49 +1,98 @@
...
+    </thead>
+    <tbody>
+      <DetourRows data={data} />
+    </tbody>
+  </Table>
+)
+
+const DetourRows = ({ data }: { data: SimpleDetour[] | null }) =>
+  data === null ? (
+    <EmptyDetourRows message="No draft detours." />
```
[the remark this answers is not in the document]
joshlarson: Yeah that's a reasonable point. I had planned to peel that out to a separate PR, but this is kind of objectively wrong right now.

Plus, doing that work in this PR will make some of the other PR's not block each other, which would be nice.

assets/src/components/detoursTable.tsx
```diff
@@ -1,49 +1,98 @@
...
+      </div>
+      <div className="d-flex justify-content-center">
+        <p className="fs-3 fw-light m-0">{message}</p>
+      </div>
+    </td>
+  </tr>
+)
+
+export const EmptyDetourTableIcon = (props: ComponentProps<"svg">) => (
+  <svg
```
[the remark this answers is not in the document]
hannahpurcell: Ohhh... I read this and then forgot about it once I saw the code out of context. Hmm, perhaps I should reply there? But how do we avoid distributing otherwise-reusable icons throughout the repo in a way that is hard to track?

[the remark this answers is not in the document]
joshlarson: Another worthwhile question: should this icon (and other icons) be moved somewhere more consistent? 

I figured it was likely that this icon would only ever be used on this page, so it wasn't necessary to create a separate home for it, but I don't feel strongly about that.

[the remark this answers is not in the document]
firestack: maybe a `skateIcons.tsx` similar to `bsIcons.tsx`? Then we can have consistent tests for them. IMO This would be for "simple" icons which contain only SVG and don't take other props.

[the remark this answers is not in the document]
hannahpurcell: That makes sense to me. Or perhaps a folder that contains a Reacty icon component per file? I prefer surfing icons by file name, but that's maybe too-personal a preference.

Also, turning my feedback to non-blocking, since this had already been discussed (and I forgot), and the requested change is mainly an organizational one at this point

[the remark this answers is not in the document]
firestack: I like the `bsIcons` route because it makes it easy to apply uniform tests to all the icons without needing to edit the tests. But I'm open to other options.

assets/src/components/detoursTable.tsx
```diff
@@ -1,49 +1,107 @@
...
+        <th className="px-3 py-4">Route and direction</th>
+        <th className="px-3 py-4 u-hide-for-mobile">Starting Intersection</th>
+        <th className="px-3 py-4 u-hide-for-mobile">On detour since</th>
+      </tr>
+    </thead>
+    <tbody>
+      <DetourRows data={data} status={status} />
+    </tbody>
+  </Table>
+)
```
firestack, line 28: nit: I'd prefer if we didn't abstract this yet, and avoid any prop drilling that we'd need to do by having this intermediate component.

assets/src/components/detoursTable.tsx
```diff
@@ -1,49 +1,107 @@
...
-        ))}
-      </tbody>
-    </Table>
+            </div>
+          </td>
+          <td className="align-middle p-3 u-hide-for-mobile">
+            {detour.intersection}
+          </td>
+          <td className="align-middle p-3 u-hide-for-mobile">
+            {timeAgoLabel(epochNowInSeconds, detour.updatedAt)}
```
firestack, line 65: issue(non-blocking): technically this is incorrect semantics and I think would readback strangely to a screen reader, but if you're okay with that we can keep it till it's an issue.

assets/src/components/detoursTable.tsx
joshlarson: Which part do you not want to abstract? I can't quite tell, but I think Github might be rendering this weird.

assets/src/components/detoursTable.tsx
joshlarson: I think Github is rendering this weird as well, but I'm reasonably certain that the part of the PR that it's currently highlighting is a pure refactor, and shouldn't change the DOM at all.

assets/src/components/detoursTable.tsx
firestack: I was thinking that `<DetourRows/>` would be inlined here

assets/src/components/detoursTable.tsx
```diff
@@ -1,49 +1,107 @@
...
+    <td colSpan={3} className="p-3 p-md-4">
+      <div className="d-flex justify-content-center mb-3">
+        <EmptyDetourTableIcon height="100px" width="100px" />
+      </div>
+      <div className="d-flex justify-content-center">
+        <p className="fs-3 fw-light m-0">{message}</p>
+      </div>
+    </td>
+  </tr>
+)
```
firestack, line 84: this comment is on the `<EmptyDetourRows/>` component
> issue(non-blocking): technically this is incorrect semantics and I think would readback strangely to a screen reader, but if you're okay with that we can keep it till it's an issue.
> https://github.com/mbta/skate/pull/2777#discussion_r1755235716

assets/src/components/detoursTable.tsx
firestack: https://github.com/mbta/skate/pull/2777/files#r1756748196

assets/src/components/detoursTable.tsx
joshlarson: I'd prefer to address this as a follow-up. No need to wait for it to become a problem, but I would like to get this merged so that it can unblock the things that come after.

Would you be up for pairing on this next time we're both in-office, since I'm still not totally sure what the problem is.

assets/src/components/detoursTable.tsx
joshlarson: [Done](https://github.com/mbta/skate/pull/2777/commits/ddf0a40aa51ec4ab37342511f600c93ab8c85dd1)

assets/src/components/detoursTable.tsx
firestack: > Would you be up for pairing on this

Yeah sure

firestack: For posterity, the core of it is that to a screen reader (or really just the accessibility tree), this would read as "a table with 1 row" but the row just says (paraphrasing)"this table is empty"

Similar to 
image

assets/src/helpers/skateIcons.tsx
firestack, line 1: can you add tests for this like bsIcons?

Imo, we should probably be putting the `bi` class on these too?

firestack (changes requested):
marking as changes requested for github overview screens

assets/src/components/detoursTable.tsx
joshlarson: @firestack would the right response here be to make the table headers (or even the entire table) be hidden to a screen reader, and replaced with just the text "No <status> detours"?

...with the `aria-hidden` attribute?

firestack: I would keep the table and headers, because otherwise it would read as "Heading Detours, Heading Drafts"

and I think we want "Heading Detours, Table, 0 Rows"

You could either use the `<caption>` element, or add the content as it's own item outside of the table. But right now, I think it'd read as "Table, 1 row" and the row would say "There are no detours"(paraphrasing)

joshlarson: Adding `aria-hidden` to the `EmptyDetourRows` makes that row disappear from the accessibility tree. So it wouldn't read the row saying "There are no detours" - it would just say "0 rows" or something like that.

But would it make more sense to get rid of the whole table, and _only_ have it say "No active detours"? So that it's something like "Heading Active Detours, No Active Detours"?

Screenshot 2024-09-23 at 1 52 17 PM

firestack (approved):

assets/src/helpers/skateIcons.tsx
joshlarson: [Done!](https://github.com/mbta/skate/pull/2777/files#diff-fe402059d452d0c04fb7c60c0689ba890cc7a01e511691964ad77b7b16335989)

PR #2781, opened 2024-09-10
fix: Need to query the table in order of latest updates

hannahpurcell:
Realized the db should be queried in order of most recently updated detours, for all cases.


firestack (approved):
I think Credo is gonna ask for a change but LGTM!

PR #2782, opened 2024-09-10
feat: Active detour panel button / icons

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208153509160251

But in addition to that, updated the button on that panel, since it's not covered in another ticket yet

### DIFF
--- a/assets/src/components/detours/activeDetourPanel.tsx
+++ b/assets/src/components/detours/activeDetourPanel.tsx
@@ -40,7 +40,7 @@ export const ActiveDetourPanel = ({
 
     <Panel.Body className="d-flex flex-column">
       <Panel.Body.ScrollArea>
-        <div className="d-flex flex-rown justify-content-between align-items-center">
+        <div className="d-flex justify-content-between align-items-center">
           {onNavigateBack && (
             <Button
               className="icon-link my-3"
--- a/assets/src/components/detours/activeDetourPanel.tsx
+++ b/assets/src/components/detours/activeDetourPanel.tsx
@@ -3,7 +3,7 @@ import { DetourDirection } from "../../models/detour"
 import { Button, ListGroup } from "react-bootstrap"
 import { Panel } from "./diversionPage"
 import { Stop } from "../../schedule"
-import { ArrowLeft } from "../../helpers/bsIcons"
+import { ArrowLeft, ExclamationTriangleFill, StopCircle } from "../../helpers/bsIcons"
 import { AffectedRoute, MissedStops } from "./detourPanelComponents"
 
 export interface ActiveDetourPanelProps {
@@ -36,17 +36,19 @@ export const ActiveDetourPanel = ({
 
     <Panel.Body className="d-flex flex-column">
       <Panel.Body.ScrollArea>
-        {onNavigateBack && (
-          <Button
-            className="icon-link my-3"
-            variant="outline-primary"
-            onClick={onNavigateBack}
-          >
-            <ArrowLeft />
-            Back
-          </Button>
-        )}
-
+        <div className="d-flex flex-rown justify-content-between align-items-center">
+          {onNavigateBack && (
+            <Button
+              className="icon-link my-3"
+              variant="outline-primary"
+              onClick={onNavigateBack}
+            >
+              <ArrowLeft />
+              Back
+            </Button>
+          )}
+          <ExclamationTriangleFill className="c-active-detour__alert-icon" />
+        </div>
         <AffectedRoute
           routeName={routeName}
           routeDescription={routeDescription}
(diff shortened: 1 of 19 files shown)


assets/src/components/detours/activeDetourPanel.tsx
```diff
@@ -85,8 +91,13 @@ export const ActiveDetourPanel = ({
       </Panel.Body.ScrollArea>
 
       <Panel.Body.Footer>
-        <Button className="flex-grow-1 m-3" onClick={onDeactivateDetour}>
-          Deactivate Detour
+        <Button
+          variant="ui-alert"
+          className="flex-grow-1 m-3 icon-button"
```
firestack, line 96: can you try `icon-link` instead?
We shouldn't need custom css for this AFAIK
```suggestion
          className="flex-grow-1 m-3 icon-link"
```

assets/src/components/detours/activeDetourPanel.tsx
```diff
@@ -36,17 +40,19 @@ export const ActiveDetourPanel = ({
...
+            <Button
+              className="icon-link my-3"
+              variant="outline-primary"
+              onClick={onNavigateBack}
+            >
+              <ArrowLeft />
+              Back
+            </Button>
+          )}
+          <ExclamationTriangleFill className="c-active-detour__alert-icon" />
```
firestack, line 54:
```suggestion
          <ExclamationTriangleFill className="fs-2" />
```

assets/src/components/detours/activeDetourPanel.tsx
```diff
@@ -36,17 +40,19 @@ export const ActiveDetourPanel = ({
...
-            className="icon-link my-3"
-            variant="outline-primary"
-            onClick={onNavigateBack}
-          >
-            <ArrowLeft />
-            Back
-          </Button>
-        )}
-
+        <div className="d-flex flex-rown justify-content-between align-items-center">
```
firestack, line 43:
```suggestion
        <div className="d-flex flex-row justify-content-between align-items-center">
```

hannahpurcell: 😂 well I guess I don't need flex-row at all! It was not necessary
```suggestion
        <div className="d-flex justify-content-between align-items-center">
```

assets/src/components/detours/activeDetourPanel.tsx
hannahpurcell: This replaces `width` and `height` css with `font-size` css, which doesn't achieve the same effect

firestack: From what I understand and see on bootstraps docs, this _should_ work, so I'd expect something about our setup is broken if this _doesn't_ work (and I confirmed locally it doesn't)

firestack: Nevermind, sorry!

assets/src/components/detours/activeDetourPanel.tsx
hannahpurcell: Oh that does get us most of the way there! Cool. Now need `text-light`, and I can do that inline.

One other thing though: the gap for `icon-link` is .375rem, while our [design system](https://www.figma.com/design/ui7Ni9uoeRrX7NLp3Hh5a6/Skate-Component-Library-(Bootstrap-5)?node-id=2658-914&t=UXiDIp8EJRIjVaNT-4) uses .5rem. It seems we should make an override for the gap of icon-link?

firestack: I'd go back to Design and ask if this is something we can standardize on. Ideally either they adopt the Bootstrap default, or we change the `$icon-link-gap` variable.

assets/src/components/detours/activeDetourPanel.tsx
firestack: If you go to a bootstrap icon, and edit the classes defined in the HTML of an icon in the Examples section to have a `fs-<x>` you'll find that it does react to `fs-<x>`.

This works because they have `width` and `height` set to `1em`. I kindof like this usage and would consider making it a utility class to make the `icon-font-square` a reusable class since our icon sizes seem to lineup with font sizes. 

Just an idea for more reusable css but I'm good with this as is I think.

assets/src/components/detours/activeDetourPanel.tsx
hannahpurcell: I chitchatted with Brian, and realized that it pretty much is standardized. All instances where we use `icon-link` as a utility in our code actually has the wrong margin, compared with the designs. The designs, whenever there is an icon with text, have a .5rem padding between the icon and text. I'm [checking](https://mbta.slack.com/archives/C024N3SKX33/p1726075371554479) with the designers now to see if this will continue to be a standard going forward

hannahpurcell: Got confirmation from both Kathleen and Brian that .5rem is the standard gap going forward

firestack (approved):

PR #2783, opened 2024-09-10
feat: Active detour panel header yellow for those in detour-pilot group

hannahpurcell:
Asana [ticket](https://app.asana.com/0/1205385723132845/1208153509160224)

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -92,6 +92,9 @@
   display: flex;
   justify-content: flex-end;
   align-items: center;
+  &.active-detour {
+    background-color: $color-lemon-100;
+  }
 }
 
 .c-diversion-panel__h1 {
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -1,3 +1,5 @@
+@use "./color/definitions" as semantic;
+
 .l-diversion-page {
   display: grid;
   // Default layout is linear
@@ -92,8 +94,12 @@
   display: flex;
   justify-content: flex-end;
   align-items: center;
+}
+
+.l-diversion-page__header,
+.l-diversion-page__panel {
   &.active-detour {
-    background-color: $color-lemon-100;
+    background-color: semantic.$active-detour;
   }
 }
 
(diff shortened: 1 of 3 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -164,11 +165,20 @@ export const DiversionPage = ({
...
     <>
       <article className="l-diversion-page h-100 border-box inherit-box">
-        <header className="l-diversion-page__header text-bg-light border-bottom">
+        <header
+          className={joinClasses([
+            "l-diversion-page__header",
+            "border-bottom",
+            snapshot.matches({ "Detour Drawing": "Active" }) &&
+            userInTestGroup(TestGroups.DetoursPilot)
+              ? "active-detour"
```
firestack, line 174: looks like `--bs-bg-opacity` isn't set correctly so it's not showing up, but I think we should be able to use `text-bg-ui-alert` ?

assets/css/_diversion_page.scss
```diff
@@ -92,6 +92,9 @@
   display: flex;
   justify-content: flex-end;
   align-items: center;
+  &.active-detour {
+    background-color: $color-lemon-100;
```
joshlarson, line 96: Idea: Since I'm going to use the same color for the background of the active detours table, could/should we make this into a semantic color, named `$active-detour` or something like that?

Non-blocking.

assets/src/components/detours/diversionPage.tsx
hannahpurcell: Ahhh I was wondering _what_ wasn't showing up, but I see that things look wonky on mobile. Is that what you're talking about? I'll fix it (but also the mobile view will be reworked further later)

firestack: I was mainly thinking that we should be able to use the bootstrap classes for background colors and that when I tried it `text-bg-ui-alert` didn't work due to `--bs-bg-opacity` defaulting to completely transparent. I know `ui-alert` isn't lemon-100, so it's fine for now.

PR #2794, opened 2024-09-13
feat: Reorganizes detour API items 

hannahpurcell:
Now that a bit more detour work is done, I had insight into how I should organize some of the model locations / route names a bit better. This PR

- stubs out the API frontend call
- moves the frontend models
- renames the routes


assets/src/models/detoursList.ts
```diff
@@ -0,0 +1,59 @@
...
+  past: nullable(array(SimpleDetourFromApi)),
+})
+
+export type GroupedDetoursFromApi = Infer<typeof GroupedDetoursFromApi>
+
+export const groupedDetoursFromData = (
+  groupedDetours: GroupedDetoursFromApi
+): GroupedSimpleDetours => ({
+  active:
+    groupedDetours.active?.map((detour) => simpleDetourFromData(detour)) ||
```
joshlarson, line 53: Question (non-blocking): Is the `|| null` necessary?

lib/skate_web/router.ex
```diff
@@ -159,9 +159,10 @@ defmodule SkateWeb.Router do
     get "/location_search/place/:id", LocationSearchController, :get
     get "/location_search/search", LocationSearchController, :search
     get "/location_search/suggest", LocationSearchController, :suggest
+    get "/detours", DetoursController, :detours
```
joshlarson, line 162: I like this - it does feel most REST-y.

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -134,7 +134,7 @@ defmodule SkateWeb.DetoursControllerTest do
     test "fetches detours from database and groups by active, past, draft", %{conn: conn} do
       author_id = populate_db_and_get_user(conn)
 
-      conn = get(conn, "/api/detours/get_detours")
+      conn = get(conn, "/api/detours")
```
joshlarson, line 137: Maybe not for this PR, but I think it's considered best practice in Phoenix to refer to URL's with `~p`, so like `conn = get(conn, ~p"/api/detours")`

Non-blocking

joshlarson: Not for this PR because it looks like it was already this way.

joshlarson (approved):

lib/skate_web/router.ex
```diff
@@ -159,9 +159,10 @@ defmodule SkateWeb.Router do
...
+    get "/detours/:detour_id", DetoursController, :detour
```
joshlarson, line 163: I think possibly the addition of this endpoint snuck in ahead of the associated controller change? 

This (and the associated change in `api.ts`) I would recommend yanking from this PR and moving into https://github.com/mbta/skate/pull/2795

hannahpurcell: That would balloon this PR quite a bit I think, so I kept the controller change out of this PR by choice, given nowhere in the UI references this route, and it doesn't cause any test failures

joshlarson: I did notice that it didn't seem to be breaking anything, which is why I approved. 

I think what I was suggesting would shrink this PR, not balloon it?

hannahpurcell: Ohhhhhhh I misread! Yank these changes from this PR to #2795, got it. I don't quite understand how to cherry pick lines in a file and commit some of those line changes to one branch but other line changes to another branch. Like, most of the changes in `api.ts` should be here, but the two route changes should be in the other PR. Thoughts @firestack?

assets/src/models/detour.ts
```diff
@@ -1,64 +1,30 @@
...
+import { any, Infer, number, string, type } from "superstruct"
 
-export const SimpleDetour = type({
-  route: string(),
-  direction: string(),
-  name: string(),
-  intersection: string(),
+export const DetourWithState = type({
+  author: string(),
+  state: any(),
```
firestack, line 8: Oh, I'm sorry I didn't see this sooner, but we don't need to define our internal structs with superstruct (and probably shouldn't)

We only need superstruct definitions for parsing boundries, so `DetourWithStateFromApi` should be `DetourWithStateData` and we should have a `type` or `interface` named `DetourWithState`

firestack: I'm a little conflicted, on one hand, this PR is not introducing new structs, but just renaming them, on the other hand I'd prefer this not to slip longer just because "that's not what this PR is for".

If you want to do that in another PR I'd be fine with that and could call this non-blocking.

assets/src/models/detoursList.ts
firestack, line 1: https://github.com/mbta/skate/pull/2794/files#r1763744778

lib/skate_web/router.ex
firestack: If you could get it into the other PR that'd be preferred but it's not in the user flow yet so I think it's fine.

firestack (changes requested):
two comments but also looks like we could merge this today!

assets/src/models/detoursList.ts
hannahpurcell: Ope forgot to check. I think it's necessary; if `active?` is not found, it will return `undefined`, and the type expects a `null`. So removing the `|| null` results in: `Type 'undefined' is not assignable to type '{ route: string; direction: string; name: string; intersection: string; updatedAt: number; }[] | null'.`

firestack: Huh! You learn something new every day! My intuition told me that `{ a: null }.a?.nested` would return `null`, but yeah it returns `undefined`!

hannahpurcell: Since I refactored the models, I did tweak this a bit! Let me know what you think

assets/src/models/detour.ts
hannahpurcell: Ahhhhhh I get it! Yeah I was very happy to make the fix in this PR. Thanks for catching

firestack (approved):
🎉

PR #2795, opened 2024-09-13
feat: Detour controller updates for single detour

hannahpurcell:
I moved around some detour models and added the new detour controller function

Depends on:
- https://github.com/mbta/skate/pull/2794


lib/skate/detours/detours.ex
```diff
@@ -141,6 +122,34 @@ defmodule Skate.Detours.Detours do
...
+      iex> get_detour_with_state!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  @spec get_detour_with_state!(integer()) :: DetourWithState.t()
+  def get_detour_with_state!(id) do
+    detour =
+      Detour
+      |> Repo.get!(id)
+      |> Repo.preload(:author)
```
firestack, line 144: suggestion: I would either replace the `Repo.get!` with `get_detour!`, or do the preload in a single query

1. 
```suggestion
      get_detour!(id)
      |> Repo.preload(:author)
```

```suggestion

      Detour
      from d in Detour, where: d.id == ^id, preload: [:author]
      |> Repo.one!()
```

assets/src/models/detoursList.ts
```diff
@@ -1,6 +1,7 @@
 import { array, Infer, nullable, number, string, type } from "superstruct"
 
 export const SimpleDetour = type({
+  uuid: number(),
```
firestack, line 4: question: do you feel strongly about naming this `uuid`?
we're not using UUID's at this point, so maybe we just call it `id`?

hannahpurcell: totally! I'm... not sure why I leaned into this naming here.

firestack (approved):

PR #2797, opened 2024-09-13
feat: Open a detour from detour list frontend

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1207527734305946

Forgot to add this! Depends on:
- https://github.com/mbta/skate/pull/2795


assets/tests/components/detours/detoursListPage.openDetour.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    
+    const { baseElement } = render(<DetourListPage />)
+
+    await userEvent.click(screen.getByText("Headsign Z"))
+
+    await waitFor(() => {
+      expect(
+        screen.getByRole("heading", { name: "Share Detour Details" })
+      ).toBeVisible()
+    })
```
firestack, line 107:
```suggestion
    expect(
      await screen.findByRole("heading", { name: "Share Detour Details" })
    ).toBeVisible()
```

assets/tests/components/detours/detoursListPage.openDetour.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+    machine.send({ type: "detour.edit.done" })
+
+    const snapshot = machine.getPersistedSnapshot()
+    machine.stop()
+
+    jest.mocked(fetchDetour).mockResolvedValue(Ok({updatedAt: 1726147775, author: "fake@email.com", state: snapshot}))
+    
+    const { baseElement } = render(<DetourListPage />)
+
+    await userEvent.click(screen.getByText("Headsign Z"))
```
firestack, line 101:
```suggestion
    await userEvent.click(await screen.findByText("Headsign Z"))
```

assets/tests/components/detours/detourListPage.test.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { neverPromise } from "../testHelpers/mockHelpers"
-import { Ok } from "../../src/util/result"
+import { DetourListPage } from "../../../src/components/detourListPage"
+import { fetchDetours } from "../../../src/api"
+import { neverPromise } from "../../testHelpers/mockHelpers"
+import { Ok } from "../../../src/util/result"
 import { render, screen } from "@testing-library/react"
 
-jest.useFakeTimers().setSystemTime(new Date("2024-08-29T20:00:00"))
+jest.spyOn(Date, "now").mockImplementation(() => 1724961600000)
```
firestack, line 9: question: what's the benefit of this change? Personally, I liked the previous way and felt it was a better way of doing things, and I'd like to update the rest of Skate to use `useFakeTimers`, was there something that necessitated this change? :o

assets/src/components/detourListPage.tsx
```diff
@@ -29,19 +60,35 @@ export const DetourListPage = () => {
...
           originalRoute={{}}
         />
       )}
+
+      {showDetourModal && stateOfDetourModal && (
+        <DetourModal
+          onClose={onCloseDetour}
+          show
+          originalRoute={{}}
+          {...(stateOfDetourModal ? { snapshot: stateOfDetourModal } : {})}
```
firestack, line 89:
```suggestion
          snapshot={stateOfDetourModal}
```

assets/src/components/detourListPage.tsx
```diff
@@ -1,21 +1,52 @@
...
+      [detourId]
+    ),
+  })
+  const detour =
+    detourResult && isOk(detourResult) ? detourResult.ok : undefined
+
+  useEffect(() => {
+    const state = isValidSnapshot(createDetourMachine, detour?.state)
+    if (isOk(state)) setStateOfDetourModal(state.ok)
+  }, [detour])
```
firestack, line 37: suggestion(non-blocking): you don't have to do this, but since our only usage of the detour result is the ok state, you could condense this into a single `useEffect` via the `useApiCall` hook

```suggestion
  const { result: stateOfDetourModal } = useApiCall({
    apiCall: useCallback(
      async () => {
        if (detourId === undefined) { return undefined }

        const detourResponse = await fetchDetour(detourId)

        if (isErr(detourResponse)) { return undefined }

        const snapshot = isValidSnapshot(createDetourMachine, detourResponse.ok.state)
        if (isErr(snapshot)) { return undefined }

        return snapshot.ok
      },
      [detourId]
    ),
  })
```

assets/src/components/detourListPage.tsx
```diff
@@ -29,19 +60,35 @@ export const DetourListPage = () => {
...
+            <DetoursTable data={detours.draft} onOpenDetour={onOpenDetour} />
+          )}
+          {detours.past && (
+            <DetoursTable data={detours.past} onOpenDetour={onOpenDetour} />
+          )}
         </>
       )}
 
-      {showDetourModal && (
+      {/* Need to be separate modals to force the modal's state machine to refresh */}
```
firestack, line 75:
```suggestion
      {/* React's state caching requires that separate modals are used to force the state machine to refresh. the `key` prop didn't reset the state either when tried with `key={detourId ?? ""}` */}
```

firestack (changes requested):
a few notes but otherwise looks like we could merge this today!

assets/src/components/detourListPage.tsx
```diff
@@ -1,21 +1,52 @@
...
+    ),
+  })
+  const detour =
+    detourResult && isOk(detourResult) ? detourResult.ok : undefined
+
+  useEffect(() => {
+    const state = isValidSnapshot(createDetourMachine, detour?.state)
+    if (isOk(state)) setStateOfDetourModal(state.ok)
+  }, [detour])
+
```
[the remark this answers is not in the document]
firestack: you could also do this for `detours` _if_ desired, but that's less relevant because it doesn't require 2+n render loops

```
  const { result: detours } = useApiCall({
    apiCall: useCallback(async () => {
      const detoursResult = await fetchDetours()

      if (isErr(detoursResult)) { return undefined }

      return detoursResult.ok
    }, []),
  })
```

assets/tests/components/detours/detourListPage.test.tsx
hannahpurcell: Oh!! Ok, so the `useFakeTimers` wasn't working for the new tests I wrote, and then I was reverse-influenced by how we use the spyOn approach everywhere else. I'll revert this change at least, and then see if I can get `useFakeTimers` to work in the new `openDetour` tests

hannahpurcell: Fixed!

assets/src/components/detourListPage.tsx
hannahpurcell: Oh yep, relic of trying to combine into one modal previously

assets/src/components/detourListPage.tsx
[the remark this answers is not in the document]
hannahpurcell: Beeeaauuutiful, yeah I love this

assets/src/components/detourListPage.tsx
hannahpurcell: Ok so, I went the two-modal method for a different reason, that I just realized I was thinking about it the wrong way. The `key` prop does reset the state, and now that I adjusted the `DetourModal` conditional better, things render well. 

I needed to delay showing the modal between the time where `detourId` exists but `stateOfDetourModal` doesn't exist yet. Updated the comment to reflect that!

firestack: Oh interesting! I was testing locally and didn't think about that and was having issues!

firestack (approved):
🍰

PR #2798, opened 2024-09-18
feat: create detour notification table

firestack:
Asana Ticket: https://app.asana.com/0/0/1208312289539633/f

Not sure if this PR is too small, but I'm fairly certain this will be the final shape of the table and that could be merged first.

This adds a field to the notifications table, because there were errors if it wasn't added, but it's intended that this table is entirely unused as of this PR, following PRs will use this table.

If this PR is too small I can roll this table creation into a PR that uses it.


hannahpurcell (approved):
Not too small! Looks good 👍

PR #2787, opened 2024-09-11
feat: Give the detours table timestamp columns different names depending on detour status

joshlarson:
# Before

Screenshot 2024-09-11 at 10 44 55 AM

# After

Screenshot 2024-09-11 at 10 45 11 AM

---

Asana Ticket: https://app.asana.com/0/1205718271156548/1208214740600737/f

Depends on:
- https://github.com/mbta/skate/pull/2786

PR #2800, opened 2024-09-19
feat: create detour status enum and "Timestamp Name"

firestack:
changes: https://github.com/mbta/skate/pull/2787 (and it's dependants)

### DIFF
--- a/assets/src/components/detoursTable.tsx
+++ b/assets/src/components/detoursTable.tsx
@@ -9,6 +9,25 @@ interface DetoursTableProps {
   data: SimpleDetour[]
 }
 
+export enum DetourStatus {
+  Draft = "draft",
+  Active = "active",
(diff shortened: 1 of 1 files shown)


assets/src/components/detoursTable.tsx
```diff
@@ -9,6 +9,25 @@ interface DetoursTableProps {
   data: SimpleDetour[]
 }
 
+export enum DetourStatus {
+  Draft = "draft",
+  Active = "active",
+  Closed = "closed"
+}
+
+export const timestampNameForStatus = (status: DetourStatus) => {
```
firestack, line 18: I'd also like to do this
```suggestion
export const timestampLabelFromStatus = (status: DetourStatus) => {
```

hannahpurcell (approved):

PR #2802, opened 2024-09-19
feat: Update route ladder dropdown

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208333281632131

~~Pushing up my current work. I'm again confused by tests; to be fixed when I get back from time off~~

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -29,6 +29,10 @@
       border-radius: 0.25rem;
       max-width: 20.5rem;
     }
+
+    .dropdown-divider {
+      border-color: $color-gray-300;
+    }
   }
 
   .c-route-ladder__dropdown-button {
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -21,7 +21,7 @@
       padding-right: 0.5rem;
     }
 
-    &.non-skate-alert .dropdown-menu {
+    &.c-route-ladder__dropdown--non-skate-alert .dropdown-menu {
       width: 20.5rem;
     }
 
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -1,3 +1,5 @@
+@use "./color/tokens_2024" as new_tokens;
+
 .c-route-ladder__header,
 .c-route-ladder__controls {
   margin: 0 0.5rem;
@@ -12,7 +14,7 @@
     padding: 6px 8px;
   }
 
-  .c-route-ladder__header__action-container {
+  .c-route-ladder__dropdown {
     display: inline-flex;
     .dropdown,
     .c-route-ladder__alert-icon {
@@ -22,12 +24,38 @@
 
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
-    width: 2.25rem;
+    width: 1.8125rem;
+    padding: 0;
+    justify-content: center;
+    align-items: center;
+    background-color: white;
+    color: black;
+    border-radius: 0.25rem;
+
+    &:active {
+      background-color: unset;
+      color: unset;
+    }
 
     &.dropdown-toggle::after {
-      // Force bootstrap carat to center even with hidden text
-      margin-left: 0;
+      // Hide bootstrap carat
+      display: none;
     }
+  }
+
+  .c-route-ladder__dropdown.non-skate-alert .dropdown-menu {
+    width: 20.5rem;
+  }
+
+  .c-route-ladder__dropdown .dropdown-menu {
+    border-radius: 0.25rem;
+    max-width: 20.5rem;
+  }
+
+  .c-route-ladder__dropdown-header-text {
+    font-size: 0.875rem;
+    font-weight: 600;
+    color: new_tokens.$gray-800;
   }
 
   .c-route-ladder__alert-icon {
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -20,6 +20,15 @@
     .c-route-ladder__alert-icon {
       padding-right: 0.5rem;
     }
+
+    &.non-skate-alert .dropdown-menu {
+      width: 20.5rem;
+    }
+
+    .dropdown-menu {
+      border-radius: 0.25rem;
+      max-width: 20.5rem;
+    }
   }
 
   .c-route-ladder__dropdown-button {
@@ -41,15 +50,6 @@
       // Hide bootstrap carat
       display: none;
     }
-  }
-
-  .c-route-ladder__dropdown.non-skate-alert .dropdown-menu {
-    width: 20.5rem;
-  }
-
-  .c-route-ladder__dropdown .dropdown-menu {
-    border-radius: 0.25rem;
-    max-width: 20.5rem;
   }
 
   .c-route-ladder__dropdown-header-text {
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -96,7 +96,7 @@ export const Header = ({
                 <Dropdown.Item className="icon-link" onClick={onClickAddDetour}>
                   <PlusSquare /> Add detour
                 </Dropdown.Item>
-                { hasAlert &&
+                {hasAlert && (
                   <>
                     <Dropdown.Divider />
                     <Dropdown.Header>
@@ -109,8 +109,7 @@ export const Header = ({
                       <a href="https://www.mbta.com/">mbta.com</a> or in IRIS.
                     </Dropdown.ItemText>
                   </>
-                }
-                
+                )}
               </Dropdown.Menu>
             </Dropdown>
           )}
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -98,7 +98,7 @@ export const Header = ({
                 </Dropdown.Item>
                 {hasAlert && (
                   <>
-                    <Dropdown.Divider />
+                    <Dropdown.Divider className="border-top-0" />
                     <Dropdown.Header>
                       <div className="c-route-ladder__dropdown-header-text">
                         Active detours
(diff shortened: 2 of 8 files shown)


assets/tests/components/routeLadder.test.tsx
```diff
@@ -259,6 +259,89 @@ describe("routeLadder", () => {
...
+        hasAlert={true}
+      />
+    )
+
+    act(() => {
+      fireEvent.click(screen.getByRole("button", { name: "28 Route Options" }))
+    })
+
+      expect(
+      await screen.findByText("This route has an active detour.")
```
joshlarson, line 295:
```suggestion
      screen.getByText(/This route has an active detour./)
```

The problem here isn't the waiting - `act(() => fireEvent...)` is running through the full round of react loops and making the dropdown visible, so that part is working just fine.

The problem is that `findByText` (and its friends `getByText`, `queryByText`, etc) do full-string matching, not substring matching, so if the text content here is `This route has an active detour. View detour details on mbta.com or IRIS.`, then `findByText` won't find it.

There are a few ways to tell `findByText` to use substring matching. I think replacing string `"This route has an active detour."` with regex `/This route has an active detour./` looks nicest, but you can also explicitly tell `findByText` to do substring matching by giving it the `exact: false` option:

```suggestion
      await screen.findByText("This route has an active detour.", { exact: false })
```

(In my original suggestion, I also replaced `findByText` with `getByText`, because the race condition isn't the problem here, and I would recommend doing that either way)

```suggestion
      screen.getByText("This route has an active detour.", { exact: false })
```

assets/tests/components/routeLadder.test.tsx
```diff
@@ -259,6 +259,89 @@ describe("routeLadder", () => {
...
+        toggleCrowding={() => {}}
+        ladderDirections={{}}
+        ladderCrowdingToggles={{}}
+        hasAlert={true}
+      />
+    )
+
+    act(() => {
+      fireEvent.click(screen.getByRole("button", { name: "28 Route Options" }))
+    })
```
joshlarson, line 292:
```suggestion
    await userEvent.click(
      screen.getByRole("button", { name: "28 Route Options" })
    )
```

When I was playing around with this on my computer, I ran into some of those annoying `An update occurred outside of act` errors. This seemed to go away when using `userEvent` instead of `fireEvent` (I don't really know why, but I do remember other reasons that `userEvent` is supposed to be better than `fireEvent`, so I try to use it when possible)

assets/tests/components/routeLadder.test.tsx
hannahpurcell: I had no ideaaaaaa it was a string matching issue! Fantastic, thanks for help and breaking down my options

assets/tests/components/routeLadder.test.tsx
hannahpurcell: Great! Thank you

assets/src/components/routeLadder.tsx
```diff
@@ -63,25 +67,49 @@ export const Header = ({
   return (
     <Card className="c-route-ladder__header">
       <Card.Body>
-        <div className="c-route-ladder__header__action-container">
+        <div
+          className={joinClasses([
+            "c-route-ladder__dropdown",
+            hasAlert && "non-skate-alert",
```
joshlarson, line 73: I'd love to get @firestack's take on this, because she knows BEM better than I do, but I _think_ that we want to name `non-skate-alert something like `c-route-ladder__dropdown--non-skate-alert` to make it clear that it's a modifier of `c-route-ladder__dropdown`.

```suggestion
            hasAlert && "c-route-ladder__dropdown--non-skate-alert",
```

assets/css/_route_ladder.scss
```diff
@@ -22,14 +24,40 @@
 
   .c-route-ladder__dropdown-button {
     height: 2.25rem;
-    width: 2.25rem;
+    width: 1.8125rem;
```
joshlarson, line 27: Can any of this be done with bootstrap utilities? I suspect the width can't, because bootstrap isn't flexible enough, but I'm thinking a bunch of the others can.

assets/css/_route_ladder.scss
```diff
@@ -22,14 +24,40 @@
...
   }
 
+  .c-route-ladder__dropdown.non-skate-alert .dropdown-menu {
+    width: 20.5rem;
+  }
+
+  .c-route-ladder__dropdown .dropdown-menu {
+    border-radius: 0.25rem;
+    max-width: 20.5rem;
+  }
```
joshlarson, line 53: Can this section be combined with the `.c-route-ladder__dropdown` section above in a sassy way to look like:

```css
  .c-route-ladder__dropdown {
    display: inline-flex;
    .dropdown,
    .c-route-ladder__alert-icon {
      padding-right: 0.5rem;
    }

    &.non-skate-alert .dropdown-menu {
      width: 20.5rem;
    }

    .dropdown-menu {
      border-radius: 0.25rem;
      max-width: 20.5rem;
    }
  }
```

assets/src/components/routeLadder.tsx
```diff
@@ -63,25 +67,49 @@ export const Header = ({
   return (
     <Card className="c-route-ladder__header">
       <Card.Body>
-        <div className="c-route-ladder__header__action-container">
+        <div
```
joshlarson, line 70: **Praise**: I very much prefer the name `...dropdown` over `...action-container`

assets/src/components/routeLadder.tsx
```diff
@@ -63,25 +67,49 @@ export const Header = ({
...
+                  </div>
+                </Dropdown.Header>
                 <Dropdown.Item className="icon-link" onClick={onClickAddDetour}>
-                  <PlusSquare />
-                  Add detour
+                  <PlusSquare /> Add detour
                 </Dropdown.Item>
+                {hasAlert && (
+                  <>
+                    <Dropdown.Divider />
```
joshlarson, line 101: Screenshot 2024-09-25 at 11 44 57 AM

The border here is a bit dark and heavy compared to the designs. I think the problem is both that it's the wrong color (`#808080` versus `gray-300`, which is `#d4d7db`) and that it's 2px instead of 1px.

You can fix the color by adding a `gray-300` mapping to `_map_overrides`, and then giving this element a `className` of `border-gray-300`. Then you can make it only 1px high by getting rid of the top or bottom border, like so:

```suggestion
                    <Dropdown.Divider className="border-gray-300 border-top-0" />
```

Result:

Screenshot 2024-09-25 at 11 44 35 AM

(Caveat: adding a `gray-300` mapping to `_map_overrides` might be _evil_ in some way. At this point, I'm tempted to do it, because I'm frustrated by bootstrap's inflexibility, and want it to have more tailwind-style utilities, but I dunno, maybe that's evil. I think the alternative is more custom CSS.)

assets/css/_route_ladder.scss
hannahpurcell: Why bulk up a lot of inline styles via utilities if this className needs to exist anyway?

assets/css/_route_ladder.scss
hannahpurcell: Sure, I've done it in the way you describe, but I've also seen folks on our team avoid it, so what makes one better than the other?

assets/src/components/routeLadder.tsx
hannahpurcell: Yesssss thanks, action-container I suppose used to hold other things at some point maybe?? But not anymore!

assets/src/components/routeLadder.tsx
hannahpurcell: Nixed my last comment, because I see the value of adding the gray-300 color change, though `_map_overrides` is (so far) just for the map. I'll find another place to make the utility

assets/css/_route_ladder.scss
joshlarson: I think the major advantage is that it reduces CSS bundle size, and is one of the major 

Caveat: I don't feel strongly about this, but I've gotten similar feedback in the past, and found the utilities to be quite nice once I wrapped my head around them. I'd be interested in @firestack's perspective on this too, because I think she understands this better than I do.

assets/src/components/routeLadder.tsx
joshlarson: I had to add another utility-class color to `_map_overrides` (`active-detour`, I think), so if/when you find a better place, let me know! 😅 

(Also, `_map_overrides` _works_, so if you want to avoid being blocked on finding the right home for this, then you could do `_map_overrides` for now as a stopgap, and move it later.)

assets/css/_route_ladder.scss
firestack: I'm not _personally_ a fan of the tailwind-style-utility classes, but in the interest of lowering the barrier to entry of styling to Skate, and using common sizes from design, I do prefer using utility classes, as long as we're trying to use bootstrap.

The benefit of something like `ps-1`(padding start size 1) is that if we decide later that 1 should be bigger or smaller, it applies everywhere, instead of needing to track down every single instance.

assets/src/components/routeLadder.tsx
hannahpurcell: I didn't prefer adding to `_map_overrides` and didn't feel strongly about keeping this out of the `_route_ladders` scss, so [here](https://github.com/mbta/skate/pull/2802/commits/6b976ac5e0c03f6285d9ab5d614eb2cd85a30c50) was the change

joshlarson (approved):
Approving because this looks great, and the out-standing CSS questions require a team-wide consensus that doesn't exist yet. Not that they shouldn't be addressed, but I'm not ready to withhold my approval for this feature on it.

assets/css/_route_ladder.scss
hannahpurcell: Ok, so in the future, I should go heavy on the inline utilities, and _additionally_ use CSS for further styling? I figured even having the styling class at all would bulk up the CSS bundle, but perhaps not.

Merging and adding this to our coffee chat list!

PR #2803, opened 2024-09-20
feat:doc: stub doc headers for database tables

firestack:
`mix docs` doesn't generate anything for these modules if `@moduledoc false` is set, but we're using `TypedEctoSchema` now which _does_ generate type information, which can be surfaced by `ExDoc`.

It's also impossible to link to modules in documentation if `@moduledoc false` is set.

I've also created a ExDoc group which is enabled via setting the option on `@moduledoc` `section: :ecto`, this should make it easier to find our database related code, and easier to navigate in the documentation.

---

It is possible to do this automatically for Ecto schemas, by setting this attribute in the shared `Skate.Schema` module. I do think this is neat, but I'm not really _certain_(i.e., let's discuss) this is a good decision?

```diff
diff --git a/lib/skate/schema.ex b/lib/skate/schema.ex
index d18c8b83e3..cde6e814cd 100644
--- a/lib/skate/schema.ex
+++ b/lib/skate/schema.ex
@@ -5,6 +5,8 @@
 
   defmacro __using__(_opts) do
     quote do
+      @module_doc section: :ecto
+
       use TypedEctoSchema
     end
   end
```

I like that this means we don't need to _remember_ to add `section: :ecto` to all new Ecto schemas, because we have a growing list of "conventions" which are difficult to keep all in your head. But on the other hand, putting the `@module_doc` in the schema macro removes the choice later to easily change the `section: :ecto` value, I looked into making the `@module_doc` setting conditional, but wasn't very successful on my first attempt.

### DIFF
--- a/lib/notifications/db/block_waiver.ex
+++ b/lib/notifications/db/block_waiver.ex
@@ -1,5 +1,7 @@
 defmodule Notifications.Db.BlockWaiver do
-  @moduledoc false
+  @moduledoc """
+  Ecto Model for `block_waivers` Database table
+  """
 
   use Skate.Schema
   import Ecto.Changeset
--- a/lib/notifications/db/bridge_movement.ex
+++ b/lib/notifications/db/bridge_movement.ex
@@ -1,5 +1,7 @@
 defmodule Notifications.Db.BridgeMovement do
-  @moduledoc false
+  @moduledoc """
+  Ecto Model for `bridge_movements` Database table
+  """
 
   use Skate.Schema
   import Ecto.Changeset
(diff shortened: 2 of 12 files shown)


joshlarson:
@firestack 

> But on the other hand, putting the @module_doc in the schema macro removes the choice later to easily change the section: :ecto value, I looked into making the @module_doc setting conditional, but wasn't very successful on my first attempt.

I'm not worried about this. If it turns out to be a problem later on, we can remove it from the `__using__` section in `schema` and put it somewhere else.

joshlarson (approved):
I went back and forth because the generated docs are pretty sparse. But you did advertise these as stubs, and they do have the potential to fix some broken links. 

I'm good with this as-is. I'm also good with you updating the macro itself as you suggested in the PR description.

PR #2804, opened 2024-09-23
tweak: Preload author and retrieve detours in one query

hannahpurcell:
No asana task, adhoc!

### DIFF
--- a/lib/skate/detours/detours.ex
+++ b/lib/skate/detours/detours.ex
@@ -11,7 +11,7 @@ defmodule Skate.Detours.Detours do
   alias Skate.Settings.User
 
   @doc """
-  Returns the list of detours.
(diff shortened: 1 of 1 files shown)


firestack (approved):
🎉 (there are credo things to fix but looks good!)

lib/skate/detours/detours.ex
```diff
@@ -11,17 +11,20 @@ defmodule Skate.Detours.Detours do
...
+  Returns the list of detours with author, sorted by updated_at
 
   ## Examples
 
       iex> list_detours()
       [%Detour{}, ...]
   """
   def list_detours do
-    Detour
+    (detour in Skate.Detours.Db.Detour)
```
firestack, line 22: woah! I didn't think the macro would allow you to do this :o

PR #2806, opened 2024-09-23
feat: Add timestamp and author to detour modal header

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208153509160249

Closing #2801; this one has collapsed commits


assets/css/_diversion_page.scss
```diff
@@ -9,10 +9,6 @@
     "map" 80vh
     / 1fr;
 
-  &__header {
-    grid-area: heading;
-  }
-
```
firestack, line 15: question/issue: why was this removed?

firestack: Seems like the intention was to get the "header" area to not be behind the panel?

This does that, but only by accident. Instead, the grid should be changed so that you can target the specific area to put the header in.

assets/src/components/detourListPage.tsx
```diff
@@ -34,7 +34,7 @@ export const DetourListPage = () => {
       if (isErr(snapshot)) {
         return undefined
       }
-      return snapshot.ok
+      return detourResponse
```
firestack, line 37:
```suggestion
      return detourResponse.ok
```

Then on lines +92 to +98
```tsx
          {...(detour
            ? {
                snapshot: detour.state,
                author: detour.author,
                updatedAt: detour.updatedAt,
              }
            : {})}
```

firestack (approved):

assets/src/components/detourListPage.tsx
hannahpurcell: Ha! Totally, thanks ✅

assets/src/components/detourListPage.tsx
```diff
@@ -115,13 +115,19 @@ export const DetourListPage = () => {
...
           onClose={onCloseDetour}
           show
           originalRoute={{}}
           key={detourId ?? ""}
-          {...(stateOfDetourModal ? { snapshot: stateOfDetourModal } : {})}
+          {...(detour
+            ? {
+                snapshot: detour.ok.state,
+                author: detour.ok.author,
+                updatedAt: detour.ok.updatedAt,
```
hannahpurcell, line 128:
```suggestion
                snapshot: detour.state,
                author: detour.author,
                updatedAt: detour.updatedAt,
```

assets/css/_diversion_page.scss
hannahpurcell: OH, I heartily agree. I find the current grid layout super confusing, given that the header is tucked behind the panel, but also agree I shouldn't just torch it for this work. That said, we decided in slack to go ahead and merge, and then pair today to sort the grid out

PR #2805, opened 2024-09-23
feat: Deactivate detour flow

joshlarson:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208153509160255/f

Depends on:
- https://github.com/mbta/skate/pull/2793

PR #2811, opened 2024-09-24
feat: Suppress "Return to regular route" button for non-dispatchers

joshlarson:
No Asana Ticket, but I noticed this while playing around with detours, and didn't think it should be lumped in with https://github.com/mbta/skate/pull/2805.

Depends on:
- https://github.com/mbta/skate/pull/2805
  - Not a real semantic conflict, but doing it in sequence avoids annoying git conflicts.


hannahpurcell (approved):
I like adding the identity check on the function for opening the deactivate modal 👍

PR #2813, opened 2024-09-25
 feat: add detour notifications to notifications module

firestack:
Step 1 of adding detour notifications to Skate. To be able to send detour notifications, we first need to create the functions and setup for storing and retrieving this info from the database.

This will be used by `notification_server` in future PR's.

The entire feature can be seen in context on the branch https://github.com/mbta/skate/compare/main...kf/asn/detour-notifications

### DIFF
--- a/lib/notifications/db/detour.ex
+++ b/lib/notifications/db/detour.ex
@@ -9,7 +9,11 @@ defmodule Notifications.Db.Detour do
   @derive {Jason.Encoder,
            only: [
              :__struct__,
-             :status
+             :status,
+             :headsign,
+             :route,
+             :direction,
+             :origin
            ]}
 
   typed_schema "detour_notifications" do
@@ -17,6 +21,12 @@ defmodule Notifications.Db.Detour do
     has_one :notification, Notifications.Db.Notification
 
     field :status, Ecto.Enum, values: [:activated]
+
+    # Derived from the associated detour
+    field :headsign, :any, virtual: true
+    field :route, :any, virtual: true
+    field :direction, :any, virtual: true
+    field :origin, :any, virtual: true
   end
 
   def changeset(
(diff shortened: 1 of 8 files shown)


test/support/factory.ex
```diff
@@ -415,8 +415,31 @@ defmodule Skate.Factory do
...
     }
   end
+
+  def with_direction(%Skate.Detours.Db.Detour{} = detour, :inbound) do
+    put_in(detour.state["context"]["routePattern"]["directionId"], 1)
+  end
+
+  def with_direction(%Skate.Detours.Db.Detour{} = detour, :outbound) do
+    put_in(detour.state["context"]["routePattern"]["directionId"], 0)
+  end
```
firestack, line 444: I just kinda did this, happy to chat about doing this a different way

lib/notifications/notification.ex
```diff
@@ -141,26 +201,17 @@ defmodule Notifications.Notification do
...
-      )
+    cutoff_time = now_fn.() - @notification_expiration_threshold
 
-    query
+    base()
+    |> select_user_read_state(user_id)
+    |> select_bridge_movements()
+    |> select_block_waivers()
+    |> select_detour_info()
+    |> where([notification: n], n.created_at > ^cutoff_time)
```
firestack, line 213: hmm, gonna need to figure out a better way to do this now... since detours can be active for longer than the cutoff time..

lib/notifications/db/detour.ex
```diff
@@ -33,4 +43,60 @@ defmodule Notifications.Db.Detour do
...
+    The "base" query that queries `Notifications.Db.Detour`'s without restriction
+    """
+    def base() do
+      from(d in Notifications.Db.Detour, as: :detour_notification, select_merge: d)
+    end
+
+    @doc """
+    Retrieves detour information for notifications from the `Notifications.Db.Detour` table
+    """
+    def get_derived_info(query \\ base()) do
```
joshlarson, line 64: **Question**: Is this something that could also be used in the `get_detours` endpoint to avoid loading and processing the full JSON blobs from each detour when loading the detour page tables?

lib/notifications/db/detour.ex
```diff
@@ -33,4 +43,60 @@ defmodule Notifications.Db.Detour do
       :status
     ])
   end
+
+  defmodule Queries do
```
joshlarson, line 47: This module feels worth testing explicitly, because I imagine we're going to re-use it a lot.

Those tests could probably be added as a fast-follow, but I wouldn't hate to see them added directly to this PR.

lib/notifications/db/detour.ex
firestack: I think so!

lib/notifications/db/notification.ex
```diff
@@ -68,4 +68,120 @@ defmodule Notifications.Db.Notification do
...
+    `Notifications.Db.Notification`'s and retrieves the Detour's
+    associated info.
+
+    ## Example
+
+    The `query` parameter defaults to
+    `Notifications.Db.Notification.Queries.base()`, but can be specified
+    explicitly
+
+        iex> Notifications.Db.Notification.Queries.base()
```
joshlarson, line 110: This looks like a doctest, but I don't think it's running, and am not sure it would pass. 

Is it worth it to set up the plumbing required to make this a runnable doctest?

firestack: We could pair on this? I don't think the contents of the return value is important, so I'd benefit from some pairing/guidance there, but I do like the idea of running doctests here

lib/notifications/db/detour.ex
firestack: Right now I feel comfortable with the level of testing done in the `notifications_test`, because every line of code in this is required and tested (IIRC, I could've missed something).

I'd be open to writing a module to test it specifically, but I don't find it very high priority and I'm having trouble thinking of what that would look like _today_. Maybe revisiting this sometime in the next week I'll be in a better space to take that on.

Alternatively, I'd be happy to pair on it!

lib/notifications/db/detour.ex
firestack: I could spend 30m exploring this, do you think it's worth it? I was worried about not having tests for a larger change like that and that does change the footprint, we could do this as a follow-up PR though

lib/notifications/db/detour.ex
firestack: Done!

lib/notifications/db/detour.ex
firestack: This was quite interesting to explore, but probably too big for this PR, I'll open it as a follow on.

joshlarson: I am very pro follow-up PR for this, and excited that this change will help make some previously-annoying things less annoying!

lib/notifications/db/detour.ex
```diff
@@ -33,4 +43,105 @@ defmodule Notifications.Db.Detour do
...
+        ...> |> Notifications.Notification.create_activated_detour_notification_from_detour()
+        ...>
+        iex> all_detour_notifications =
+        ...>   Notifications.Db.Detour.Queries.base()
+        ...>   |> Skate.Repo.all()
+        ...>
+        iex> match?(
+        ...>   [
+        ...>     %Notifications.Db.Detour{
+        ...>     }
```
joshlarson, line 73:
```suggestion
        ...>     %Notifications.Db.Detour{}
```

Sooooper non-blocking, but I think it looks nicer collapsed

lib/notifications/db/detour.ex
```diff
@@ -33,4 +43,105 @@ defmodule Notifications.Db.Detour do
...
+      from(
+        [detour_notification: dn] in query,
+        left_join: ad in assoc(dn, :detour),
+        as: :associated_detour,
+        select_merge: %{
+          route: ad.state["context"]["route"]["name"],
+          origin: ad.state["context"]["routePattern"]["name"],
+          headsign: ad.state["context"]["routePattern"]["headsign"],
+
+          # Ecto can't figure out how to index a JSON map via another JSON value
```
joshlarson, line 119: **Praise** This explanation is very helpful for this query, both for what that `CAST` is doing and for why it's there in the first place.

lib/notifications/db/detour.ex
firestack: agreed, thanks!

lib/notifications/notification.ex
joshlarson: Do we even need/want to limit detour lifespan? I figured that a detour would just be active until it's deactivated.

test/notifications/notification_test.exs
```diff
@@ -315,4 +322,77 @@ defmodule Notifications.NotificationTest do
...
+        :detour
+        |> insert()
+        |> Notifications.Notification.create_activated_detour_notification_from_detour()
+      end
+
+      # assert it is in the database
+      assert count == Skate.Repo.aggregate(Notifications.Db.Detour, :count)
+    end
+
+    test "creates a unread notification for all users" do
```
joshlarson, line 341:
```suggestion
    test "creates an unread notification for all users" do
```

Nit: Grammar

test/notifications/notification_test.exs
```diff
@@ -315,4 +322,77 @@ defmodule Notifications.NotificationTest do
...
+        )
+        |> Notifications.Notification.create_activated_detour_notification_from_detour()
+
+      detour =
+        Notifications.Db.Notification
+        |> Skate.Repo.get!(detour.id)
+        |> Skate.Repo.preload(:users)
+
+      # assert all users have a notification that is unread
+      assert Kernel.length(detour.users) == number_of_users
```
joshlarson, line 360: Nit: I think this variable represents the notification, not the detour itself

```suggestion
      detour_notification =
        Notifications.Db.Notification
        |> Skate.Repo.get!(detour.id)
        |> Skate.Repo.preload(:users)

      # assert all users have a notification that is unread
      assert Kernel.length(detour_notification.users) == number_of_users
```

test/support/factory.ex
joshlarson: I personally like it at first glance.

joshlarson (approved):
Made a couple of nits, and I think this looks good!

lib/notifications/notification.ex
firestack: You're correct, but block waivers and bridge movements don't agree and do require this, so I need a way to only apply this for non detour notifications

PR #2814, opened 2024-09-25
feat: Update activate detour modal to match hi-fi designs

joshlarson:
# Before

Screenshot 2024-09-26 at 11 11 21 AM

# After

Screenshot 2024-09-26 at 11 11 36 AM

---

Asana Ticket: https://app.asana.com/0/1148853526253420/1208390832296805/f

### DIFF
--- a/assets/src/components/detours/activateDetourModal.tsx
+++ b/assets/src/components/detours/activateDetourModal.tsx
@@ -84,7 +84,7 @@ const SurroundingModal = ({
 
 interface StepSubtitleProps extends PropsWithChildren {}
 const StepSubtitle = ({ children }: StepSubtitleProps) => (
-  <h5 className="fs-5 fw-semibold mb-3 mt-0">{children}</h5>
+  <h5 className="fs-5 fw-semibold mb-3 mt-1">{children}</h5>
 )
 
 const SelectingDuration = ({
@@ -103,6 +103,7 @@ const SelectingDuration = ({
     <Form>
       {possibleDurations.map((duration) => (
         <Form.Check
+          className="mb-2"
           onChange={() => {
             onSelectDuration(duration)
           }}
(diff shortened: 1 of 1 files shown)


hannahpurcell (commented):
Great! I like having the StepSubtitle. 

Some things to consider: 
- I think there needs to be a close button in the top right of the modal (in addition to the cancel button)
- Seems like there needs to be more padding between the radio options
- Validation needs to be done. Do that here?

joshlarson:
@hannahpurcell I think I've addressed your feedback, and this is ready for another 👀

hannahpurcell:
> I was honestly thinking that the easiest way to validate would be to ship to skate-dev and have Brian do validation there, with any fixes being fast-follow PR's.

Oh I meant form validation! I should have attached [a link](https://www.figma.com/design/XGmmwVROXdOvnWtJeUFC50/Skate-%7C-Unplanned-Detours-%7C-Handoff?node-id=3678-2705&t=RFMXKhBGPaGLeWyR-4)

firestack (approved):

joshlarson:
Ooohhhh

I left that out because I'm pretty sure that's an impossible state to get to. The `Next` button is unclickable when nothing is selected, but the only way to get the validation errors is by clicking on it when nothing is selected.

firestack:
> I left that out because I'm pretty sure that's an impossible state to get to. The `Next` button is unclickable when nothing is selected, but the only way to get the validation errors is by clicking on it when nothing is selected.

Oh oops sorry I didn't see @hannahpurcell's review. IIRC disabled buttons are a bad UX practice and they should do something when clicked and provide a message, so I'm pretty certain we should do this, if only to satisfy the designs. I personally like it being in the disabled state, but can we do that in addition to showing a error state?

joshlarson:
I talked a bit with Hannah about this in our 1:1.

I'd prefer to merge this as-is, and treat validation as a follow-up, especially since I'm uncertain about the level-of-effort associated with making a disabled button clickable, and treat validation (error feedback, disabled/enabled buttons, etc) as a follow-up.

firestack (approved):

PR #2815, opened 2024-09-25
feat: add detour notifications to notifications server 

firestack:
Step 2 of adding detour notifications to Skate. This creates the notification API for having the notification server create a new notification for a detour.

Asana Ticket: https://app.asana.com/0/0/1208393164852685/f

Depends on:
- #2813

### DIFF
--- a/lib/notifications/notification_server.ex
+++ b/lib/notifications/notification_server.ex
@@ -46,6 +46,7 @@ defmodule Notifications.NotificationServer do
 
   If the `:notify_finished` option is present, a `{:new_notification, detour: detour.id}` message
   is sent to the process referred to by the `:notify_finished` value.
+  This option has mainly been useful for testing code to avoid `Process.sleep()` calls.
   """
   @spec detour_activated(detour :: Skate.Detours.Db.Detour.t(), keyword()) :: :ok
   def detour_activated(
--- a/lib/notifications/notification_server.ex
+++ b/lib/notifications/notification_server.ex
@@ -36,6 +36,28 @@ defmodule Notifications.NotificationServer do
     GenServer.cast(server, {:bridge_movement, bridge_movement})
   end
 
+  @doc """
+  Creates a new Detour Activated Notification for a `Skate.Detours.Db.Detour`.
+
+  ## Options
+
+  If the `:server` option is present, the notification is sent to the process
+  referred to by the `:server` value.
+
+  If the `:notify_finished` option is present, a `{:new_notification, detour: detour.id}` message
+  is sent to the process referred to by the `:notify_finished` value.
+  """
+  @spec detour_activated(detour :: Skate.Detours.Db.Detour.t(), keyword()) :: :ok
+  def detour_activated(
+        %Skate.Detours.Db.Detour{} = detour,
+        options \\ []
+      ) do
+    server = Keyword.get(options, :server, default_name())
+    notify_finished = Keyword.get(options, :notify_finished, nil)
+
+    GenServer.cast(server, {:detour_activated, detour, notify_finished})
+  end
+
   def subscribe(user_id, server \\ default_name()) do
     registry_key = GenServer.call(server, :subscribe)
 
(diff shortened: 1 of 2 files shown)


lib/notifications/notification_server.ex
```diff
@@ -36,6 +36,29 @@ defmodule Notifications.NotificationServer do
...
+  ## Options
+
+  If the `:server` option is present, the notification is sent to the process
+  referred to by the `:server` value.
+
+  If the `:notify_finished` option is present, a `{:new_notification, detour: detour.id}` message
+  is sent to the process referred to by the `:notify_finished` value.
+  """
+  @spec detour_activated(detour :: Skate.Detours.Db.Detour.t()) :: :ok
+  @spec detour_activated(detour :: Skate.Detours.Db.Detour.t(), keyword()) :: :ok
```
hannahpurcell, line 51: This is so silly, but I didn't not realize until now that you can stack multiple specs for conditional parameters 😅 I've needed that before! That said, since you provide a default value for options, I don't think it's conditional anymore, so can the top spec be eliminated?

lib/notifications/notification_server.ex
```diff
@@ -36,6 +36,29 @@ defmodule Notifications.NotificationServer do
...
+  @doc """
+  Creates a new Detour Activated Notification for a `Skate.Detours.Db.Detour`.
+
+  ## Options
+
+  If the `:server` option is present, the notification is sent to the process
+  referred to by the `:server` value.
+
+  If the `:notify_finished` option is present, a `{:new_notification, detour: detour.id}` message
+  is sent to the process referred to by the `:notify_finished` value.
```
hannahpurcell, line 48: question: I don't quite see the situations where these options would be provided, and I checked your subsequent PR and don't see examples there either. Why do detours need these additional options, but the other notifications don't? (Not a request for more documentation. I'm glad these comments are here, but I still just don't understand the big picture.)

Reading some of the later comments: is :notify_finished partly included to support testing?

test/notifications/notification_server_test.exs
```diff
@@ -613,4 +613,55 @@ defmodule Notifications.NotificationServerTest do
       refute String.contains?(log, "new_notification")
     end
   end
+
+  # rename to activated_detour?
```
hannahpurcell, line 617: I don't mind the current naming as-is. If this is a comment for some future, fine to leave it, but seems like this can be removed.

hannahpurcell (approved):
I'm less proficient in reviewing GenServer work, but it helped that the notification_server already existed for other notification types, so I could follow the set pattern for detour notifs. So I only have a couple questions, but none of them block approval

test/notifications/notification_server_test.exs
firestack: Oh yes, thank you!
This used to be named `detour/3` and I changed it to `detour_activated/2` and didn't notice this comment was left over

lib/notifications/notification_server.ex
firestack: So I originally had this function written as `detour/3` => `detour(detour, server, notify_finished)`

This followed the other "new notification" functions but with a new `notify_finished` argument. As I was starting to work this this, I found myself not certain about the order of arguments (should `server` come first because it may be used more? or will `notify_finished` be used more?) That pointed me to think that I was actually approaching this wrong and the other signatures are what should be changed to use keyword lists, so I went with keywords. I struggled with _when_ to change the other signatures though, which is why I've only made this change for detours :/ .

But that said, I'm also not entirely happy with the `notify_finished` argument either. It was only useful for the tests for `notification_server` despite me _wishing_ it was useful for more tests, like the [`detours_controller` tests](https://github.com/mbta/skate/pull/2821/files#diff-6b5f7db841fd66bb4a09b0870661a72d957ad6cf263cccf090ebb4a251bc0687). I couldn't find a way to nicely inject that argument into the call for tests, and I also haven't found any places in the code where `notify_finished` useful for non-testing code, which is frustrating.
The only thing which made me think that it's worth keeping around, is that the `Skate.Detours.TripModificationPublisher`'s `:on_connect` option is essentially exactly the same as the `notify_finished` argument, and has also not been useful outside of tests, but has been a needed and elegant solution to the issue of waiting for an event to happen before continuing ([tests]; but ideally, maybe, in the future, additionally non-test code).

lib/notifications/notification_server.ex
firestack: Removed!

test/notifications/notification_server_test.exs
firestack: https://github.com/mbta/skate/compare/76a38d89c7a3879197589be193fb8d65d5e28cb4..f6923945f764b44ad4a0da4d65097f226e087e25

lib/notifications/notification_server.ex
hannahpurcell: Thank you so much!! Amazing and thoughtful reply. I like knowing it's primarily a testing argument right now, and I'm ok with having it in the hopes it will prove helpful for some future non-test code. But even if we don't, it's valuable to have it for testing.

firestack: I used to have a comment along these lines, but I dropped it apparently, thoughts?
```suggestion
  If the `:server` option is present, the notification is sent to the process
  referred to by the `:server` value.

  If the `:notify_finished` option is present, a `{:new_notification, detour: detour.id}` message
  is sent to the process referred to by the `:notify_finished` value.
  This option has mainly been useful for testing code to avoid `Process.sleep()` calls.
```

hannahpurcell: Oh yes, I do like that final line for clarity! Thanks

PR #2816, opened 2024-09-25
feat: Added StepperBar to activate detour flow

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208153509160239


assets/src/components/stepperBar.tsx
```diff
@@ -0,0 +1,187 @@
...
+)
+
+const StopDots = ({
+  stepWidth: stepWidth,
+  numDots: numDots,
+}: {
+  stepWidth: number
+  numDots: number
+}) => {
+  let dots = []
```
joshlarson, line 104: The build is being ornery about this because `dots` can be declared as a `const` here (technically I guess you can add things to a `const` array 🤯).

assets/src/components/stepperBar.tsx
```diff
@@ -0,0 +1,187 @@
...
+          stepCount={currentStep}
+          isPartial={true}
+        />
+        <UnfilledStepper
+          stepWidth={stepWidth}
+          stepCount={totalSteps - currentStep}
+        />
+      </>
+    )
+  }
```
joshlarson, line 34: Nit (non-blocking): You can get rid of the `let` here by using a ternary:


```suggestion
  const stepper =
    currentStep == totalSteps ? (
      <FilledStepper stepWidth={stepWidth} stepCount={currentStep} />
    ) : (
      <>
        <FilledStepper
          stepWidth={stepWidth}
          stepCount={currentStep}
          isPartial={true}
        />
        <UnfilledStepper
          stepWidth={stepWidth}
          stepCount={totalSteps - currentStep}
        />
      </>
    )
```

joshlarson (approved):
This looks great! 

The build is being grumpy about a `let` statement, and I did make one cosmetic suggestion, but that's non-blocking.

assets/src/components/stepperBar.tsx
hannahpurcell: Ahh yep! I do not make any changes to the variable itself

```suggestion
  const dots = []
```

PR #2817, opened 2024-09-25
feat: Contents for past detour page

joshlarson:
# Before

Screenshot 2024-09-25 at 3 56 20 PM

# After

Screenshot 2024-09-26 at 11 59 31 AM

---

This doesn't _exactly_ match the hi-fi designs, but I think it's still worth merging so that we don't have the obnoxious experience of having all the content disappear when a detour is deactivated.

The remaining things that I'm thinking of at the moment, which I think should be follow-up PR's are:
- Copy Details button in the header
- Created by and `Last used` in the header (kinda sorta blocked by https://github.com/mbta/skate/pull/2806, since that does this hi-fi task for the active detour panel)

Asana Ticket: https://app.asana.com/0/1205385723132845/1208391763317540/f

### DIFF
--- a/assets/src/components/detours/pastDetourPanel.tsx
+++ b/assets/src/components/detours/pastDetourPanel.tsx
@@ -8,7 +8,7 @@ import { AffectedRoute, MissedStops } from "./detourPanelComponents"
 
 export interface PastDetourPanelProps {
   directions?: DetourDirection[]
-  connectionPoints?: string[]
+  connectionPoints: [string, string]
   missedStops?: Stop[]
   routeName: string
   routeDescription: string
@@ -19,7 +19,7 @@ export interface PastDetourPanelProps {
 
 export const PastDetourPanel = ({
   directions,
-  connectionPoints,
+  connectionPoints: [connectionPointStart, connectionPointEnd],
   missedStops,
   routeName,
   routeDescription,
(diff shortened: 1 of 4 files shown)


assets/src/components/detours/pastDetourPanel.tsx
```diff
@@ -1,12 +1,87 @@
...
+import { DetourDirection } from "../../models/detour"
+import { Stop } from "../../schedule"
+import { ArrowLeft } from "../../helpers/bsIcons"
+import { Button, ListGroup } from "react-bootstrap"
+import { AffectedRoute, MissedStops } from "./detourPanelComponents"
 
-export const PastDetourPanel = () => (
+export interface PastDetourPanelProps {
+  directions?: DetourDirection[]
+  connectionPoints?: string[]
```
firestack, line 11: suggestion(non-blocking): Hmm, could we make it explicit how many points we expect?
```suggestion
  connectionPoints?: [string, string]
```

firestack: I think you could then de-structure it out of the params if you felt like it, but I wouldn't say you _have_ to

assets/src/components/detours/pastDetourPanel.tsx
```diff
@@ -1,12 +1,87 @@
...
+          routeName={routeName}
+          routeDescription={routeDescription}
+          routeOrigin={routeOrigin}
+          routeDirection={routeDirection}
+        />
+
+        {connectionPoints && (
+          <section className="pb-3">
+            <h2 className="c-diversion-panel__h2">Connection Points</h2>
+            <ListGroup as="ul">
```
firestack, line 59: note: I do find it annoying that the bootstrap react library doesn't default to `ul` >.>

firestack: I guess, I mean that I don't like that it defaults to a `div`, given it could be an `ol`

firestack (approved):

assets/src/components/detours/pastDetourPanel.tsx
joshlarson: Huh - i didn't know you could do that.

joshlarson: And since `connectionPoints` are actually always provided here, [I can destructure the value too](https://github.com/mbta/skate/pull/2817/commits/29420917b5680a1eef64aec6e3b933097e3affdd).

(I wonder if that can also be done for the `activeDetourPanel`, but that's a question for a different PR.)

PR #2820, opened 2024-09-26
feat: make `categorize_detour/2` public

firestack:


lib/skate/detours/detours.ex
```diff
@@ -95,19 +95,25 @@ defmodule Skate.Detours.Detours do
 
   @type detour_type :: :active | :draft | :past
 
+  @doc """
```
joshlarson, line 98: Non-blocking, but wouldn't hate this as a fast-follow: This is a good candidate for doctests.

lib/skate/detours/detours.ex
```diff
@@ -95,19 +95,25 @@ defmodule Skate.Detours.Detours do
...
+  Takes a `Skate.Detours.Db.Detour` struct and a `Skate.Settings.Db.User` id
+  and returns a `t:detour_type/0` based on the state of the detour.
+
+  returns `nil` if the state is indeterminate.
```
joshlarson, line 102: I think it actually returns `nil` for detours that are drafts belonging to someone else.

firestack: How about this?

```suggestion
  returns `nil` if the state is not one of `t:detour_type/0`.
```


I don't agree that this would _only_ return `nil` for drafts belonging to someone else, but I could add that as a clause

```suggestion
  if it is in a draft state, but does not belong to the provided user, returns `nil`
  returns `nil` if the state is not one of `t:detour_type/0`.
```
```suggestion
  returns `nil` if the state is not one of `t:detour_type/0`, or the draft does not belong to the provided user.
```

lib/skate/detours/detours.ex
joshlarson: Let's not block this PR on that - I'll queue it up as a fast-follow and see if I can get it done in the 16 hours remaining to me.

lib/skate/detours/detours.ex
joshlarson: I'm okay with that, but I'm reasonably sure that given the pattern-matching that's currently in place, the only scenario where `nil` will get returned is the scenario where it's a draft and you're not the author.

(This is based on 👇, which effectively says that a draft is anything that isn't active or past, since it doesn't check anything about the detour other than the author.)

```elixir
def categorize_detour(%{author_id: author_id}, user_id) when author_id == user_id, do: :draft
```

firestack: Oooh, yeah okay.

joshlarson (approved):
I disabled auto-merge so that you can respond to https://github.com/mbta/skate/pull/2820#discussion_r1777083384, but I think this looks good once that's addressed.

lib/skate/detours/detours.ex
firestack: https://github.com/mbta/skate/compare/427f1a3da80c6902959878d13207d060fbb36dfc..3676f43393db0641437d03c3ad7345868dce02d4

PR #2819, opened 2024-09-26
feat(ts/notifications): create notification card for detours

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1208393164852689/f

This creates the frontend part of notifications work, it's independent of other work because the parsing will not be triggered until the detour notification data is live and being sent.

Todo:
- [x] Tests(?)

PR #2825, opened 2024-09-26
feat: Open the detour modal from a notification

joshlarson:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207752079129768/f

### DIFF
--- a/assets/src/components/notificationCard.tsx
+++ b/assets/src/components/notificationCard.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from "react"
+import React, { ReactNode, useCallback, useState } from "react"
 import { useRoute, useRoutes } from "../contexts/routesContext"
 import {
   BlockWaiverNotification,
@@ -13,6 +13,13 @@ import { CardBody, CardProperties, CardReadable } from "./card"
 import { fullStoryEvent } from "../helpers/fullStory"
 import { RoutePill } from "./routePill"
 import inTestGroup, { TestGroups } from "../userInTestGroup"
+import { useApiCall } from "../hooks/useApiCall"
+import { fetchDetour } from "../api"
+import { createDetourMachine } from "../models/createDetourMachine"
+import { isValidSnapshot } from "../util/isValidSnapshot"
+import { isErr } from "../util/result"
+import { DetourModal } from "./detours/detourModal"
+import { DetourId } from "../models/detoursList"
 
 export const NotificationCard = ({
   notification,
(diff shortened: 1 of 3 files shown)


assets/tests/factories/notification.ts
```diff
@@ -84,3 +85,23 @@ export const bridgeLoweredNotificationFactory = Factory.define<
...
+
+const detourActivatedNotificationContentFactory =
+  Factory.define<DetourNotification>(() => ({
+    $type: NotificationType.Detour,
+    status: "activated",
+    detourId: 79,
+    headsign: "79 - Wooded Mountains",
+    route: "79",
+    direction: "Inbound",
+    origin: "Maple Knoll",
```
firestack, line 97: question: is it possible to not hardcode these values?

firestack: It's a bit of a struggle when tests assume hardcoded values from factories and then in the future we decide to make them generated, so I'd prefer to start with generated values.

I'd be open to pulling in faker @mbta/skate-developers

assets/src/components/notificationCard.tsx
firestack, line 1: note(non-blocking): I would like to change the `<NotificationCard/>` component into a `switch` which then renders some other component, like, `<DetourNotificationCard/>` to make it a bit simpler to understand what's going on (like the activate detour modal!) I'd be open to that happening in this PR, or a followup, or just noting it as a future tech debt item.

Based on the work you did for this @joshlarson does that seem worth it or do you think that it's beneficial to keep it this way for now? I'm worried this would be too early of a "component-ization"

assets/tests/factories/notification.ts
joshlarson: I agree - I did this just to get _something_ working, but I'd rather get non-hardcoded values in before merging.

assets/src/components/notificationCard.tsx
joshlarson: I think this is a good time to component-ize, but I'm on the fence about whether I want to do it in this PR versus teasing the different notification cards apart as a "fast"(indeterminate-speed)-follow.

assets/tests/factories/notification.ts
hannahpurcell: This was done in https://github.com/mbta/skate/pull/2819

assets/src/components/notificationCard.tsx
hannahpurcell: Given this PR has already changed hands, I think I'll consider this a "future do"

PR #2826, opened 2024-09-27
feat: Make Deactivate Modal match Hi-Fi designs

joshlarson:
# Before

Screenshot 2024-09-27 at 12 00 55 PM

# After

Screenshot 2024-09-27 at 5 49 12 PM

---

~One thing that's missing is that the button still has the wrong text (should say `Return to regular route` rather than `Confirm`). That can either be part of this PR or a follow-up.~

Asana Ticket: https://app.asana.com/0/1205526445275136/1208403460973162/f

### DIFF
--- a/assets/src/components/detours/activeDetourPanel.tsx
+++ b/assets/src/components/detours/activeDetourPanel.tsx
@@ -95,7 +95,7 @@ export const ActiveDetourPanel = ({
         {onOpenDeactivateModal && (
           <Button
             variant="ui-alert"
-            className="flex-grow-1 m-3 icon-link text-light"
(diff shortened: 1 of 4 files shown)


hannahpurcell:
> One thing that's missing is that the button still has the wrong text

I would include that in this PR. I'll start working on that!

joshlarson:
> > One thing that's missing is that the button still has the wrong text
>
> I would include that in this PR. I'll start working on that!

Ope! I forgot to update the description, but I think I actually did fix that before signing off last week.

hannahpurcell (approved):
Looks good! Made 2 very tiny tweaks: one to center the content of the button on the detour panel, one to add the closebutton to the modal

PR #2828, opened 2024-09-27
tweak: Move alert icon to route pill

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208333281632130

Talked with Kathleen async and decided a few things that are not captured in the handoff doc (given that the doc was medium-fi):

1. Make the route pills dynamically sized for the route ladders. (Rationale: if the 32/33 route had an alert, that's very wide. If we made all the pills that wide, it would look funny. So instead, give all pills 1.5rem padding.)
2. Minimum margin around the routePill: .5rem.
3. Alert + route number can be centered in the pill.

### DIFF
--- a/assets/css/_route_ladder.scss
+++ b/assets/css/_route_ladder.scss
@@ -65,22 +65,16 @@
   .c-route-ladder__alert-icon {
     display: flex;
     align-items: center;
-    width: 1.6rem;
 
     cursor: pointer;
 
     svg {
-      height: 1.6rem;
-      width: 1.6rem;
+      height: 1rem;
+      width: 1.125rem;
     }
   }
 
-  .c-route-ladder__route-pill {
-    grid-column: 2/3;
-  }
-
   .c-route-ladder__close-button-container {
-    grid-column: 3/4;
     margin: 2px 0 2px auto;
   }
 }
--- a/assets/css/_route_pill.scss
+++ b/assets/css/_route_pill.scss
@@ -11,15 +11,22 @@
   font-family: $font-family-route-pill;
 
   display: flex;
-  flex-direction: column;
+  flex-direction: row;
   justify-content: center;
+  align-items: center;
 
   &--large-format {
-    width: 4.5rem;
+    min-width: 4.5rem;
     height: 2.25rem;
     border-radius: 1.875rem;
     font-size: $h3-font-size;
     font-weight: 600;
+  }
+
+  &--dynamic-size {
+    min-width: unset;
+    padding: 0 1.5rem;
+    margin: 0 0.5rem;
   }
 }
 
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -20,11 +20,8 @@ import { LoadableTimepoints, Route, RouteId } from "../schedule.d"
 import IncomingBox from "./incomingBox"
 import Ladder from "./ladder"
 import Loading from "./loading"
-import Tippy from "@tippyjs/react"
-import { tagManagerEvent } from "../helpers/googleTagManager"
 import inTestGroup, { TestGroups } from "../userInTestGroup"
 import {
-  ExclamationTriangleFill,
   PlusSquare,
   ThreeDotsVertical,
 } from "../helpers/bsIcons"
@@ -75,6 +72,7 @@ export const Header = ({
         >
           {showDropdown && (
             <Dropdown className="border-box inherit-box">
+              {/* This is the thing getting labeled!! */}
               <Dropdown.Toggle
                 className="c-route-ladder__dropdown-button d-none d-sm-flex"
                 aria-labelledby={joinTruthy([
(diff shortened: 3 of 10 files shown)


firestack:
Hmm, I don't agree with moving the alert icon _into_ the route pill component, can you instead make the route pill component take either a `routeName` OR `children`?

I'm imagining something like this

```tsx
<RoutePill routeName={...} />
```

```tsx
<RoutePill>
    {hasAlert && (
      <Tippy
        content="Active detour"
        trigger="click"
        onShow={() => tagManagerEvent("alert_tooltip_clicked")}
      >
        <div className="c-route-ladder__alert-icon" aria-label="Route Alert">
          <ExclamationTriangleFill />
        </div>
      </Tippy>
    )}
	<RoutePill.RouteName routeName={...} />
</RoutePill>
```
OR name the component which takes the `children` `<RoutePill.Container/>` and then make `<RoutePill/>` internally use `<RoutePill.Container><RoutePill.RouteName/><RoutePill.Container/>`
```tsx
<RoutePill.Container>
    {hasAlert && (
      <Tippy
        content="Active detour"
        trigger="click"
        onShow={() => tagManagerEvent("alert_tooltip_clicked")}
      >
        <div className="c-route-ladder__alert-icon" aria-label="Route Alert">
          <ExclamationTriangleFill />
        </div>
      </Tippy>
    )}
	<RoutePill.RouteName routeName={...} />
</RoutePill>
```

hannahpurcell:
> Hmm, I don't agree with moving the alert icon into the route pill component, can you instead make the route pill component take either a routeName OR children?

Ahhh totally. I did it simply, without using `.Container` or `.RouteName`

assets/css/_route_ladder.scss
```diff
@@ -65,22 +65,16 @@
...
+      width: 1.125rem;
     }
   }
 
-  .c-route-ladder__route-pill {
-    grid-column: 2/3;
-  }
-
   .c-route-ladder__close-button-container {
-    grid-column: 3/4;
```
firestack, line 83: question: why was this deleted? it seems that the grid is still intact, so I'd assume that the only thing this changes is making the grid column it occupies implicit from DOM order instead of explicitly defined here.

firestack (approved):
one question but otherwise lgtm

assets/css/_route_ladder.scss
firestack: > And all I really need the grid to do here is allow the center cell to take the space it needs, and the first and third cell to have the same width. Is this the right way to do that?

That is the correct way to do that. For example, just so that we can precisely talk about what's happening, I'll give the columns names based on their numbers
```
1: 1fr
2: auto
3: 1fr
```

If you had this on a parent element (named `grid-container`), with 3 children without any styling
```html
<grid-container>
	<child-1/>
	<child-2/>
	<child-3/>
</grid-container>
```
and with this grid defined on the `grid-container`
```css
grid-container {
	/* inline or "regular" grid, doesn't matter */
	/* display: inline-grid; */
	/* display: grid; */
	grid-template-columns: 1fr auto 1fr;
}
```

`child-1` would be in grid column `1`, `child-2` would be in column `2`, and `child-3` in `3`.

If you rearranged them
```html
<grid-container>
	<child-1/>
	<child-3/>
	<child-2/>
</grid-container>
```
`child-1` would be in `1`, `child-2` would be in column `3`, and `child-3` would be in column `2`.

If instead these children had styling telling them which column to be in, 
```css
child-1 {
	grid-column: 1/2;
}

child-2 {
	grid-column: 2/3;
}

child-3 {
	grid-column: 3/4;
}
```

Then the style is now independent of the DOM order, which is good for two reasons.
A: It allows us to rearrange the order of things in the accessibility tree, and 

B: we don't need to worry that somehow accidentally the elements will get rearranged without noticing and the style then breaks because, from the first re-arranging example, `child-3` would not be stretched to `auto` (with the route ladders, this means the close button would be stretched).

A is beneficial because to a screen reader, right now reading the route ladder headers would read as "options, route <name>, remove" but honestly, it doesn't really make sense to hear that there are options, before you know what route you're operating on and the order _should_ be "route <name>, options, remove" which would have a example DOM order like this
```html
<grid-container>
	<child-2/> <!-- route -->
	<child-1/> <!-- options -->
	<child-3/> <!-- close -->
</grid-container>
```

So all-in-all, I feel more comfortable being explicit with grid styling then I do about having grid styling be implicit by DOM order.

hannahpurcell: Ohhhhhh interesting!! My accessibility expertise is stale / nonexistent, so I didn't realize the accessibility tree could be ordered differently from the DOM. The example you provide in regards to the route ladder header readout was the biggest "ohhhh" realization for me. Thanks, I'll tweak it

firestack: Just noting this from DMs, I wrote all that to point out what we _could_ do in the future by keeping the `grid-colunm` property explicit, and what happens if it's not explicit. Just to be clear, I'm **not** advocating for changing the DOM order _in this PR_.

PR #2831, opened 2024-09-30
feat(ex/notifications/notification_server): send notification to all connected Skate instances

firestack:
To be able to synchronize multiple Skate instances in production, we need to setup
distributed Elixir.

Following the way https://github.com/mbta/ride_along did it, this uses `DNSCluster`
to find the other Skate instances, and if the variable is not present, it does nothing
so this is safe to deploy to production before the corresponding DevOps PR.

Depends on:
- #2841 
- #2838


firestack:
I uhh, didn't mean to put this up for review just yet, as I was looking into if we could add tests. But we could merge without tests because I'm unsure about writing tests for distributed Elixir, I've been looking at https://elixirschool.com/en/lessons/advanced/otp_distribution#conditionally-excluding-tests-with-tags-11 but am still not certain how to do so.

hannahpurcell (approved):
Ok! Iiiiii think this makes sense to me and can't think of any feedback, so let's see it in action?

config/dev.exs
```diff
@@ -19,7 +19,7 @@ config :skate, Skate.Detours.TripModificationPublisher, start: false
 config :skate, SkateWeb.Endpoint,
   https: [
     ip: {0, 0, 0, 0},
-    port: 4000,
+    port: System.get_env("PORT") || 4000,
```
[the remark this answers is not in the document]
firestack: > But why not commit those changes to make my dev life easier? Hence my curiosity about what configurations do / do not get committed to the dev config file

I don't _personally_ think we should be using http for local dev, because all of Skate happens in https in prod, so I ignore the message from Phoenix because I think it's wrong. Additionally, we now use features in Skate that you _cannot_ use in a insecure environment (meaning served over http), this was first introduced when we started using the location information from the browser because browsers don't let you access that without https.

Having said this, if you agree, I think I should add a comment noting this :)

> 👀 what is the purpose of this change to only dev?

- For `test` envs, we don't need to think about this (afaik).
- For `prod`, this seems to already be configured to [use the port variable if it's present](https://github.com/mbta/skate/blob/03652e9cba76a91a8a224d0ffff440d984e4cb97/config/prod.exs#L26), but, this happens in `prod.exs` not `runtime.exs` so this variable is only respected at build time, not when it's deployed.
- For `dev`, when running multiple instances of Skate, it's necessary to change the port because two instances cannot use the same port, which is why I added this now.
 
I'd be open to considering moving this change into `runtime.exs` but I think we'd need to consider it a bit more than this change because that could have production implications, unless we gated it on `Mix.env() == :dev`.

lib/notifications/notification_server.ex
```diff
@@ -120,6 +139,37 @@ defmodule Notifications.NotificationServer do
...
+    #  in `NotificationServer`. To do this, we'd need to implement the same
+    # filtering mechanism that this module has implemented. For now, we'll
+    # send messages to other Skate instances letting them know about new
+    # Notifications.
+
+    # Skate instances currently do not "specialize", and therefore we need to
+    # send the notification to all instances
+    for node <- Node.list() do
+      Node.spawn(node, fn ->
+        server = Keyword.get(options, :server, default_name())
```
paulswartz, line 166: suggestion: you shouldn't need `Node.spawn` here. You can target the named server on another node with `{process_name, node}`: https://github.com/mbta/ride_along/blob/1d393329af28c00f080d37b730b1ad725e70a59d/lib/ride_along/rider_notifier.ex#L60
```elixir
server = Keyword.get(options, :server, default_name())
for node <- Node.list() do
  GenServer.cast({server, node}, {:detour_activated_notification, notification_id})
end
```

lib/skate/application.ex
```diff
@@ -38,6 +38,7 @@ defmodule Skate.Application do
         end ++
         [
           {Phoenix.PubSub, name: Skate.PubSub},
+          {DNSCluster, Application.get_env(:skate, DNSCluster) || [query: :ignore]},
```
paulswartz, line 41: suggestion: rather than the extra query here, I'd throw this into `config/config.exs`:
```elixir
config :skate, DNSCluster, query: :ignore
```

firestack: I'd thought of that but was on the fence, I think that's the right call though, thanks!

firestack: Done! [`057f2eb` (#2831)](https://github.com/mbta/skate/pull/2831/commits/057f2eb598097af9aa5642d8de7754eb03cf783a)

lib/notifications/notification_server.ex
firestack: Awesome, that works!
https://github.com/mbta/skate/compare/c9cf2debf1e309c4140e4b43753b2d8b26356a97..37026f15ff15bba582f0b85b6ad6e158bba2142d

If you have any thoughts on how to automatically test this, let me know? 
I was looking at https://elixirschool.com/en/lessons/advanced/otp_distribution#conditionally-excluding-tests-with-tags-11 but that didn't mention anything about automatically setting up a second test instance, and when I tried to do so, it seemed like I'd have to run a second real instance of Skate for the tests to run against.

paulswartz: I don't have any good answers for this, unfortunately. For RideAlong, I tested the code paths manually and tried to keep them simple.

If you don't have something already, you could test calling the `cast` and making sure is has the right effect?

lib/notifications/notification_server.ex
```diff
@@ -120,6 +139,35 @@ defmodule Notifications.NotificationServer do
...
 
+  defp broadcast_notification_to_other_instances(
+         %Notifications.Notification{
+           id: notification_id,
+           content: %Notifications.Db.Detour{}
+         },
+         options \\ []
+       )
+       when not is_nil(notification_id) do
+    server = Keyword.get(options, :server, default_name())
```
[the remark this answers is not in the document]
paulswartz: It needs to know the name of the server somehow, but there are bunch of ways:
- require a static name
- store the name in app config
- store the name in the GenServer state

lib/notifications/notification_server.ex
firestack: >> If you have any thoughts on how to automatically test this, let me know?
>
> I don't have any good answers for this, unfortunately. For RideAlong, I tested the code paths manually and tried to keep them simple.

This is what I've been doing locally
```sh
PORT=4001 iex \
	--sname 1@localhost \
	--cookie test \
	-S mix phx.server
```
```sh
iex \
	--sname 2@localhost \
	--cookie test \
	-S mix phx.server
```
And then run this on `2@localhost` shell
```ex
Node.connect(:"1@localhost")
```
(which means I've not been able to test `DNSCluster`, because it gets mad when I don't run it in release mode, as well as I don't think I can put a socket address (IP + Port) into what would be returned from DNS for `localhost`)

<details>
<summary>extra</summary>

or of course the corresponding call for the other server
```ex
Node.connect(:"2@localhost")
```

</details>

And then I open a browser window to both servers and create and activate a detour and confirm that the other browser window on the other server.

lib/notifications/notification_server.ex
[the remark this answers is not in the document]
firestack: Since ideally we'd move to using Phoenix Channel PubSub anyway, I'll put it in the app config, since then tests _could_ change the value.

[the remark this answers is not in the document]
firestack: Changed my mind, I think this is better https://github.com/mbta/skate/compare/37026f15ff15bba582f0b85b6ad6e158bba2142d..d4601dcbf421efc6fe8097d8cda8f9b112739755

paulswartz:
Code looks good (modulo the formatting issue) but I think it'd be good to put it up in one of the dev environments as well to make sure it works before merging.

firestack:
Checked dev-blue and confirmed no changes to the functionality of sending notifications, but I've not been able to test and confirm that multiple instances _work_.

firestack:
Ran a test yesterday and found that things weren't working. This seems to be for a few reasons
1. Elixir was not configured to use "names" instead of "short names", I don't know what this defaults too, but DNSCluster wouldn't even try to connect due to this. This is now set with a link to the documentation as well.
2. We did not have the Erlang EPMD and RPC ports opened, and the RPC ports weren't configured, they are as of commit "[fix: expose EPMD and Erlang RPC ports](https://github.com/mbta/skate/pull/2831/commits/4a6055b43b803a18b5c073b06fee248f042cecbd)"
3. AWS Security groups in devops aren't configured to allow communication yet, I've created a PR for this and am checking that it plans properly, but I'm going to deploy before that's deployed to make sure DNSCluster complains about not finding anything because it's not configured, or some other issue.
4. I'm not sure that I have `RELEASE_NODE` specified correctly as of this comment, I'm hoping DNSCluster will log some info about that.

hannahpurcell (approved):
👍

PR #2833, opened 2024-10-01
cleanup(ex/notifications/detour): remove changeset function

firestack:
just a little cleanup, I had originally created this function because that's what other notifications did. I think that was actually incorrect because we don't create detour notifications from external input, unlike bridge notifications or block waivers which get their data from external sources and need to validate the data.

To make it clear that detour notifications are not created from user input, I've removed the changeset function so we're not tempted to use it without question, and instead halve to think about it when/if we re-add the function


hannahpurcell (approved):
Got it, and appreciate the in-code comment

PR #2835, opened 2024-10-01
feat: Open the detour modal from a notification

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1148853526253420/1207752079129768/f

Essentially reopening https://github.com/mbta/skate/pull/2825 but with conflicts resolved and merging to main now

Sorry for the lingering broken test! Again, unsure how to force a wait in a way that would resolve the issue


assets/src/components/notificationCard.tsx
```diff
@@ -43,52 +57,117 @@ export const NotificationCard = ({
...
+          />
+        )}
+      </CardReadable>
+      {detourId && (
+        <DetourNotificationModal
+          show={showDetourModal}
+          detourId={detourId}
+          onClose={onCloseDetour}
         />
       )}
```
firestack, line 124: Hmm, I think we should gate this also on `showDetourModal`? because otherwise I think this is going to fetch and load every detour in the background whenever there's a notification?
```suggestion
      {detourId && showDetourModal && (
        <DetourNotificationModal
          show={showDetourModal}
          detourId={detourId}
          onClose={onCloseDetour}
        />
      )}
```

firestack: this _miiiiight_ fix the test too.

assets/tests/components/notificationCard.test.tsx
```diff
@@ -263,7 +272,7 @@ describe("NotificationCard", () => {
         />
       </RoutesProvider>
     )
-    expect(baseElement).toMatchSnapshot()
+    await waitFor(() => expect(baseElement).toMatchSnapshot())
   })
```
firestack, line 276: if https://github.com/mbta/skate/pull/2835#discussion_r1786716063 fixes this test failiure, I think we should try removing the async here as well, as I don't _think_ it's needed?

firestack (approved):
LGTM, approving since comments need to be resolved to merge, and if those are resolved (and tests pass) I'm good with this.

assets/tests/components/notificationCard.test.tsx
hannahpurcell: Yes! Fab, thank you

assets/src/components/notificationCard.tsx
hannahpurcell: Aha, thank you!! This was needed, as was a conditional around the <DetourModal/> to ensure the fetch for a detour returned before rendering.

firestack (approved):

PR #2837, opened 2024-10-03
feat: add detour deactivated notifications 

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1208409921049870/f

This implements deactivated notifications through the entire stack, from the notification internals, all the way to the deactivated card.

TODO:
- [x] Failing CI: Checks
	- There's something up with `eslint` not properly detecting that all possible branches of the `switch` are handled. Maybe I did something wrong here and I'd love that pointed out, but if I add an explicit `break` to make `eslint` happy, typescript then complains that the break is "unreachable" and "unnecessary". Still looking into this.

Depends On:
- ~#2831~
- #2836

### DIFF
--- a/assets/src/components/notificationCard.tsx
+++ b/assets/src/components/notificationCard.tsx
@@ -6,6 +6,7 @@ import {
   BlockWaiverReason,
   NotificationType,
   isBlockWaiverNotification,
+  DetourNotificationStatus,
 } from "../realtime"
 import { Route } from "../schedule"
 import { formattedTime } from "../util/dateTime"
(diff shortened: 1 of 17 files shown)


assets/src/components/notificationCard.tsx
```diff
@@ -182,20 +190,48 @@ const description = (
...
     case NotificationType.Detour: {
-      return (
-        <>
-          <div className="d-flex flex-row gap-2">
-            <RoutePill routeName={notification.content.route} />
-            <div>
-              <div className="fw-semibold">{notification.content.headsign}</div>
-              <div className="fw-normal text-body-secondary">
-                From {notification.content.origin.split(" - ")[0]}
+      switch (notification.content.status) {
```
hannahpurcell, line 193: The returns for both the `Activated` and `Deactivated` cases are identical. Is this switch needed?

firestack: Oh y'know, that's true, I'll remove the switch

lib/skate/detours/detours.ex
```diff
@@ -237,28 +243,49 @@ defmodule Skate.Detours.Detours do
...
-       )
-       when previous_status != :active do
-    if categorize_detour(detour, user_id) == :active do
-      Notifications.NotificationServer.detour_activated(detour)
-    end
+       ) do
+    send_notification(%{
+      next_detour: new_record,
+      previous: categorize_detour(previous_record, user_id),
+      next: categorize_detour(new_record, user_id)
```
hannahpurcell, line 263: non-blocking: What do you think of making `previous` / `next` into `previous_status` / `next_status` for unambiguity?

PR #2838, opened 2024-10-04
fix: split up build and deploy steps

firestack:
When deploying #2831 I was having difficulties with getting the deployment to succeed due to our database issues. It was getting pretty frustrating to have to build the docker container just to attempt a redeploy. I tried a few ways of parallelizing this, but so far this seems like the best option.

I was having trouble getting a deploy to work for testing this, so this also depends on #2841 
Depends on:
- #2841

### DIFF
--- a/.github/workflows/deploy-base.yml
+++ b/.github/workflows/deploy-base.yml
@@ -18,37 +18,24 @@ on:
       SLACK_WEBHOOK:
         required: true
 
+
 jobs:
-  deploy:
-    name: Deploy
+  build:
+    name: Build/push Docker Image
     runs-on: ubuntu-latest
     permissions:
       id-token: write
       contents: read
     environment: ${{ inputs.env }}
-    env:
-      ECS_CLUSTER: skate
-      ECS_SERVICE: skate-${{ inputs.env }}
+    outputs:
+      docker-tag-suffix: ${{ steps.build-push.outputs.docker-tag-suffix }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-base.yml
```diff
@@ -18,37 +18,24 @@ on:
...
     environment: ${{ inputs.env }}
-    env:
-      ECS_CLUSTER: skate
-      ECS_SERVICE: skate-${{ inputs.env }}
+    outputs:
+      docker-tag-suffix: ${{ steps.build-push.outputs.docker-tag-suffix }}
+      sentry-release: ${{ steps.version-ids.outputs.sentry-release }}
     steps:
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v4
```
hannahpurcell, line 34: The configure step went away, without an equivalent step needing to take its place?

hannahpurcell (approved):
Looks good, and tests well in your dev deployments. One question I don't understand, but otherwise, this seems reasonable.

.github/workflows/deploy-base.yml
firestack: I think it was unnecessary to begin with at the start of this PR because [`build-push-ecr`](https://github.com/mbta/actions/blob/main/build-push-ecr/action.yml#L36-L41) does it for us

PR #2840, opened 2024-10-04
feat: Active detour details are in a drawer on mobile

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208452134851701

I feel like I did things in a maybe janky way, but it's working! So I want to take another pass next week, but posting for early review.

- [ ] tests!


hannahpurcell:
Breaking this up

PR #2841, opened 2024-10-07
fix: update `tzdata` 

firestack:
We've been having deploy issues due to this dependency being out of date.

I thought that this required removing `fast_local_datetime` because I figured it pinned the dependency, but that doesn't seem to be the case. We were only using it to generate a ISO timestamp for the report filenames though, so I figured it was better to remove it and it was pretty simple.


hannahpurcell (approved):
👍

PR #2842, opened 2024-10-07
tweak: reorganize diversion page

hannahpurcell:
I found it mildly easier to read extracting the detourPanel logic out of the rest of the JSX tree, but this isn't expressly necessary if you prefer it the old way!

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -396,7 +396,7 @@ export const DiversionPage = ({
 
   return (
     <>
-      <article className={`l-diversion-page h-100 border-box inherit-box`}>
+      <article className="l-diversion-page h-100 border-box inherit-box">
         <header
           className={joinClasses([
             "l-diversion-page__header",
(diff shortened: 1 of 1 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -177,9 +177,226 @@ export const DiversionPage = ({
...
+      )
+    } else {
+      return <></>
+    }
+  }
+
   return (
     <>
-      <article className="l-diversion-page h-100 border-box inherit-box">
+      <article className={`l-diversion-page h-100 border-box inherit-box`}>
```
firestack, line 399: I don't _think_ it's an issue but I don't think this change is necessary?

firestack (approved):
one thing but otherwise LGTM!

assets/src/components/detours/diversionPage.tsx
hannahpurcell: Ah totally right, this is a consequence of splitting up the work -- in the following branch, I insert a classname based on isMobile, and forgot to revert the interpolation here. I'll roll back the interpolation for posterity, and then the following PR will bring it back
```suggestion
      <article className="l-diversion-page h-100 border-box inherit-box">
```

PR #2843, opened 2024-10-07
feat: Simple drawer component

hannahpurcell:
Currently only going to be used for the Active Detour mobile drawer, but can be applied to the shuttles / search map pages for later consistency, if design team approves


assets/css/_drawer_tab.scss
```diff
@@ -13,7 +13,6 @@
 }
 
 .c-drawer-tab__tab-button {
-  @include button-transparent;
```
hannahpurcell, line 16: Brian requested that the drawerTab not turn green when focused. I didn't know the history of adding this `button-transparent`, so lemme know if this is a no-go

firestack: We could make this more specific though, and remove this but add it specifically for the ladder pages and shuttle page

hannahpurcell: Oh thanks for reminding me, I left out a crucial detail! I did ask Brian if he would be open to reverting the green tab on the shuttle and ladder pages and he said yes, please. So with that in mind, how does it feel to just remove that style?

firestack: Hmm, then it has no focus states, which isn't _great_. But it's already not great elsewhere, so sure.

assets/src/components/drawer.tsx
```diff
@@ -0,0 +1,31 @@
...
+      >
+        <DrawerTab isVisible={drawerOpen} toggleVisibility={toggleDrawer} />
+        {children}
+      </div>
+      <button onClick={toggleDrawer} className="c-drawer__backdrop-button" />
+    </>
+  )
+}
+
+export default Drawer
```
firestack, line 31: issue: we're trying to move away from default exports

assets/src/components/drawer.tsx
```diff
@@ -0,0 +1,31 @@
...
+import DrawerTab from "../components/drawerTab"
+import { joinClasses } from "../helpers/dom"
+
+interface Props {
+  children?: ReactElement<HTMLElement>
+}
+
+const Drawer = ({ children }: Props): ReactElement<HTMLDivElement> => {
+  const [drawerOpen, setDrawerOpen] = useState(true)
+  const toggleDrawer = () => setDrawerOpen((drawerOpen) => !drawerOpen)
```
firestack, line 11: I'd personally prefer that the default drawer does not contain it's state, and instead has a wrapper component, with interfaces like `<Drawer.WithState startOpen onToggle />` `<Drawer open onToggle={() => {}} />`

Happy to commit this if you're open to it

```diff
diff --git a/assets/src/components/drawer.tsx b/assets/src/components/drawer.tsx
index 73c8f4a486..6fe4aed078 100644
--- a/assets/src/components/drawer.tsx
+++ b/assets/src/components/drawer.tsx
@@ -1,31 +1,54 @@
-import React, { ReactElement, useState } from "react"
+import React, { ComponentProps, PropsWithChildren, useState } from "react"
 import DrawerTab from "../components/drawerTab"
 import { joinClasses } from "../helpers/dom"
 
 interface Props {
-  children?: ReactElement<HTMLElement>
-}
-
-const Drawer = ({ children }: Props): ReactElement<HTMLDivElement> => {
-  const [drawerOpen, setDrawerOpen] = useState(true)
-  const toggleDrawer = () => setDrawerOpen((drawerOpen) => !drawerOpen)
+  open?: boolean
+  onToggleOpen?: () => void
+}
+
+const Drawer = ({
+  open = true,
+  onToggleOpen,
+  children,
+}: PropsWithChildren<Props>) => (
+  <>
+    <div
+      className={joinClasses([
+        "c-drawer",
+        ...(open
+          ? ["c-drawer--visible"]
+          : ["c-drawer--hidden", "u-hideable--hidden"]),
+      ])}
+    >
+      <DrawerTab
+        isVisible={open}
+        toggleVisibility={onToggleOpen || (() => {})}
+      />
+      {children}
+    </div>
+    <button onClick={onToggleOpen} className="c-drawer__backdrop-button" />
+  </>
+)
+
+interface DrawerWithStateProps {
+  startOpen?: boolean
+  onToggleOpen: Props["onToggleOpen"]
+}
+
+const DrawerWithState = (props: PropsWithChildren<DrawerWithStateProps>) => {
+  const [drawerOpen, setDrawerOpen] = useState(props.startOpen)
 
   return (
-    <>
-      <div
-        className={joinClasses([
-          "c-drawer",
-          ...(drawerOpen
-            ? ["c-drawer--visible"]
-            : ["c-drawer--hidden", "u-hideable--hidden"]),
-        ])}
-      >
-        <DrawerTab isVisible={drawerOpen} toggleVisibility={toggleDrawer} />
-        {children}
-      </div>
-      <button onClick={toggleDrawer} className="c-drawer__backdrop-button" />
-    </>
+    <Drawer
+      open={drawerOpen}
+      onToggleOpen={() => setDrawerOpen((drawerOpen) => !drawerOpen)}
+    >
+      {props.children}
+    </Drawer>
   )
 }
 
+Drawer.WithState = DrawerWithState
+
 export default Drawer
```

hannahpurcell: Ok, thinking about it! Why is your preference that Drawer doesn't manage its state? In your suggestion, state is scooted to a wrapper that effectively does the same thing, and can only ever be a wrapper for Drawer, so what is the purpose of that split? Also, why would `onToggleOpen` be an optional prop for `Drawer`? The purpose of a Drawer in Skate is always to be toggle-able, so this seems like it should be a mandatory prop

firestack: > Why is your preference that Drawer doesn't manage its state?

Because 19 times out of 20, design will ask us to control the state programmatically based on some other state[^1], and you can't control it if the state is burried within the component. IMO, it's also just better to separate state from a component if the state isn't essential to the core functionality. It just leaves more wiggle room up to the caller and is usually fairly easy to do, as shown by how simple the wrapper component is.

(side note: This also makes it easy to split out into using a context instead of a wrapper component as well.)

[^1]: See the following examples
    - Ladder page remembers the last state of the drawer across refreshes
    - So does the Shuttle page
    - The map page opens the drawer when you select a new vehicle
    - the map page closes the drawer if on mobile when you click the "go to map" button from the VPP
    - the map page closes the drawer if you select a search result while on mobile
    
	I'm sure there are more cases 😅
    From experience, I've found it's easier to start with a stateless component that provides a nice and simple wrapper rather than lumping the states together with the component from the start. Sometimes the refactors can become a bit painful once a component becomes used a lot.

---

But, let's say we did put the state inside the component,  

Then we're asked to open or close the drawer based on if we're on mobile once the page loads.  
We would probably add a "`startOpen`" prop.

But then we get another requirement that says we want to change the state if they change the window size, 

but then, `useState` ignores our `startOpen` prop because it's already rendered and `startOpen` only _initializes_ the state, but doesn't _change_ the state,

so then we end up with this pattern anyway, and we have to change a bunch of interfaces.


> Also, why would `onToggleOpen` be an optional prop for `Drawer`? The purpose of a Drawer in Skate is always to be toggle-able, so this seems like it should be a mandatory prop

This is just habit for testing, _usually_ it's kinda annoying when a prop is mandatory and it doesn't _need_ to be for _something_ to work. But we've started creating wrapper components in tests which provide default values so that we don't need to specify every prop[^2]. so the optional `onToggleOpen` is more taste and preference and I could go either way.

[^2]: I'd still make it optional because writing those wrapper components can be super annoying sometimes, but it's fine if we don't

hannahpurcell: I followed up on slack and forgot to follow up here! But this is a fantastic write-up, and I'm open to your changes, so would you like to commit what you proposed above?

firestack: Yep! Does this look okay? [`86aecd6` (#2843)](https://github.com/mbta/skate/pull/2843/commits/86aecd618bb9dd29dfdf730c0842a05d2ac30b6e)

assets/src/components/drawer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <button onClick={onToggleOpen} className="c-drawer__backdrop-button" />
+  </>
+)
+
+interface DrawerWithStateProps {
+  startOpen?: boolean
+  onToggleOpen: Props["onToggleOpen"]
+}
+
+const DrawerWithState = (props: PropsWithChildren<DrawerWithStateProps>) => {
```
hannahpurcell, line 39:
```suggestion
const Drawer = ({
  open = true,
  onToggleOpen,
  children,
}: PropsWithChildren<Props>) => (
  <>
    <div
      className={joinClasses([
        "c-drawer",
        ...(open
          ? ["c-drawer--visible"]
          : ["c-drawer--hidden", "u-hideable--hidden"]),
      ])}
    >
      <DrawerTab
        isVisible={open}
        toggleVisibility={onToggleOpen || (() => {})}
      />
      {children}
    </div>
    <button onClick={onToggleOpen} className="c-drawer__backdrop-button" />
  </>
)

interface DrawerWithStateProps {
  startOpen?: boolean
  onToggleOpen: Props["onToggleOpen"]
}

export const DrawerWithState = (props: PropsWithChildren<DrawerWithStateProps>) => {
```

assets/src/components/drawer.tsx
```diff
@@ -0,0 +1,54 @@
...
+      />
+      {children}
+    </div>
+    <button onClick={onToggleOpen} className="c-drawer__backdrop-button" />
+  </>
+)
+
+interface DrawerWithStateProps {
+  startOpen?: boolean
+  onToggleOpen: Props["onToggleOpen"]
```
hannahpurcell, line 36: Is this prop needed? It doesn't seem to be used

firestack: Oh shoot this was supposed to be used, but now that you mention it, yeah let's just remove it, if the caller wants to know when this changes, then they can just use `<Drawer>`

firestack:
```suggestion
```

assets/src/components/drawer.tsx
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+    <button onClick={onToggleOpen} className="c-drawer__backdrop-button" />
+  </>
+)
+
+interface DrawerWithStateProps {
+  startOpen?: boolean
+}
+
+export const DrawerWithState = (
```
firestack, line 38: I don't think this should be exported, and `Drawer` should be exported?

hannahpurcell: I saw your changes to the later PR first, so I fixed it [there](https://github.com/mbta/skate/pull/2844/commits/b8b1d4b873bc11666f13b650c780089eb2f8b69d)

firestack (approved):

PR #2847, opened 2024-10-10
feat(ts/components/vehicleMarker): use `ReactMarker` for `VehicleMarker` Icon and Label

firestack:
(Note: PR Reviewable by commits)

This is the refactor of converting the Vehicle Markers to use `ReactMarker` which has many benefits, but a large one being that when a vehicle updates, the entire element is not recreated and replaced (which is likely better for both performance, but also for Devtools in browsers)

We could probably combine the label and vehicle icon, but that's out of scope for this PR.

The snapshot changes are not _required_, but to do so requires creating a custom `DivIcon`. They have no effect on the final rendering in browser from the tests I've done. If we do want to avoid this (and I do!) see https://github.com/mbta/skate/pull/2848


hannahpurcell (approved):
Beautiful! Love the refactor and can't find a thing I'd change

hannahpurcell:
Approval given, but kinda want to review the follow-up PR before merging if that's ok?

PR #2848, opened 2024-10-10
feat(ts/components/reactDivIcon): create custom `DivIcon` for React portals [optional]

firestack:
This implements the changes required to remove the diffs on the `VehicleIcon` if we need it for #2847.


hannahpurcell (approved):
Ok this is way more magical to me, but the comments help guide me through it so I feel comfortable approving 👍

PR #2849, opened 2024-10-10
feat: Button icons for detour creation

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208153509160209/f


assets/src/helpers/bsIcons.tsx
```diff
@@ -321,6 +358,25 @@ export const PlusSquare = (props: SvgProps) => (
   </svg>
 )
 
+/**
+ * @returns https://icons.getbootstrap.com/icons/card-checklist/
```
firestack, line 362: issue: This seems like the wrong link?

assets/tests/components/detours/diversionPage.test.tsx
```diff
@@ -375,14 +375,14 @@ describe("DiversionPage", () => {
     })
 
     expect(
-      screen.getByRole("button", { name: "Review Detour", hidden: true })
+      screen.getByRole("button", { name: "Review", hidden: true })
```
firestack, line 378: oh instead of changing the text on these selectors, can you instead make the change to use `reviewDetourButton` from selectors?

assets/src/helpers/bsIcons.tsx
hannahpurcell:
```suggestion
 * @returns https://icons.getbootstrap.com/icons/power/
```

assets/tests/components/detours/diversionPage.test.tsx
firestack: - https://github.com/mbta/skate/pull/2849/files#diff-1535a72a38a9500ad346d1b3d9735aff2c788a2d97cfbf49f2aeaac7aefb26cfR1215
- https://github.com/mbta/skate/pull/2849/files#diff-1535a72a38a9500ad346d1b3d9735aff2c788a2d97cfbf49f2aeaac7aefb26cfR385

firestack (approved):
one issue(oh it's resolved already!), one improvement, otherwise looks good 👍🏻

assets/tests/components/detours/diversionPage.test.tsx
hannahpurcell: Oh of course!!

hannahpurcell: Ok I've been playing with this a bit, and I see that this particular assertion includes `hidden: true` in the button `get`, and I'm wondering why? What's the risk of excluding that option

firestack: If the tests pass, I don't _think_ there is a risk?

From the docs

> `hidden`: If `true` includes elements in the query set that are usually excluded from the accessibility tree. `role="none"` or `role="presentation"` are included in either case.

This would mean that if we were hiding the button via the `hidden` prop, then the query would fail, because looking at this from an accessibility perspective, you can't find a element that is hidden. So this is an escape hatch to let you do so.

hannahpurcell: Sorry for the delayed reply! Yeah, I can't get the test to pass for the one with hidden set, and that's partly because the assertion is a `.not.toBeVisible()`, but I also wondered why the `hidden` was included for this test and not the others.

firestack (approved):

PR #2852, opened 2024-10-11
feat: Move and style copyButton, add to active and past detour views

hannahpurcell:
Add yourself to the copy-button user group to see this on the active / past detour views, because I haven't yet done the copy logic. 

The button will be visible on the View Draft Detour panel without being added to the user group.

Now depends on:
- https://github.com/mbta/skate/pull/2853

^ only because the copy change will prevent the button from being too cramped on the "View Draft Detour" page


assets/css/_diversion_page.scss
```diff
@@ -89,7 +93,7 @@
       }
     }
 
-    .l-diversion-page-panel__header .c-diversion-panel__back-button {
+    .l-diversion-page-panel__header .c-diversion-panel__outline-button--back {
```
firestack, line 96: question: why change the name of this button? IMO the "outline" info doesn't seem necessary?

assets/css/_diversion_page.scss
```diff
@@ -161,3 +165,14 @@
...
+
+.c-diversion-panel__outline-button {
+  height: 2rem;
+  border-radius: 0.2rem;
+  font-size: 0.875rem;
+  padding: 0.375rem 0.5rem;
+
+  &--copy {
+    margin: 0.78125rem 0;
+  }
```
firestack, line 177: I think this margin should be handled via the `.align-items-baseline` class on the parent flex

assets/src/components/detours/detourPanelComponents.tsx
```diff
@@ -64,3 +71,26 @@ export const AffectedRoute = ({
...
+    rootClose
+    rootCloseEvent="mousedown"
+    overlay={
+      <Popover>
+        <Popover.Body>Copied to clipboard!</Popover.Body>
+      </Popover>
+    }
+  >
+    <Button
+      className="c-diversion-panel__outline-button c-diversion-panel__outline-button--copy icon-link"
```
firestack, line 88: Between these two suggestions
- https://github.com/mbta/skate/pull/2852/files#r1806786402
- https://github.com/mbta/skate/pull/2852/files#r1806794925

I don't think most of the classes are needed?
```suggestion
      className="icon-link"
```

assets/css/_diversion_page.scss
hannahpurcell: You're totally right. I renamed at one point, assuming I'd use the `--back` as an extension of outline button, as in [these lines](https://github.com/mbta/skate/pull/2852/files#diff-a84a6c4e5f6eb9053ec860f02bebc5f2e4b0e544b079b3c814e92b6eaa6bea0aR169-R178), but not needed. (Plus, I see those might need tweaking as well)

assets/css/_diversion_page.scss
```diff
@@ -161,3 +165,14 @@
...
+}
```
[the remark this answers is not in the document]
hannahpurcell: Oh my gosh, I totally reinvented the wheel. Thanks!!

firestack (approved):

PR #2853, opened 2024-10-11
tweak: Update detour panel titles / buttons to match hi-fi language

hannahpurcell:
Adhoc, but here's the [figma link](https://www.figma.com/design/XGmmwVROXdOvnWtJeUFC50/Skate-%7C-Unplanned-Detours-%7C-Handoff?node-id=3666-13075&t=dJLTPrFTZyzHvNWX-4)

### DIFF
--- a/assets/tests/components/mapPage.test.tsx
+++ b/assets/tests/components/mapPage.test.tsx
@@ -1850,7 +1850,7 @@ describe("<MapPage />", () => {
         })
       )
 
-      expect(screen.getByText("Create Detour")).toBeVisible()
+      expect(screen.getByText("Draw Detour")).toBeVisible()
     })
 
     test("dismisses the detour modal on escape key", async () => {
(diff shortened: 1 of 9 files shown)


assets/tests/components/mapPage.test.tsx
```diff
@@ -1850,7 +1850,7 @@ describe("<MapPage />", () => {
         })
       )
 
-      expect(screen.getByText("Create Detour")).toBeVisible()
+      expect(screen.getByText("Draw Detour")).toBeVisible()
```
firestack, line 1853: suggestion: do you feel like pulling this out into it's own selector too? Seems like this had to be changed in a few places as well

firestack (changes requested):
Looks good! One suggestion

assets/tests/components/mapPage.test.tsx
hannahpurcell: Oh hey! I already had a selector that would work for this, as well as other tests, but I didn't use it universally. Updated ✔️

PR #2858, opened 2024-10-16
fix: Uneditable detour directions

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208559338827067

I'm stuck on getting a proper test set up for this. When I revert the `extendedDirections` -> `directions` code, the test still passes, so it's not a correct test.


hannahpurcell:
This is newer: https://github.com/mbta/skate/pull/2861

PR #2859, opened 2024-10-17
feat(ts/components/markers): move vehicle marker into own file

firestack:
I moved this out into it's own file to make adding shuttle colors simpler, figured I'd put this up for review so it's not merged together with shuttle colors.


hannahpurcell (approved):
👍

PR #2864, opened 2024-10-21
refactor:test(ex/conn_case): split `conn` creation into helper function

firestack:
I was working on upgrading some of Skate to latest Phoenix for a personal project, and ended up refactoring `ConnCase` to see if I could make it any cleaner. I liked this, but it's not necessary, let me know what you think!

### DIFF
--- a/test/support/conn_case.ex
+++ b/test/support/conn_case.ex
@@ -14,8 +14,6 @@ defmodule SkateWeb.ConnCase do
   """
 
   use ExUnit.CaseTemplate
-  import Plug.Test
-  alias Skate.Settings.User
 
   using do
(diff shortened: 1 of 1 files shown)


hannahpurcell (commented):
I'm not sure that I find this easier to read than previously, but I'm not opposed to it if you find it easier to use.

For me, I mildly prefer the single user / resource creation so that it didn't need to be a reusable function. 

However, one thing I don't understand about the old file, and which you fix in your changes, is the second `User.upsert(username, email)` that happens for each case in the `cond do`. That does seem unneeded given the `user = User.upsert(username, email)` that happens previously.

PR #2865, opened 2024-10-21
refactor:test(ex): use `DataCase.setup_sandbox` acording to `phx.new` template

firestack:
This allows other test cases to configure the database via a shared function


hannahpurcell (approved):
👍 Thanks for the links in the PR description, and the comment updates in each of the test support files

PR #2868, opened 2024-10-21
cleanup(ex/test): general cleanup related to upgrading to the latest Phoenix version and Template

firestack:
Reviewable by commit for more info.

### DIFF
--- a/test/support/data_case.ex
+++ b/test/support/data_case.ex
@@ -41,7 +41,7 @@ defmodule Skate.DataCase do
   end
 
   @doc """
-  A helper that transform changeset errors to a map of messages.
+  A helper that transforms changeset errors into a map of messages.
 
       assert {:error, changeset} = Accounts.create_user(%{password: "short"})
       assert "password is too short" in errors_on(changeset).password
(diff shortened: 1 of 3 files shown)


test/support/data_case.ex
```diff
@@ -49,8 +49,8 @@ defmodule Skate.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, fn {key, value}, acc ->
-        String.replace(acc, "%{#{key}}", to_string(value))
+      Regex.replace(~r"%{(\w+)}", message, fn _, key ->
+        opts |> Keyword.get(String.to_existing_atom(key), key) |> to_string()
```
firestack, line 53: https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_ecto/data_case.ex#L50-L55

Dunno much about this change, but I don't think we use this function yet.

hannahpurcell: Good enough for me. Thanks for the link!

hannahpurcell (approved):
👍

PR #2871, opened 2024-10-21
fix: Avoid using non-unique stop names as key

hannahpurcell:
The 1 has two stops called Mass Ave & Albany! One in Cambridge, one in Boston. This suppresses some rowdy console errors

### DIFF
--- a/assets/src/components/detours/detourMap.tsx
+++ b/assets/src/components/detours/detourMap.tsx
@@ -269,7 +269,7 @@ export const DetourMap = ({
             <>
               {uniqBy(stops, (stop) => stop.id).map((stop) => (
                 <StopMarkerWithStopCard
-                  key={stop.name}
+                  key={stop.id}
                   stop={stop}
                   zoomLevel={zoomLevel}
                   interactionStatesDisabled={false}


assets/src/components/detours/detourMap.tsx
```diff
@@ -269,7 +269,7 @@ export const DetourMap = ({
             <>
               {uniqBy(stops, (stop) => stop.id).map((stop) => (
                 <StopMarkerWithStopCard
-                  key={stop.name}
+                  key={stop.id}
```
firestack, line 272: Can you add the missed stop `index` to this key? Until we have `stop_sequence` this could also error on routes like (route pattern) `217-1-0`

https://www.mbta.com/schedules/217/line?from=global-search&query=217&schedule_direction%5Bdirection_id%5D=0&schedule_direction%5Bvariant%5D=217-1-0

image

I created a Detour in Skate Dev which shows this.

firestack (changes requested):
https://github.com/mbta/skate/pull/2871/files#r1809095958

assets/src/components/detours/detourMap.tsx
hannahpurcell: I've checked out the skate dev detour, but I don't quite understand how this would error if key was changed to stop.id. I made the same detour locally with no error.  

Also, were you looking for an example where a stop is hit multiple times on the same route? I don't think this is one. There is a loop, but stops are not revisited

Local screenshot matches skate-dev, no console errors
Screenshot 2024-10-22 at 8 22 16 AM

firestack: ooooh I thought this was a different part of the code, thanks for pointing that out. This change is good then 👍🏻.

---

It's `missed stops` that I was thinking of, which produces wrong results in production.

Notice that `North Quincy` is only listed once.

## Dev:
image

## Local patch:
image

firestack (approved):

assets/src/components/detours/detourMap.tsx
hannahpurcell: Ohhhh I see. That's a good bug. I'm curious whether design would say the stop should be listed twice, or the missed stop count should be 7. Because the redundancy doesn't _seem_ like it would be valuable, but not certain

firestack: I personally think it's valuable to riders, because it's possible that the stop could be missed on the first or second pass, but not the other. I think this also matters for GTFS-RT, so obscuring the link between operations data and rider data doesn't seem worth it to me.

PR #2875, opened 2024-10-22
feat(ex/test/conn-case): make default user dynamic from factory 

firestack:
#2874 
> There are a few tests which assume what user information they're going to get, but https://github.com/mbta/skate/pull/2864 "revealed" that we hardcode the user for ConnCase, but with https://github.com/mbta/skate/pull/2873 we can now generate dynamic users, so this fixes the tests so that we can dynamically create users.

Depends on:
- ~https://github.com/mbta/skate/pull/2864~
- #2873 
- #2874


firestack:
oops 🙃

PR #2881, opened 2024-10-24
feat(ts/detours): add `data-fs-element` properties to detours flow

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1208610656881565/f

---

Superceeds: #2870

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -410,6 +410,7 @@ export const DiversionPage = ({
   return (
     <>
       <article
+        data-fs-element="Detours"
         className={`l-diversion-page l-diversion-page--${displayType} h-100 border-box inherit-box`}
       >
         <header
--- a/assets/src/components/detours/activateDetourModal.tsx
+++ b/assets/src/components/detours/activateDetourModal.tsx
@@ -66,7 +66,11 @@ const SurroundingModal = ({
         Cancel
       </Button>
       {onActivate ? (
-        <Button variant="primary" onClick={onActivate}>
+        <Button
+          variant="primary"
+          onClick={onActivate}
+          data-fs-element="Confirm Activate Detour"
+        >
           Activate detour
         </Button>
       ) : (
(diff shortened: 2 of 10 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -410,6 +410,7 @@ export const DiversionPage = ({
   return (
     <>
       <article
+        data-fs-element="Detours"
```
hannahpurcell, line 413: This is the only one that gives me pause, as it is generic, but perhaps that's exactly what we want! Whenever a detour modal is opened, no user actions taken, the "Detours" event occurs

hannahpurcell (approved):
Looks good to me! Love how much use we can get out of data-fs-element for our needs.

Would it be good to write a list of the elements and their action and share them with the rest of the product team to verify these names are ideal, or we can skip that and just tweak it later based on feedback?

Example:
Activate detour button in confirmation modal: "Confirm Activate Detour"
Start detour button, to open confirmation modal: "Begin Activate Detour"

firestack:
> Looks good to me! Love how much use we can get out of data-fs-element for our needs.
> 
> Would it be good to write a list of the elements and their action and share them with the rest of the product team to verify these names are ideal, or we can skip that and just tweak it later based on feedback?
> 
> Example: Activate detour button in confirmation modal: "Confirm Activate Detour" Start detour button, to open confirmation modal: "Begin Activate Detour"
> 
> Might be best just to ship and tweak based on feedback

We can do this in fullstory. As long as we keep the name the same, the selector can be renamed and described in fullstory.

assets/src/components/detours/diversionPage.tsx
firestack: This is really for scoping. If we ever had a "Undo" button elsewhere in the application, the `data-fs-element="Undo"` selector that FS uses would trigger for both of those. I _think_ it's _better_ to use more generic names and scope them via this type of "context" selector, since fullstory allows you to compose them

PR #2883, opened 2024-10-24
tweak: Resize editable directions textbox

hannahpurcell:
Brian requested that the textarea for the editable directions resize to the contained text, to avoid ever having a double scrollbar on the detour details panel

Built on:
- https://github.com/mbta/skate/pull/2880

PR #2884, opened 2024-10-25
tweak: Active and Past detours use edited directions

hannahpurcell:
Edited directions are parsed for the active / past detours pages

Built on
- https://github.com/mbta/skate/pull/2883

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -261,6 +261,11 @@ export const DiversionPage = ({
         <DetourFinishedPanel
           onNavigateBack={editDetour}
           detourText={textArea}
+          connectionPoints={[
+            connectionPoints?.start?.name ?? "N/A",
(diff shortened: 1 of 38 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -49,6 +49,10 @@ const displayFieldsFromRouteAndPattern = (
   return { routeName, routeDirection, routeOrigin, routeDescription, shape }
 }
 
+const parseIntoDirectionsList = (directions: string) => {
+  return directions.split("\n").map((direction) => ({ instruction: direction }))
+}
+
```
firestack, line 55: I think we should approach this differently and do this in the state machine, hence my hesitancy to approve https://github.com/mbta/skate/pull/2880

I think we should be doing this "parsing" when the `detour.share.edit-directions` event is fired, and assigning that value directly to `directions`, instead of keeping an intermediate `editedDirections` value.

firestack (approved):

PR #2886, opened 2024-10-28
feat: fix deployment

firestack:


hannahpurcell:
Is this just a re-org, or were deploys somehow broken, and how does this fix it?

firestack:
> Is this just a re-org, or were deploys somehow broken, and how does this fix it?

So I _think_ this will fix the "double approval" we were seeing on prod deploys.
I _think_ that'll be entirely handled by https://github.com/mbta/skate/pull/2886/files#diff-c4bb66f9eb59a6e5f596c0a49ffedadf4ccdbc5cd8d4a3ef1a410679f96afb66L29

I noticed this because when deploying branches in PR's I'd get "double notifications" in the PR logs. So instead of _one_ of the deployment messages[^1] for a commit push, I'd see two, for the same deployment.

[^1]: > @[firestack](https://github.com/firestack) firestack [requested a deployment](https://github.com/mbta/skate/actions/runs/11599965303/job/32302931798) to dev-green[ October 30, 2024 16:23 ](https://github.com/mbta/skate/pull/2886#event-15017382608)— with [GitHub Actions](https://github.com/apps/github-actions) In progress

I do seem to have resolved that issue in this PR.

The concurrency thing was just me trying to refine the build process. When I originally did #2838 I tried to put the concurrency "flag" on the `deploy-base` jobs, but it was complaining that it was conflicting with another definition elsewhere, so I punted on it. I realized recently that it was because the `deploy-<env>` definitions were specifying the concurrency flag in that file, which was marking all of the jobs as blocking the other. Because we can build an image while deploying a different one, this just optimizes the build pipeline.

hannahpurcell (approved):
Willing to give it a shot!

PR #2888, opened 2024-10-29
feat(ts/detour-entrypoints): add entrypoints for add detour

firestack:
---
  - https://app.asana.com/0/0/1208610656881548
  - https://app.asana.com/0/0/1208610656881549

### DIFF
--- a/assets/src/components/detours/detourDropdown.tsx
+++ b/assets/src/components/detours/detourDropdown.tsx
@@ -54,6 +54,7 @@ export const DetourDropdown = ({
               zoom,
             })
           }}
+          data-fs-element="Add Detour on Trip"
         >
           Start a detour on route {routeName}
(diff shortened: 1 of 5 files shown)


assets/src/components/detours/detourDropdown.tsx
```diff
@@ -54,6 +54,7 @@ export const DetourDropdown = ({
               zoom,
             })
           }}
+          data-fs-element="Add Detour on Trip"
```
hannahpurcell, line 57: non-blocking: Perhaps, this instead, given we're not adding a detour for only one trip, but just via the vehicle entrypoint
```suggestion
          data-fs-element="Add Detour via Vehicle Marker"
```

hannahpurcell (approved):

PR #2889, opened 2024-10-29
feat(ts/detour/fs-element-steps): add element labels to next step in activate modal

firestack:
Not quite happy with this, but it works <shrug>. LMK if you have any thoughts on how to better organize and structure this.


assets/src/components/detours/diversionPage.tsx
```diff
@@ -276,6 +276,21 @@ export const DiversionPage = ({
...
+                    "Share Detour": { Activating: "Selecting Duration" },
+                  },
+                }) &&
+                  "Select Duration") ||
+                (snapshot.matches({
+                  "Detour Drawing": {
+                    "Share Detour": { Activating: "Selecting Reason" },
+                  },
+                }) &&
+                  "Select Reason") ||
```
hannahpurcell, line 291: non-blocking: Would these element ids be better as past tense, given that when the next button is clicked, the action already happened?
```suggestion
                  "Selected Duration") ||
                (snapshot.matches({
                  "Detour Drawing": {
                    "Share Detour": { Activating: "Selecting Reason" },
                  },
                }) &&
                  "Selected Reason") ||
```

hannahpurcell (approved):

assets/src/components/detours/diversionPage.tsx
firestack: 1. These labels can be changed in fullstory UI if desired by our team
2. I think given how this appears in fullstory[^1], it's more correct to use present tense. I could alternatively name these `Confirm Reason` or `Confirm Duration`

[^1]: 
image

firestack:
```suggestion
                  "Confirm Duration") ||
                (snapshot.matches({
                  "Detour Drawing": {
                    "Share Detour": { Activating: "Selecting Reason" },
                  },
                }) &&
                  "Confirm Reason") ||
```

hannahpurcell: Got it! I do prefer "Confirm" then (just so in the future we don't think these actions are attached to the radio buttons, which would be a different meaning of "Select Reason")

PR #2895, opened 2024-11-04
chore(ex): configure check_origin for all transports on the `SkateWeb.Endpoint`

firestack:
This configuration has always confused me a little, mainly that we had to "implement" it ourselves.

So I tracked this down and confirmed that this can be configured via Application config
- https://hexdocs.pm/phoenix/1.7.11/Phoenix.Endpoint.html#socket/3-common-configuration
	- > `:check_origin` - if the transport should check the origin of requests when the origin header is present. [...]. **_Defaults to `:check_origin` setting at endpoint configuration_**.

I do not know if this option was available at the time of https://github.com/mbta/skate/pull/36/ but now we can configure this via `SkateWeb.Endpoint`.

I loaded this up on dev-green and saw no changes or issues when loading the application.

This also seems to match other "modern" TID applications.

---

Depends on:
- #2894

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -31,12 +31,11 @@ config :skate, SkateWeb.Endpoint,
     port: {:system, "STATIC_PORT"},
     path: {:system, "STATIC_PATH"}
   ],
-  cache_static_manifest: "priv/static/cache_manifest.json"
-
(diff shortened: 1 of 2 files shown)


hannahpurcell (approved):
👍 yeah, this makes sense

PR #2896, opened 2024-11-04
 refactor:test(ex/skate_web/detour_controller): use `:user` factory 

firestack:
Depends on:
- #2897

### DIFF
--- a/lib/skate/detours/detours.ex
+++ b/lib/skate/detours/detours.ex
@@ -193,36 +193,33 @@ defmodule Skate.Detours.Detours do
   end
 
   @doc """
-  Update or create a detour given a user id & detour id.
+  Update or insert a detour given a user id and a XState Snapshot.
   """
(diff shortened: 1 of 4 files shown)


hannahpurcell (approved):
The diff is larger than I think it's supposed to be -- seems like the only change is line 14 of the test? Nothing looks out of place from `main`, though! so ✅

hannahpurcell:
Whoops, I didn't think about auto-merge, but hopefully all is well!

firestack:
> The diff is larger than I think it's supposed to be -- seems like the only change is line 14 of the test? Nothing looks out of place from `main`, though! so ✅

Correct, looks like I didn't rebase this so it reported wrong.



> Whoops, I didn't think about auto-merge, but hopefully all is well!

Looks like it turned out okay though! https://github.com/mbta/skate/commit/bf5f673731ce850d761bb34fee2644aa39867ff1

PR #2897, opened 2024-11-04
init(ex/skate/snapshot_serde): create deserialize function

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1208675099752884/f

### DIFF
--- a/lib/skate/detours/detours.ex
+++ b/lib/skate/detours/detours.ex
@@ -193,36 +193,33 @@ defmodule Skate.Detours.Detours do
   end
 
   @doc """
-  Update or create a detour given a user id & detour id.
+  Update or insert a detour given a user id and a XState Snapshot.
   """
-  def update_or_create_detour_for_user(user_id, uuid, attrs \\ %{}) do
-    user = User.get_by_id!(user_id)
-
-    previous_record = uuid != nil && Skate.Repo.get(Detour, uuid)
-
-    detour =
-      case uuid do
-        nil ->
-          create_detour_for_user(user_id, attrs)
-
-        _ ->
-          Repo.insert(
-            Detour.changeset(%Detour{author: user, id: uuid}, attrs),
-            returning: true,
-            conflict_target: [:id],
-            on_conflict: {:replace, [:state, :updated_at]}
-          )
+  def upsert_from_snapshot(author_id, %{} = snapshot) do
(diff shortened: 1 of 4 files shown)


lib/skate/detours/detours.ex
```diff
@@ -193,36 +193,33 @@ defmodule Skate.Detours.Detours do
...
   """
-  def update_or_create_detour_for_user(user_id, uuid, attrs \\ %{}) do
-    user = User.get_by_id!(user_id)
-
-    previous_record = uuid != nil && Skate.Repo.get(Detour, uuid)
-
-    detour =
-      case uuid do
-        nil ->
-          create_detour_for_user(user_id, attrs)
```
hannahpurcell, line 206: Huh! I'm realizing that this case wasn't necessary, given the Repo.insert would handle that, right?

firestack: Affirmative! I tried to figure it[^1] out when we first added this but ran into issues, but it went smoothly this time so 🤷🏻‍♀️ 

[^1]: "it" being `upsert` with a single call without a `case`

lib/skate_web/controllers/detours_controller.ex
```diff
@@ -11,15 +11,11 @@ defmodule SkateWeb.DetoursController do
...
-  def update_snapshot(conn, %{"snapshot" => %{"context" => context} = snapshot}) do
-    uuid = Map.get(context, "uuid")
-
+  def update_snapshot(conn, %{"snapshot" => snapshot}) do
     %{id: user_id} = AuthManager.Plug.current_resource(conn)
 
     {:ok, %Skate.Detours.Db.Detour{id: returned_uuid}} =
-      Detours.update_or_create_detour_for_user(user_id, uuid, %{
-        state: snapshot
-      })
```
hannahpurcell, line 22: I wonder why I previously wrapped the snapshot like this? I see that your changes assign state properly in the db, so... great! Much simpler 😊

lib/skate/detours/detours.ex
hannahpurcell, line 206: `create_detour_for_user` isn't being used anymore. What do you think: keep it or ditch it?

hannahpurcell (approved):
Looks great! The associated question is non-blocking

PR #2899, opened 2024-11-07
feat: Detour serializer

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208675099752885

The second commit is completely optional, and maybe not even proper. I enjoyed using map_diff to help print the difference between `state` and `serialized_snapshot` on [line 29](https://github.com/mbta/skate/blob/46b0d593e4895fb643188328fc51469d51a83e07/lib/skate/detours/snapshot_serde.ex#L29) of `snapshot_serde.ex`. Not used in production, though, so might not be logical to commit it. On the other hand, I can see it being valuable to our production logs if we _did_ want it.


firestack:
> The second commit is completely optional, and maybe not even proper. I enjoyed using map_diff to help print the difference between `state` and `serialized_snapshot` on [line 29](https://github.com/mbta/skate/blob/46b0d593e4895fb643188328fc51469d51a83e07/lib/skate/detours/snapshot_serde.ex#L29) of `snapshot_serde.ex`. Not used in production, though, so might not be logical to commit it. On the other hand, I can see it being valuable to our production logs if we _did_ want it.

I'm curious to see this! But I don't see any usage of `MapDiff` in the PR currently?

hannahpurcell:
> I'm curious to see this! But I don't see any usage of `MapDiff` in the PR currently?

Yeah, I suppose I should have left my log in there as explanation, especially since I've had subsequent merges. I'll add it back in for discussion / testing. I can remove if we deem it too noisy for Splunk

lib/skate/detours/snapshot_serde.ex
```diff
@@ -27,4 +31,280 @@ defmodule Skate.Detours.SnapshotSerde do
...
+
+  defp validate_serialized_snapshot(
+         serialized_snapshot,
+         %Detour{id: id, state: state}
+       ) do
+    if serialized_snapshot === state do
+      serialized_snapshot
+    else
+      log_fallback(id)
+      Logger.info(MapDiff.diff(state, serialized_snapshot))
```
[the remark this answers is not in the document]
firestack:
```diff
Commit ID: 8cb73595acaa415b993f8d09b343d4ae386510d8
Change ID: oqpysvvsyklookkyrrulxvrrkxxqszxt
Author: Kayla Firestack <kfirestack@mbta.com> (2024-11-12 08:15:32)
Committer: Kayla Firestack <kfirestack@mbta.com> (2024-11-12 13:02:31)

    (no description set)

diff --git a/lib/skate/detours/snapshot_serde.ex b/lib/skate/detours/snapshot_serde.ex
index 394948d1d1..364d21a2ec 100644
--- a/lib/skate/detours/snapshot_serde.ex
+++ b/lib/skate/detours/snapshot_serde.ex
@@ -55,8 +55,9 @@
     if serialized_snapshot === state do
       serialized_snapshot
     else
-      log_fallback(id)
-      Logger.info(MapDiff.diff(state, serialized_snapshot))
+      Logger.error(
+        "Serialized detour doesn't match saved snapshot. Falling back to snapshot for detour_id=#{id} diff=#{inspect(MapDiff.diff(state, serialized_snapshot))}"
+      )
       state
     end
   end
@@ -299,8 +300,6 @@
 
   defp snapshot_children_from_detour(_), do: nil
 
-  defp log_fallback(field) when is_binary(field),
-    do: Logger.warning("Unexpected detour structure. Using snapshot for field: #{field}")
 
   defp log_fallback(id) when is_integer(id),
     do:
```

test/skate_web/controllers/detours_controller_test.exs
firestack, line 1: will review this file in ~1h

firestack (changes requested):

lib/skate/detours/snapshot_serde.ex
```diff
@@ -27,4 +31,280 @@ defmodule Skate.Detours.SnapshotSerde do
...
+      "waypoints" => waypoints_from_detour(detour),
+      "nearestIntersection" => nearestintersection_from_detour(detour),
+      "detourShape" => detourshape_from_detour(detour),
+      "finishedDetour" => finisheddetour_from_detour(detour),
+      "editedDirections" => editeddirections_from_detour(detour),
+      "selectedDuration" => selectedduration_from_detour(detour),
+      "selectedReason" => selectedreason_from_detour(detour)
+    }
+    |> Enum.filter(fn {_, v} -> v != nil end)
+    |> Enum.into(%{})
```
firestack, line 93: suggestion(non-blocking): Love it! just wondering, what do you think about using [`Map.filter/2`](https://hexdocs.pm/elixir/Map.html#filter/2)?

```suggestion
    |> Map.filter(fn {_, v} -> v != nil end)
```

https://github.com/devonestes/fast-elixir#filtering-maps-code
says `:maps.filter/2` is faster, by up to 1.3x, but `:maps.filter/2` has the fn first, and the map argument second, which wouldn't work well with Elixir's pipes. (I know we're not at the point of chasing down every bit of performance right now, I just use it as a reference)

lib/skate/detours/snapshot_serde.ex
```diff
@@ -27,4 +31,280 @@ defmodule Skate.Detours.SnapshotSerde do
...
+  defp uuid_from_detour(%Detour{
+         state: %{
+           "context" => %{
+             "uuid" => uuid
+           }
+         }
+       }) do
+    log_fallback("uuid")
+    uuid
+  end
```
firestack, line 118: issue: I don't think we should fallback to the snapshot here, because the `id` is the authoritative field.

What do you think?

```suggestion
  defp uuid_from_detour(%Detour{id: id}) do
     id
  end
```

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -223,6 +223,66 @@ defmodule SkateWeb.DetoursControllerTest do
...
+      conn = get(conn, "/api/detours/#{detour_id}")
+
+      assert %{
+               "data" => %{
+                 "author" => "test_user@test.com",
+                 "state" => ^snapshot,
+                 "updated_at" => _
+               }
+             } = json_response(conn, 200)
+    end
```
firestack, line 285: issue: I'm noticing that these tests are only _happening_ to fail because the snapshot is a partial snapshot that doesn't include some of the fields that the serializer adds.

Maybe we should use the `build(:detour)` factory to generate snapshots and update that factory to add the extra required snapshot fields like `children`, `history`, `status`, and `value`?

And then to make these tests fail, we could insert a snapshot into the DB, but then change the `uuid` in the snapshot so that the snapshot `context.uuid` doesn't match the `Detour.id`, which would fail the equivalency check.

This feedback is fairly late, so happy to pair on this if it'd help.

firestack: For now, we may be able to skip it if it was important to merge this sooner than later, but I think we'd need to do it anyway?

lib/skate/detours/snapshot_serde.ex
firestack: or at the very least we should prioritize the `id` field.

lib/skate/detours/snapshot_serde.ex
hannahpurcell: Agh rats, I did look up `Map.filter` vs `Enum.filter |> Enum.into` and interpreted that the latter was more efficient, but can't find my primary source 🫠  But I'm more open to :maps.filter. I did consider that

lib/skate/detours/snapshot_serde.ex
hannahpurcell: That's probably right! So, no fallback function here at all; the function will just fail

test/skate_web/controllers/detours_controller_test.exs
hannahpurcell: Should be fixed now!

test/support/factories/detour_factory.ex
```diff
@@ -38,12 +38,16 @@ defmodule Skate.DetourFactory do
...
+          "children" => %{},
+          "historyValue" => %{},
+          "status" => "active"
         }
       end
 
       def with_id(%Skate.Detours.Db.Detour{} = detour, id) do
-        %{detour | state: with_id(detour.state, id)}
+        detour_with_id = %{detour | id: id}
+        %{detour_with_id | state: with_id(detour.state, id)}
```
firestack, line 50: nit: I think this can be done in a single statement?

```suggestion
       %{
          detour
          | id: id,
            state: with_id(detour.state, id)
        }
```

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -224,64 +224,66 @@ defmodule SkateWeb.DetoursControllerTest do
...
+
+      retrieved_detour = Detours.get_detour!(detour_id)
+      # Changing the status is a sure way to force a fallback, as it should always be "active"
+      edited_snapshot = put_in(detour_snapshot["status"], nil)
+      edited_detour = %{retrieved_detour | state: edited_snapshot}
 
       log =
         CaptureLog.capture_log(fn ->
-          get(conn, "/api/detours/#{detour_id}")
+          Skate.Detours.SnapshotSerde.serialize(edited_detour)
```
firestack: Sorry, could you do this via `get` instead of calling `Skate.Detours.SnapshotSerde.serialize` directly?

Same here [`99c5e17` (#2899)](https://github.com/mbta/skate/pull/2899/commits/99c5e172d003962725f1e6c8674d64992c4a8434#diff-6b5f7db841fd66bb4a09b0870661a72d957ad6cf263cccf090ebb4a251bc0687R286)

It's a _little_ strange to have a test in our controller tests that doesn't call the `get api/detours/:id` endpoint, but I do like that these tests _are_ done in this suite and I don't think it's worth pulling it out into a test suite for `         Skate.Detours.SnapshotSerde` specifically.

firestack (changes requested):
Looks great! one note, but otherwise I think this is ready to go!

test/skate_web/controllers/detours_controller_test.exs
hannahpurcell: Ok, this is what I was struggling with. When do I edit the detour into something that is invalid? If I add to the db via put, and retrieve from the db via get, I have no opportunity to create `edited_detour` to force the logged error

firestack: Something like this?
```diff
Commit ID: 2974bc649c2d65fba1acb1062e066c0a31cb6898
Change ID: ytznlvkvoxysnvtrvllxpspkmkrnyotr
Author: Kayla Firestack <kfirestack@mbta.com> (2024-11-19 09:23:14)
Committer: Kayla Firestack <kfirestack@mbta.com> (2024-11-19 09:26:19)

    (no description set)

diff --git a/test/skate_web/controllers/detours_controller_test.exs b/test/skate_web/controllers/detours_controller_test.exs
index 7ad3f7169f..0c02ce4b30 100644
--- a/test/skate_web/controllers/detours_controller_test.exs
+++ b/test/skate_web/controllers/detours_controller_test.exs
@@ -252,16 +252,17 @@
         |> build()
         |> with_id(detour_id)
 
-      put(conn, "/api/detours/update_snapshot", %{"snapshot" => detour_snapshot})
+      conn = put(conn, "/api/detours/update_snapshot", %{"snapshot" => detour_snapshot})
 
-      retrieved_detour = Detours.get_detour!(detour_id)
       # Changing the status is a sure way to force a fallback, as it should always be "active"
-      edited_snapshot = put_in(detour_snapshot["status"], nil)
-      edited_detour = %{retrieved_detour | state: edited_snapshot}
+      detour_id
+      |> Detours.get_detour!()
+      |> Skate.Detours.Detours.change_detour(%{state: put_in(detour_snapshot["status"], nil)})
+      |> Skate.Repo.update()
 
       log =
         CaptureLog.capture_log(fn ->
-          Skate.Detours.SnapshotSerde.serialize(edited_detour)
+          get(conn, "/api/detours/#{detour_id}")
         end)
 
       assert log =~
```

hannahpurcell: Ohhhh I see, a re-insert to the db with edited state only. Yeah, can do!

firestack (approved):
Perfect thank you!

PR #2907, opened 2024-12-11
tweak: Move user filter for drafts

hannahpurcell:
`user_id` is only needed in `db_detour_to_detour` for the sake of filtering out draft detours not created by the current user. That can and should be done in the parent function, not in `categorize_detour`. Refactored to simplify. Useful for the detours channel work

### DIFF
--- a/lib/skate/detours/detours.ex
+++ b/lib/skate/detours/detours.ex
@@ -55,7 +55,7 @@ defmodule Skate.Detours.Detours do
     %{
       active: Map.get(detours, :active),
       draft:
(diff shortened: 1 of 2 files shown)


lib/skate/detours/detours.ex
```diff
@@ -48,18 +48,19 @@ defmodule Skate.Detours.Detours do
   def grouped_detours(user_id) do
     detours =
       list_detours()
-      |> Enum.map(fn detour -> db_detour_to_detour(detour, user_id) end)
+      |> Enum.map(fn detour -> db_detour_to_detour(detour) end)
```
firestack, line 51:
```suggestion
      |> Enum.map(&db_detour_to_detour/1)
```

PR #2908, opened 2024-12-11
feat: Detours channel

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208600990461875

Depends on:
- https://github.com/mbta/skate/pull/2906
- https://github.com/mbta/skate/pull/2907

To try it out, I call 
```
const skateDetours = useActiveDetoursByRoute(socket, selectedRouteIds)
console.log("skatedetours: ", skateDetours)
```
in `routeLadders.tsx`, and `useActiveDetours(socket)`, `usePastDetours(socket)`, and `useDraftDetours(socket)` in `detourListPage.tsx`.

Needs: 
- [x] tests


lib/skate/detours/detours.ex
```diff
@@ -114,9 +130,16 @@ defmodule Skate.Detours.Detours do
...
-  def categorize_detour(_detour_context, nil = _user_id), do: :draft
-  def categorize_detour(%{author_id: author_id}, user_id) when author_id == user_id, do: :draft
-  def categorize_detour(_, _), do: nil
+    @doc """
+  Takes a `Skate.Detours.Db.Detour` struct and a `Skate.Settings.Db.User` id
+  and returns a `t:detour_type/0` based on the state of the detour.
+
+  otherwise returns `nil` if it is a draft but does not belong to the provided
+  user
+  """
```
firestack, line 139: This @doc comment seems incorrect?

lib/skate_web/channels/detours_channel.ex
```diff
@@ -0,0 +1,62 @@
...
+  def join_authenticated("detours:active", _message, socket) do
+    SkateWeb.Endpoint.subscribe("detours:active")
+    %{id: user_id} = Guardian.Phoenix.Socket.current_resource(socket)
+    detours = Detours.grouped_detours(user_id)[:active]
+    {:ok, %{data: detours}, socket}
+  end
+
+  @impl SkateWeb.AuthenticatedChannel
+  def join_authenticated("detours:active" <> route_id, _message, socket) do
+    SkateWeb.Endpoint.subscribe("detours:active" <> route_id)
```
firestack, line 20: I _think_ this is better? otherwise I think this would match on things like `detours:activejoafjoifwaiojfewijowef` where `route_id` would contain `joafjoifwaiojfewijowef`
```suggestion
  def join_authenticated("detours:active:" <> route_id, _message, socket) do
    SkateWeb.Endpoint.subscribe("detours:active:" <> route_id)
```

lib/skate/detours/detours.ex
hannahpurcell: Oh this was a very weird merging situation. Didn't want to delete the categorize_detour functions either. That should be in another PR

```suggestion
  def categorize_detour(_detour_context, nil = _user_id), do: :draft
  def categorize_detour(%{author_id: author_id}, user_id) when author_id == user_id, do: :draft
  def categorize_detour(_, _), do: nil
  
```

lib/skate/detours/detours.ex
```diff
@@ -117,6 +133,10 @@ defmodule Skate.Detours.Detours do
   def categorize_detour(_detour_context, nil = _user_id), do: :draft
   def categorize_detour(%{author_id: author_id}, user_id) when author_id == user_id, do: :draft
   def categorize_detour(_, _), do: nil
+  
+  @spec get_detour_route(detour :: map()) :: String.t()
+  def get_detour_route(%{state: %{"context" => %{"route" => %{"name" => route_name}}}}),
```
hannahpurcell, line 138:
```suggestion
  defp get_detour_route(%{state: %{"context" => %{"route" => %{"name" => route_name}}}}),
```

lib/skate_web/channels/detours_channel.ex
hannahpurcell: Absolutely!! This was a mistake, and in fact, the frontend expects the colon.

firestack: I believe it still would've worked, `route_id` would just be `:66` instead of `66`.

assets/src/hooks/useDetours.ts
```diff
@@ -0,0 +1,244 @@
...
+    channel.on("deactivated", ({ data: data }) => handleDeactivated(data))
+  channel.on("auth_expired", reload)
+
+  channel
+    .join()
+    .receive("ok", ({ data: data }: { data: SimpleDetour[] }) => {
+      const detoursMap = data.reduce(
+        (acc, detour) => ({ ...acc, [detour.id]: detour }),
+        {}
+      )
```
firestack, line 36: suggestion(non-blocking): small syntax thing that I think makes it easier to read
```suggestion
      const detoursMap = Object.fromEntries(data.map(v => [v.id, v]))
```

lib/skate_web/channels/detours_channel.ex
```diff
@@ -0,0 +1,62 @@
...
+  @impl SkateWeb.AuthenticatedChannel
+  def join_authenticated("detours:active:" <> route_id, _message, socket) do
+    SkateWeb.Endpoint.subscribe("detours:active:" <> route_id)
+    %{id: user_id} = Guardian.Phoenix.Socket.current_resource(socket)
+
+    detours = Detours.active_detours_by_route(route_id)
+    {:ok, %{data: detours}, socket}
+  end
+
+  # Past
```
firestack, line 27: `user_id` seems unused?
```suggestion
```

test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -352,7 +352,7 @@ defmodule SkateWeb.DetoursControllerTest do
     test "will not return detours from other users", %{conn: conn} do
       current_user_id = populate_db_and_get_user(conn)
 
-      other_user = build(:user)
+      other_user = Skate.Settings.User.upsert("secondary_user_uid", "secondary_user@test.com")
```
firestack, line 355:
```suggestion
      other_user = insert(:user)
```

test/skate_web/channels/detours_channel_test.exs
```diff
@@ -0,0 +1,294 @@
...
+      |> Guardian.Phoenix.Socket.put_current_resource(%{id: authenticated_user_id})
+
+    start_supervised({Phoenix.PubSub, name: Skate.PubSub})
+
+    populate_db(conn)
+
+    {:ok, %{conn: conn, socket: socket}}
+  end
+
+  def populate_db(conn) do
```
firestack, line 29: nit(non-blocking): btw I'm running into issues with this pattern in `detours_controller_test` with `populate_db_and_get_user` and would prefer that we try and do these in the tests themselves, but I'll leave that up to you, as we can refactor later.

firestack: I'd also prefer we don't hardcode Ids, but the whole (insert into db, pull out, update snapshot with id, put back into db) rigamarole is annoying, so no worries. Just a small concern.

test/skate_web/controllers/detours_controller_test.exs
hannahpurcell: OHHHHHHHhhhh I see, awesome

test/skate_web/channels/detours_channel_test.exs
hannahpurcell: Mmmm I see I see. Let's talk when you get back. I can see the frustration with this, but want to confirm the better pattern (if there is one) before trying to edit

lib/skate/detours/detours.ex
```diff
@@ -112,6 +131,10 @@ defmodule Skate.Detours.Detours do
 
   def categorize_detour(_detour_context), do: :draft
 
+  @spec get_detour_route(detour :: map()) :: String.t()
+  defp get_detour_route(%{state: %{"context" => %{"route" => %{"name" => route_name}}}}),
```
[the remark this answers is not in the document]
hannahpurcell: ahh! Thanks for catching that!

test/support/factories/detour_factory.ex
```diff
@@ -74,12 +75,23 @@ defmodule Skate.DetourFactory do
...
+                "route" => %{"name" => _, "id" => _, "directionNames" => direction_names}
+              }
+            } = state,
+            route
+          ) do
+        put_in(state["context"]["route"], %{
+          "name" => route,
+          "id" => route,
+          "directionNames" => direction_names
+        })
```
firestack, line 94: I think we should have separate functions for `with_route` and `with_route_id` because in this form, it's not possible to have name be SL3 and id be 743, which the bug which introduced this change is based on them differing.

test/skate_web/channels/detours_channel_test.exs
```diff
@@ -112,13 +91,21 @@ defmodule SkateWeb.DetoursChannelTest do
     test "subscribes to draft detours with initial detours", %{conn: conn, socket: socket} do
       %{id: authenticated_user_id} = SkateWeb.AuthManager.Plug.current_resource(conn)
 
+      :detour |> build() |> with_id(1) |> insert()
+      :detour |> build() |> with_id(2) |> insert()
+      :detour |> build() |> with_id(3) |> insert()
+
+      put(conn, ~p"/api/detours/update_snapshot", %{
+        "snapshot" => :detour_snapshot |> build() |> with_id(4)
+      })
```
firestack, line 100: question: can you explain why this can't use the `:detour |> [...] |> insert()` method?

assets/src/hooks/useDetours.ts
```diff
@@ -182,8 +197,10 @@ const subscribeByRoute = (
 
   channel
     .join()
-    .receive("ok", ({ data: data }: { data: SimpleDetour[] }) => {
-      const detoursMap = Object.fromEntries(data.map((v) => [v.id, v]))
+    .receive("ok", ({ data: data }: { data: SimpleDetourData[] }) => {
+      const detoursMap = Object.fromEntries(
+        data.map((v) => [v.id, simpleDetourFromData(v)])
+      )
```
firestack, line 203: issue: since we're needing to rewrite what `useChannel` does so we can support multiple message types ("activated" | "deactivated" etc), this change to use `{ data: data }: { data: SimpleDetourData[] }` is actually hiding an issue that `data` may not actually be a `SimpleDetourData` object. The underlying library has `data` typed as `any`, but it should really be `unknown` which is why we're not getting errors here redefining the type as `SimpleDetourData[]`.

This should instead be using superstruct to validate the `data` before using it.

assets/src/hooks/useDetours.ts
```diff
@@ -21,17 +25,25 @@ const subscribe = (
...
   channel.on("auth_expired", reload)
 
   channel
     .join()
-    .receive("ok", ({ data: data }: { data: SimpleDetour[] }) => {
-      const detoursMap = Object.fromEntries(data.map((v) => [v.id, v]))
+    .receive("ok", ({ data: data }: { data: SimpleDetourData[] }) => {
+      const detoursMap = Object.fromEntries(
+        data.map((v) => [v.id, simpleDetourFromData(v)])
+      )
```
firestack, line 46: Example for https://github.com/mbta/skate/pull/2908/files/8acf7072c0392401f45e97dfc4e47d1e83b134a8..ecd135c15c69f5510318f45962d851c35238520e#r1909378792

```suggestion
        .receive("ok", ({ data: unknownData }: { data: unknown }) => {
      const data = create(unknownData, array(SimpleDetourData))

      const detoursMap = Object.fromEntries(
        data.map((v) => [v.id, simpleDetourFromData(v)])
      )
```

firestack (changes requested):
[Finished review]

test/support/factories/detour_factory.ex
hannahpurcell: Oh my goodness OF COURSE! Ok and I need to add a test for that specifically. Smh for missing that, but thank you

assets/src/hooks/useDetours.ts
hannahpurcell: Ohooooooo I see! Thanks for the example, too. Ok I'll need to do this for each of the `.receive("ok",`s and each of the `channel.on`s

hannahpurcell: https://github.com/mbta/skate/pull/2908/commits/69b649c28079d73cc8e1de3e8bb0e84fcabd46a5

assets/src/hooks/useDetours.ts
hannahpurcell: https://github.com/mbta/skate/pull/2908/commits/69b649c28079d73cc8e1de3e8bb0e84fcabd46a5

test/support/factories/detour_factory.ex
hannahpurcell: Well... maybe a test isn't the most useful in this case. But properly setting id / name is the right change to make at least

test/skate_web/channels/detours_channel_test.exs
firestack:
```suggestion
    test "subscribes to draft detours with initial detours", %{conn: conn, socket: socket, user: user} do
      %{id: authenticated_user_id} = SkateWeb.AuthManager.Plug.current_resource(conn)

      :detour |> build() |> with_id(1) |> insert()
      :detour |> build() |> with_id(2) |> insert()
      :detour |> build() |> with_id(3) |> insert()

      :detour |> build(author: user) |> with_id(4)
      # OR
      :detour |> build(author_id: authenticated_user_id) |> with_id(4)
```

firestack: I see! I think this should work
https://github.com/mbta/skate/pull/2908/files/8acf7072c0392401f45e97dfc4e47d1e83b134a8..ecd135c15c69f5510318f45962d851c35238520e#r1910687982
```suggestion
    :detour |> build(user: user) |> with_id(4)
```

firestack: Since `author` is set by the factory, I'm not _certain_ that you could use the `author_id` as is, or if you'd need to explicitly set `author` to `nil` in the case of using `author_id` or not.
```
      :detour |> build(author_id: authenticated_user_id, author: nil) |> with_id(4)

Comment (firestack) [reply test/skate_web/channels/detours_channel_test.exs]: But since you can get the current user from the test arguments it's probably easier to just use `author: user`

Comment (hannahpurcell) [reply test/skate_web/channels/detours_channel_test.exs:158]: OH BEAUTIFUL! Yes it sure does work, thank you 🚀

File: test/support/factories/detour_factory.ex [hunk 13]
```diff
@@ -73,6 +75,43 @@ defmodule Skate.DetourFactory do
...
+      def with_route(
+            %{"context" => %{"route" => %{"directionNames" => direction_names}}} = state,
+            %{name: route_name, id: route_id}
+          ) do
+        put_in(state["context"]["route"], %{
+          "id" => route_id,
+          "name" => route_name,
+          "directionNames" => direction_names
+        })
+      end
```
firestack, line 91: nit(non-blocking): `with_route_name` and `with_route_id` are unused here, I'd prefer to invoke them separately when needed, but you may consider this to avoid needing to populate the other keys present, like `direction_names`.
```suggestion
      def with_route(%Skate.Detours.Db.Detour{} = detour, %{name: _, id: _} = route) do
        %{detour | state: with_route(detour.state, route)}
      end

      def with_route(
            %{"context" => %{"route" => %{}}} = state,
            %{name: route_name, id: route_id}
          ) do
        state
        |> with_route_id(route_id)
        |> with_route_name(route_name)
      end
```

firestack (approved):
One nit, but LGTM 👍🏻

PR #2909, opened 2024-12-11
feat: delete draft detour UI

bfauble:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208932353926619/f

image
image
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ RUN apk add --no-cache --update curl
 WORKDIR /root
 
 ADD \
-  --checksum=sha256:390fdc813e2e58ec5a0def8ce6422b83d75032899167052ab981d8e1b3b14ff2 \
+  --checksum=sha256:05c6b4a9bf4daa95c888711481e19cc8e2e11aed4de8db759f51f4a6fce64917 \
   https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
   aws-cert-bundle.pem
 
--- a/assets/src/hooks/useDetours.ts
+++ b/assets/src/hooks/useDetours.ts
@@ -1,0 +1,280 @@
+import { Channel, Socket } from "phoenix"
+import {
+  SimpleDetour,
+  SimpleDetourData,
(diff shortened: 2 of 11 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -365,6 +373,25 @@ export const DiversionPage = ({
...
+                send({ type: "detour.delete.delete-modal.delete-draft" })
+              }
+              onCancel={() =>
+                send({ type: "detour.delete.delete-modal.cancel" })
+              }
+              routeName={routeName || "??"}
+              routeDescription={routeDescription || "??"}
+              routeOrigin={routeOrigin || "??"}
+              routeDirection={routeDirection || "??"}
+            />
```
firestack, line 393: nit(non-blocking): what do you think of doing this via composistion instead of taking the various props?

```
          affectedRoute={
            <AffectedRoute
              routeName={routeName ?? "??"}
              routeDescription={routeDescription ?? "??"}
              routeOrigin={routeOrigin ?? "??"}
              routeDirection={routeDirection ?? "??"}
            />
          }
```

assets/src/components/detours/detourPanels/detourFinishedPanel.tsx
```diff
@@ -81,15 +83,26 @@ export const DetourFinishedPanel = ({
...
-        {onActivateDetour && (
-          <Button
-            className="m-3 flex-grow-1 icon-link justify-content-center"
-            onClick={onActivateDetour}
-            data-fs-element="Begin Activate Detour"
-          >
-            <BsIcons.Power />
-            Start Detour
-          </Button>
+        {onDeleteDetour && (
```
hannahpurcell, line 86: The `onActivateDetour` check has been lost, so users who do have "activation" permissions will lose access to the activation button if they don't have the "deletion" permissions

firestack (commented):
Looks good! Just need to get CI to pass but then I think we can merge 👍🏻

PR #2910, opened 2024-12-18
feat(ex/skate/detours): add `activated_at` column

firestack:
To be able to show the activated at time in various places within Skate, as well as for sorting, we need to start recording the time of activation for a detour.
This PR
- adds the database column
- backfills the database column so that currently active detours are marked as having been active since the last modification time
- starts recording the activation time in the state machine when a detour is activated

---

Asana Ticket: https://app.asana.com/0/0/1208989312907929/f

### DIFF
--- a/test/skate_web/controllers/detours_controller_test.exs
+++ b/test/skate_web/controllers/detours_controller_test.exs
@@ -82,6 +82,23 @@ defmodule SkateWeb.DetoursControllerTest do
     end
 
     @tag :authenticated
+    test "adds `activated_at` field when provided", %{conn: conn} do
+      %Skate.Detours.Db.Detour{id: id, state: snapshot, activated_at: nil} = insert(:detour)
+
+      activated_at_time =
+        DateTime.utc_now() |> Skate.DetourFactory.browser_date() |> Skate.DetourFactory.db_date()
+
+      put(conn, ~p"/api/detours/update_snapshot", %{
+        "snapshot" => snapshot |> activated(activated_at_time) |> with_id(id)
+      })
+
+      Process.sleep(10)
+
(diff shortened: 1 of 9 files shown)


test/skate_web/controllers/detours_controller_test.exs
```diff
@@ -81,6 +81,22 @@ defmodule SkateWeb.DetoursControllerTest do
       assert Skate.Repo.aggregate(Notifications.Db.Detour, :count) == 1
     end
 
+    @tag :authenticated
+    test "adds `activated_at` field when provided", %{conn: conn} do
+      setup_notification_server()
```
hannahpurcell, line 86: So far the notification server setup is only used for notification tests. Is there a reason the server is needed for this test?

hannahpurcell (approved):
One question, otherwise looks great!

test/skate_web/controllers/detours_controller_test.exs
firestack: this is not needed, thanks for catching this!!
[`fc6cd74ae62fbeb10f88788d5ca0ee719d668ef7..ebe052e5db79e714d76bac510240110a3d9d644e`#diff-6b5f7db841](https://github.com/mbta/skate/compare/fc6cd74ae62fbeb10f88788d5ca0ee719d668ef7..ebe052e5db79e714d76bac510240110a3d9d644e#diff-6b5f7db841fd66bb4a09b0870661a72d957ad6cf263cccf090ebb4a251bc0687)L86

hannahpurcell (approved):
Oooo your date truncation notes are super helpful! Still looks good to me. Thanks for posting for re-review

PR #2911, opened 2024-12-21
 feat: add `activated-at` column to frontend detours list 

firestack:
This uses the work in #2910 to fix the Activated At column to use the _real_ activated at time, in the Activated Detours table.

Asana Ticket: https://app.asana.com/0/0/1208989312907930/f
Depends on:
- #2910

### DIFF
--- a/assets/src/components/detoursTable.tsx
+++ b/assets/src/components/detoursTable.tsx
@@ -1,8 +1,8 @@
 import React from "react"
 import { Table } from "react-bootstrap"
 import { RoutePill } from "./routePill"
-import { useCurrentTimeSeconds } from "../hooks/useCurrentTime"
-import { timeAgoLabel } from "../util/dateTime"
+import { useCurrentTime } from "../hooks/useCurrentTime"
+import { timeAgoLabel, timeAgoLabelFromDate } from "../util/dateTime"
 import { SimpleDetour } from "../models/detoursList"
 import { EmptyDetourTableIcon } from "../helpers/skateIcons"
 import { joinClasses } from "../helpers/dom"
@@ -55,7 +55,11 @@ export const DetoursTable = ({
     </thead>
     <tbody>
       {data ? (
-        <PopulatedDetourRows data={data} onOpenDetour={onOpenDetour} />
+        <PopulatedDetourRows
+          status={status}
+          data={data}
+          onOpenDetour={onOpenDetour}
+        />
       ) : (
         <EmptyDetourRows message={`No ${status} detours.`} />
       )}
@@ -65,12 +69,15 @@ export const DetoursTable = ({
 
 const PopulatedDetourRows = ({
   data,
+  status,
   onOpenDetour,
 }: {
   data: SimpleDetour[]
+  status: DetourStatus
   onOpenDetour: (detourId: number) => void
 }) => {
-  const epochNowInSeconds = useCurrentTimeSeconds()
+  const epochNow = useCurrentTime()
+  const epochNowInSeconds = epochNow.valueOf() / 1000
 
   return (
     <>
@@ -91,7 +98,9 @@ const PopulatedDetourRows = ({
             {detour.intersection}
           </td>
           <td className="align-middle p-3 u-hide-for-mobile">
-            {timeAgoLabel(epochNowInSeconds, detour.updatedAt)}
+            {status === DetourStatus.Active && detour.activatedAt
+              ? timeAgoLabelFromDate(detour.activatedAt, epochNow)
+              : timeAgoLabel(epochNowInSeconds, detour.updatedAt)}
           </td>
         </tr>
       ))}
--- a/assets/src/util/dateTime.ts
+++ b/assets/src/util/dateTime.ts
@@ -98,9 +98,19 @@ export const timeAgoLabel = (
   epochNowInSeconds: number,
   epochTime: number
 ): string => {
-  const duration = epochNowInSeconds - epochTime
-  const diffHours = Math.floor(duration / 3_600)
-  const diffMinutes = Math.floor((duration % 3_600) / 60)
+  const second = 1
+  const minute = second * 60
+  const hour = minute * 60
 
-  return diffHours >= 1 ? `${diffHours} hours ago` : `${diffMinutes} min ago`
+  const duration = epochNowInSeconds - epochTime
+
+  if (duration > 1 * hour) {
+    return `${Math.floor(duration / hour)} hours ago`
+  }
+
+  if (duration > 1 * minute) {
+    return `${Math.floor(duration / minute)} min ago`
+  }
+
+  return "Just now"
 }
--- a/assets/src/util/dateTime.ts
+++ b/assets/src/util/dateTime.ts
@@ -114,3 +114,8 @@ export const timeAgoLabel = (
 
   return "Just now"
 }
+
+export const timeAgoLabelFromDate = (start: Date, end: Date) => {
+  const second = 1000
+  return timeAgoLabel(end.valueOf() / second, start.valueOf() / second)
+}
--- a/lib/skate/detours/detour.ex
+++ b/lib/skate/detours/detour.ex
@@ -30,6 +30,59 @@ defmodule Skate.Detours.Detour do
       :author_id,
       :status
     ]
+
+    def from(
+          status,
+          %{
+            state: %{
+              "context" => %{
+                "route" => %{"name" => route_name, "directionNames" => direction_names},
+                "routePattern" => %{"headsign" => headsign, "directionId" => direction_id},
+                "nearestIntersection" => nearest_intersection
+              }
+            }
+          } = db_detour
+        ) do
+      direction = Map.get(direction_names, Integer.to_string(direction_id))
+
+      %__MODULE__{
+        id: db_detour.id,
+        route: route_name,
+        direction: direction,
+        name: headsign,
+        intersection: nearest_intersection,
+        updated_at: timestamp_to_unix(db_detour.updated_at),
+        author_id: db_detour.author_id,
+        status: status
+      }
+    end
+
+    def from(_status, _attrs), do: nil
+
+    # Converts the db timestamp to unix
+    defp timestamp_to_unix(db_date) do
+      db_date
+      |> DateTime.from_naive!("Etc/UTC")
+      |> DateTime.to_unix()
+    end
+  end
+
+  defmodule ActivatedDetourDetails do
+    @moduledoc """
+    Extended information for active detours
+    """
+
+    @type t :: %__MODULE__{
+            activated_at: DateTime.t(),
+            details: Detailed.t()
+          }
+
+    @derive Jason.Encoder
+
+    defstruct [
+      :activated_at,
+      :details
+    ]
   end
 
   defmodule WithState do
--- a/lib/skate/detours/detours.ex
+++ b/lib/skate/detours/detours.ex
@@ -4,6 +4,7 @@ defmodule Skate.Detours.Detours do
   """
 
   import Ecto.Query, warn: false
+  alias Skate.Detours.Detour.ActivatedDetourDetails
   alias Skate.Repo
   alias Skate.Detours.Db.Detour
   alias Skate.Detours.SnapshotSerde
@@ -67,7 +68,10 @@ defmodule Skate.Detours.Detours do
       list_detours()
       |> Enum.map(&db_detour_to_detour/1)
       |> Enum.filter(& &1)
-      |> Enum.group_by(fn detour -> detour.status end)
+      |> Enum.group_by(fn
+        %ActivatedDetourDetails{details: %{status: status}} -> status
+        %{status: status} -> status
+      end)
 
     %{
       active: Map.get(detours, :active, []),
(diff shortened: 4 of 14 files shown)


assets/src/components/detourListPage.tsx
```diff
@@ -75,12 +84,53 @@ export const DetourListPage = () => {
             visibility="All Skate users"
             classNames={["d-flex"]}
           />
-          <DetoursTable
```
hannahpurcell, line 78: Replacing `DetoursTable` gets rid of the empty table state, which we need in case a miracle happens and there are absolutely no active detours

hannahpurcell: It doesn't seem like it would be too complex to use the `status === DetourStatus.Active` to determine whether the extra column should be added, which should be all we need for this feature. Is there something else that was changed with the replacement?

firestack: > Replacing `DetoursTable` gets rid of the empty table state, which we need in case a miracle happens and there are absolutely no active detours

Good catch, Thanks!

lib/skate/detours/detours.ex
```diff
@@ -61,29 +65,9 @@ defmodule Skate.Detours.Detours do
...
-      updated_at: timestamp_to_unix(db_detour.updated_at),
-      author_id: db_detour.author_id,
-      status: categorize_detour(db_detour)
-    }
+  @spec db_detour_to_detour(status :: detour_type(), Detour.t()) :: DetailedDetour.t() | nil
+  def db_detour_to_detour(%{} = db_detour) do
+    db_detour_to_detour(categorize_detour(db_detour), db_detour)
   end
 
   def db_detour_to_detour(invalid_detour) do
```
hannahpurcell, line 73: This has kinda been replaced by `db_detour_to_detour/2` on lines 101-7, because the only reason this function would execute is if the detour is simply not a map at all (probably if it's nil?) I'm not sure this function would still be useful in that case, but if it is, then the error message should be changed

lib/skate/detours/detour.ex
```diff
@@ -30,6 +30,59 @@ defmodule Skate.Detours.Detour do
...
+          %{
+            state: %{
+              "context" => %{
+                "route" => %{"name" => route_name, "directionNames" => direction_names},
+                "routePattern" => %{"headsign" => headsign, "directionId" => direction_id},
+                "nearestIntersection" => nearest_intersection
+              }
+            }
+          } = db_detour
+        ) do
```
hannahpurcell, line 45: I like moving this logic here! I'm trying to think of an alternative name but don't have a great one... Maybe an explanatory comment would be good though? I see `from` and assumed it was something like a kernel operation I didn't know about

assets/src/util/dateTime.ts
```diff
@@ -104,3 +104,21 @@ export const timeAgoLabel = (
 
   return diffHours >= 1 ? `${diffHours} hours ago` : `${diffMinutes} min ago`
 }
+
+const second = 1000
+const minute = second * 60
+const hour = minute * 60
+
+export const timeAgoLabelForDates = (start: Date, end: Date) => {
```
hannahpurcell, line 112: What's the purpose / difference between this and `timeAgoLabel`? The input / output seems spiritually the same (even if they take different input types)

assets/src/components/detourListPage.tsx
firestack: > It doesn't seem like it would be too complex to use the `status === DetourStatus.Active` to determine whether the extra column should be added, which should be all we need for this feature. Is there something else that was changed with the replacement?

I think doing this logic within the component would make it unreasonably messy, and would require a larger refactor to clean up the component to accept different data types, in my opinion, this way was just generally simpler and straight forward (and _in my opinion_ points in the direction that we should not have made the table into it's own component when we did).

I tried mocking up what it would be like to do this within the existing table component to decipher which types are being used at which points, which I think introduces unnecessary conditional logic to access the types.

```diff
Commit ID: 1738660c49f3a596fd94026a3f0a3bd56f8379ed
Change ID: xpltoxzkmrnpwsqvmylwtvlskrotnluw
Author   : Kayla Firestack <kfirestack@mbta.com> (2025-01-09 09:36:38)
Committer: Kayla Firestack <kfirestack@mbta.com> (2025-01-09 10:01:41)

    (no description set)

diff --git a/assets/src/components/detourListPage.tsx b/assets/src/components/detourListPage.tsx
index 4a8a847d39..2e21a7a35e 100644
--- a/assets/src/components/detourListPage.tsx
+++ b/assets/src/components/detourListPage.tsx
@@ -84,57 +84,12 @@
             visibility="All Skate users"
             classNames={["d-flex"]}
           />
-          <Table
-            hover={!!detours.active}
-            className={"mb-5 c-detours-table"}
-            variant={"active-detour"}
-          >
-            <thead className="u-hide-for-mobile">
-              <tr>
-                <th className="px-3 py-4">Route and direction</th>
-                <th className="px-3 py-4 u-hide-for-mobile">
-                  Starting Intersection
-                </th>
-                <th className="px-3 py-4 u-hide-for-mobile">
-                  {timestampLabelFromStatus(DetourStatus.Active)}
-                </th>
-                <th className="px-3 py-4 u-hide-for-mobile">Est. Duration</th>
-              </tr>
-            </thead>
-            <tbody>
-              {detours.active
-                ? detours.active.map((detour, index) => (
-                    <tr
-                      key={index}
-                      onClick={() => onOpenDetour(detour.details.id)}
-                    >
-                      <td className="align-middle p-3">
-                        <div className="d-flex">
-                          <RoutePill routeName={detour.details.route} />
-                          <div className="c-detours-table__route-info-text d-inline-block">
-                            <div className="pb-1 fs-4 fw-semibold">
-                              {detour.details.name}
-                            </div>
-                            <div className="c-detours-table__route-info-direction fs-6">
-                              {detour.details.direction}
-                            </div>
-                          </div>
-                        </div>
-                      </td>
-                      <td className="align-middle p-3 u-hide-for-mobile">
-                        {detour.details.intersection}
-                      </td>
-                      <td className="align-middle p-3 u-hide-for-mobile">
-                        {timeAgoLabelForDates(detour.activatedAt, epochNow)}
-                      </td>
-                      <td className="align-middle p-3 u-hide-for-mobile">
-                        {detour.estimatedDuration}
-                      </td>
-                    </tr>
-                  ))
-                : null}
-            </tbody>
-          </Table>
+          <DetoursTable
+            data={detours.active}
+            status={DetourStatus.Active}
+            onOpenDetour={onOpenDetour}
+            classNames={["mb-5"]}
+          />
           {userInTestGroup(TestGroups.DetoursPilot) && (
             <>
               <Title
diff --git a/assets/src/components/detoursTable.tsx b/assets/src/components/detoursTable.tsx
index f8fe3d9824..77af66700c 100644
--- a/assets/src/components/detoursTable.tsx
+++ b/assets/src/components/detoursTable.tsx
@@ -1,19 +1,30 @@
 import React from "react"
 import { Table } from "react-bootstrap"
 import { RoutePill } from "./routePill"
-import { useCurrentTimeSeconds } from "../hooks/useCurrentTime"
-import { timeAgoLabel } from "../util/dateTime"
-import { SimpleDetour } from "../models/detoursList"
+import useCurrentTime, { useCurrentTimeSeconds } from "../hooks/useCurrentTime"
+import { timeAgoLabel, timeAgoLabelForDates } from "../util/dateTime"
+import { ActivatedDetour, SimpleDetour } from "../models/detoursList"
 import { EmptyDetourTableIcon } from "../helpers/skateIcons"
 import { joinClasses } from "../helpers/dom"
+import { ActivateDetour } from "./detours/activateDetourModal"
 
-interface DetoursTableProps {
-  data: SimpleDetour[] | undefined
+interface DetoursTableBaseProps {
   onOpenDetour: (detourId: number) => void
-  status: DetourStatus
   classNames?: string[]
 }
 
+type DetoursTableProps = DetoursTableBaseProps &
+  (
+    | {
+        data: ActivatedDetour[] | undefined
+        status: DetourStatus.Active
+      }
+    | {
+        data: SimpleDetour[] | undefined
+        status: DetourStatus.Closed | DetourStatus.Draft
+      }
+  )
+
 export enum DetourStatus {
   Draft = "draft",
   Active = "active",
@@ -51,11 +62,26 @@
         <th className="px-3 py-4 u-hide-for-mobile">
           {timestampLabelFromStatus(status)}
         </th>
+        {status === DetourStatus.Active && (
+          <th className="px-3 py-4 u-hide-for-mobile">Est. Duration</th>
+        )}
       </tr>
     </thead>
     <tbody>
       {data ? (
-        <PopulatedDetourRows data={data} onOpenDetour={onOpenDetour} />
+        status === DetourStatus.Active ? (
+          <PopulatedDetourRows
+            status={status}
+            data={data}
+            onOpenDetour={onOpenDetour}
+          />
+        ) : (
+          <PopulatedDetourRows
+            status={status}
+            data={data}
+            onOpenDetour={onOpenDetour}
+          />
+        )
       ) : (
         <EmptyDetourRows message={`No ${status} detours.`} />
       )}
@@ -65,36 +91,67 @@
 
 const PopulatedDetourRows = ({
   data,
+  status,
   onOpenDetour,
 }: {
-  data: SimpleDetour[]
   onOpenDetour: (detourId: number) => void
-}) => {
-  const epochNowInSeconds = useCurrentTimeSeconds()
+} & (
+  | {
+      data: ActivatedDetour[]
+      status: DetourStatus.Active
+    }
+  | {
+      data: SimpleDetour[]
+      status: DetourStatus.Closed | DetourStatus.Draft
+    }
+)) => {
+  const epochNow = useCurrentTime()
+  const epochNowInSeconds = epochNow.valueOf() / 1000
 
   return (
     <>
-      {data.map((detour, index) => (
-        <tr key={index} onClick={() => onOpenDetour(detour.id)}>
-          <td className="align-middle p-3">
-            <div className="d-flex">
-              <RoutePill routeName={detour.route} />
-              <div className="c-detours-table__route-info-text d-inline-block">
-                <div className="pb-1 fs-4 fw-semibold">{detour.name}</div>
-                <div className="c-detours-table__route-info-direction fs-6">
-                  {detour.direction}
+      {data.map((detour, index) => {
+        let simpleDetour: SimpleDetour
+        if (status == DetourStatus.Active) {
+          simpleDetour = (detour as ActivatedDetour).details
+        } else {
+          simpleDetour = detour as SimpleDetour
+        }
+
+        return (
+          <tr key={index} onClick={() => onOpenDetour(simpleDetour.id)}>
+            <td className="align-middle p-3">
+              <div className="d-flex">
+                <RoutePill routeName={simpleDetour.route} />
+                <div className="c-detours-table__route-info-text d-inline-block">
+                  <div className="pb-1 fs-4 fw-semibold">
+                    {simpleDetour.name}
+                  </div>
+                  <div className="c-detours-table__route-info-direction fs-6">
+                    {simpleDetour.direction}
+                  </div>
                 </div>
               </div>
-            </div>
-          </td>
-          <td className="align-middle p-3 u-hide-for-mobile">
-            {detour.intersection}
-          </td>
-          <td className="align-middle p-3 u-hide-for-mobile">
-            {timeAgoLabel(epochNowInSeconds, detour.updatedAt)}
-          </td>
-        </tr>
-      ))}
+            </td>
+            <td className="align-middle p-3 u-hide-for-mobile">
+              {simpleDetour.intersection}
+            </td>
+            <td className="align-middle p-3 u-hide-for-mobile">
+              {status === DetourStatus.Active
+                ? timeAgoLabelForDates(
+                    (detour as ActivatedDetour).activatedAt,
+                    epochNow
+                  )
+                : timeAgoLabel(epochNowInSeconds, simpleDetour.updatedAt)}
+            </td>
+            {status === DetourStatus.Active && (
+              <td className="align-middle p-3 u-hide-for-mobile">
+                {(detour as ActivatedDetour).estimatedDuration}
+              </td>
+            )}
+          </tr>
+        )
+      })}
     </>
   )
 }
```

firestack: If you see a better way to have done this though, I'd be happy to pair or chat about a diff or suggestion!

assets/src/util/dateTime.ts
firestack: The main difference is the addition of "Just now" but I made a separate function for "development speed" sake, I'll see about condensing these!

lib/skate/detours/detour.ex
firestack: I personally would like to keep this convention (and do it more 😅), I _did_ pick it up from Rust [which got inspiration from some other language IIRC] (there's a From/Into convention, where you can do 
```rs
1.into<String>()
// or 
String::from(1)
```
and I think this is a good way to do conversions.

firestack: It could be named `from_detour_snapshot` though. I think `from` is still the best descriptor/word though.

assets/src/util/dateTime.ts
firestack: how's this? https://github.com/mbta/skate/compare/08685fb499b5cbfb8a076b3d605cd401047bfb5d..700b88734064884c7f03bdc1dbb6a3392794a6e6

lib/skate/detours/detours.ex
firestack: I'm still thinking about this, but I'm going to re-request review while I consider this, to unblock the other conversations in case I can iterate on those faster.

assets/src/util/dateTime.ts
firestack: I am a little surprised by the snapshot change, so I'm gonna review that tomorrow (grr, snapshots >.>)

assets/src/components/detourListPage.tsx
hannahpurcell: Hm I seeeeeeeee. So it seems primarily tricky because SimpleDetour is , and the typing could be made consistent if the activation details were optional attributes on SimpleDetour?

```
SimpleDetour {
  id: DetourId
  route: string
  direction: string
  name: string
  intersection: string
  updatedAt: number
  activatedAt?: Date
}
```

This is what I was kinda thinking about when I suggested that activation info be nested within SimpleDetour as opposed to a wrapper around SimpleDetour. But I get that may have made things hairier on the backend. Hrmm. How do you feel about parsing the differently-typed data from the backend into a consistent "SimpleDetour" structure on the frontend?

lib/skate/detours/detour.ex
firestack: just a note: I'd recommend leaning on Elixir LS and VSCode LSP tooling for that kind of search, because you can do a "Find all references" search which uses the information from the compiler to find _actual_ references based on the compiled code, rather than text search.

assets/src/util/dateTime.ts
hannahpurcell: Yeah, looks good! Just... grr snapshots

firestack: Okay fixed the snapshots! It was the "seconds" value 😅

lib/skate/detours/detours.ex
firestack: Removed the function for now, I think we'll be coming back though here for the refactor anyway so I think that'll be fine for now https://github.com/mbta/skate/compare/752d652b63447cf46f6a5a6aabcdb0f3eb36a98f..f88875b154cd8218a2eadca2adce531963439e29

hannahpurcell (commented):
Finishing review, to revisit once [bug](https://github.com/mbta/skate/pull/2911/files#r1908841790) is fixed

assets/src/components/detourListPage.tsx
firestack: Resolved 👍🏻 thanks for the suggestion https://github.com/mbta/skate/compare/f88875b154cd8218a2eadca2adce531963439e29..eb4f28a56ebc9958b9069b5bc35f299e86f76581

hannahpurcell: Oh beautiful, that looks awesome

assets/src/components/detoursTable.tsx
```diff
@@ -65,12 +69,15 @@ export const DetoursTable = ({
...
+  status,
   onOpenDetour,
 }: {
   data: SimpleDetour[]
+  status: DetourStatus
   onOpenDetour: (detourId: number) => void
 }) => {
-  const epochNowInSeconds = useCurrentTimeSeconds()
+  const epochNow = useCurrentTime()
+  const epochNowInSeconds = epochNow.valueOf() / 1000
```
hannahpurcell, line 80: non-blocking: what's the purpose of this change? Looks like `useCurrentTime` uses the same underlying logic as `useCurrentTimeSeconds` (other than the divide by seconds)

hannahpurcell (approved):
Alright, great job! We talked about the data model synchronously and all the changes across test factories, etc, make sense. We reserve the right to re-structure some of these types if and when we need to after the db refactor!

PR #2917, opened 2025-01-02
tweak: Is the "_" supposed to be there when there is no variant number?

hannahpurcell:
Tiny thing, but want to ask design!

Screenshot 2025-01-02 at 3 57 25 PM


hannahpurcell:
Worth talking about, but not ready to be completed at this time

PR #2915, opened 2025-01-02
feat: Base work to get detours in route ladder dropdown

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208748637161894

To be covered in a future PR:
- [ ] Read-only view
- [ ] Route pattern id in detour name

Needed:
- [ ] Tests

PR #2919, opened 2025-01-03
tweak: Added hook for the recurrent pattern of loading one detour

hannahpurcell:
Built off of https://github.com/mbta/skate/pull/2908. This pattern will be used again in https://github.com/mbta/skate/pull/2915, so it made sense to make a hook for it.


firestack (approved):
lgtm!

PR #2922, opened 2025-01-14
fix: update checksum for aws cert bundle

hannahpurcell:
Here we are again! Updating the checksum to unblock deploys

Prior work: https://github.com/mbta/skate/pull/2901


firestack (approved):
Thanks for fixing this!

PR #2924, opened 2025-01-17
feat(ts/diversionPage): remove timestamp from diversion page headers

firestack:


firestack:
Wasn't sure what to do for tests here, anyone has any thoughts for what should be tested (as I'm not sure we should be testing the _absence_ of something, let me know)

hannahpurcell:
> Wasn't sure what to do for tests here

Hmm! Yeah I see what you mean. It'd be peculiar to have a test to check absence of something. Maybe once it's been added to the side panel, perhaps that test could query for all instances of "On detour since" and verify it's visible, but also something else that has been updated about the element between the two versions? Is there a way to check its parent element? There is a way to check elements contained by the selected one, but that might not be enough

hannahpurcell (commented):
I'm comfortable approving, but awaiting discussion on tests

firestack:
> Well hm, another way of thinking about it. There should have been tests to check the text variation between the 3 detour statuses, and those weren't written. If those existed, how would we have updated the test to pass here?

I _think_ we probably would've had 3 tests and would remove one with the removal of the feature? We could've also inverted the test so that we _expect_ it to fail.

hannahpurcell:
> I _think_ we probably would've had 3 tests and would remove one with the removal of the feature?

Yeah, I agree that we'd probably just remove a test. I think my philosophy, then, would be not introducing a test to verify an omission. So approving!

hannahpurcell (approved):

PR #2925, opened 2025-01-17
feat: delete draft detour backend

bfauble:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208845212421698/f

### DIFF
--- a/assets/tests/components/detours/diversionPage.delete.test.tsx
+++ b/assets/tests/components/detours/diversionPage.delete.test.tsx
@@ -127,7 +127,7 @@ describe("Detours Page: Open a Detour", () => {
     await userEvent.click(cancelButton.get())
 
     expect(
-      await screen.queryAllByText("Are you sure you want to delete this draft?")
-    ).toStrictEqual([])
+      screen.queryByText("Are you sure you want to delete this draft?")
+    ).not.toBeInTheDocument()
   })
 })


assets/src/api.ts
```diff
@@ -556,6 +557,23 @@ export const fetchDetour = (
     ErrStruct: never(),
   }).then((v) => map(v, detourStateFromData))
 
+export const deleteDetour = (
+  id: number | undefined
```
firestack, line 561: question: when would `id` be `undefined`?

assets/src/components/detours/diversionPage.tsx
```diff
@@ -171,7 +182,11 @@ export const DiversionPage = ({
     }
   }
 
-  const detourPanel = () => {
+  const detourPanel: ({
+    deleteDetourCallback,
+  }: {
+    deleteDetourCallback: () => void
+  }) => React.JSX.Element = () => {
```
firestack, line 189: question: What's the benefit of getting this via a parameter instead of pulling it from the scope?

assets/src/hooks/useDetours.ts
```diff
@@ -170,6 +179,13 @@ export const useDraftDetours = (socket: Socket | undefined) => {
     })
   }
 
+  const handleDeleted = (detour_id: number) => {
```
firestack, line 182: nit: as much as I love snake_case, we should be using camelCase for variables on the TS side.

I'm honestly baffled we don't have a lint for this.

assets/src/models/createDetourMachine.ts
firestack, line 1: These changes don't quite make sense to me, and part of that is on me for not seeing this in #2909 

As is, when the `detour.delete.delete-modal.delete-draft` event is sent, the state machine will go through the following states

- `"Detour Drawing"."Share Detour"."Deleting"."Confirming"` [Start state]
- `"Detour Drawing"."Share Detour"."Deleting"."Deleted"`
- `"Detour Drawing"."Share Detour"."Deleting"."Done"`
- `"Detour Drawing"."Share Detour"."DoneNoSave"`
- `"Detour Drawing"."Active"`

firestack: Oh oops this posted before I meant it to.

firestack: I think you might want something like this instead

```
            Deleting: {
              on: {
                "detour.delete.delete-modal.cancel": {
                  target: "Reviewing",
                },
                "detour.delete.delete-modal.delete-draft": {
                  tags: "no-save",
                  target: "Deleted"
                }
              },
            },
            Deleted: { tags: "no-save" },
```

firestack: Let's discuss though!

firestack: x-posting/summarizing from slack: it may be better to setup a top level state that halt's the machine and use an `id` to go to that state from the nested state.

firestack: I think there's still an issue with the way this is constructed, it shouldn't be necessary to delete everything locally, and I think that's a symptom of the flow being incorrect. 

```diff
diff --git a/assets/src/models/createDetourMachine.ts b/assets/src/models/createDetourMachine.ts
index b2693d8c67..6ba87bd293 100644
--- a/assets/src/models/createDetourMachine.ts
+++ b/assets/src/models/createDetourMachine.ts
@@ -210,22 +210,6 @@
       finishedDetour: undefined,
       detourShape: undefined,
     }),
-    "detour.deleted": assign({
-      uuid: undefined,
-      route: undefined,
-      routePattern: undefined,
-      routePatterns: undefined,
-      waypoints: [],
-      startPoint: undefined,
-      endPoint: undefined,
-      nearestIntersection: undefined,
-      detourShape: undefined,
-      finishedDetour: undefined,
-      editedDirections: undefined,
-      selectedDuration: undefined,
-      selectedReason: undefined,
-      activatedAt: undefined,
-    }),
   },
 }).createMachine({
   id: "Detours Machine",
@@ -669,15 +653,10 @@
                 },
                 "detour.delete.delete-modal.delete-draft": {
                   tags: "no-save",
-                  target: "Deleted",
-                  actions: "detour.deleted",
+                  target: "#Deleted",
                 },
               },
             },
-            Deleted: {
-              target: "Done",
-              tags: "no-save",
-            },
             Done: { type: "final" },
           },
         },
@@ -708,6 +687,11 @@
           },
         },
         Past: {},
+        Deleted: {
+          id: "Deleted",
+          type: "final",
+          tags: "no-save",
+        },
       },
     },
     SaveState: {
```

firestack (commented):
Looks good! I think you said you were adding tests? So I'll just await those 👍🏻

assets/tests/components/detours/diversionPage.delete.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+  test("calls API with correct detour ID", async () => {
+    jest
+      .mocked(fetchDetours)
+      .mockResolvedValue(Ok(detourListFactoryWithDraft.build()))
+
+    render(<DetourListPage />)
+
+    await userEvent.click(await screen.findByText("Draft Detour 123"))
+    expect(fetchDetour).toHaveBeenCalledWith(123)
+  })
```
firestack, line 61: question: do you think we need this test if this specific case is covered by `detoursListPage.openDetour.test.tsx`?

assets/tests/components/detours/diversionPage.delete.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    expect(
+      await screen.findByText("Are you sure you want to delete this draft?")
+    ).toBeVisible()
+
+    await userEvent.click(cancelButton.get())
+
+    expect(
+      await screen.queryAllByText("Are you sure you want to delete this draft?")
+    ).toStrictEqual([])
+  })
```
firestack, line 132: Can we use regular `queryByText`? and the RTL library provides us with `not.toBeInTheDocument`, which works with the result of `queryBy` IIRC.

And I don't _think_ `queryBy` functions return a promise.
```suggestion
    expect(
      screen.queryByText("Are you sure you want to delete this draft?")
    ).not.toBeInTheDocument()
  })
```

firestack (approved):

PR #2926, opened 2025-01-22
feat(ex/formatter): configure migrations formatter in directory

firestack:
This makes `.formatter.exs` match `phx.new` template, and moves formatter config for specific directories into local files in those directories, to match "new" phoenix conventions.make `.formatter.exs` match `phx.new` template

This also configures the async_migrations folder to be formatted, and formats those files.


hannahpurcell (approved):
Assigning myself since I had some spare time and this was an easy read! ✅

PR #2928, opened 2025-01-23
feat: Change detour duration modal

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1208845212421705

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -2,6 +2,7 @@ import React, {
   ComponentProps,
   ComponentPropsWithoutRef,
   PropsWithChildren,
+  useCallback,
   useContext,
   useState,
 } from "react"
@@ -33,6 +34,7 @@ import { Drawer } from "../drawer"
 import { isMobile } from "../../util/screenSize"
 import { AffectedRoute } from "./detourPanelComponents"
 import { ChangeDuration } from "./changeDurationModal"
+import { deleteDetour } from "../../api"
 
 const displayFieldsFromRouteAndPattern = (
   route: Route,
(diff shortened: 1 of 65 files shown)


assets/tests/factories/detourStateMachineFactory.ts
```diff
@@ -34,8 +34,15 @@ export const detourInProgressFactory = Factory.define<DetourWithState>(() => {
   }
 })
 
-export const activeDetourFactory = Factory.define<DetourWithState>(() => {
-  // Stub out a detour machine, and start a detour-in-progress
+interface DetourWithStateTransientParams extends DetourWithState {
```
firestack, line 37: question: is the `extend` nessicary?

assets/src/components/detours/diversionPage.tsx
```diff
@@ -458,6 +461,32 @@ export const DiversionPage = ({
...
             />
           ) : null}
+          {snapshot.matches({
+            "Detour Drawing": { Active: "Changing Duration" },
+          }) ? (
+            <ChangeDuration.Modal
+              onCancel={() =>
+                send({ type: "detour.active.change-duration-modal.cancel" })
+              }
+              onNext={() =>
```
firestack, line 471: nit(non-blocking?): Could we convert `onActivate` to `onDone` and use that rather than `onNext`?

assets/tests/factories/detourStateMachineFactory.ts
hannahpurcell: ahhh I assumed that all params were needed in the object, and that's not correct. you're right, omitting!

```suggestion
interface DetourWithStateTransientParams {
```

assets/src/components/detours/diversionPage.tsx
hannahpurcell: Mmm I see! I used the next step because there was already a concept of `nextStepLabel` that I could use for the `fs` label. So I'd have to introduce more props to turn onActivate into a generic button, but I'm open to that

hannahpurcell: Will wait to discuss this further and apply this next week if needed. Making note of it!

PR #2929, opened 2025-01-23
fix(ex/snapshot_serde): fix "activatedAt" not present in old snapshots at serialize time

firestack:
I don't think this is a full solution, but I think this addresses the bug well enough that we can ship other things by tomorrow.


hannahpurcell (approved):
Fab! 🎉

PR #2930, opened 2025-01-23
fix: useActiveDetours instead of -ByRoute to avoid db-related failures

hannahpurcell:
Active detours by route is failing with the long db queries; just useActiveDetours instead right now.

### DIFF
--- a/assets/src/components/routeLadders.tsx
+++ b/assets/src/components/routeLadders.tsx
@@ -9,8 +9,8 @@ import RoutesContext from "../contexts/routesContext"
 import useTimepoints from "../hooks/useTimepoints"
 import { SocketContext } from "../contexts/socketContext"
 import useAlerts from "../hooks/useAlerts"
-import { useActiveDetoursByRoute } from "../hooks/useDetours"
-import { DetourId } from "../models/detoursList"
+import { DetoursMap, useActiveDetours } from "../hooks/useDetours"
+import { DetourId, SimpleDetour } from "../models/detoursList"
 
 export const findRouteById = (
   routes: Route[] | null,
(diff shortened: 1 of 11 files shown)


assets/src/components/routeLadders.tsx
```diff
@@ -63,15 +63,25 @@ const RouteLadders = ({
   const alerts = useAlerts(socket, selectedRouteIds)
   const routesWithAlerts: RouteId[] = []
 
-  const skateDetours = useActiveDetoursByRoute(socket, selectedRouteIds)
+  // TODO: once DB is normalized, we can open individual channels for each route ladder
```
firestack, line 66:
```suggestion
  // TODO: once DB is optimized for querying by route and status, we can open individual channels for each route ladder
```

assets/tests/components/routeLadders.test.tsx
```diff
@@ -114,11 +114,9 @@ describe("RouteLadders", () => {
...
         .mocked(useTimepoints)
         .mockImplementationOnce(() => timepointsByRouteId)
-      jest.mocked(useActiveDetoursByRoute).mockReturnValue({
-        "1": {
-          "1": simpleDetourFactory.build({ id: 1 }),
-          "2": simpleDetourFactory.build({ id: 2 }),
-        },
+      jest.mocked(useActiveDetours).mockReturnValue({
+        "1": simpleDetourFactory.build({ id: 1 }),
+        "2": simpleDetourFactory.build({ id: 2 }),
```
firestack, line 119: Hmm, I'm a little surprised that this is working without `routeId` specified, do you know what I'm missing/why this works without it?

hannahpurcell: Oooo yes, and it's totally an accident. The factory successfully generated route 1 for the first one and route 2 for the second one, but if the tests were reordered and the factory generated 3 or 4, this would break. Thanks!

hannahpurcell: I added the route id: https://github.com/mbta/skate/pull/2930/commits/af3dd74a4c4a10f0e04a09de2904d0cf7a656ecf

PR #2932, opened 2025-01-23
fix(ts/activeDetourPanel): fix time ago label

firestack:
The arguments were reversed 🙃 
related to #2923

### DIFF
--- a/assets/src/components/detours/detourPanels/activeDetourPanel.tsx
+++ b/assets/src/components/detours/detourPanels/activeDetourPanel.tsx
@@ -106,7 +106,7 @@ export const ActiveDetourPanel = ({
               On detour since
             </dt>
             <dd aria-labelledby={dlActiveSinceId}>
-              {timeAgoLabelFromDate(currentTime, activatedAt)}
+              {timeAgoLabelFromDate(activatedAt, currentTime)}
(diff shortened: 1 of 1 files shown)


hannahpurcell (commented):
😅 good catch! was there a test missing?

firestack:
> 😅 good catch! was there a test missing? (Don't mind approving if we run out of time between now and tomorrow morning)

I don't _think_ there's an easy way to test this as this is more integration between the backend and frontend? The tests we have only really can test in the 

> 😅 good catch! was there a test missing? (Don't mind approving if we run out of time between now and tomorrow morning)

I've been trying to figure out if there could've been a test, I think the biggest issue is that there isn't a way to fast forward the time from when the detour is activated in the test to when it's viewed. so the test that asserted the functionality of this specific invocation only asserted that it was `"Just now"` and not some specific amount of time, like `10 min ago`.

hannahpurcell (approved):
I'll leave myself a note to keep noodling on the test, but makes sense that there isn't an easy check here 👍

firestack:
> I'll leave myself a note to keep noodling on the test, but makes sense that there isn't an easy check here 👍

I'll keep thinking too, but merge for now

PR #2934, opened 2025-01-24
fix(ts/routeLadder): show dropdown on mobile breakpoint

firestack:
## Before:
before change image

## After:
after change image


hannahpurcell (approved):
thanks!

PR #2936, opened 2025-01-24
fix(ts/createDetourMachine): create `activatedAt` date when invoked

firestack:
Currently, the date is being created when the machine starts, _not_ when the action is fired 🤣 😭 whoops.


hannahpurcell (approved):
❗ wow!

PR #2942, opened 2025-01-27
fix: Temporarily turn off dropdown in route ladder for non-dispatch

hannahpurcell:


assets/tests/components/routeLadder.test.tsx
```diff
@@ -367,38 +367,38 @@ describe("routeLadder", () => {
...
-      />
-    )
-
-    await userEvent.click(
-      screen.getByRole("button", { name: /28 Route Options/ })
-    )
-
-    expect(screen.queryByText("Add detour")).not.toBeInTheDocument()
-  })
+  // test("does not render 'Add detour' in the dropdown if not an admin", async () => {
```
firestack, line 370: Would using `.failing` work here?
```suggestion
  // test.failing("does not render 'Add detour' in the dropdown if not an admin", async () => {
```

PR #2944, opened 2025-01-28
fix: Updating the spacing on detour panel

hannahpurcell:
Related to the UAT work [here](https://app.asana.com/0/1205385723132845/1209237355450159). Updates the spacing of items on the active detour panel, which had some overlap with other detour panels.

Screenshot 2025-01-28 at 5 22 45 PM
Screenshot 2025-01-28 at 5 22 57 PM
Screenshot 2025-01-28 at 5 23 06 PM
Screenshot 2025-01-28 at 5 23 21 PM

### DIFF
--- a/assets/css/_diversion_page.scss
+++ b/assets/css/_diversion_page.scss
@@ -97,6 +97,11 @@
       display: none;
     }
   }
+
+  ol,
+  ul {
+    margin-top: unset;
+  }
 }
 
 .l-diversion-page-panel {
@@ -156,14 +161,15 @@
   font-weight: var(--font-weight-semi);
 }
 
-.c-diversion-panel__h2 {
+.c-diversion-panel__section-header {
   @include font-size($h4-font-size);
-  // font-size: 1.125rem; // h4 font size
-  // font-weight: var(--font-weight-semi);
+  font-weight: var(--font-weight-semi);
+  margin: 0 0 map-get($spacers, 2) 0;
 }
 
 .c-diversion-panel__help-text {
   line-height: 1.5;
+  margin-top: map-get($spacers, 2);
 }
 
 .c-active-detour__alert-icon {
@@ -188,6 +194,7 @@
 
   div {
     margin-bottom: map-get($spacers, 1);
+    line-height: 1.5rem;
   }
 
   // Remove leading margin from `<dd/>`'s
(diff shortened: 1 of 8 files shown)


assets/src/components/detours/detourPanelComponents.tsx
```diff
@@ -71,10 +71,10 @@ export const AffectedRoute = ({
         <p className="my-0 c-diversion-panel__description">
           {routeDescription}
         </p>
-        <p className="my-0 text-muted c-diversion-panel__origin py-1">
+        <p className="my-1 text-muted c-diversion-panel__origin">
           {routeOrigin}
         </p>
-        <p className="my-0 small c-diversion-panel__direction py-1">
+        <p className="my-1 small c-diversion-panel__direction">
```
firestack, line 77: _praise_: oh awesome!

assets/css/_diversion_page.scss
```diff
@@ -97,6 +97,11 @@
       display: none;
     }
   }
+
+  ol,
+  ul {
+    margin-top: unset;
+  }
```
firestack, line 104: note(non-blocking: I think this is fine for now): This is my biggest uncertainty, mainly because we expect to have the reboot handle a lot of this for us.

question: Could you explain why to use `unset` instead of `0`?
```suggestion
  ol,
  ul {
    margin-top: 0;
  }
```

assets/css/_diversion_page.scss
```diff
@@ -188,6 +194,7 @@
 
   div {
     margin-bottom: map-get($spacers, 1);
+    line-height: 1.5rem;
```
firestack, line 197: [Prefer unitless numbers for line-height values](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values)
```suggestion
    line-height: 1.5;
```

and maybe add a note that we expect the bootstrap reboot to handle this in the future?
```suggestion
    // Expected to be controlled by bootstrap css "reboot" in the future
    line-height: 1.5;
```

firestack (approved):

assets/css/_diversion_page.scss
hannahpurcell: 🫤 Emmm I have no justification for this one! I started with unset to get a sense of what stacked margin modifiers were getting applied to the element, and suppose I didn't revisit it? `0` is correct for my uses here. I agree that reboot will change all of this, which is bothersome

PR #2946, opened 2025-01-29
fix: speed up Detour queries by returning subsets of JSON snapshot

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1209252334767151/f

Depends on:
- #2954

### DIFF
--- a/lib/notifications/db/detour.ex
+++ b/lib/notifications/db/detour.ex
@@ -67,6 +67,12 @@ defmodule Notifications.Db.Detour do
       from(d in Notifications.Db.Detour, as: :detour_notification, select_merge: d)
     end
 
+    def with_detour(query \\ base()) do
+      with_named_binding(query, :detour, fn query, binding ->
+        join(query, :left, [detour_notification: n], assoc(n, ^binding), as: ^binding)
+      end)
+    end
+
     @doc """
     Retrieves detour information for notifications from the `Notifications.Db.Detour` table
 
@@ -77,7 +83,7 @@ defmodule Notifications.Db.Detour do
         ...> |> Notifications.Notification.create_activated_detour_notification_from_detour()
         ...>
         iex> all_detour_notifications =
-        ...>   Notifications.Db.Detour.Queries.get_derived_info()
+        ...>   Notifications.Db.Detour.Queries.select_detour_notification_info()
         ...>   |> Skate.Repo.all()
         ...>
         iex> [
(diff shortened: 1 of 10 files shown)


lib/notifications/db/detour.ex
```diff
@@ -92,42 +98,13 @@ defmodule Notifications.Db.Detour do
...
-            )
-        }
-      )
+    def select_detour_notification_info(query \\ base()) do
+      query
+      |> with_detour()
+      |> Skate.Detours.Db.Detour.Queries.select_route_name(:route)
+      |> Skate.Detours.Db.Detour.Queries.select_route_pattern_name(:origin)
+      |> Skate.Detours.Db.Detour.Queries.select_route_pattern_headsign(:headsign)
+      |> Skate.Detours.Db.Detour.Queries.select_direction(:direction)
```
firestack, line 107: I'm not quite happy with this, but I wasn't sure about changing the virtual field names. would be happy to discuss though.

lib/skate/detours/db/detour.ex
```diff
@@ -24,4 +47,115 @@ defmodule Skate.Detours.Db.Detour do
...
+      order_by(query, desc: :updated_at)
+    end
+
+    def with_author(query \\ base(), key \\ :author) do
+      from([detour: d] in query,
+        join: a in assoc(d, :author),
+        as: :author,
+        select_merge: %{^key => a}
+      )
+    end
```
firestack, line 74: note: still not certain about this

lib/skate/detours/db/detour.ex
```diff
@@ -24,4 +47,115 @@ defmodule Skate.Detours.Db.Detour do
...
+
+    def select_estimated_duration(query \\ base(), key \\ :estimated_duration) do
+      select_merge(query, [detour: d], %{
+        ^key => d.state["context"]["selectedDuration"]
+      })
+    end
+
+    def select_state_value(query \\ base(), key \\ :state_value) do
+      select_merge(query, [detour: d], %{^key => %{"value" => d.state["value"]}})
+    end
```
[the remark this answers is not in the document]
hannahpurcell: Since we're still planning on chatting about state value for other reasons, I don't have a strong preference using one or the other, so this is fine to me!

lib/skate/detours/db/detour.ex
```diff
@@ -24,4 +47,115 @@ defmodule Skate.Detours.Db.Detour do
...
+        #   ]
+        #
+        # But, Postgres _is_ able to do this, _if_ we get the types correct.
+        # A JSON value in Postgres is either of type JSON or JSONB, but
+        # - indexing a JSON array requires an `INTEGER`,
+        # - accessing a JSON map, requires Postgres's `TEXT` type.
+        #
+        # So because we know the `directionId` will correspond to the keys in
+        # `directionNames`, casting the `directionId` to `TEXT` allows us to
+        # access the `directionNames` JSON map
```
hannahpurcell, line 135: Wowwww this is so interesting. Ok, thanks for the thorough comment!

hannahpurcell (commented):
Had to learn a little bit about several new things to me (virtual fields and ecto query composition mainly), but I feel like I'm following everything here. It's clean and clear.

The note on `with_author`, I don't see what could be wrong with it. The note on `select_detour_notification_info`, I see what you're saying but think what you did is still a solid approach

lib/skate/detours/db/detour.ex
firestack: I figured out what I'd prefer in https://github.com/mbta/skate/pull/2948

lib/notifications/db/detour.ex
firestack: https://github.com/mbta/skate/pull/2948 is why I'm not quite happy with this, as I'm not sure how to hide these functions from the public API while also allowing renaming for situations like this.

hannahpurcell: yeah, I don't have any better ideas here, unfortunately

hannahpurcell (approved):
I've done several comb-throughs at this point and don't have any recommendations (I'm afraid from my own limitations here, but thankfully Brian is also reviewing.) I know the virtual fields may be changed based on today's talks? But I'm not withholding approval in the meantime

firestack:
@mbta/skate-developers I had to make a few changes when I rebased (namely the introduction of the `:state` column and the Telemetry Tests) so just putting this back up for review due to those changes.

PR #2948, opened 2025-01-29
proposal: change to `list_detours` and other detour_admin endpoint to request fields explicitly

firestack:
Some ideas on how we should structure the "Read" subset of CRUD API's in Skate, but specifically for Detours in this PR.

I think this could be merged into #2946 if people like it.

### DIFF
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -61,6 +61,23 @@ defmodule Skate.Detours.Db.Detour do
       from(Skate.Detours.Db.Detour, as: :detour, select: [])
     end
 
+    def select_virtual_fields(query \\ base(), fields) when is_list(fields) do
+      Enum.reduce(fields, query, fn field, query ->
+        case field do
+          :route_id -> select_route_id(query)
+          :route_name -> select_route_name(query)
+          :route_pattern_id -> select_route_pattern_id(query)
+          :route_pattern_name -> select_route_pattern_name(query)
+          :headsign -> select_route_pattern_headsign(query)
+          :direction -> select_direction(query)
+          :nearest_intersection -> select_starting_intersection(query)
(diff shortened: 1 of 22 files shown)


lib/skate/detours/db/detour.ex
```diff
@@ -75,22 +117,31 @@ defmodule Skate.Detours.Db.Detour do
 
     def select_detour_list_info(query \\ base()) do
       query
-      |> select_merge([
+      |> preload([
+        :author
+      ])
```
firestack, line 122: note: I _feel_ like there should be a better way to tell if we need to preload the author, (and preload using a join instead) but I wasn't able to figure it out in time.

firestack: [`ab0e852` (#2948)](https://github.com/mbta/skate/pull/2948/commits/ab0e852919f6ccc2b8b71699dab884df324e2c75)

firestack:
> more of a general question for discussion, why do we have `Detailed`, `ActivatedDetourDetails`, and `WithState` structs all for `Detour`?
> 
> especially in the case of `ActivatedDetourDetails` it seems excessive to add 2 fields and add an additional complexity of nesting the actual `Detour` info within `:details`

`Detailed` and `WithState` exist because we didn't have a way to have the caller (i.e., the Detours Controller) specify the fields they needed, and there was not a way to query the JSON, so the Detours Context was left to do the mapping and de-structuring of the snapshot. We also use Superstruct on the frontend to ensure that the returned data matches the expected shape so that types work on the frontend, and we encoded the various shapes into structs for type safety.

`ActivatedDetourDetails` exists because Elixir doesn't have struct extensions, so there's not a non-macro way to add these two fields, for just the places that need it.

In hindsight, I'd remove all of these / move them into the controller instead of having them as details of the `Detours` context. I'd be happy to do that, but I think that's out of scope for this work.

PR #2949, opened 2025-01-31
Update detour serde validation check

bfauble:
Asana Ticket: https://app.asana.com/0/1205385723132845/1209213931559675/f

### DIFF
--- a/.github/workflows/deploy-base.yml
+++ b/.github/workflows/deploy-base.yml
@@ -71,7 +71,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: getsentry/action-release@v1.10.2
+      - uses: getsentry/action-release@v1
         env:
           SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
           SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}


lib/skate/detours/snapshot_serde.ex
```diff
@@ -56,19 +56,28 @@ defmodule Skate.Detours.SnapshotSerde do
...
 
   defp validate_serialized_snapshot(
-         serialized_snapshot,
-         %Detour{id: id, state: state} = detour
+         %{"context" => serialized_context} = serialized_snapshot,
+         %Detour{id: id, state: %{"context" => state_context} = state} = detour
        ) do
-    if serialized_snapshot === state do
+    relevant_state_keys = Map.keys(serialized_snapshot)
+    relevant_context_keys = Map.keys(serialized_context)
```
hannahpurcell, line 63: question: Where was the `center` value for originalRoute sneaking into the snapshot? When reviewing this PR I tried to look at some existing detours that are mismatching like https://skate.mbta.com/api/detours/570 or https://skate.mbta.com/api/detours/604, but they must be mismatching for another field because `center` is not present

hannahpurcell: We talked about it in slack and have an example here: https://skate-dev.mbtace.com/detours_admin/126. That makes sense that these fields prompting the mismatch appear at the top level alongside the other main attributes of the snapshot. 

IF we ever had a schema change nested _within_ one of the main context keys, this solution wouldn't cover that. I think I'm ok with that right now, but something to keep in mind? What do you think?

hannahpurcell: Good way of clarifying. No, but I was worried that
```
# serialized one
%{
  context: %{
    a: %{
      b: "b",
    }
  }
}
and
# snapshot
%{
  context: %{
     a: %{
      b: "b",
      c: "c"
    }
  }
}
```
would return as non-matching and would fall back to the snapshot, and `c` would stick around in the same way that `center` stuck around. But actually in talking this out, I realize that won't be true. For example, if for whatever reason, we realize we don't need to pass _everything_ related to a routePattern into the db (like let's say we omit sortOrder or timeDescription), then `routepattern_from_detour` would return the snapshot with updated schema and so we'd be fine. Is this logical? All in all, I think your approach is right

hannahpurcell: I left with a question, but I think I'm comfortable with my understanding with this. Going to approve

hannahpurcell (approved):
✅

hannahpurcell (approved):
Looks good!

PR #2952, opened 2025-02-04
feat(ex/detours): add detour `:status` field, and populate from changeset

firestack:
This adds a new database column and starts populating it from the changeset function.
Once we deploy this and have all instances become aware of the field. We'll then
deploy a data migration which adds this to previously made detours, so that there
are no running instances without this change to the changeset function.

---

Asana Ticket: https://app.asana.com/0/1203014709808707/1209213931559673

### DIFF
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -26,7 +26,20 @@ defmodule Skate.Detours.Db.Detour do
   def changeset(detour, attrs) do
     detour
     |> cast(attrs, [:state, :activated_at])
-    |> validate_required([:state])
+    |> add_status()
+    |> validate_required([:state, :status])
     |> foreign_key_constraint(:author_id)
+  end
+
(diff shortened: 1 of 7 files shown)


lib/skate/detours/db/detour.ex
```diff
@@ -21,7 +26,20 @@ defmodule Skate.Detours.Db.Detour do
...
+    |> add_status()
+    |> validate_required([:state, :status])
     |> foreign_key_constraint(:author_id)
   end
+
+  defp add_status(changeset) do
+    case fetch_change(changeset, :state) do
+      {:ok, state} ->
+        # Once this column is added for all detours, `categorize_detour` logic
+        # should be moved here and should not be needed anymore
```
hannahpurcell, line 38: Great note, roger that 👍

hannahpurcell (approved):
Not a single note, looks great!

PR #2954, opened 2025-02-05
feat: Log db performance (query / query time)

hannahpurcell:
Asana Ticket: https://app.asana.com/0/1205385723132845/1209319191705031


mix.exs
```diff
@@ -97,6 +97,7 @@ defmodule Skate.MixProject do
       {:sobelow, "~> 0.13", only: [:dev, :test], runtime: false},
       {:ssl_verify_fun, "~> 1.1"},
       {:stream_data, "~> 1.1.1", only: :test},
+      {:telemetry_test, "~> 0.1.0"},
```
firestack, line 100: suggestion: should this be `only: :test`?
```suggestion
      {:telemetry_test, "~> 0.1.0", only: :test},
```

firestack, line 100: could we make `:telemetry` an explicit dependency?
Going off of https://github.com/beam-telemetry/telemetry?tab=readme-ov-file#installation


```suggestion
      {:telemetry, "~> 1.0"},
      {:telemetry_test, "~> 0.1.0"},
```

firestack: And, it seems like the `:telemetry` library comes with a test helper, although not as nice as this library.
Do you think the `:telemetry_test` library is worth pulling in vs using `:telemetry_test.attach_event_handlers(self(), [[:skate, :repo, :query]])`
```suggestion
      {:telemetry, "~> 1.2.0"},
```

We could probably write our own setup function to make it nicer while still using the default `:telemetry` package?

```elixir
  setup tags do
    if events = tags[:telemetry_listen] do
      ref = :telemetry_test.attach_event_handlers(self(), events)

      on_exit(fn -> :telemetry.detach(ref) end)

      # maybe not needed, just added for completeness
      %{telemetry_ref: ref}
    end
  end
```

```diff
diff --git a/test/skate/telemetry_test.exs b/test/skate/telemetry_test.exs
index 300e4b1b3c..cff3337942 100644
--- a/test/skate/telemetry_test.exs
+++ b/test/skate/telemetry_test.exs
@@ -1,20 +1,27 @@
 defmodule Skate.TelemetryTest do
   @moduledoc false
   use Skate.DataCase
-  import TelemetryTest
   import Skate.Factory
   import ExUnit.CaptureLog, only: [capture_log: 2]
   import Test.Support.Helpers, only: [set_log_level: 1]
 
   alias Skate.Detours.Detours
 
-  setup [:telemetry_listen]
+  setup tags do
+    if events = tags[:telemetry_listen] do
+      ref = :telemetry_test.attach_event_handlers(self(), events)
+
+      on_exit(fn -> :telemetry.detach(ref) end)
+
+      %{telemetry_ref: ref}
+    end
+  end
 
   defp detour_fixture do
     insert(:detour)
   end
 
-  @tag telemetry_listen: [:skate, :repo, :query]
+  @tag telemetry_listen: [[:skate, :repo, :query]]
   test "logs exception info" do
     set_log_level(:info)
     detour = detour_fixture()
@@ -24,12 +31,9 @@
         assert Skate.Repo.preload(Detours.list_detours(), :author) == [detour]
       end)
 
-    assert_receive {:telemetry_event,
-                    %{
-                      event: [:skate, :repo, :query],
-                      measurements: %{decode_time: _, query_time: _, queue_time: _, total_time: _},
-                      metadata: %{query: _, result: _, source: "detours"}
-                    }}
+    assert_receive {[:skate, :repo, :query], _ref,
+                    %{decode_time: _, query_time: _, queue_time: _, total_time: _},
+                    %{query: _, result: _, source: "detours"}}
 
     assert log =~ "Telemetry for Detours query"
   end

```

lib/skate/telemetry.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  def handle_event([:skate, :repo, :query], measurements, metadata, _config) do
+    if metadata.source == "detours" do
+      {:ok, result} = metadata.result
+
+      Logger.info(fn ->
+        "Telemetry for Detours query, connection_id=#{result.connection_id} num_rows=#{result.num_rows} decode_time=#{measurements.decode_time} query_time=#{measurements.query_time} total_time=#{measurements.total_time} query=#{metadata.query}"
+      end)
+    end
+  end
```
firestack, line 25: I'd like to pattern match away the `if` and `{:ok, result} = metadata.result`
```suggestion
  def handle_event(
        [:skate, :repo, :query],
        measurements,
        %{source: "detours", result: {:ok, result}} = metadata,
        _config
      ) do
    Logger.error(fn ->
      "Telemetry for Detours query, connection_id=#{result.connection_id} num_rows=#{result.num_rows} decode_time=#{measurements.decode_time} query_time=#{measurements.query_time} total_time=#{measurements.total_time} query=#{metadata.query}"
    end)
  end

  def handle_event(_event, _measurements, _metadata, _config), do: nil
```


but what do you think of pattern matching everything?
```suggestion
  def handle_event(
        [:skate, :repo, :query],
        %{decode_time: decode_time, query_time: query_time},
        %{
          source: "detours",
          result: {:ok, %{connection_id: connection_id, num_rows: num_rows}},
          query: query
        },
        _config
      ) do
    Logger.error(fn ->
      "Telemetry for Detours query, connection_id=#{connection_id} num_rows=#{num_rows} decode_time=#{decode_time} query_time=#{query_time} total_time=#{total_time} query=#{query}"
    end)
  end

  def handle_event(_event, _measurements, _metadata, _config), do: nil
```

test/skate/telemetry_test.exs
```diff
@@ -0,0 +1,36 @@
...
+
+    assert_receive {:telemetry_event,
+                    %{
+                      event: [:skate, :repo, :query],
+                      measurements: %{decode_time: _, query_time: _, queue_time: _, total_time: _},
+                      metadata: %{query: _, result: _, source: "detours"}
+                    }}
+
+    assert log =~ "Telemetry for Detours query"
+  end
```
firestack, line 35: related to https://github.com/mbta/skate/pull/2954/files#r1945313804
```suggestion
  import Skate.Factory
  import ExUnit.CaptureLog, only: [capture_log: 2]
  import Test.Support.Helpers, only: [set_log_level: 1]

  alias Skate.Detours.Detours

  setup tags do
    if events = tags[:telemetry_listen] do
      ref = :telemetry_test.attach_event_handlers(self(), events)

      on_exit(fn -> :telemetry.detach(ref) end)

      %{telemetry_ref: ref}
    end
  end

  defp detour_fixture do
    insert(:detour)
  end

  @tag telemetry_listen: [[:skate, :repo, :query]]
  test "logs exception info" do
    set_log_level(:info)
    detour = detour_fixture()

    log =
      capture_log([level: :info], fn ->
        assert Skate.Repo.preload(Detours.list_detours(), :author) == [detour]
      end)

    assert_receive {[:skate, :repo, :query], _ref,
                    %{decode_time: _, query_time: _, queue_time: _, total_time: _},
                    %{query: _, result: _, source: "detours"}}

    assert log =~ "Telemetry for Detours query"
  end
```

lib/skate/telemetry.ex
```diff
@@ -0,0 +1,26 @@
...
+        %{}
+      )
+  end
+
+  def handle_event([:skate, :repo, :query], measurements, metadata, _config) do
+    if metadata.source == "detours" do
+      {:ok, result} = metadata.result
+
+      Logger.info(fn ->
+        "Telemetry for Detours query, connection_id=#{result.connection_id} num_rows=#{result.num_rows} decode_time=#{measurements.decode_time} query_time=#{measurements.query_time} total_time=#{measurements.total_time} query=#{metadata.query}"
```
firestack, line 22: I think we should quote `query` so that Splunk can grab all the contents of the `query` value using it's `key="value"` extractor, I'm not actually _certain_ that it works with single quotes though[^1]? 

```suggestion
        "Telemetry for Detours query, connection_id=#{result.connection_id} num_rows=#{result.num_rows} decode_time=#{measurements.decode_time} query_time=#{measurements.query_time} total_time=#{measurements.total_time} query='#{metadata.query}'"
```

[^1]: we'd need to use single quotes though because the query has double quotes in it.

lib/skate/telemetry.ex
firestack, line 1: I like this! And I like that this captures every query for detours for our DB!

But I'm left wondering if using `:telemetry` is the right option?
I'm really stuck here and not sure how to make a decision here :/ 

I think for informing the detours database changes this makes sense, but I'm not sure about long term.

firestack (changes requested):
Left my thoughts, sorry it took so long!

mix.exs
hannahpurcell: Oh yeah, big oversight that `:telemetry` wasn't imported directly, but I like the mods to the test even better. https://github.com/mbta/skate/pull/2954/commits/3942db2b3ba9756953196be9b84736d83b7e2d7e

lib/skate/telemetry.ex
hannahpurcell: What are the cons to using `:telemetry` long-term? I don't know the downsides here. Our other currently available option is the duration util, or something else new

lib/skate/telemetry.ex
hannahpurcell: Ahhhh definitely! 🚀

lib/skate/telemetry.ex
hannahpurcell: Yeah, super! Works for me. Accepted with a slight tweak of the `measurements` field to re-include the total_time

lib/skate/telemetry.ex
hannahpurcell: (Don't worry about writing it out, we can discuss tomorrow morning!)

mix.lock
```diff
@@ -77,8 +77,9 @@
...
-  "telemetry": {:hex, :telemetry, "1.2.1", "68fdfe8d8f05a8428483a97d7aab2f268aaff24b49e0f599faa091f1d4e7f61c", [:rebar3], [], "hexpm", "dad9ce9d8effc621708f99eac538ef1cbe05d6a874dd741de2e689c47feafed5"},
+  "telemetry": {:hex, :telemetry, "1.3.0", "fedebbae410d715cf8e7062c96a1ef32ec22e764197f70cda73d82778d61e7a2", [:rebar3], [], "hexpm", "7015fc8919dbe63764f4b4b87a95b7c0996bd539e0d499be6ec9d7f3875b79e6"},
   "telemetry_registry": {:hex, :telemetry_registry, "0.3.1", "14a3319a7d9027bdbff7ebcacf1a438f5f5c903057b93aee484cca26f05bdcba", [:mix, :rebar3], [{:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "6d0ca77b691cf854ed074b459a93b87f4c7f5512f8f7743c635ca83da81f939e"},
+  "telemetry_test": {:hex, :telemetry_test, "0.1.2", "122d927567c563cf57773105fa8104ae4299718ec2cbdddcf6776562c7488072", [:mix], [{:telemetry, "~> 1.2", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "7bd41a49ecfd33ecd82d2c7edae19a5736f0d2150206d0ee290dcf3885d0e14d"},
```
firestack, line 82: Can you run `mix deps.unlock telemetry_test` and remove this?
(it may just be safe to delete this line.)

lib/skate/telemetry.ex
```diff
@@ -0,0 +1,33 @@
...
+  require Logger
+
+  def setup_telemetry do
+    _ =
+      :telemetry.attach(
+        "skate",
+        [:skate, :repo, :query],
+        &Skate.Telemetry.handle_event/4,
+        %{}
+      )
```
firestack, line 14: nit(non-blocking): not doing anything with this assignment/match
```suggestion
    :telemetry.attach(
      "skate",
      [:skate, :repo, :query],
      &Skate.Telemetry.handle_event/4,
      %{}
    )
```

lib/skate/telemetry.ex
firestack: > What are the cons to using `:telemetry` long-term? I don't know the downsides here. Our other currently available option is the duration util, or something else new

I don't know either, it's my unfamiliarity with it that makes me question it, maybe we can tag @bfauble here?

hannahpurcell: I don't know if I mentioned this yet, but I used the other apps in our /mbta repo for reference when doing this, so I know we have a solid precedent for using `:telemetry` here, if that assuages any concerns

mix.lock
hannahpurcell: Oh shoot, thanks for catching this

lib/skate/telemetry.ex
```diff
@@ -0,0 +1,32 @@
...
+        %{decode_time: decode_time, query_time: query_time, total_time: total_time},
+        %{
+          source: "detours",
+          result: {:ok, %{connection_id: connection_id, num_rows: num_rows}},
+          query: query
+        },
+        _config
+      ) do
+    Logger.info(fn ->
+      "Telemetry for Detours query, connection_id=#{connection_id} num_rows=#{num_rows} decode_time=#{decode_time} query_time=#{query_time} total_time=#{total_time} query='#{query}'"
```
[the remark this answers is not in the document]
firestack: At that point I think we'd _maybe_ be better off setting `:log` for Ecto, and I'm not sure we want to log _everything_? That could be a lot of logs and I'm not sure what our spunk upper bound is.

If we did this, I'd also like source to be something we can easily use in a Splunk query so `source=#{source}`, which may be worth it?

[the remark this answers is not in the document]
firestack: We do have `mfa` metadata logged, so it is possible to drill down through logs by module and function, so `mfa=Skate.Telemetry` or `mfa=Skate.Telemetry.handle_event`?

https://github.com/mbta/skate/blob/main/config/config.exs#L144

Maybe `handle_event` should be `repo_telemetry` or something of the sort?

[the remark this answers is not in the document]
hannahpurcell: + yep was just going to note the same thing. Current output example:
```
[info] mfa=Skate.Telemetry.handle_event/4 Telemetry for Detours query, connection_id=90613 num_rows=33 decode_time=9108 query_time=351594349 total_time=352624333 query='SELECT d0."id", d0."state", d0."author_id", d0."activated_at", d0."inserted_at", d0."updated_at" FROM "detours" AS d0 ORDER BY d0."updated_at" DESC'
```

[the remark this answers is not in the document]
hannahpurcell: Ok! And seems like `telemetry` as a whole doesn't raise any concerns. I'll make a less ambiguous function name, add source=detours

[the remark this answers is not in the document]
hannahpurcell: [tweak: make log more descriptive](https://github.com/mbta/skate/pull/2954/commits/aef226855f97b4528dc7f9dc295f43aec501aff8)

firestack (approved):

PR #2955, opened 2025-02-05
Draft: interface separation discussion

bfauble:


assets/src/models/detoursList.ts
```diff
@@ -60,10 +60,24 @@ export interface GroupedSimpleDetours {
...
 
+const MyNormalizedType = type({
+  my_id: detourId,
+  my_route: string(),
+  my_via_variant: string(),
+  my_direction: string(),
+  my_name: string(),
+  my_intersection: string(),
+  my_updated_at: number(),
+  my_activated_at: coerce(date(), string(), (dateStr) => new Date(dateStr)),
```
hannahpurcell, line 71: So if we're dealing with a draft detour (never activated), this and estimated_duration would be null? (I could see past detours retaining their activated_at and estimated_duration because there's no reason to delete them, as long as status is also included somewhere in this normalized type)

hannahpurcell: Would this model also include the state blob, since some endpoints also return that to the frontend?

PR #2961, opened 2025-02-11
tweak: make use of status field in detour list queries

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209213931559671

### DIFF
--- a/lib/skate_web/controllers/detours_controller.ex
+++ b/lib/skate_web/controllers/detours_controller.ex
@@ -32,11 +32,13 @@ defmodule SkateWeb.DetoursController do
   def detours(conn, _params) do
     %{id: user_id} = AuthManager.Plug.current_resource(conn)
 
-    detours = %{
-      active: Detours.detours_for_user(user_id, :active),
-      draft: Detours.detours_for_user(user_id, :draft),
-      past: Detours.detours_for_user(user_id, :past)
-    }
+    detours =
+      [:active, :draft, :past]
+      |> Enum.map(fn status ->
+        Task.async(fn -> {status, Detours.detours_for_user(user_id, status)} end)
+      end)
+      |> Task.await_many()
+      |> Map.new()
 
     json(conn, %{data: detours})
   end


lib/skate/detours/detours.ex
```diff
@@ -65,34 +67,33 @@ defmodule Skate.Detours.Detours do
...
-      draft:
-        detours
-        |> Map.get(:draft, [])
-        |> Enum.filter(fn detour -> detour.author_id == user_id end),
-      past: Map.get(detours, :past, [])
+      active: detours_for_user(user_id, :active),
+      draft: detours_for_user(user_id, :draft),
+      past: detours_for_user(user_id, :past)
     }
   end
```
firestack, line 75: suggestion: since this is only used in the detours controller, how do you feel about pushing this implementation into the detours controller so we're not having to concern ourselves with keeping this interface available at this level?

lib/skate/detours/db/detour.ex
```diff
@@ -162,6 +162,7 @@ defmodule Skate.Detours.Db.Detour do
         :author_id,
         :activated_at,
         :updated_at,
+        :status,
```
firestack, line 165: suggestion: can we remove `:state_value` with the addition of this?

lib/skate/detours/detours.ex
```diff
@@ -65,34 +67,33 @@ defmodule Skate.Detours.Detours do
...
+    |> Enum.map(&db_detour_to_detour/1)
+    |> Enum.reject(&is_nil/1)
+  end
+
+  defp apply_user_and_status_filter(query, user_id, :draft) do
+    where(query, [detour: d], d.status == :draft and d.author_id == ^user_id)
+  end
+
+  defp apply_user_and_status_filter(query, _user_id, status) do
+    where(query, [detour: d], d.status == ^status)
```
firestack, line 90: thought(non-blocking): what do you think of putting these in `Skate.Detours.Db.Detour.Queries`? I'm not that well versed in Elixir, so I'll defer to you here, but my thinking is that query building functions would belong in the `Queries` module.

firestack (approved):

lib/skate/detours/db/detour.ex
firestack: I'd like to see if we can do this but otherwise approved!

firestack (changes requested):

lib/skate/detours/db/detour.ex
firestack: It was used to make the `categorize_detour` call cheaper https://github.com/mbta/skate/blob/bf-query-with-status-field/lib/skate/detours/detours.ex#L123-L125

lib/skate/detours/detours.ex
```diff
@@ -40,59 +40,35 @@ defmodule Skate.Detours.Detours do
...
-  @doc """
-  Returns the detours grouped by active, draft, and past.
-
-  ## Examples
+  def detours_for_user(user_id, status) do
+    Skate.Detours.Db.Detour.Queries.select_detour_list_info()
+    |> apply_user_and_status_filter(user_id, status)
+    |> Repo.all()
+    |> Enum.map(&db_detour_to_detour/1)
+    |> Enum.reject(&is_nil/1)
```
firestack, line 57: _praise_: love it! 🎉

lib/skate/detours/db/detour.ex
firestack: Let's chat about this? https://github.com/mbta/skate/pull/2965

firestack: Good for now in this PR though

firestack (approved):

firestack:
A thought, now that listing detours is 3 separate sequential DB queries, we'll probably want to get https://github.com/mbta/skate/pull/2962 in so that we can parallelize the queries

firestack:
> > A thought, now that listing detours is 3 separate sequential DB queries, we'll probably want to get #2962 in so that we can parallelize the queries
> 
> made a small update to make the 3 calls async which should help a little bit but the queries overall are already quite a bit faster even w/o.

Love it!

firestack (approved):
ceremonial re-approval for new changes]

PR #2962, opened 2025-02-12
feat: Populate detours list with channels

hannahpurcell:
Asana ticket: https://app.asana.com/0/1205385723132845/1208390246838825

In terms of tests, I'm not sure what else to cover? Ideally, I'd like to verify some integration flow -- like activating a detour triggers the channel to be affected -- but I struggled with that.

Known issue: preventing a drafted detour from appearing in the channel and passing through the `handleDrafted` function before it has all it's returned values (like nearestIntersection). To be discussed in subsequent PR.

Also: have confirmed with Brian that the spinner is great!

### DIFF
--- a/assets/src/components/detourListPage.tsx
+++ b/assets/src/components/detourListPage.tsx
@@ -1,7 +1,7 @@
-import React, { useCallback, useState } from "react"
+import React, { useContext, useState } from "react"
 import { DetoursTable, DetourStatus } from "./detoursTable"
 import userInTestGroup, { TestGroups } from "../userInTestGroup"
-import { Button } from "react-bootstrap"
+import { Button, Spinner } from "react-bootstrap"
 import {
   GlobeAmericas,
   LockFill,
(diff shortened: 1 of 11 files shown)


assets/src/hooks/useDetours.ts
```diff
@@ -94,15 +94,15 @@ export const useActiveDetours = (
   enable: boolean = true
 ) => {
   const topic = "detours:active"
-  const [activeDetours, setActiveDetours] = useState<DetoursMap>({})
+  const [activeDetours, setActiveDetours] = useState<DetoursMap | null>(null)
```
firestack, line 97: suggestion/question(non-blocking): what do you think about using `undefined` instead of `null`? 

To me, `undefined` means it hasn't been initializd yet, which correlates with a "loading" state, where `null` _to me_ would mean that the channel connected and got `null` as a response. 

I'm also _curious_ about something like this but not sure about it...
```suggestion
  const [activeDetours, setActiveDetours] = useState<DetoursMap | "loading">("loading")
```
I wish we had better types for "loading" states :/

assets/src/components/detourListPage.tsx
```diff
@@ -10,19 +10,39 @@ import {
...
+    ? Object.values(activeDetoursMap).sort(
+        (a, b) => b.activatedAt - a.activatedAt
+      )
+    : null
+  const draftDetours: SimpleDetour[] | null = draftDetoursMap
+    ? Object.values(draftDetoursMap).sort((a, b) => b.updatedAt - a.updatedAt)
+    : null
+  const pastDetours: SimpleDetour[] | null = pastDetoursMap
+    ? Object.values(pastDetoursMap).sort((a, b) => b.updatedAt - a.updatedAt)
+    : null
```
firestack, line 44: nit(non-blocking): could we simplify these?

```suggestion
  const activeDetours = activeDetoursMap
    && Object.values(activeDetoursMap).sort(
        (a, b) => b.activatedAt - a.activatedAt
      )
  const draftDetours = draftDetoursMap
    && Object.values(draftDetoursMap).sort((a, b) => b.updatedAt - a.updatedAt)
  const pastDetours = pastDetoursMap
    && Object.values(pastDetoursMap).sort((a, b) => b.updatedAt - a.updatedAt)
```

firestack (changes requested):
Looks good! A few comments but nothing that big, ready to approve after dialoging on this.

assets/src/hooks/useDetours.ts
hannahpurcell: Hmmm I think I'm going to opt for `undefined` (good catch, thanks). Using "loading" led TS to struggle with the `{ ...activeDetours` logic because it couldn't ascertain whether it was working DetoursMap or a string (even if I incorporated conditionals to assert that `activeDetours != "loading"`)

assets/src/components/detourListPage.tsx
hannahpurcell: Oh of course, ha! Thanks 👍

firestack (approved):
waiting to resolve the issue with `nearestIntersection` but otherwise lgtm 👍🏻

PR #2964, opened 2025-02-14
tweak: The detours dropdown should totally be behind DetoursOnLadder flag

hannahpurcell:
So that we can toggle on the flag for all users when ready

### DIFF
--- a/assets/src/components/routeLadder.tsx
+++ b/assets/src/components/routeLadder.tsx
@@ -83,7 +83,7 @@ export const Header = ({
             hasAlert && "c-route-ladder__dropdown--non-skate-alert",
           ])}
         >
-          {isAdmin && (
+          {(isAdmin || inTestGroup(TestGroups.DetoursOnLadder)) && (
             <Dropdown className="border-box inherit-box">
               <Dropdown.Toggle
                 className="c-route-ladder__dropdown-button d-flex"
@@ -116,7 +116,7 @@ export const Header = ({
                 )}
                 {hasAlert && (
                   <>
-                    <Dropdown.Divider className="border-top-0" />
+                    {isAdmin && <Dropdown.Divider className="border-top-0" />}
                     <Dropdown.Header>
                       <div className="c-route-ladder__dropdown-header-text">
                         Active detours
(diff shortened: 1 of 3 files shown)


assets/src/components/routeLadder.tsx
firestack, line 1: issue: seems like the "Active Detours" header is missing for the "no active detours"

Screenshot from PR: 
image

Screenshot from Figma:
image

firestack: The padding is also really high, I'm wondering if we can remove the `pb-4` from the `ItemText`'s this or a followup PR?

hannahpurcell: Oooohhh yep, thanks for the catch! And maaaaaaybe the padding change should be a follow-up, but I'm kinda partial to doing it here

firestack: I'm good either way 👍🏻

hannahpurcell: re: padding, looks like I chose pb-4 because the goal padding was 32px [in the design](https://www.figma.com/design/XGmmwVROXdOvnWtJeUFC50/Unplanned-Detours-%7C-Handoff?node-id=6383-8181&t=ewppeFUs9yR943hJ-4), but I think there are some things wonky with that number (the line-height of that text is off.) But pb-3 looks good, so I think I'm sticking with that. After fix:

Screenshot 2025-02-18 at 3 43 24 PM
Screenshot 2025-02-18 at 3 44 19 PM

hannahpurcell: re: the [missing active detours text fix](https://github.com/mbta/skate/pull/2964/commits/9c8976cc72bf296c22e4faece252411518a00db4), here's how it looks in all cases

Admin and Detours-on-ladder
admin and detours-on-ladder

Admin and no Detours-on-ladder
admin and no detours-on-ladder

Inspector and Detours-on-ladder
inspector and detours-on-ladder

And the inspector and no detours-on-ladder is just no dropdown at all

firestack (approved):

PR #2965, opened 2025-02-18
cleanup(ex/db/detour): remove `state_value`

firestack:
PR for `:state_value` removal thread https://github.com/mbta/skate/pull/2961#discussion_r1952756606


firestack:
> if possible can we try to remove any calls to `categorize_detour` that aren't happening before an update/create in the DB?

Sure yeah! I'd prefer to have that in it's own PR, as this should only be concerned with the `:state_value` IMO.

See https://github.com/mbta/skate/pull/2967

PR #2966, opened 2025-02-19
fix: update serde module to handle activatedAt no longer being stored in the state prop

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209254939270955

---
  - https://app.asana.com/0/0/1209285582564817

### DIFF
--- a/test/skate/detours/snapshot_serde_test.exs
+++ b/test/skate/detours/snapshot_serde_test.exs
@@ -5,7 +5,7 @@ defmodule Skate.Detours.SnapshotSerdeTest do
   alias Skate.Detours.SnapshotSerde
 
   describe "compare_snapshots" do
-    test "true: compares activate_at prop with activatedAt in serialized detour snapshot" do
+    test "true: compares activated_at prop with activatedAt in serialized detour snapshot" do
       activated_at = Skate.DetourFactory.browser_date()
 
       %{id: id, state: snapshot} =


lib/skate/detours/snapshot_serde.ex
```diff
@@ -64,12 +64,36 @@ defmodule Skate.Detours.SnapshotSerde do
...
+      reconsile_activated_at(scoped_state, serialized_snapshot)
+
+    diff = MapDiff.diff(scoped_state, serialized_snapshot)
+
+    matches = Map.get(diff, :changed) == :equal
 
     {matches, diff, scoped_state}
   end
 
+  defp reconsile_activated_at(
```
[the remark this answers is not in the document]
hannahpurcell: As it is written, if the serialized state has an `activatedAt`, but the state blob doesn't, the state blob will have the `activatedAt` added back in. Then in `validate_serialized_snapshot`, if the other fields match, the serialized version is returned and the modified state blob is ditched. But if the other fields don't match, the state blob becomes the main source of truth, and then `fix_snapshot_activated_at` function is run and the mismatch error is logged.

Two thoughts from this: I don't think the `compare_snapshots` function should transform either of objects it is comparing. I'm trying to think through alternative approaches but haven't decided on one. 

Secondly, I think it may be odd to have 2 "fixing functions" related to `activatedAt`. Could there be just one?

Edit: oh I just had another thought that is even more foundational than the earlier thoughts: if `activatedAt` has been removed from the state blob, then it shouldn't be returned in the serialized function anyway, right? Because the serializer is just supposed to build the state blob. Otherwise, this function is re-inserting the field to the state blob when it shouldn't be there?

hannahpurcell (commented):
Early thoughts

lib/skate/detours/snapshot_serde.ex
```diff
@@ -82,17 +87,16 @@ defmodule Skate.Detours.SnapshotSerde do
 
   defp validate_serialized_snapshot(%Detour{id: id} = detour) do
     serialized_snapshot = serialize_snapshot(detour)
-    {matches, diff, scoped_state} = compare_snapshots(detour, serialized_snapshot)
+    {matches, diff} = compare_snapshots(detour, serialized_snapshot)
 
     if matches do
```
hannahpurcell, line 92: This if/else can be polished into just an `if` that handles logging now that both cases return `serialized_snapshot`

lib/skate/detours/snapshot_serde.ex
```diff
@@ -82,17 +87,16 @@ defmodule Skate.Detours.SnapshotSerde do
...
       serialized_snapshot
     else
-      scoped_state = fix_snapshot_activated_at(scoped_state, detour)
       Sentry.capture_message(mismatch_message(id), extra: %{diff: diff})
 
       Logger.error("#{mismatch_message(id)} #{diff_details(diff)}")
 
-      scoped_state
```
hannahpurcell, line 95: Just confirming: if the serialized-from-db-data detour does not equal the original snapshot we have for that detour, we're falling back to the serialized version? That's a divergence from the original intention of the validation logic. It's not necessarily a wrong change, I'm just trying to figure out which one is the right approach, and I don't know

lib/skate/detours/snapshot_serde.ex
```diff
@@ -64,10 +60,19 @@ defmodule Skate.Detours.SnapshotSerde do
       |> Map.take(relevant_state_keys)
       |> Map.put("context", scoped_state_context)
 
-    matches = serialized_snapshot === scoped_state
-    diff = MapDiff.diff(state, serialized_snapshot)
+    diff = diff_snapshots(scoped_state, serialized_snapshot)
```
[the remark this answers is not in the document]
hannahpurcell: Ahaaaaaa ok thanks

lib/skate/detours/snapshot_serde.ex
hannahpurcell: ok got it!! Perfect explanation, thank you

PR #2968, opened 2025-02-24
refactor(ex/skate/detours): use changeset to check change to `status` in `send_notification`

firestack:
Currently, `Skate.Detours.SnapshotSerde.deserialize` "clobbers" anything in the database because the created changeset isn't doing any real change tracking because the detour object isn't pulled from the database first.

This starts tracking changes appropriately, and then uses the created changeset to check if `:status` is changed and uses _that_ change to trigger sending & creation of notifications.


firestack:
Merged into #2967

PR #2969, opened 2025-02-26
fix: correct `ArrowDownLeftSquare` icon

firestack:
Turns out, the data for the `ArrowDownLeftSquare` was actually the `ArrowUpRightSquare` data.

image

---

Asana Ticket: https://app.asana.com/0/1148853526253437/1209227790219716

### DIFF
--- a/assets/src/helpers/bsIcons.tsx
+++ b/assets/src/helpers/bsIcons.tsx
@@ -50,14 +50,14 @@ export const ArrowDownLeftSquare = (props: SvgProps) => (
     width="16"
     height="16"
     fill="currentColor"
-    className="bi bi-arrow-up-right-square"
+    className="bi bi-arrow-down-left-square"
     viewBox="0 0 16 16"
     aria-hidden
     {...props}
   >
     <path
       fillRule="evenodd"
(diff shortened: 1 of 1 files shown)


assets/src/helpers/bsIcons.tsx
```diff
@@ -57,7 +57,7 @@ export const ArrowDownLeftSquare = (props: SvgProps) => (
   >
```
meagharty, line 57: nit: should we update the `className` on the svg to be `bi bi-arrow-down-left-square` for correctness?

meagharty (approved):
Looks good as is on dev-blue with the updated path!

A nit comment on the `className` that I don't think requires re-review

assets/src/helpers/bsIcons.tsx
firestack: Thanks for catching this!! I slipped up and should've re-done the [instructions](https://github.com/mbta/skate/blob/main/assets/src/helpers/bsIcons.tsx#L3-L18) 

Fixed in https://github.com/mbta/skate/compare/deaaad2e81aa16ad93898a540ff366a4ae8816f2..c18322e37739dc1f6cfbb4f3c3a688152516acaa

PR #2971, opened 2025-02-27
fix: snapshot race condition bug

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209459414962855

### DIFF
--- a/assets/tests/components/detours/diversionPage.saveDetour.test.tsx
+++ b/assets/tests/components/detours/diversionPage.saveDetour.test.tsx
@@ -2,7 +2,11 @@ import { jest, describe, test, expect, beforeEach } from "@jest/globals"
 import { act, fireEvent, render, screen, waitFor } from "@testing-library/react"
 import React from "react"
 import "@testing-library/jest-dom/jest-globals"
-import { fetchNearestIntersection, fetchRoutePatterns, putDetourUpdate } from "../../../src/api"
+import {
+  fetchNearestIntersection,
+  fetchRoutePatterns,
+  putDetourUpdate,
+} from "../../../src/api"
 import {
   DiversionPage as DiversionPageDefault,
   DiversionPageProps,
@@ -32,7 +36,9 @@ jest.mock("../../../src/userTestGroups")
 
 beforeEach(() => {
   jest.mocked(fetchRoutePatterns).mockReturnValue(neverPromise())
-  jest.mocked(fetchNearestIntersection).mockReturnValue(Promise.resolve("Returned Intersection"))
+  jest
+    .mocked(fetchNearestIntersection)
+    .mockReturnValue(Promise.resolve("Returned Intersection"))
   jest.mocked(getTestGroups).mockReturnValue([])
   jest.mocked(putDetourUpdate).mockReturnValue(neverPromise())
 })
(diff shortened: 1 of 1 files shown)


assets/src/hooks/useDetour.ts
```diff
@@ -26,29 +26,10 @@ export const useDetour = (input: UseDetourInput) => {
...
-          actorRef
-            .getSnapshot()
-            .can({ type: "detour.save.set-uuid", uuid: uuid.ok })
-        ) {
-          actorRef.send({ type: "detour.save.set-uuid", uuid: uuid.ok })
-        }
-      })
     })
 
     return () => {
```
hannahpurcell, line 35: Question for everyone: I think the `localStorage` save is not used anymore. That was our first stage of setting up a sort of autosave feature, so if Skate freezes up and needs a refresh, the detour in progress is retained, even before we had the db. Now I don't think it's necessary, right? In which case this whole hook could go?

firestack: Yeah I'd remove the entire hook.

assets/src/models/createDetourMachine.ts
```diff
@@ -231,11 +256,22 @@ export const createDetourMachine = setup({
...
+      invoke: {
+        id: "detourApi",
+        src: "detourApiActor",
+      },
+      always: {
+        actions: ["detour.auto-save", assign({ saved: true })],
+        guard: ({ context }) => {
+          return !context.saved && !!context.nearestIntersection
+        },
+      },
```
hannahpurcell, line 274: So, when the first point is selected (aka the startpoint), will the autosave run or fail to run because `nearestIntersection` doesn't finish by the time the autosave logic executes? Let me walk through the steps to see:

It looks like the actions in the `always` happen after the other actions, which is good. So if a state machine is at the initial `Editing` state, they'll start in `Pick Start Point`. If the user engages with the `DetourMap` and clicks a start point, the `detour.edit.place-waypoint-on-route` event is sent to the state machine, which runs the `detour.add-start-point` action, which assigns `{saved: false}`. So, then the `always` logic runs, but the guard prevents any saving from happening because `nearestIntersection` is not populated.

Then the state machine steps forward to the target of that event and enters the "Place Waypoint" state. Immediately `fetch-nearest-intersection` is invoked. I'm not totally sure about the next step. Does the `onDone` literally wait until the fetch finishes? And then once the `onDone` is complete, the `always` portion is re-executed? If the `onDone` did have a chance to finish, then by the time the `always` re-executes, saved = false and nearestIntersection is set, so that autosave should work. Otherwise, it won't.

I had to write that all out to understand this logic, but then in testing: the detour still saves on the first waypoint placed. So it seems like the `onDone` doesn't execute until the nearestIntersection is populated, which is GREAT. Love it.

assets/src/models/createDetourMachine.ts
```diff
@@ -12,6 +19,11 @@ import {
 import { DetourShape, FinishedDetour } from "./detour"
 import { fullStoryEvent } from "../helpers/fullStory"
 import { type, optional, coerce, date, string } from "superstruct"
+import { detourApiActor } from "../state/detourApiActor"
+
+const tracked = () => ({
+  saved: false,
+})
```
hannahpurcell, line 26: I think I'd like a different word here than `tracked`? Maybe `autosave`? Open for debate

firestack:
issues: 
- I can't get this to work for old/existing detours, I think because the `detourApiActor` doesn't exist in old snapshots; when it's restored from a snapshot, it isn't recreated in the machine.
- This doesn't work with our _existing_ backend save logic, when I create a new detour, it is always a draft, and cannot be reopened. I think this is due to just sending `context` to the `detourApiActor` and not the entire serialized snapshot.

assets/src/models/createDetourMachine.ts
```diff
@@ -40,6 +52,7 @@ export const createDetourMachine = setup({
       activatedAt?: Date
 
       editedSelectedDuration?: string
+      saved: boolean
```
firestack, line 55: issue: making this required without adding this to the `serde` module I think would cause issues?

```suggestion
      saved?: boolean
```

PR #2972, opened 2025-02-27
docs: Upload script from winter 25 hastus hotfix

hannahpurcell:
Is this the place for it?


documentation/scripts/hastus_export_hotfix_winter_25.ex
```diff
@@ -0,0 +1,63 @@
+#/usr/bin/env elixir
+
+# run with mix run hastus_export_bugfix_winter_25.ex --no-mix-exs
+
+# TL;DR: This script was needed to fix an issue with the HASTUS export for Winter 2025.
+# The Transit Data export was run before the Skate export, and the trip_ids for `td_lynn_ns_trips`
+# were edited in-between. So this script used the all_trips.txt file generated by TD
+# to update the trip ids of Skate's trips.csv and output a new_trips.csv file.
+
+# Extra info: https://www.notion.so/mbta-downtown-crossing/Proposal-to-reorganize-quarterly-HASTUS-exports-for-Transit-Tech-teams-184f5d8d11ea80acb242fb8a524c5093?pvs=4#185f5d8d11ea8072a4ebd9547c2d40af
```
meagharty, line 10: This makes sense to me to put here! The discussion in Slack didn't bring up any strong opinions re:cruft or not, so I'm approving ✅ 

Even if Skate isn't regularly performing the HASTUS export and upload with the new process outlined in Notion here, it's still possible we'd need to do a hotfix on occasion.

meagharty (approved):

documentation/scripts/hastus_export_hotfix_winter_25.ex
hannahpurcell: Ok great, thanks for the thoughtful response, and I agree 👍

PR #2973, opened 2025-02-28
fix: require nearest intersection before saving detour/snapshot

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209459414962855


hannahpurcell (approved):
Oh yeah, simple and sweet! 👍 thanks for putting up the quick fix even while thinking through the longer fix

firestack (approved):

PR #2975, opened 2025-03-11
feat(ts/createDetourMachine): default `fetchNearestIntersection` result to an Em Dash on error

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1209629085185045

### DIFF
--- a/assets/tests/components/detours/diversionPage.saveDetour.test.tsx
+++ b/assets/tests/components/detours/diversionPage.saveDetour.test.tsx
@@ -38,9 +38,7 @@ jest.mock("../../../src/userTestGroups")
 beforeEach(() => {
   jest.mocked(fetchRoutePatterns).mockReturnValue(neverPromise())
   jest.mocked(getTestGroups).mockReturnValue([])
-  jest
-    .mocked(fetchNearestIntersection)
-    .mockResolvedValue("Returned Intersection")
+  jest.mocked(fetchNearestIntersection).mockReturnValue(neverPromise())
   jest.mocked(putDetourUpdate).mockReturnValue(neverPromise())
 })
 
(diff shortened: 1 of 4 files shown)


assets/tests/components/detours/diversionPage.saveDetour.test.tsx
```diff
@@ -83,7 +83,7 @@ describe("DiversionPage autosave flow", () => {
     })
 
     await waitFor(() => {
-      expect(putDetourUpdate).toHaveBeenCalledTimes(3)
+      expect(putDetourUpdate).toHaveBeenCalledTimes(5)
```
firestack, line 86: not quite sure why this changed, I tried to figure out if there was a way to do "at least 2 times" but didn't see anything.

assets/tests/components/detours/diversionPage.saveDetour.test.tsx
```diff
@@ -40,7 +40,7 @@ beforeEach(() => {
   jest.mocked(getTestGroups).mockReturnValue([])
   jest
     .mocked(fetchNearestIntersection)
-    .mockResolvedValue("Returned Intersection")
+    .mockReturnValue(neverPromise())
```
firestack, line 43: The intersection isn't used in these tests and it's just easier to have the promise never return, because otherwise you run into "An update inside a test was not wrapped in act(...)."

assets/tests/components/detours/diversionPage.saveDetour.test.tsx
firestack: okay actually I think this changed due to doing `                    "set.nearest-intersection-fallback"` in `Pick Start Point`

meagharty (approved):
Make sense to me!

I'm still getting up to speed on xstate but your comment on the test update was helpful, thank you!

PR #2976, opened 2025-03-12
feat: add API key configuration for Swiftly Service Adjustments API

firestack:
Asana Ticket: https://app.asana.com/0/0/1209654531318891

I just lightly setup variables to use for an imaginary module. Happy to change any names. Getting this setup so we can open a Devops PR.

### DIFF
--- a/.env.1p.skate
+++ b/.env.1p.skate
@@ -11,6 +11,10 @@ SKATE_HASTUS_URL="op://$_1P_VAULT/Skate Data URLs/SKATE_HASTUS_URL"
 SWIFTLY_REALTIME_VEHICLES_URL="op://$_1P_VAULT/Swiftly API/hostname"
 SWIFTLY_AUTHORIZATION_KEY="op://$_1P_VAULT/Swiftly API/credential"
 
+SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/hostname"
+SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/credential"
+SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/agency"
+
 ## Map Tileset URL's
 BASE_TILESET_URL="op://$_1P_VAULT/Skate Map Tileset URLs/BASE_TILESET_URL"
 SATELLITE_TILESET_URL="op://$_1P_VAULT/Skate Map Tileset URLs/SATELLITE_TILESET_URL"
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -36,6 +36,11 @@ config :skate, Skate.OpenRouteServiceAPI,
   api_key: System.get_env("OPEN_ROUTE_SERVICE_API_KEY"),
   client: Skate.OpenRouteServiceAPI.Client
 
+config :skate, Swiftly.API.ServiceAdjustments,
+  api_base_url: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL"),
+  api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
+  agency_id: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
+
 config :skate, SkateWeb.AuthManager, secret_key: System.get_env("GUARDIAN_SECRET_KEY")
 
 pool_size =


config/runtime.exs
```diff
@@ -36,6 +36,11 @@ config :skate, Skate.OpenRouteServiceAPI,
   api_key: System.get_env("OPEN_ROUTE_SERVICE_API_KEY"),
   client: Skate.OpenRouteServiceAPI.Client
 
+config :skate, Swiftly.API.ServiceAdjustments,
```
meagharty, line 39: Does the [code anti-pattern of Namespace trespassing](https://hexdocs.pm/elixir/1.18.3/code-anti-patterns.html#namespace-trespassing) apply here? Would it be a concern if we added a `Swiftly` package with a module called `Swiftly.API.ServiceAdjustments`? Should we add the `Skate` unique prefix to avoid module name clashes? 

I noticed OpenRouteServiceAPI above isn't set up the same way, it's Skate.OpenRouteServiceAPI. There are other modules in `skate/lib` that don't have the `Skate` prefix though.

.env.1p.skate
```diff
@@ -11,6 +11,10 @@ SKATE_HASTUS_URL="op://$_1P_VAULT/Skate Data URLs/SKATE_HASTUS_URL"
 SWIFTLY_REALTIME_VEHICLES_URL="op://$_1P_VAULT/Swiftly API/hostname"
 SWIFTLY_AUTHORIZATION_KEY="op://$_1P_VAULT/Swiftly API/credential"
 
+SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/hostname"
```
meagharty, line 14: I added these values to 1Password in the Skate vault

config/runtime.exs
firestack: Good question! Because we want to pull these modules out into their own application at some point, my thought is that "namespace trespassing" doesn't apply, and this is closer to doing something like an umbrella application, but under the same `/lib`. 

Other Skate modules _have_ done "namespace tresspasing" (See, `Realtime`, `Schedule`(?), `Api`(?), `TrainVehicles`, `Report`, `Notifications` (<.<), and maybe `Concentrate`) and I agree that we shouldn't I'd like to avoid doing this in the future, and possibly fix these in the future.

I think `Skate.OpenRouteServiceAPI` makes sense because it's _intended_ for Skate to use ORS directly. Whereas it is not intended in the long run that Skate talks directly to Swiftly, and we plan remove it at some point.

meagharty: Sounds good!

config/runtime.exs
```diff
@@ -36,6 +36,11 @@ config :skate, Skate.OpenRouteServiceAPI,
...
+  api_base_url: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL"),
+  api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
+  agency_id: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
```
firestack, line 42: @meagharty based on the work for the API module, I'm considering the following changes

```suggestion
  base_url:
    (with(
       base_url_var when is_binary(base_url_var) <-
         System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL"),
       {:ok, uri} <- URI.new(base_url_var)
     ) do
       uri
     else
       _ ->
         nil
     end),
  api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
```

meagharty: Why are you suggesting to add the changes to `base_url`, and why here in this file? 

It looks like we have the value for `SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL` set as a string in ECS

meagharty (approved):
Looks good to me!

If you feel strongly that we should ship this as a commit as part of the Swiftly module task, then feel free to put the link to this PR in that task.

config/runtime.exs
firestack: For an example, I want the function `Swiftly.API.ServiceAdjustments.get_adjustments_v1/1` to take [`Keyword`](https://hexdocs.pm/elixir/Keyword.html) list as it's argument, and I'm expecting that the keyword list matches up with the keys in this config.

I think that the `get_adjustments_v1` function should expect that `base_url` is a well formed `URI`, so it's easier to just expect that callers construct a valid `URI` beforehand so that the module doesn't need to concern itself with also validating URL's, which makes it reasonable to push the validation and creation of the `URI` struct into the `runtime.exs` config IMO.

PR #2980, opened 2025-03-21
feat: add detour creation request module to send detour info to swiftly

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209713487040472/f

### DIFF
--- a/.env.1p.skate
+++ b/.env.1p.skate
@@ -11,6 +11,10 @@ SKATE_HASTUS_URL="op://$_1P_VAULT/Skate Data URLs/SKATE_HASTUS_URL"
 SWIFTLY_REALTIME_VEHICLES_URL="op://$_1P_VAULT/Swiftly API/hostname"
 SWIFTLY_AUTHORIZATION_KEY="op://$_1P_VAULT/Swiftly API/credential"
 
+SWIFTLY_SERVICE_ADJUSTMENTS_API_BASE_URL="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/hostname"
+SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/credential"
+SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY="op://$_1P_VAULT/Swiftly API Service Adjustments Development Credentials/agency"
+
 ## Map Tileset URL's
 BASE_TILESET_URL="op://$_1P_VAULT/Skate Map Tileset URLs/BASE_TILESET_URL"
 SATELLITE_TILESET_URL="op://$_1P_VAULT/Skate Map Tileset URLs/SATELLITE_TILESET_URL"
--- a/assets/src/components/notificationCard.tsx
+++ b/assets/src/components/notificationCard.tsx
@@ -176,6 +176,15 @@ export const title = (notification: Notification) => {
         case "raised":
           return "Chelsea St Bridge Raised"
       }
+      break
+    }
+
+    case NotificationType.DetourExpiration: {
+      if (notification.content.expiresIn > 0) {
+        return `Detour duration - ${notification.content.expiresIn} min warning`
+      } else {
+        return "Detour duration reached"
+      }
     }
   }
 }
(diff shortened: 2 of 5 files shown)


lib/skate/detours/db/detour.ex
```diff
@@ -126,6 +133,8 @@ defmodule Skate.Detours.Db.Detour do
           :direction -> select_direction(query)
           :nearest_intersection -> select_starting_intersection(query)
           :estimated_duration -> select_estimated_duration(query)
+          :direction_id -> select_direction_id(query)
+          :coordinates -> select_coordiantes(query)
```
meagharty, line 137: nit: typo `coordinates` (function below too)

lib/skate/detours/request.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Skate.Detours.Request do
+  @moduledoc """
+  Module for constructing request objects for detours.
+  """
+
+  defmodule Swiftly.CreateDetour do
```
meagharty, line 6: Relating to what we discussed in standup, is this the piece that may need to be re-visited once the Swiftly module is ready?

lib/skate/detours/request.ex
```diff
@@ -0,0 +1,75 @@
...
+         %{
+           routeShortName: detour.route_id,
+           direction: Integer.to_string(detour.direction_id),
+           shape: parse_shape(detour)
+         }
+       ]
+     }}
+  end
+
+  def to_swiftly(detour) do
```
meagharty, line 55: What about deactivation?

lib/skate/detours/db/detour.ex
```diff
@@ -247,5 +262,11 @@ defmodule Skate.Detours.Db.Detour do
         ^key => d.state["context"]["selectedDuration"]
       })
     end
+
+    def select_coordiantes(query \\ base(), key \\ :coordinates) do
+      select_merge(query, [detour: d], %{
+        ^key => d.state["context"]["detourShape"]["ok"]["coordinates"]
```
meagharty, line 268: Do we need to handle an error case where "detourShape" is not "ok"?

lib/swiftly/api/requests.ex
```diff
@@ -0,0 +1,51 @@
...
+    {:ok,
+     %CreateAdjustmentRequestV1{
+       feedId: "TEMP_FEED_ID",
+       feedName: "TEMP_FEED_NAME",
+       notes: Integer.to_string(detour.id),
+       details: %DetourV0CreationDetailsV1{
+         adjustmentType: :DETOUR_V0,
+         detourRouteDirectionDetails: [
+           %{
+             routeShortName: detour.route_id,
```
meagharty, line 23: Using `detour.route_id` seems okay to me for this pilot but for my own understanding:

IIRC we already know that we'd prefer to send Swiftly `routeId` instead of `routeShortName` given our GTFS. is that the thinking here?

I saw the virtual field for `route_name` available  on `Skate.Detours.Db.Detour`. Is there a particular reason why you didn't use that here? Will it [be route_long_name](https://github.com/mbta/skate/blob/3c670abebfba400e56670ded35b840d37cf7f81d/lib/schedule/gtfs/route.ex#L79) in some cases anyway?

lib/swiftly/api/requests.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Swiftly.Api.Requests do
+  @moduledoc """
+  Module for constructing request objects for swiftly api requests.
```
meagharty, line 3: Since we discussed fetching the list of swiftly service adjustments to determine which adjustment to delete when a skate detour is ended/deactivated, I can see that being a follow-up PR once the API module is ready. Do you have thoughts, like if it should be a subtask or a new task?

meagharty (approved):
This is looking good to me! A few questions that I'd like follow-up on but I don't see them as merge-blocking.

lib/swiftly/api/requests.ex
firestack: just fyi: I checked in Swiftly and creating a Detour in their UI for the SL1 seems to use `SL1` as the `routeShortName`

lib/skate/detours/db/detour.ex
meagharty: yeah that seems good, I don't think we can recover. we should have logging for those request failures 👍

lib/swiftly/api/requests.ex
meagharty: yeah I'd consider this as part of this same task for now (but another PR), we can chat tomorrow

PR #2984, opened 2025-04-02
feat: update swiftly service adjustments when a detour is activated or deactivated

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209612240132141/f

### DIFF
--- a/lib/swiftly/api/service_adjustments.ex
+++ b/lib/swiftly/api/service_adjustments.ex
@@ -53,7 +53,9 @@ defmodule Swiftly.API.ServiceAdjustments do
       {:ok, %HTTPoison.Response{status_code: 200, body: response_body}} ->
         body = Jason.decode!(response_body)
 
-        Logger.info("adjustment_created_in_swiftly adjustment_id=#{Map.get(body, "adjustmentId")}")
+        Logger.info(
+          "adjustment_created_in_swiftly adjustment_id=#{Map.get(body, "adjustmentId")}"
+        )
 
         {:ok, Swiftly.API.ServiceAdjustments.AdjustmentIdResponse.load(body)}
 


test/swiftly/api/service_adjustments_test.exs
```diff
@@ -12,7 +12,7 @@ defmodule Swiftly.API.ServiceAdjustmentsTest do
 
   @default_arguments [
     client: @mock_client_module,
-    base_url: URI.parse("https://localhost"),
+    base_url: URI.parse("https://localhost/adjustments"),
```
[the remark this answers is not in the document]
firestack: oops I'm sorry I forgot about this, this change was intentional and I'm realizing I think I should've approached this differently in #2976 and https://github.com/mbta/devops/pull/2667

After starting to work on #2981 I realized it was better to use the "true baseurl" rather than our made-up base-url which included `adjustments`. So I'd prefer we update elsewhere rather than updating here & the tests/implementation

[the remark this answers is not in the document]
firestack: I can make those changes in devops/1password, but I want to make sure we as a team are okay with that @mbta/skate-developers

config/runtime.exs
```diff
@@ -49,7 +49,9 @@ config :skate, Swiftly.API.ServiceAdjustments,
          nil
      end),
   api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
-  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
+  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY"),
+  feedId: "TEMP_FEED_ID",
```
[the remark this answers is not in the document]
meagharty: I reviewed and approved the other PR first given the dependency

lib/swiftly/api/requests.ex
```diff
@@ -1,14 +1,14 @@
...
   """
 
   alias Skate.Detours.Db.Detour
-  alias Swiftly.Api.Requests.{CreateAdjustmentRequestV1, DetourV0CreationDetailsV1}
+  alias Swiftly.API.ServiceAdjustments.{CreateAdjustmentRequestV1, DetourV0CreationDetailsV1}
 
   require Logger
 
-  @spec to_swiftly(Detour.t()) :: {:ok, DetourV0CreationDetailsV1.t()} | :error
+  @spec to_swiftly(Detour.t()) :: {:ok, CreateAdjustmentRequestV1.t()} | :error
```
meagharty, line 11: good catch 👍

lib/swiftly/api/requests.ex
```diff
@@ -20,8 +20,8 @@ defmodule Swiftly.Api.Requests do
          beginTime: parse_begin_time(detour),
          detourRouteDirectionDetails: [
            %{
-             routeShortName: detour.route_name,
-             direction: Integer.to_string(detour.direction_id),
+             routeShortName: detour.state["context"]["route"]["name"],
```
meagharty, line 23: Noting here that we'll want to fix this for feedId and feedName

Also noting: I think the added code for configuring feedId and feedName by environment in #2985 supports opts (over request body), so you could remove feedId and feedName here

lib/swiftly/api/service_adjustments/adjustment_with_status_v1.ex
```diff
@@ -24,4 +24,16 @@ defmodule Swiftly.API.ServiceAdjustments.AdjustmentWithStatusV1 do
     :validity,
     :validityReason
   ]
+
+  def load(adjustment) do
+    fields =
+      Enum.map(__MODULE__.__info__(:struct), fn %{field: field} -> Atom.to_string(field) end)
```
meagharty, line 30: ✨ 
https://hexdocs.pm/elixir/Module.html#c:__info__/1

lib/swiftly/api/service_adjustments.ex
```diff
@@ -118,6 +122,7 @@ defmodule Swiftly.API.ServiceAdjustments do
     |> client.request()
     |> case do
       {:ok, %HTTPoison.Response{status_code: 204}} ->
+        "adjustment_deleted_in_swiftly adjustment_id=#{adjustment_id}"
```
meagharty, line 125: Did you mean to log this?

lib/swiftly/api/requests.ex
```diff
@@ -1,55 +1,58 @@
...
      }}
   end
 
-  def to_swiftly(detour) do
-    Logger.warning(
-      "detour_not_active_to_swiftly detour_id=#{detour.id} status=#{inspect(detour.status)}"
-    )
-
-    :error
+  defp parse_begin_time(%Detour{activated_at: activated_at}) when not is_nil(activated_at) do
```
meagharty, line 35: I was thinking we didn't need to provide beginTime if it's approximately "now" in this flow as-is but this isn't a problem to be more specific if we have it

lib/skate/detours/detours.ex
```diff
@@ -208,6 +208,7 @@ defmodule Skate.Detours.Detours do
       {:ok, %Detour{} = new_record} ->
         broadcast_detour(new_record, author_id)
         send_notification(detour_changes, new_record)
+        update_swiftly(detour_changes, new_record)
```
meagharty, line 211: writing aloud for myself (correct me if I'm off anywhere): so if the update to swiftly unexpectedly causes a crash, the detour update was already committed to the database, good 👍 

on the user side for the user modifying the detour, what's the consequence here when this "then" is not evaluated? https://github.com/mbta/skate/blob/54396431fb4b746a01197563ab3f52165468c29a/assets/src/hooks/useDetour.ts#L41
I'm imagining that's minor

meagharty (commented):
This is looking good! I'll take another look tomorrow

lib/swiftly/api/service_adjustments.ex
meagharty: Oh, yeah, my question was more did you mean to pass this string to a `Logger` call here?

lib/skate/detours/detours.ex
meagharty: Okay makes sense ty! 

Remaining question is around the feature flag that we discussed in standup, it's a bit more scope here so we can discuss how to tackle in Slack if you want (implementation, in this PR or separate PR that we'd merge in, etc)

meagharty:
Started testing this in `skate-dev-blue` but testing the happy path with the feature flag on is blocked by https://github.com/mbta/devops/pull/2741

lib/swiftly/api/service_adjustments.ex
```diff
@@ -50,8 +50,12 @@ defmodule Swiftly.API.ServiceAdjustments do
     }
     |> client.request()
     |> case do
-      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
-        {:ok, Jason.decode!(body, keys: :atoms!)}
+      {:ok, %HTTPoison.Response{status_code: 200, body: response_body}} ->
+        body = Jason.decode!(response_body)
+
+        Logger.info("adjustment_created_in_swiftly adjustment_id=#{Map.get(body, :adjustmentId)}")
```
meagharty, line 56: This logs an empty string for adjustment_id as-is because we don't parse the keys as atoms until `load` below

Can we call load before or log from "adjustmentId"?

meagharty (changes requested):
Testing in skate-dev-blue looked good for activate/deactivate! I found one thing to fix-up.

meagharty (approved):

PR #2985, opened 2025-04-03
feat(ex/swiftly/service-adjustments): add feed info to application config and service adjustments API module

firestack:
Asana Task: https://app.asana.com/0/0/1209845589952414

---

Take your pick of implementation. I think the "refactored" commit is "nicer", but maybe the conversion from snake_case to camelCase isn't worth the time. It's easy to drop that commit so happy to do so if desired.


lib/swiftly/api/service_adjustments.ex
```diff
@@ -38,7 +38,10 @@ defmodule Swiftly.API.ServiceAdjustments do
       method: :post,
       url: fetch_base_url(opts) |> URI.append_path("/adjustments") |> URI.to_string(),
       params: params,
-      body: Jason.encode!(body),
+      body:
```
meagharty, line 41: Should we update the tests for the additions? It looks like we check the params like agency already

config/runtime.exs
```diff
@@ -49,7 +49,10 @@ config :skate, Swiftly.API.ServiceAdjustments,
          nil
      end),
   api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
-  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
+  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY"),
+  feed_id: "skate.#{Application.get_env(:skate, :environment_name, nil)}.service-adjustments",
```
meagharty, line 53: Should use `System.get_env("ENVIRONMENT_NAME", "missing-env")` like we do above? Since that application config is defined here in runtime.exs
 
Same below on line 55

meagharty (commented):
Github is having a time for me at the moment with loading new commits at the moment, sending this out for now

image

lib/swiftly/api/service_adjustments.ex
firestack: Yep! sorry took me a bit to realize this could be added after I opened this PR

- [`d8167b8` (#2985)](https://github.com/mbta/skate/pull/2985/commits/d8167b813d5d820d8bd201d0f702209a5848f146)
- [`9711c17` (#2985)](https://github.com/mbta/skate/pull/2985/commits/9711c178dee21d54e90a9ee69bf12c0de1a56a6a)

firestack:
> Github is having a time for me at the moment with loading new commits at the moment, sending this out for now
> 
> image

Sorry!
Those commits are intentionally empty, both in content and description. 

This is more of an issue with the `jj` "mega-merge" workflow conflicting with GitHub's assumptions. _in the past_ (haven't checked recently), GitHub doesn't like when the head of a PR is a merge commit with more than one parent, so I got in the habit of creating empty merge commit +1 empty commit when pushing to GitHub.

I can look into other solutions if this is an issue!

The benefit here _for me_ is it allows me to work on tests and implementation separately, and test different "stages" of test changes against different "stages" of implementation. (it also allows me to absorb changes and force push without affecting the commit id's of other "branches", but I've been trying to tweak how I work since we don't have something like Gerrit for code reviews which would make tracking force-push/commit amending workflows better.)

config/runtime.exs
firestack: yeah actually I'm gonna agree with you here, I'll push up something in a moment.

firestack: https://github.com/mbta/skate/compare/9eb0c50b4bfbd1b11c6051acf97ed14fc392f2be..ddcbf46ef0031c3b44e830143a729153bc6e35c5

config/runtime.exs
```diff
@@ -49,7 +49,10 @@ config :skate, Swiftly.API.ServiceAdjustments,
          nil
      end),
   api_key: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_API_KEY"),
-  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY")
+  agency: System.get_env("SWIFTLY_SERVICE_ADJUSTMENTS_AGENCY"),
+  feed_id: "skate.#{System.get_env("ENVIRONMENT_NAME", "missing-env")}.service-adjustments",
+  feed_name:
+    "Skate (#{System.get_env("ENVIRONMENT_NAME", "missing-env")}) Service Adjustments Feed"
```
firestack, line 55: one note: currently `ENVIRONMENT_NAME` has a `skate-` prefix. so these are duplicated
(ex:
- `skate.skate-dev-blue.service-adjustments`
- `Skate (skate-dev-blue) Service Adjustments Feed`

)

I could remove the prefixes, but I'm not actually sure it's worth it or matters. I was at first thinking we'd want a pattern match for removing the prefix, but I realize while writing this comment we could do `String.replace_prefix("skate-", "")`

Let me know 👍🏻

meagharty: Yeah the prefix doesn't seem worth changing to me, I'm okay with as-is. Your other suggestion would be okay too if you prefer that.

Should we define `System.get_env("ENVIRONMENT_NAME", "missing-env")` in a variable above line 7 in this file to reduce duplication? Since we don't see this pilot code living here this way long-term, I don't feel strongly about that.

lib/swiftly/api/service_adjustments.ex
meagharty: The tests look good to me! 👍

meagharty (approved):
looks good! I had one comment that's not blocking about adding a variable, I accidentally submitted that outside of the review (sorry about that)

good to know about the jj workflow in case I see that again

config/runtime.exs
firestack: > Should we define `System.get_env("ENVIRONMENT_NAME", "missing-env")` in a variable above line 7 in this file to reduce duplication? Since we don't see this pilot code living here this way long-term, I don't feel strongly about that.

Good question, I think right now, no.

PR #2987, opened 2025-04-07
feat: add `stop_times` to `Trip` info

firestack:
Asana Ticket: https://app.asana.com/0/1203014709808707/1209871468833666/f

### DIFF
--- a/assets/src/minischedule.d.ts
+++ b/assets/src/minischedule.d.ts
@@ -1,5 +1,5 @@
 import { RunId } from "./realtime"
-import { BlockId, DirectionId, RouteId, TripId, ViaVariant } from "./schedule"
+import { BlockId, DirectionId, RouteId, StopId, TimepointId, TripId, ViaVariant } from "./schedule"
 
 export type Activity = Break | Piece
 
@@ -53,6 +53,13 @@ export interface Trip {
   endTime: Time
   startPlace: string
   endPlace: string
+  stopTimes: StopTime[]
+}
+
+export interface StopTime {
+  time: Time
+  stopId: StopId,
+  timepointId: TimepointId | null
 }
 
 export type Time = number
--- a/assets/src/minischedule.d.ts
+++ b/assets/src/minischedule.d.ts
@@ -1,5 +1,13 @@
 import { RunId } from "./realtime"
-import { BlockId, DirectionId, RouteId, StopId, TimepointId, TripId, ViaVariant } from "./schedule"
+import {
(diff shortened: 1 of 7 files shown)


assets/src/minischedule.d.ts
```diff
@@ -53,6 +61,13 @@ export interface Trip {
   endTime: Time
   startPlace: string
   endPlace: string
+  stopTimes: StopTime[]
+}
+
+export interface StopTime {
```
meagharty, line 67: Why did you decide to define `StopTime` in minischedule over schedule? 

I suppose I was expecting `schedule.d.ts` if that can be thought as "GTFS schedule" originated data but I'm not sure that's the case.

assets/src/models/minischeduleData.ts
```diff
@@ -59,6 +68,13 @@ interface TripData {
   end_time: Time
   start_place: string | null
   end_place: string | null
+  stop_times: StopTimeData[]
+}
+
+interface StopTimeData {
```
meagharty, line 74: Do we need this type if import StopTime already?

assets/tests/factories/stopTimeFactory.ts
```diff
@@ -0,0 +1,8 @@
+import { Factory } from "fishery"
```
meagharty, line 1: oooh nice!

assets/src/models/minischeduleData.ts
firestack: Good question, this is related to the case conventions I mentioned yesterday.

`StopTime` uses TypeScript case conventions (camelCase) whereas the `*Data` variants of structures use the Elixir case conventions (snake_case)

So in keeping with existing conventions in Skate, we need both the "Regular" type and the "Data" type defined.

assets/src/minischedule.d.ts
firestack: Mainly because of locality. I think minischedules is the only thing mainly using this right now, and because `Trip` is defined in here and I needed to add `StopTime` to `Trip` I figured I'd define `StopTime` locally with it's usage in `Trip`.

meagharty: Okay, makes sense. It looks like this Trip has `RunId`, defined in `realtime.d.ts`, so `Trip` doesn't belong in `schedule.d.ts` as-is either. I don't have a better suggestion atm, this works.

meagharty (approved):

PR #2988, opened 2025-04-08
feat: add create/delete in swiftly option in detour admin pages

bfauble:
Asana ticket: https://app.asana.com/0/1205385723132845/1209913555044913/f

* Show adjustment info if present in swiftly
* Show Delete button if detour status is `:past` and adjustment still present in swiftly
* Show Create button if detour status is `:active` and adjustment is not present in swiftly

### DIFF
--- a/assets/src/minischedule.d.ts
+++ b/assets/src/minischedule.d.ts
@@ -1,5 +1,13 @@
 import { RunId } from "./realtime"
-import { BlockId, DirectionId, RouteId, TripId, ViaVariant } from "./schedule"
+import {
+  BlockId,
+  DirectionId,
+  RouteId,
+  StopId,
+  TimepointId,
+  TripId,
+  ViaVariant,
+} from "./schedule"
 
 export type Activity = Break | Piece
 
@@ -53,6 +61,13 @@ export interface Trip {
   endTime: Time
   startPlace: string
   endPlace: string
+  stopTimes: StopTime[]
+}
+
+export interface StopTime {
+  time: Time
+  stopId: StopId
+  timepointId: TimepointId | null
 }
 
 export type Time = number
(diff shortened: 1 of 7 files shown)


lib/skate/detours/detours.ex
```diff
@@ -376,6 +390,17 @@ defmodule Skate.Detours.Detours do
...
 
+  defp get_swiftly_adjustment_from_response(adjustments_response, detour) do
+    adjustments_response
+    |> Map.get(:adjustments, [])
+    |> Enum.filter(fn adjustment ->
+      Map.get(adjustment, :status) == "ACTIVE" &&
+        Map.get(adjustment, :adjustmentType) == "DETOUR_V0" &&
+        Map.get(adjustment, :notes) == Integer.to_string(detour.id)
+    end)
+    |> Enum.at(0)
```
firestack, line 401: nit/suggestion(non-blocking): just curious, is there a reason to prefer `Map.get` over pattern matching?

```suggestion
    detour_id = Integer.to_string(detour.id)

    adjustments_response
    |> Map.get(:adjustments, [])
    |> Enum.filter(fn
      %{status: "ACTIVE", adjustmentType: "DETOUR_V0", notes: ^detour_id} -> true
      _ -> false
    end)
    |> Enum.at(0)
```

lib/skate/detours/detours.ex
```diff
@@ -352,18 +357,27 @@ defmodule Skate.Detours.Detours do
...
+  defp update_swiftly(%Detour{status: :active} = detour, :create, true) do
+    {:ok, adjustment_request} = Swiftly.API.Requests.to_swiftly(detour)
+
+    service_adjustments_module().create_adjustment_v1(
+      adjustment_request,
+      build_swiftly_opts()
+    )
+  end
+
+  defp update_swiftly(%Detour{status: :past} = detour, :delete, true) do
```
firestack, line 372: suggestion: is it still useful to keep the third argument that's always `true` here? Or could we simplify this and drop these instances of `update_swiftly/3` to `update_swiftly/2` calls?

PR #2989, opened 2025-04-14
feat(ts/components/minischedule): add timepoint stop information to run and block schedules

firestack:
Getting this PR up so we can discuss the implementation, while I work out the style still.

---

Asana Ticket: https://app.asana.com/1/15492006741476/project/1205385723132845/task/1208502301998476

TODO:
- [x] Create/update style for new timepoint rows
- [ ] Add any missing tests

---
  - https://app.asana.com/0/0/1210002892244198

### DIFF
--- a/assets/css/_minischedule.scss
+++ b/assets/css/_minischedule.scss
@@ -67,6 +67,10 @@
     border-top: 1px solid $color-component-light;
   }
 
+  // The timepoint-row and layover rows are very similar at the moment, both
+  // with their borders and padding/height, so timepoint-rows are tagging along
+  // with some layover row styling, _for now_.
+  &.c-minischedule__timepoint-row,
   &.c-minischedule__layover-row {
     border-top-style: none;
     padding-top: 0;
@@ -76,6 +80,20 @@
     .c-minischedule__right-text {
       border-top: 1px solid $color-component-light;
       padding-top: 0.25rem;
+    }
+  }
+
+  &.c-minischedule__timepoint-row {
+    .c-minischedule__left-text {
+      // Indent the stop text slightly to visually associate the stops with the trip
+      margin-left: 0.25rem;
+    }
+
+    // Use block layout mode to remove any issues from inline layout
+    // (i.e., vertical layout offsets from `line-height`)
+    .c-minischedule__svg--revenue,
+    svg {
+      display: block;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


firestack:
@bfauble just giving a heads up/rundown of what changed

1. Added a refactor which uses `findIndex` and `map` instead of `reduce` to make it a little simpler, and also make it easier to swap out the logic if we implemented this using StopSequence instead.
2. Added a refactor which pulls it out into it's own component, I'm not certain about this so happy to drop this commit
3. Added the CSS for this feature
4. Added the feature flag for this feature (and updated tests to use it)
	5. not so sure if it's worth adding tests for the absence of the flag. Running it locally, I confirmed the snapshot changes are undone if the flag is not present <shrug>

assets/css/_minischedule.scss
```diff
@@ -88,6 +106,19 @@
   padding: 0.25rem;
   width: 1rem;
   z-index: 2;
+
+  .c-minischedule__timepoint-row & {
+    // Make the icon smaller
```
meagharty, line 111: I can see how the re-sizing of the icons requires significantly more work with the current css and icons. This is a good start for what we want!

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -687,16 +696,95 @@ const RevenueTrip = ({
...
+          overloadOffset={overloadOffset}
+          timeBasedStyle={timeBasedStyle}
+          nextStop={nextStop}
+        />
+      )}
+    </>
+  )
+}
+
+const TimePointRows = ({
```
meagharty, line 727: This wrapping a (set of) Row mirrors existing code in this file, cleaner and makes sense to me.

assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -559,6 +561,11 @@ const TripSchedule = ({
             timeBasedStyle={onRouteTimeBasedStyle}
             activeStatus={onRouteActiveStatus}
             overloadOffset={overloadOffset}
+            nextStop={
+              (isVehicleInScheduledService(vehicleOrGhost) &&
```
meagharty, line 565: **question** (not blocking): Do we have existing tests for this realtime vehicle data in this panel? 

I confirmed this looked good on skate-dev-green for a vehicle with realtime data.

meagharty (approved):

assets/src/components/propertiesPanel/minischedule.tsx
firestack: As far as I can tell, none of the existing tests test with realtime data. I don't see any calls to `update` in the test file which seems to be the way to re-render the react tree (since this file is mainly using `react-test-renderer` and not `@testing-library/react`). Would be good for us to have, maybe as a followup?

meagharty: I'm going to resolve this so it's not blocking

PR #2993, opened 2025-04-23
feat(minischedule): use timepoint name for timepoints in run and block panels

meagharty:
**Asana Ticket**: [Add timepoint name field for timepoints](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1208502301998475?focus=true)

### DIFF
--- a/assets/src/components/propertiesPanel/minischedule.tsx
+++ b/assets/src/components/propertiesPanel/minischedule.tsx
@@ -26,10 +26,11 @@ import {
 } from "../../hooks/useMinischedule"
 import {
   AsDirected,
-  Block,
   Break,
   Piece,
   Run,
+  ScheduleRun,
+  ScheduleBlock,
   Time,
   Trip,
 } from "../../minischedule"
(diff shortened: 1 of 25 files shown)


assets/src/components/propertiesPanel/minischedule.tsx
```diff
@@ -74,10 +89,13 @@ export const MinischeduleBlock = ({ vehicleOrGhost }: Props): ReactElement => {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     vehicleOrGhost.tripId!
   )
+  const timepoints = useTimepointsByIdForRoute(vehicleOrGhost.routeId)
```
firestack, line 92: issue: Hmm, I'm not sure that we can do this by `routeId`. Have you checked the results for runs or blocks that span multiple routes and variants?

For instance, run `138-1046` does both route 16 and 39, which we expect to show in a single view. If we source this for only a single route, and only the active route of the vehicle, that means we'd show the ID for timepoints which don't overlap with the route id and variant of the current trip.

Run `123-1434` does variants of route `16`, `16_9` and `16_5` which also have different timepoints (though a lot of overlap), which points out the possibility of a single route having multiple variants without timepoint overlap, which would mean they're missing from this call.

Which I think means if we wanted to do this as a seprate API call, we'd either need a `useTimepointsForRunId`/`useTimepointsForBlockId` or make multiple calls with a `useTimepointsForTripId` or `useTimepointsForRoutePattern`?

IMO, between the prop drilling, and the complexity of cases that need to be handled, I wonder if doing this in the `ScheduleController` or elsewhere on the back-end would be better?

firestack: > I'm not sure this is an issue, shouldn't all those route variants (16, 16_9 and 16_5) be covered by call for the timepoints for the routeId of `16`?

Ah, I think you're right there, I had assumed that "timepoints for route" didn't handle variants or did some sort of reduction that made it not apply to all routes since it's used for route ladders.

meagharty: I moved fetching the timepoints for a run or block to the backend 👍

PR #2996, opened 2025-04-29
fix(timepoints): filter schedule pieces to scheduled trips with stop times

meagharty:
Asana Ticket: [Handle AsDirected activities](https://app.asana.com/1/15492006741476/project/1148853526253420/task/1210114300058018?focus=true)

### DIFF
--- a/test/schedule_test.exs
+++ b/test/schedule_test.exs
@@ -1142,27 +1142,11 @@ defmodule ScheduleTest do
     setup do
       pid =
         Schedule.start_mocked(%{
-          hastus: %{
-            "activities.csv" => [
-              "schedule_id;area;run_id;start_time;end_time;start_place;end_place;activity_type;activity_name",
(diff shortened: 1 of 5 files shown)


test/schedule_test.exs
```diff
@@ -1145,6 +1145,8 @@ defmodule ScheduleTest do
           hastus: %{
             "activities.csv" => [
               "schedule_id;area;run_id;start_time;end_time;start_place;end_place;activity_type;activity_name",
+              "schedule;123;4567;00:00;00:00;start;start;Sign-on;Sign-on",
```
meagharty, line 1148: This data isn't required since we only need `gtfs` => [`checkpoints.txt`, `trips.txt`, and `stop_times.txt`] for the timepoint lookups. `Schedule.AsDirected` is produced in the `piece` below.

Removing extra test data in a commit to be pushed up.

firestack (approved):
🧼

PR #2999, opened 2025-05-08
fix(ex/notifications/bridge): remove comma in log

firestack:
In Splunk, the logfmt parser they have doesn't really handle this comma very well, and logfmt formatted logs should be key=value pairs with spaces between.

---

No Ticket.

### DIFF
--- a/lib/notifications/bridge.ex
+++ b/lib/notifications/bridge.ex
@@ -119,7 +119,7 @@ defmodule Notifications.Bridge do
       |> do_get_datetime()
 
     Logger.info(
-      "bridge_response status=#{inspect(raw_status)}, lowering_time=#{inspect(lowering_time)}"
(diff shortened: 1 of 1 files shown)


lib/notifications/bridge.ex
```diff
@@ -119,7 +119,7 @@ defmodule Notifications.Bridge do
       |> do_get_datetime()
 
     Logger.info(
-      "bridge_response status=#{inspect(raw_status)}, lowering_time=#{inspect(lowering_time)}"
+      "bridge_response status=#{inspect(raw_status)} lowering_time=#{inspect(lowering_time)}"
```
meagharty, line 122: Was the status field previously parsed as `status="Lowered",` instead of the expected `status="Lowered"`?

meagharty (approved):
Thanks for fixing this up!

PR #3003, opened 2025-05-22
Create Detour Scheduler Module

bfauble:
Asana Ticket: [Create Detour Scheduler Module](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1209653841420225?focus=true)


firestack (commented):
Leaving my in-progress thoughts before heading out for today, in case you need to find another reviewer, otherwise I'll finish reviewing when we're back on 2025/05/27 👍🏻

lib/skate/detours/notification_scheduler.ex
```diff
@@ -0,0 +1,43 @@
...
+      ) do
+    DetourExpirationNotification
+    |> Skate.Repo.get_by(detour_id: detour.id)
+    |> DetourExpirationNotification.update_changeset(%{
+      detour: detour,
+      estimated_duration: estimated_duration,
+      expires_at: expires_at
+    })
+    |> Skate.Repo.insert_or_update()
+  end
```
firestack, line 40: suggestion(non-blocking): https://hexdocs.pm/elixir/code-anti-patterns.html#complex-extractions-in-clauses

```suggestion
  def detour_activated(
        %Detour{status: :active} = detour,
        expires_at
      ) do
    %{estimated_duration: estimated_duration} = detour

    %{detour: detour, estimated_duration: estimated_duration, expires_at: expires_at}
    |> DetourExpirationNotification.create_changeset()
    |> Skate.Repo.insert()
  end

  def detour_activated(_, _), do: :error

  def detour_deactivated(%Detour{status: :past} = detour) do
    DetourExpirationNotification
    |> Skate.Repo.get_by!(detour_id: detour.id)
    |> Skate.Repo.delete()
  end

  def detour_deactivated(_), do: :error

  def detour_duration_changed(
        %Detour{status: :active} = detour,
        expires_at
      ) do
    %{estimated_duration: estimated_duration} = detour

    DetourExpirationNotification
    |> Skate.Repo.get_by(detour_id: detour.id)
    |> DetourExpirationNotification.update_changeset(%{
      detour: detour,
      estimated_duration: estimated_duration,
      expires_at: expires_at
    })
    |> Skate.Repo.insert_or_update()
  end
```

test/skate/detours/notification_scheduler_test.exs
```diff
@@ -0,0 +1,79 @@
...
+      new_expires_at = DateTime.add(expires_at, 20, :minute)
+
+      assert {:ok, updated_detour_expiration_notification} =
+               NotificationScheduler.detour_duration_changed(detour, new_expires_at)
+
+      assert new_expires_at == updated_detour_expiration_notification.expires_at
+      assert detour_expiration_notification.id == updated_detour_expiration_notification.id
+    end
+
+    test "creates record if previous notification did not exist becuase of 'Until further notice' estimated duration" do
```
firestack, line 63: typo:

```suggestion
    test "creates record if previous notification did not exist because of 'Until further notice' estimated duration" do
```

lib/skate/detours/notification_scheduler.ex
firestack: a little more clear
```diff
Commit ID: b8c5fc1b380cdd3f936098e3bfc2323348c5de06
Change ID: vovmyrmwwrzxppypmwzpruztolkvtksm
Author   : Kayla Firestack <kfirestack@mbta.com> (2025-05-28 10:00:05)
Committer: Kayla Firestack <kfirestack@mbta.com> (2025-05-28 10:02:15)

    (no description set)

diff --git a/lib/skate/detours/notification_scheduler.ex b/lib/skate/detours/notification_scheduler.ex
index 44755daef2..d19000eebc 100644
--- a/lib/skate/detours/notification_scheduler.ex
+++ b/lib/skate/detours/notification_scheduler.ex
@@ -7,9 +7,11 @@
   require Logger
 
   def detour_activated(
-        %Detour{status: :active, estimated_duration: estimated_duration} = detour,
+        %Detour{status: :active} = detour,
         expires_at
       ) do
+    %{estimated_duration: estimated_duration} = detour
+
     %{detour: detour, estimated_duration: estimated_duration, expires_at: expires_at}
     |> DetourExpirationNotification.create_changeset()
     |> Skate.Repo.insert()
@@ -26,9 +28,11 @@
   def detour_deactivated(_), do: :error
 
   def detour_duration_changed(
-        %Detour{status: :active, estimated_duration: estimated_duration} = detour,
+        %Detour{status: :active} = detour,
         expires_at
       ) do
+    %{estimated_duration: estimated_duration} = detour
+
     DetourExpirationNotification
     |> Skate.Repo.get_by(detour_id: detour.id)
     |> DetourExpirationNotification.update_changeset(%{
```

priv/repo/migrations/20250522210823_add_detour_expiration_notifications_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Skate.Repo.Migrations.AddDetourExpirationNotificationsTable do
+  use Ecto.Migration
+
+  def change do
+    create table("detour_expiration_notifications") do
+      add :expires_at, :utc_datetime_usec, null: false
+      add :detour_id, references(:detours), null: false
```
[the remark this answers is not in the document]
firestack: My _expectation/assumption_ is that if you somehow _updated_ the Primary key on a record, `on_update: :update_all` would configure postgres to cascade the change to all FK's referencing that record.

I don't know if we should or shouldn't use it, but if my assumption is correct, then it would prevent our data from becoming inconsistent.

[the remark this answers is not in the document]
firestack: https://www.postgresql.org/docs/current/ddl-constraints.html
https://hexdocs.pm/ecto_sql/Ecto.Migration.html#references/2

lib/skate/detours/notification_scheduler.ex
firestack, line 1: question(non-blocking): I'm curious to chat about the implementation direction here and tradeoffs:

I'm wondering if having the changeset functions do the logic for figuring out if something should be created or modified is the right direction, what do you think of this alternate implementation?

```diff
Commit ID: c088656480bd708a9d223c7839481f2241d816ad
Change ID: pxtkzlokorqxxwlrnsoxrnvvysxlqttq
Author   : Kayla Firestack <kfirestack@mbta.com> (2025-06-02 09:21:17)
Committer: Kayla Firestack <kfirestack@mbta.com> (2025-06-02 13:19:58)

    (no description set)

diff --git a/lib/skate/detours/db/detour.ex b/lib/skate/detours/db/detour.ex
index 624b60e81f..d09f7e98c0 100644
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -42,6 +42,7 @@
     field :estimated_duration, :string, virtual: true
 
     # -------------------------------------------------------
+    has_many :expiration_tasks, Skate.Detours.Db.DetourExpirationTask
   end
 
   def changeset(detour, attrs) do
diff --git a/lib/skate/detours/db/detour_expiration_task.ex b/lib/skate/detours/db/detour_expiration_task.ex
index 7aa4c4a4c5..938e59cf14 100644
--- a/lib/skate/detours/db/detour_expiration_task.ex
+++ b/lib/skate/detours/db/detour_expiration_task.ex
@@ -15,17 +15,10 @@
     timestamps()
   end
 
-  def create_changeset(%{detour: %Skate.Detours.Db.Detour{} = detour} = attrs) do
-    %__MODULE__{}
-    |> cast(attrs, [:expires_at])
-    |> put_assoc(:detour, detour)
-  end
-
-  def update_changeset(nil, attrs) do
-    create_changeset(attrs)
-  end
-
-  def update_changeset(detour_expiration_task, attrs) do
-    cast(detour_expiration_task, attrs, [:expires_at])
+  def changeset(expiration_task, params) do
+    expiration_task
+    |> cast(params, [:expires_at])
+    |> validate_required([:expires_at, :detour_id])
+    |> assoc_constraint(:detour)
   end
 end
diff --git a/lib/skate/detours/notification_scheduler.ex b/lib/skate/detours/notification_scheduler.ex
index a0c043a16e..337acfbcbb 100644
--- a/lib/skate/detours/notification_scheduler.ex
+++ b/lib/skate/detours/notification_scheduler.ex
@@ -1,6 +1,6 @@
 defmodule Skate.Detours.NotificationScheduler do
   @moduledoc """
-  Interface for managing detour expiration notifiactions.
+  Interface for managing detour expiration notifications.
   """
 
   alias Skate.Detours.Db.{Detour, DetourExpirationTask}
@@ -10,8 +10,12 @@
         %Detour{status: :active, estimated_duration: estimated_duration} = detour,
         expires_at
       ) do
-    %{detour: detour, estimated_duration: estimated_duration, expires_at: expires_at}
-    |> DetourExpirationTask.create_changeset()
+    detour
+    |> new_task_from_detour()
+    |> DetourExpirationTask.changeset(%{
+      expires_at: expires_at,
+      estimated_duration: estimated_duration
+    })
     |> Skate.Repo.insert()
   end
 
@@ -29,10 +33,9 @@
         %Detour{status: :active, estimated_duration: estimated_duration} = detour,
         expires_at
       ) do
-    DetourExpirationTask
-    |> Skate.Repo.get_by(detour_id: detour.id)
-    |> DetourExpirationTask.update_changeset(%{
-      detour: detour,
+    (Skate.Repo.get_by(DetourExpirationTask, detour_id: detour.id) ||
+       new_task_from_detour(detour))
+    |> DetourExpirationTask.changeset(%{
       estimated_duration: estimated_duration,
       expires_at: expires_at
     })
@@ -40,4 +43,8 @@
   end
 
   def detour_duration_changed(_, _), do: :error
+
+  defp new_task_from_detour(detour) do
+    Ecto.build_assoc(detour, :expiration_tasks)
+  end
 end
```

I have two reasons for why I came up with the above implementation
1. the case `update_changeset(nil, attrs)` wouldn't ever be reached by "normal" Ecto changeset functions, i.e., if for some reason we used `cast_assoc` somewhere else in the codebase:
	```elixir
	|> cast_assoc(:expiration_tasks, with: &DetourExpirationTask.update_changeset/2)
	```
	I don't think it's possible for Ecto to provide `nil` as a first parameter, which makes me _wonder_ if we should follow that design.

2. When reading https://hexdocs.pm/ecto/Ecto.Changeset.html#put_assoc/4-example-adding-a-comment-to-a-post it shows that there are a few ways to associate records, but one that stands out to me is [`Ecto.build_assoc/3`](https://hexdocs.pm/ecto/Ecto.html#build_assoc/3). I like `build_assoc` because it doesn't require that the changeset directly access values in the `params` to put an association, and instead allows the changeset to ignore configuring associations at all, and can focus on validating that the associations exist, which pushes the logic for whether something is _allowed_ to be associated up to the caller.  

	`put_assoc` otherwise feels _slightly_ "dangerous" as it can (attempt to) erase associations if provided `nil`, update the associated record when you might otherwise not want to, or create records. Whereas validating the association and building that association before the changeset function means that you can enforce with `assoc_constraint` and `validate_required` that the association exists and that the associated record already exists and _will not_ be created when inserting the new associated record (a `DetourExpirationTask` in this case)

I don't think either of these are **hard** blockers for merging, but I am looking to improve my own usage of Ecto and am curious what you think.

lib/skate/detours/notification_scheduler.ex
```diff
@@ -0,0 +1,43 @@
...
+      ) do
+    DetourExpirationTask
+    |> Skate.Repo.get_by(detour_id: detour.id)
+    |> DetourExpirationTask.update_changeset(%{
+      detour: detour,
+      estimated_duration: estimated_duration,
+      expires_at: expires_at
+    })
+    |> Skate.Repo.insert_or_update()
+  end
```
firestack, line 40: question: Do you anticipate this implementation changing when working on [Implement Notification Creation Job](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210329527498513)?

Right now the code (`_changed` and `_deactivated`) assume there is only one `ExpirationTask` per detour, but because we need to send both the "30 minutes remaining" and "0 minutes remaining" after they have been reached, I had _assumed_ that a `ExpirationTask` would be created for each "set point"[^1] and deleted as the tasks are completed. I'm okay to defer this to that engineering task if that was your plan, just wanted to ask just in case[^2].

Thoughts?

[^1]: for lack of a better word

[^2]: also feedback for myself, I think I wrote these implementation tickets a little backwards, and splitting this work out into it's own "create crud interface" prerequisite task may have not been the right call since it's building bottom up from a spec in a ticket for some "imaginary" use case rather than top down, working from the UX requirements. So that's on me 🫠 I'll keep this in mind next time and seek feedback from the team here.

firestack (approved):
Left a few questions but aside from anything that may arise from chatting about them, LGTM.

PR #3006, opened 2025-05-29
Notify Scheduler when Detour is Activated

bfauble:
Asana Ticket: [Notify Scheduler when Detour is Activated](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1209653841420228?focus=true)


lib/skate/detours/detours.ex
```diff
@@ -426,4 +428,53 @@ defmodule Skate.Detours.Detours do
...
+  end
+
+  defp do_calculate_expiration_timestamp(
+         %Detour{
+           estimated_duration: n_hours
+         } = detour
+       )
+       when is_binary(n_hours) do
+    hours =
+      n_hours
```
meagharty, line 470: I like that we could add or tweak "x hour(s)" values without modifying this code.

Do you have thoughts around the expected values handled here and the values only being defined [on the frontend](https://github.com/mbta/skate/blob/a1f9334cdeee053566d9019ac3b1a74b710af425/assets/src/components/detours/activateDetourModal.tsx#L5-L16)? I don't think we have these values defined on the BE.

This isn't blocking, the current FE values are handled here.

lib/skate/detours/detours.ex
```diff
@@ -296,17 +296,19 @@ defmodule Skate.Detours.Detours do
...
-  defp send_notification(
+  defp process_notifications(
          %Ecto.Changeset{
            data: %Detour{status: :draft},
            changes: %{status: :active}
          },
          %Detour{} = detour
        ) do
     Notifications.NotificationServer.detour_activated(detour)
+    expires_at = calculate_expiration_timestamp(detour)
```
meagharty, line 307: Are there tests for this? Not sure if there's anything in the other branch

meagharty: We discussed synchronously that [a follow up task](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210440024819575?focus=true) will add more test coverage around calculating the expiration timestamp including service end time, etc.

PR #3008, opened 2025-05-30
fix(amend/adr/5): use `Duration` where `DateTime` was accidentally used

firestack:
I had intended when writing the ADR that `expires_at` would be a `Duration`, but it seems I missed that when originally writing this ADR.

The ticket for [Create New Notification Type on Backend](https://app.asana.com/1/15492006741476/project/1203014709808707/task/1209653841420226?focus=true) specifies `Duration`, and that's how I have currently implemented it, but if we need to discuss the path forward I understand, since I missed this.

Related-To: #3009

### DIFF
--- a/documentation/adr/0005-detour-reminder-toasts-architecture.md
+++ b/documentation/adr/0005-detour-reminder-toasts-architecture.md
@@ -498,14 +498,14 @@ iex> Skate.Notifications.create_detour_expiration_notification(%{
 ...>   # store the _current_ estimated duration at this point in time
 ...>   estimated_duration: detour.estimated_duration,
 ...>   # 
-...>   expires_at: %DateTime{}
+...>   expires_at: %Duration{}
 ...> })
 
 %Skate.Notifications.Notification{
   detour_expiration: %Skate.Notifications.DetourExpiration{
     detour_id: ^detour.id,
     estimated_duration: ^detour.estimated_duration
-    expires_at: %DateTime{},
+    expires_at: %Duration{},
     # [...]
   },
   # [...]
--- a/documentation/adr/0005-detour-reminder-toasts-architecture.md
+++ b/documentation/adr/0005-detour-reminder-toasts-architecture.md
@@ -498,14 +498,14 @@ iex> Skate.Notifications.create_detour_expiration_notification(%{
 ...>   # store the _current_ estimated duration at this point in time
 ...>   estimated_duration: detour.estimated_duration,
 ...>   # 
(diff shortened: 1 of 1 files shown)


firestack:
Adding @meagharty since they originally reviewed the ADR in #2998

documentation/adr/0005-detour-reminder-toasts-architecture.md
```diff
@@ -498,14 +498,14 @@ iex> Skate.Notifications.create_detour_expiration_notification(%{
 ...>   # store the _current_ estimated duration at this point in time
 ...>   estimated_duration: detour.estimated_duration,
 ...>   # 
-...>   expires_at: %DateTime{}
+...>   expires_at: %Duration{}
```
meagharty, line 501: Should this be expires_in if it's a duration?

```suggestion
...>   expires_in: %Duration{}
```

Understanding the why of this change: beyond the frontend displaying the expiration as duration, is there another reason for using a Duration here? Compared to something like a string representation of the duration (considering that we are configuring these to a limited set of values on the backend as-is)?

This seems reasonable enough to me though. Looking at the accompanying PR, it doesn't seem like it adds much complexity.

firestack: > Should this be expires_in if it's a duration?

Whoops 🙃 Yes this is correct, fixed in d0d571b8

firestack: > Understanding the why of this change: beyond the frontend displaying the expiration as duration, is there another reason for using a Duration here? Compared to something like a string representation of the duration (considering that we are configuring these to a limited set of values on the backend as-is)?

I had made the assumption that Duration would be the most useful and versatile type to use, since you can [encode & decode Duration from/to a ISO-8602 string](https://hexdocs.pm/elixir/Duration.html#to_iso8601/1), use it to [`shift` a DateTime](https://hexdocs.pm/elixir/DateTime.html#shift/3), or [add and subtract Durations](https://hexdocs.pm/elixir/Duration.html#add/2), as well as directly access the units[^1]. 

I figured it would be used to backdate the `created_at` time when creating a notification and make configuring the `NotificationScheduler` via `config.exs` declarative and easy to read. From there, I just passed the type onto the Notifications system as well because I didn't see a reason to change the type.

[^1]: At least until Postgres poked a few holes in that as noted in the ["Usage of %Duration{} in Schema" section of the linked PR](https://github.com/mbta/skate/pull/3009)

meagharty (approved):

PR #3009, opened 2025-05-30
feat: create detour expiration notification type

firestack:
Asana Ticket: [Create New Notification Type on Backend](https://app.asana.com/1/15492006741476/task/1210418717846181)

# Summary
This does just about everything to setup the database side of notifications, except for returning the new notifications from `unexpired_notifications_for_user`. 

If a `DetourExpiration` notification is returned to the frontend before the Superstruct definition is completed on the frontend, then all notifications will fail to work, so that has been left out of this to unblock this from being merged or needing any sort of rollback.

## Usage of `%Duration{}` in Schema

This also configures `PostgrexTypes` per the documentation, so that we can use `%Duration{}`(as per the ideas in the ADR[^adr-mistake]).

[^adr-mistake]: turns out, I also had a bit of a "typo" in the original ADR, which I've opened a [PR to correct the mistake](https://github.com/mbta/skate/pull/3008), but if we need to reconsider usage of `%Duration{}` since it wasn't originally in the ADR that would be understandable.

I believe this is the _correct_ type to use for this, but I'm not sure if it'll currently be the most _ergonomic_ type, because Postgres stores limited info in it's `interval` type (<a href="https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT"><q>PostgreSQL stores interval values as months, days, and microseconds</q></a>) and due to this postgrex converts the fields in a duration into this limited set and fails to reconstruct it as the original units.

So for example, when a Duration is created with 30 minutes, it comes out of the database as 1800 seconds on retrieval. 
```elixir
iex> Duration.new!(minute: 30)
%Duration{minute: 30}

# becomes
iex> ExampleDatabase.get()
%Duration{second: 1800, microsecond: {0, 6}}
```

I had assumed that the `Duration` API would have functions for converting between units or ["balancing"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration#duration_balancing) the duration, but that doesn't seem to exist, like it does in [C#'s `TimeSpan`](https://learn.microsoft.com/en-us/dotnet/api/system.timespan.totalminutes?view=net-9.0), or [Temporal's `round()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration/round).

I expect that it would be easy[^2] to implement "balancing" as needed[^1], but it's disappointing to realize that this doesn't exist _now_. But, at the risk of speculating too much, I think it's still worth using the `Duration` type, as those functions could be either:
1. added in future Elixir versions
2. implemented as a Library
3. or done well enough with our own implementation ad-hoc.[^postgrex-reference-implementation]

[^postgrex-reference-implementation]: Here's some examples from Postgrex of what implementing this would _maybe_ look like:
	1. https://github.com/elixir-ecto/postgrex/blob/412b55567b6f0f3feb587e38466fcab047581c0f/lib/postgrex/extensions/interval.ex#L80-L82
	2. https://github.com/elixir-ecto/postgrex/blob/412b55567b6f0f3feb587e38466fcab047581c0f/lib/postgrex/extensions/interval.ex#L68-L76
	3. https://github.com/elixir-ecto/postgrex/blob/412b55567b6f0f3feb587e38466fcab047581c0f/lib/postgrex/extensions/interval.ex#L31-L44

[^1]: i.e., convert 1800 seconds to 30 minutes, or instead change implementations to check for 1800 seconds instead of 30 minutes

[^2]: in our _limited_ use case, which is 30 minutes or 0, the problem set is not "easy", as mentioned in the [Temporal docs on Duration Balancing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration#duration_balancing)

<details>
<summary>Enabling Duration Type Docs</summary>

1. > For the `:duration` type, you may need to enable [Duration](https://hexdocs.pm/elixir/Duration.html) support in your adapter. For information on how to enable it in Postgrex, see their [HexDocs page](https://hexdocs.pm/postgrex/readme.html#data-representation).
	> &mdash; https://hexdocs.pm/ecto/Ecto.Schema.html#module-primitive-types
1. https://hexdocs.pm/postgrex/Postgrex.Types.html#define/3

</details>

### DIFF
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -54,6 +54,58 @@ defmodule Notifications.Notification do
   ]
 
   @doc """
+  Creates a new detour expiration notification
+
+  ## Example
+  ### Creating a notification
+      iex> %Skate.Detours.Db.Detour{id: detour_id} = detour = Skate.Factory.insert(:detour)
+      iex>
+      iex> {:ok, result} = Notifications.Notification.create_detour_expiration_notification(%{
+      ...>   detour: detour,
+      ...>   expires_in: Duration.new!(minute: 30),
+      ...>   estimated_duration: "1 hour",
+      ...>   notification: %{}
+      ...> })
+      iex>
+      iex> %Notifications.Db.DetourExpiration{
+      ...>   detour_id: ^detour_id,
+      ...>   notification: %Notifications.Db.Notification{}
+      ...> } = result
+
+  ### Creating a notification and backdating the notification `created_at` time
+      iex> %Skate.Detours.Db.Detour{id: detour_id} = detour = Skate.Factory.insert(:detour)
+      iex>
+      iex> created_at =
+      ...> DateTime.utc_now()
+      ...> |> DateTime.shift(minute: 30)
+      ...> |> DateTime.to_unix()
+      iex>
+      iex> {:ok, result} = Notifications.Notification.create_detour_expiration_notification(%{
+      ...>   detour: detour,
+      ...>   expires_in: Duration.new!(minute: 30),
+      ...>   estimated_duration: "1 hour",
+      ...>   notification: %{
+      ...>     created_at: created_at
+      ...>   }
+      ...> })
+      iex>
+      iex> %Notifications.Db.DetourExpiration{
+      ...>   detour_id: ^detour_id,
+      ...>   notification: %Notifications.Db.Notification{created_at: ^created_at}
+      ...> } = result
+  """
+  def create_detour_expiration_notification(params) when is_map(params) do
+    create_detour_expiration_notification(Map.get(params, :detour), params)
+  end
+
+  def create_detour_expiration_notification(%Skate.Detours.Db.Detour{} = detour, params) do
+    detour
+    |> Ecto.build_assoc(:detour_expiration_notifications)
+    |> Notifications.Db.DetourExpiration.changeset(params)
+    |> Skate.Repo.insert()
+  end
+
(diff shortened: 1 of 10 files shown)


lib/notifications/notification.ex
```diff
@@ -53,6 +53,58 @@ defmodule Notifications.Notification do
...
+      iex> {:ok, result} = Notifications.Notification.create_detour_expiration_notification(%{
+      ...>   detour: detour,
+      ...>   expires_in: Duration.new!(minute: 30),
+      ...>   estimated_duration: "1 hour",
+      ...>   notification: %{}
+      ...> })
+      iex>
+      iex> %Notifications.Db.DetourExpiration{
+      ...>   detour_id: ^detour_id,
+      ...>   notification: %Notifications.Db.Notification{created_at: created_at}
```
meagharty, line 72: There's a warning from this line about `created_at` variable being unused

```
warning: variable "created_at" is unused (if the variable is not meant to be used, prefix it with an underscore)
└─ (for doctest at) lib/notifications/notification.ex:72: Notifications.NotificationTest."doctest Notifications.Notification.create_detour_expiration_notification/1 (1)"/1
```

meagharty (approved):
Small fix-ups / suggestions that don't require re-review from my perspective

This looks good to me with the ADR changes & the doctests 👍

PR #3012, opened 2025-06-09
Open Detour Modal when Notification is Clicked

meagharty:
Asana Ticket: [Open Detour Modal when Notification is Clicked](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1209653841420232?focus=true)

See comments below. 

This makes understanding the NotificationCard component easier by separating out the business logic by domain and resolves the read state bug with detour notifications.

### DIFF
--- a/assets/src/components/notificationCard.tsx
+++ b/assets/src/components/notificationCard.tsx
@@ -13,8 +13,8 @@ import { DetourNotificationCard } from "./notificationCards/detourNotificationCa
 interface NotificationCardProps {
   notification: Notification
   currentTime: Date
-  setNotificationRead: (notification: Notification) => void
-  setNotificationSelected: (notification: Notification) => void
+  onRead: (notification: Notification) => void
+  onSelect: (notification: Notification) => void
   hideLatestNotification?: () => void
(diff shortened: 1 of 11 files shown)


assets/src/components/notificationCard.tsx
```diff
@@ -57,276 +42,50 @@ export const NotificationCard = ({
...
-      <CardReadable
-        currentTime={currentTime}
-        title={<>{title(notification)}</>}
-        style="kiwi"
-        isActive={isUnread}
-        openCallback={() => {
-          if (notification.content.$type === NotificationType.Detour) {
-            setShowDetourModal(true)
-          } else {
-            openVPPForCurrentVehicle(notification)
```
meagharty, line 75: This function was setting the read state of the notification **and** setting the notification as the selected notification in state (which evidently opens the VPP for current vehicle for BlockWaivers).

Detour notifications currently aren't marked as read. This sort of bug lead me to separating out these notification types. The component `CardReadable` is still used by all notifications. Open to more iteration here.

I kept detour activated / deactivated notifications in the same component as detour expiration since they share a domain and are similar.

PR #3014, opened 2025-06-11
feat: Parse Detour Expiration Notification on the frontend

firestack:
Asana Ticket: [Parse Detour Expiration Notification on the frontend](https://app.asana.com/1/15492006741476/project/1203014709808707/task/1210353029759135?focus=true)

Related-to: #3011 

> [!NOTE]
> This PR has one commit from #3011 since it's required for the functionality to work.

This adds superstruct definitions for the expiration notifications and also hides these notifications if not in the `detour-expiration-notifications` test group.

I decided to go with encoding the `expiresIn` field in the ISO8601 duration format for flexibility and also to collapse the amount of data that would need to be sent as JSON. I'm open to chaining this implementation if preferred though. I considered sending exactly what the front end requests (`0 | 30`), but _I_ think ISO8601 gets out of the way enough that if we relax the requirement of `0 | 30` in the future this would be more extensible.


assets/src/components/notificationCard.tsx
firestack, line 1: note: I expect this'll conflict with #3012 so I'll fix this up and rebase after. I can also inline the function instead of pulling it out.

firestack: Resolved with https://github.com/mbta/skate/compare/46fe114f7040bebd3055b3496350c934d2792ef6..d81547fe433707a502ac56593489f5d9525843fa

lib/notifications/notification.ex
```diff
@@ -379,6 +379,44 @@ defmodule Notifications.Notification do
   defp content_from_db_notification(%DbNotification{
          detour_expiration: %Notifications.Db.DetourExpiration{} = detour_expiration
        }) do
-    detour_expiration
+    # `Jason` doesn't know how to encode a `%Duration{}`, or tuples
```
meagharty, line 382: Did you consider [implementing an encoder](https://hexdocs.pm/jason/readme.html#encoders) for Duration?

assets/src/models/notificationData.ts
```diff
@@ -80,6 +81,39 @@ export const DetourNotificationData = type({
...
+// as the Regular Expression for ISO8601 Durations is _somewhat_ straightforward
+//
+// For now, we know we only have two valid values for `expiresIn` so this
+// matches the exact strings the ISO8601 Durations should take the form
+// of, and reduces the amount of total data that would otherwise be sent
+// if we encoded durations as objects, and makes this easier to switch
+// to Temporal in the future.
+const detourExpiresInFromISO8601Duration = coerce(
+  enums([0, 30]),
+  enums(["PT0S", "PT30M"]),
```
meagharty, line 95: I was expecting this to be already be in the form of "0 | 30" since we don't currently do any calculations on the frontend with the ISO8601 value (and we don't have Temporal or a library like it). Do you see that changing?

meagharty (approved):

lib/notifications/notification.ex
meagharty: As-is this works in terms of the coercing in these functions using `Duration.to_iso8601`, or if you decide to change it to `0 | 30` here, a change in these functions for that

assets/src/models/notificationData.ts
firestack: Addressed in [`7986718` (#3014)](https://github.com/mbta/skate/pull/3014/commits/79867187309b4cc99be22b039d45df5fda4cb2b7)

Thanks for your feedback!

PR #3015, opened 2025-06-12
test: clicking a notification card marks it as read

meagharty:
Asana Ticket: [Mark Notification as Read when Clicked](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210353029759140?focus=true)

I looked briefly at changing the role of these readable Cards to something other than button, like dialog (while keeping the buttons as buttons). That change affected a lot of other code paths and tests though, so I didn't pursue it further.

### DIFF
--- a/assets/tests/components/notificationCard.test.tsx
+++ b/assets/tests/components/notificationCard.test.tsx
@@ -12,6 +12,7 @@ import {
   blockWaiverNotificationFactory,
   bridgeLoweredNotificationFactory,
   bridgeRaisedNotificationFactory,
+  detourActivatedNotificationContentFactory,
   detourActivatedNotificationFactory,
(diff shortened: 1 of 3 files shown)


assets/tests/components/notificationCard.test.tsx
```diff
@@ -394,4 +404,67 @@ describe("NotificationCard", () => {
       }
     }
   )
+
+  test.each<{
```
meagharty, line 408: We could refactor these tests to be organized file-wise by notification content.

I already spent some time refactoring the content types to clarify that code, given the existing differences in behavior. Refactoring the tests feels more like the pursuit of cleanliness than clarity so I'm thinking we can defer this.

PR #3023, opened 2025-06-26
feat: Use http_stage library

arkadyan:
This replaces our almost identical local implementation for a GenStage provider that fetches HTTP resources.

Data is now received as parsed JSON rather than binary.

No Asana ticket. Up to you if you want to take this or leave it.

Example of where we transitioned Alerts Viewer to use this library: https://github.com/mbta/alerts_viewer/pull/236

In DRAFT mode because it could probably use more testing for insurance. Running it locally, it appears that we are successfully retrieving and parsing data. I'm seeing occasional errors, but they seem unrelated (I'm seeing them on `main` as well).

---
  - https://app.asana.com/0/0/1212078175864597

### DIFF
--- a/lib/concentrate/pipeline/vehicle_positions_pipeline.ex
+++ b/lib/concentrate/pipeline/vehicle_positions_pipeline.ex
@@ -42,10 +42,12 @@ defmodule Concentrate.Pipeline.VehiclePositionsPipeline do
           headers: %{
             "Authorization" => opts[:swiftly_authorization_key]
           },
-          params: %{
-            unassigned: "true",
-            verbose: "true"
-          }
+          get_opts: [
+            params: %{
+              unassigned: "true",
+              verbose: "true"
+            }
+          ]
         )
       else
         nil


lib/concentrate/producer/http/state_machine.ex
```diff
@@ -1,392 +0,0 @@
...
-  * `fallback_url`: (optional) URL to use once the `content_warning_timeout` is hit.
-  * `get_opts`: (optional) values to pass as options to HTTPoison
-  """
-  require Logger
-  @default_timeout 15_000
-  @default_fetch_after 5_000
-  @default_content_warning_timeout 600_000
-
-  defstruct url: "",
-            get_opts: [
```
meagharty: RE: `get_opts` 

The default receive timeout for `Req` seems fine as a replacement here, it's still `15_000`.

It looks like the pool checkout timeout is a separate configuration and default value (`:pool_timeout`, `5000`) but I'm not convinced we need to override the default pooling configuration.

Pooling information for Req: https://hexdocs.pm/req/Req.Steps.html#run_finch/1-http-1-pools

PR #3040, opened 2025-07-11
fix: restore logging for block waiver notifications

meagharty:
Asana Ticket: [Add logging for block waiver notifications](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210775001583863?focus=true)

I pulled this out of Kayla's work in progress branch for block waivers so we can unblock deploys to production.

### DIFF
--- a/lib/notifications/notification.ex
+++ b/lib/notifications/notification.ex
@@ -155,6 +155,8 @@ defmodule Notifications.Notification do
          {:ok,
           %Notifications.Db.BlockWaiver{
             reason: reason,
+            block_id: block_id,
+            service_id: service_id,
             route_ids: route_ids,
             run_ids: run_ids,
             trip_ids: trip_ids,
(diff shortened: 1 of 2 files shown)


lib/notifications/notification.ex
```diff
@@ -151,9 +151,35 @@ defmodule Notifications.Notification do
...
+            route_ids: route_ids,
+            run_ids: run_ids,
+            trip_ids: trip_ids,
+            start_time: start_time,
+            end_time: end_time,
+            notification: %{created_at: created_at}
+          }}
+       ),
+       do:
+         "result=notification_created" <>
```
meagharty, line 169: There is a change from `new_notification` (currently seen in prod logging) to `result=notification_created` (seen now in skate-dev-blue logging). I can add a subtask to make sure the prod alert is updated.

PR #3049, opened 2025-07-28
Revert revert block waiver changes for notifications

meagharty:
Asana Ticket: [🐞 Follow up on BlockWaiver changes](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210812417103597?focus=true)

Resolves `(Ecto.ConstraintError) constraint error when attempting to insert struct: notifications_users_pkey` if a user has multiple route tabs with overlapping block waiver routes

### DIFF
--- a/lib/skate/settings/user.ex
+++ b/lib/skate/settings/user.ex
@@ -71,12 +71,7 @@ defmodule Skate.Settings.User do
   end
 
   def list_users_with_route_ids(route_ids) do
-    Skate.Repo.all(from(u in users_for_route_ids_query(route_ids), select: [:id]))
-  end
-
-  @spec user_ids_for_route_ids([String.t()]) :: [DbUser.id()]
-  def user_ids_for_route_ids(route_ids) do
-    Skate.Repo.all(from(u in users_for_route_ids_query(route_ids), select: u.id))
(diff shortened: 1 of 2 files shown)


lib/notifications/notification.ex
```diff
@@ -266,51 +297,6 @@ defmodule Notifications.Notification do
...
-               db_record
-             else
-               _ -> Skate.Repo.rollback(nil)
-             end
-           end) do
-        {:ok, db_record} when not is_nil(db_record) ->
-          db_record
-
-        {:error, _} ->
-          Skate.Repo.one!(
```
meagharty, line 297: I'm not sure if changes are still needed for `get`, in this case on `Repo.rollback`.

It's possible we receive real-time updates to existing block waivers and we still want to return something. But this doesn't return the new data or commit the new data, so that doesn't seem too valuable? Same for `on_conflict: :nothing` above on `insert/2`

I'm going to see if any new errors come up as this stews on skate-dev-green.

meagharty: The issue, fixed in this PR, prevented a notification from ever being created (the observed bug was missing notifications). The different issue that I described in the last comment is a new logging. It's logged at the "warn" level. 

I'm wondering if the re-processing of block waiver data would be better resolved in a different part of the application, before the create notification function is called. For now we could move forward with these notification changes as-is.

PR #3056, opened 2025-08-20
feat: SOC and Miles Remaining on the VPP

bfauble:
Asana Ticket: [SOC and Miles Remaining on the VPP](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1210774132483625?focus=true)

image
image

currently deployed on dev-blue, currently no live data until https://github.com/mbta/busloc/pull/433 is merged in so will always display unknown info. 

Can mimic live data by using the vehicle factory in assets/src/models/ladderVehicles.ts:59 with
```
const myvehicle = vehicleFactory.build({
    routeId: "1",
    stateOfCharge: { value: 80, time: new Date(), milesRemaining: 160 },
})

const vehiclesOnLadder: VehicleOnLadder[] = [
    ...vehiclesAndGhosts,
    myvehicle,
  ].map((vehicleOrGhost) =>
```
and update assets/src/components/propertiesPanel.tsx:60 to
```
(liveVehicle !== null || isLoggedOut(mostRecentVehicle)) ? (
```
 then you should be able to select the vehicle on the route ladder to open the vpp.
 
 Will need to create the new test group "state-of-charge" and enable in admin to view the battery info panel.

### DIFF
--- a/assets/src/models/vehicleData.ts
+++ b/assets/src/models/vehicleData.ts
@@ -130,7 +130,7 @@ const baseVehicleData = {
   end_of_trip_type: EndOfTripTypeData,
   block_waivers: array(BlockWaiverData),
   crowding: nullable(CrowdingData),
-  state_of_charge: optional(nullable(StateOfCharge)),
+  state_of_charge: nullable(StateOfCharge),
 }
 
 export const VehicleData = type(baseVehicleData)
--- a/lib/concentrate/parser/gtfs_realtime_enhanced.ex
+++ b/lib/concentrate/parser/gtfs_realtime_enhanced.ex
@@ -153,21 +153,18 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
   end
 
   defp decode_state_of_charge(%{
-         "state_of_charge_percentage" => nil,
-         "state_of_charge_timestamp" => nil
-       }),
-       do: nil
+         "state_of_charge_percentage" => pct,
+         "state_of_charge_timestamp" => unix
+       })
+       when is_integer(pct) and is_integer(unix) do
+    %{
+      value: pct,
+      time: unix
+    }
+  end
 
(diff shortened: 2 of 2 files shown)


assets/src/models/vehicleData.ts
```diff
@@ -130,7 +130,7 @@ const baseVehicleData = {
   end_of_trip_type: EndOfTripTypeData,
   block_waivers: array(BlockWaiverData),
   crowding: nullable(CrowdingData),
-  state_of_charge: optional(StateOfCharge),
+  state_of_charge: optional(nullable(StateOfCharge)),
```
paulswartz, line 133: question: I would have though that this would be the same as `crowding` on line 132. Can you say more about why this one needs to be both nullable and optional?

lib/concentrate/parser/gtfs_realtime_enhanced.ex
```diff
@@ -152,11 +152,22 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
...
+  defp decode_state_of_charge(vp) do
+    value = Map.get(vp, "state_of_charge_percentage")
+    time = Map.get(vp, "state_of_charge_timestamp")
+
+    if value && time do
+      %{
+        value: Map.get(vp, "state_of_charge_percentage"),
+        time: Map.get(vp, "state_of_charge_timestamp")
+      }
+    end
```
paulswartz, line 170: suggestion: I think this might be backwards? I think you'd want the default case to be where both values are present, and the fallback to be `nil`:
```elixir
defp decode_state_of_charge(
  %{
    "state_of_charge_percentage" => pct, 
    "state_of_charge_timestamp" => unix
  }) when is_integer(pct) and is_integer(unix) do
  %{
    value: pct
    time: unix
  }
end
defp decode_state_of_charge(_vp) do
  nil
end
```

PR #3058, opened 2025-08-25
feat: Filter Battery Info by Allow List

bfauble:
Asana Ticket: [Filter Battery Info by Allow List](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1211138824486400?focus=true)

### DIFF
--- a/assets/src/components/propertiesPanel/stateOfChargeInfo.tsx
+++ b/assets/src/components/propertiesPanel/stateOfChargeInfo.tsx
@@ -8,22 +8,30 @@ import {
   Battery81to100Icon,
   BatteryUnknownIcon,
 } from "../../helpers/icon"
-import { StateOfCharge } from "../../models/stateOfCharge"
+import {
+  StateOfCharge,
+  StateOfChargeMissing,
+  StateOfChargeUnknown,
+} from "../../models/stateOfCharge"
 import useCurrentTime from "../../hooks/useCurrentTime"
 import { timeAgoLabelFromDate } from "../../util/dateTime"
 
-const getIcon = (stateOfCharge: StateOfCharge | null) => {
+const getIcon = (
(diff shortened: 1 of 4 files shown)


paulswartz (commented):
question: is there a reason this should be in Skate rather than Busloc? For example, we do this for assignment information: https://github.com/mbta/busloc/blob/master/config/config.exs#L69-L70

paulswartz:
If we're still testing the functionality, is it worth doing this now at all? It seems like we could come back to this when we're closer to showing BEB data to inspectors/dispatchers (and put it in Busloc at that point).

lib/concentrate/parser/gtfs_realtime_enhanced.ex
```diff
@@ -152,21 +165,30 @@ defmodule Concentrate.Parser.GTFSRealtimeEnhanced do
     Map.get(vp, "occupancy_status")
   end
 
-  defp decode_state_of_charge(%{
-         "state_of_charge_percentage" => pct,
-         "state_of_charge_timestamp" => unix
-       })
+  defp decode_state_of_charge(
```
paulswartz, line 168: suggestion: based on our conversation, I think this needs to include the case where the vehicle_id is in `@soc_vehicle_allow_list` but there's not SOC information, and return `%{time: nil, value: nil}`. Since the list is hardcoded, you could do something like
```elixir
defp decode_state_of_charge(vp, vehicle_id) when vehicle_id in @soc_vehicle_allow_list do
  case vp do
    %{
         "state_of_charge_percentage" => pct,
         "state_of_charge_timestamp" => unix
       } when is_integer(pct) and is_integer(unix) ->
        %{value: pct, time: unix}
    _ -> %{value: nil, time: nil}
  end
end

defp decode_state_of_charge(_vp, _vehicle_id) do
  nil
end
```

paulswartz (commented):
suggestion: could also use a test for the "BEB but no SOC data" case.

PR #3168, opened 2026-04-06
feat: Add custom date detour duration

stevejrowe:
Asana Ticket: [Detour Duration | Add Date as Duration Option](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1213232885542016?focus=true)

* Decided to not relay the time as a duration like in the story - it would be messy.
  * Instead, selected dates are sent and stored as ISO8601 date strings
  * formatted dates could be parsed pretty easily in elixir with Timex, but would have been messy to convert back to JS date objects without a date library on the front end.
* Flatpickr was not really operable with bootstraps modal auto-focus - they were not functioning in a helpful way anyway, so I turned that option off with `enforceFocus={false}` until we get around to fixing it.
* When a person selects the date select, there is now a state where selectedDuration is undefined - so I updated some state and types to reflect that.


lib/skate/detours/detours.ex
```diff
@@ -810,17 +810,25 @@ defmodule Skate.Detours.Detours do
...
+        |> String.split()
+        |> Enum.at(0)
+        |> String.to_integer()
+
+      detour
+      |> Map.get(:activated_at)
+      |> DateTime.add(hours, :hour)
+    end
+
+    if duration =~ ~r/\d{4}-\d{2}-\d{2}/ do
```
bfauble: should this be a separate branch of the above if statement? if line 8916 evaluates to true, this seems like it would always evaluate to nil and this function would return nil since the last line is what is implicitly returned.

Related, can you add some tests around this calculation with the new date input.

stevejrowe: I updated the conditional to be more explicit and added some tests 👍

assets/src/util/dateTime.ts
```diff
@@ -146,3 +169,7 @@ export const isUpdatedAfterActivated = ({
     dateFromEpochSeconds(updatedAt).valueOf() > activatedAt.valueOf() + bufferMs
   )
 }
+
+export const toIsoDateString = (date: Date) => {
+  return date.toISOString().split("T")[0]
```
bfauble: is this going to cause any issues with this returning a UTC timestamp and potentially pushing the time into a different day inadvertently?

stevejrowe: Oops, yes it would. I'll fix this.

stevejrowe: [e0215a9](https://github.com/mbta/skate/pull/3168/commits/e0215a9f32dd64b3a3f53103dd5c4afb295805a8)
Turns out flatpickr has a little date formatter that can be used independently, and it works in this scenario.

```
const d = new Date('2026-04-23T01:00:00.000Z')
flatpickr.formatDate(d, "Y-m-d")
> '2026-04-22'
```

bfauble (approved):

PR #3172, opened 2026-04-21
feat: text only - draw detour panel

stevejrowe:
Asana Ticket: [Text Only Detours | Draw Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1214013307977447?focus=true)

Also consolidated a duplicate userInTestGroup import in `assets/src/components/detours/diversionPage.tsx`


assets/src/models/createDetourMachine.ts
```diff
@@ -803,6 +808,9 @@ export const createDetourMachine = setup({
                 },
               },
             },
+            "Type Detour": {
+              // state for typing directions
```
bfauble, line 812: Just want to make sure I'm seeing the right endpoint of the implementation so far. Should just show the button and send the action to get to this state, but then the text input etc will be added in a follow up ticket, correct? 
,

stevejrowe: That's correct 👍

bfauble (approved):

PR #3179, opened 2026-04-27
chore(deps): Remove UUID dependency

stevejrowe:
After merging in https://github.com/mbta/skate/pull/3177, I realized we shouldn't need UUID. There is a native method to generate v4 UUID in Javascript.

Browser support: https://caniuse.com/mdn-api_crypto_randomuuid

### DIFF
--- a/assets/tests/setup.tsx
+++ b/assets/tests/setup.tsx
@@ -2,6 +2,7 @@
 // @ts-nocheck
 import failOnConsole from "jest-fail-on-console"
 import React from "react"
+import { randomUUID } from "node:crypto"
 
 const ignoreList = [/.*components\/app\.test\.tsx/]
 const ignoreNameList = []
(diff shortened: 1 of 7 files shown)


assets/tests/setup.tsx
```diff
@@ -57,6 +58,14 @@ document.createElementNS = function (namespaceURI, qualifiedName) {
   return element
 }
 
+Object.defineProperty(global, "crypto", {
+  value: {
+    randomUUID: randomUUID,
+  },
+  writable: true,
+  configurable: true,
```
bfauble: are these two props needed? As far as I can tell they're for re-assigning the prop after this initial assignment.

stevejrowe: Fixed this, thanks!

bfauble (approved):

PR #3199, opened 2026-05-28
Fix: Vehicles missing from route ladder

stevejrowe:
Asana Ticket: [Vehicles missing from route ladder](https://app.asana.com/1/15492006741476/project/1148853526253420/task/1205857827225827?focus=true)

I'd like to make sure _this_ doesn't mess with performance in any way.

### DIFF
--- a/assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
+++ b/assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
@@ -64,13 +64,13 @@ const useRouteVehicles = (
     | (VehicleInScheduledService | Ghost)[]
     | undefined = routeId === null ? undefined : vehiclesByRouteId[routeId]
   // If we haven't already fetched this route, open a new channel.
-  const { socket } = useContext(SocketContext)
-  const newVehiclesAndGhosts: (VehicleInScheduledService | Ghost)[] | null =
-    useVehiclesForRoute(
-      socket,
-      existingVehiclesAndGhosts === undefined ? routeId : null
-    )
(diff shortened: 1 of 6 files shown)


assets/src/components/propertiesPanel/vehiclePropertiesPanel.tsx
```diff
@@ -63,14 +61,8 @@ const useRouteVehicles = (
   const existingVehiclesAndGhosts:
     | (VehicleInScheduledService | Ghost)[]
     | undefined = routeId === null ? undefined : vehiclesByRouteId[routeId]
-  // If we haven't already fetched this route, open a new channel.
-  const { socket } = useContext(SocketContext)
-  const newVehiclesAndGhosts: (VehicleInScheduledService | Ghost)[] | null =
-    useVehiclesForRoute(
-      socket,
-      existingVehiclesAndGhosts === undefined ? routeId : null
```
bfauble, line 71: were you able to find any other 'pages' where this component is being used? This line makes me think that there's a non-ladder entry point that doesn't necessarily have the `routeId` present yet and that's why it's effectively doing a fresh fetch with the current filters

stevejrowe: Yup, although it isn't much use in this one case, but is a good example of this case: they can open up a vehicle from the map view - so this was originally needed to make the minimap display correctly.

The idea is that in cases like that, where they select a vehicle not in the route ladder, selectedVehicleOrGhost in app.tsx opens the socket instead.

bfauble (approved):

PR #3200, opened 2026-06-01
feat: Text only detours - type detour panel

stevejrowe:
Asana Ticket: [Text Only Detours | Type Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1214013307977452?focus=true)

### Styling updates
1. hide valid form validation styling
2. add monospace font for read-only forms.

### New front end features
1. Add disabled state to map
2. Floating text alert over map
3. Consolidate use of Directions, ConnectionPoints, and MissedStops components
4. Add TypeDetourPanel

### Backend
1. Add `is_text_only` (bool) field to db and SimpleDetour type (list view)
2. Add `typed_detours` (map) field to db
3. Only update swiftly if is_text_only is false


assets/css/_diversion_page.scss
```diff
@@ -123,10 +123,18 @@
     overflow-y: auto;
   }
 
-  .form-control {
```
lubeso, line 126: the `.form-control` was split into `.tall-input`, `.medium-input`, and `.small-input`?

would consider consolidating them like so?:

```suggestion
  .tall-input,
  .medium-input,
  .small-input {
```

lubeso (approved):
🥳

assets/css/_diversion_page.scss
stevejrowe: They each should have different min-heights, so combining the selectors doesn't really work here. Let me know if I'm misunderstanding though.

PR #3223, opened 2026-07-10
fix: Text only detours - style updates

stevejrowe:
Asana Ticket: [Text Only Detours | Type Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1214013307977452?focus=true)

https://mbta.slack.com/archives/C09RWAP5ZHP/p1783538654278689
2- Text only banner breaks the side panel (alerts are too wide by default)
3- Showing regular route alert has tiny text (regular font size missing in leaflet container)
4- can we had a hair more space between the black dashes? (updated dash-array)
5- The warning banner should be Inter first like the other things/banners. (updated leaflet-container to be inter by default)
6- It seems to be missing the hint text and has weird spacing under the input fields.(Feedback must be a sibling of Control + autosized field had extraneous margin-bottom)

* Black border around the whole frame on focus (pre-existing bug, not fixed)

### DIFF
--- a/assets/tests/components/detours/detoursListPage.openDetour.test.tsx
+++ b/assets/tests/components/detours/detoursListPage.openDetour.test.tsx
@@ -68,8 +68,8 @@ describe("Detours Page: Open a Detour", () => {
     // Click an arbitrary detour from the list
     await userEvent.click(await screen.findByText("Headsign Z"))
     await screen.findByText("Base Map")
-    // Render modal based on mocked value, which is a detour-in-progress
 
+    // Render modal based on mocked value, which is a detour-in-progress
     expect(viewDraftDetourHeading.get()).toBeVisible()
 
     // Finally, check snapshot
--- a/assets/tests/components/detours/detoursListPage.openDetour.test.tsx
+++ b/assets/tests/components/detours/detoursListPage.openDetour.test.tsx
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, jest, test } from "@jest/globals"
 
 import React from "react"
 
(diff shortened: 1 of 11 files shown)


assets/tests/components/detours/detoursListPage.openDetour.test.tsx
```diff
@@ -67,12 +67,13 @@ describe("Detours Page: Open a Detour", () => {
 
     // Click an arbitrary detour from the list
     await userEvent.click(await screen.findByText("Headsign Z"))
-
+    await screen.findByText("Base Map")
     // Render modal based on mocked value, which is a detour-in-progress
+
```
lubeso, line 71: > in case the whitespace wasn't intentional

```suggestion
```

assets/tests/components/detours/detoursListPage.openDetour.test.tsx
```diff
@@ -67,12 +67,13 @@ describe("Detours Page: Open a Detour", () => {
 
     // Click an arbitrary detour from the list
     await userEvent.click(await screen.findByText("Headsign Z"))
-
+    await screen.findByText("Base Map")
```
lubeso, line 70: for my benefit, when does the 'base map' text appear? i was going through the `dev-blue` environment and i didn't see anything with that text pop up

lubeso (approved):
🎈

assets/tests/components/detours/detoursListPage.openDetour.test.tsx
stevejrowe: It's a hidden header
Image

PR #3224, opened 2026-07-13
feat: Text only detours - detours table

stevejrowe:
Asana Ticket: [Text Only Detours | Detours Table](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1213996836942138?focus=true)

* add "c-" prefix for the custom component class
* add white-space nowrap to shared class
* add text-only pill to table


lubeso (approved):
🥷 🎨 🖌️

PR #3227, opened 2026-07-16
feature: save active detours to s3 for lamp + alerts teams

lubeso:
Asana Ticket: [Save active detours to S3 for LAMP + Alerts Teams](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1214933171602153?focus=true)


lib/skate/detours/detours.ex
```diff
@@ -191,6 +191,44 @@ defmodule Skate.Detours.Detours do
   defp handle_detour_updated(changeset, new_record, author_id) do
     broadcast_detour(new_record, author_id)
     process_notifications(changeset, new_record)
+
+    # conditionally trigger s3 export job
+    s3_export_job_trigger =
```
stevejrowe: Can this use a pattern that is more consistent with the rest of the file? Specifically, Skate typically tries to have one purpose per function, so this could be `trigger_s3_export` or something like that.

Secondly, we can use [patterns and guards](https://elixir.hexdocs.pm/patterns-and-guards.html) for the changeset and the final detour to either trigger an effect or nil. (see update_swiftly and process_notifications as examples - they are also keying in on similar conditions)

lib/skate/detours/detours.ex
```diff
@@ -191,6 +191,44 @@ defmodule Skate.Detours.Detours do
...
+                nil
+            end
+          end
+      end
+
+    if !is_nil(s3_export_job_trigger) do
+      %{trigger: s3_export_job_trigger}
+      |> Skate.AlertsManager.ActiveDetours.S3Exporter.new()
+      |> Oban.insert()
+    end
```
stevejrowe: Going off of the previous request, instead of triggering this when the result isn't nil, how about make this a function, and call it using the value.

Pseudocode, but this pattern could be:
```
defp trigger_s3_export(pattern) do
  log(_your descriptive string_)
  do_s3_export()
end

defp do_s3_export() do
 Oban.insert(Skate.AlertsManager.ActiveDetours.S3Exporter.new())
end
```

lib/skate/alerts_manager/active_detours/s3_exporter.ex
```diff
@@ -0,0 +1,241 @@
+defmodule Skate.AlertsManager.ActiveDetours.S3Exporter do
```
stevejrowe, line 1: Why is this in AlertsManager?

lib/skate/alerts_manager/active_detours/s3_exporter.ex
```diff
@@ -0,0 +1,241 @@
...
+
+  def handle_telemetry_event(_, _, _, _), do: nil
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: args} = _) do
+    detours =
+      Detour
+      |> Ecto.Query.where(status: :active)
+      |> Ecto.Query.order_by(desc: :activated_at)
+      |> Skate.Repo.all()
```
stevejrowe, line 94: Thoughts on including this with the other common queries in lib/skate/detours/detours.ex

There are common queries and filters defined there.

lib/skate/alerts_manager/active_detours/s3_exporter.ex
```diff
@@ -0,0 +1,241 @@
...
+      :error -> {:error, :missing_s3_bucket}
+      # aws s3 operation failed
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  @doc """
+  Convert a detour into a map.
+  """
+  @spec convert!(Detour.t()) :: map()
```
stevejrowe, line 139: There are different Detour struct variants defined in lib/skate/detours/detour.ex

I think we could define a Report Detour struct there, similar to how it was done here: https://github.com/mbta/skate/pull/3187/changes#diff-610f0f43343dd249fae1b02cf9660c717fd1abc48663129ebbaeccd701788693R166

stevejrowe (changes requested):

lib/skate/alerts_manager/active_detours/s3_exporter.ex
lubeso: i personally don't like that established practice of defining common queries and filters, so i'd rather leave this snippet as it is...

lib/skate/detours/detours.ex
lubeso: i purposely avoided that practice, because i don't like having many private function definitions... however, i'll abide by the suggestion

lib/skate/detours/detours.ex
lubeso: i'll move this into its own function

lib/skate/alerts_manager/active_detours/s3_exporter.ex
lubeso: it seemed reasonable since Alerts are a consumer of this information

do you have a different suggestion? so i know where to move the code to

lib/skate/alerts_manager/active_detours/s3_exporter.ex
lubeso: yes i saw that implementation

i'd rather not define yet another detour struct, especially since the map representation this function builds is specific to this s3 export

what are the benefits of defining a new struct?

lib/skate/alerts_manager/active_detours/s3_exporter.ex
stevejrowe: skate/detours

lib/skate/detours/s3_exporter.ex
```diff
@@ -0,0 +1,241 @@
...
+      :error -> {:error, :missing_s3_bucket}
+      # aws s3 operation failed
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  @doc """
+  Convert a detour into a map.
+  """
+  @spec convert!(Detour.t()) :: map()
```
stevejrowe: It could be nice to have definitions of all of our detour types in one place

lubeso: i tried, but i'm having trouble coming up with a good way to define the detour struct/s in that file...

if (so far) the only place this struct is referenced is in the `lib/skate/detours/s3_exporter.ex` file, then it doesn't feel right to instead define the struct in `lib/skate/detours/detour.ex` and alias it instead...

the other challenge i'm facing is that the struct fields are different based on whether the detour is text-only... in an OOP language, i could just use inheritance/composition, but i'm not sure how to accomplish a similar outcome in elixir?

would you be able to try and provide an implementation i can start from?

thanks for your patience...

stevejrowe: Sure! So the struct itself doesn't bother with inheritance or defining when a property is optional. It also uses the patterns in this file for mapping data to to a struct. It' retains your error raising on missed_stops and connection_points, but instead of a pipeline of merging objects, it uses a base_report private function to define the common fields and establish the shared struct.

In detour.ex:

```
defmodule Skate.Detours.Detour do
  # ...
  alias Skate.Detours.Db.Detour
  # ...

  defmodule Report do
    @type t :: %__MODULE__{
            id: integer(),
            route_id: String.t(),
            reason: String.t(),
            nearest_intersection: String.t() | nil,
            estimated_duration: String.t(),
            activated_at: integer(),
            updated_at: integer(),
            direction_id: integer(),
            missed_stops: [String.t()] | nil,
            connection_points: [String.t()] | nil,
            missed_stops_text_only: String.t() | nil,
            connection_points_text_only: String.t() | nil,
            route_segments:
              %{
                before_detour: [Util.Location.From.t()],
                after_detour: [Util.Location.From.t()],
                bypassed_segment: [Util.Location.From.t()],
                detour_segment: [Util.Location.From.t()]
              }
              | nil
          }

    @derive Jason.Encoder
    defstruct [
      :id,
      :route_id,
      :reason,
      :nearest_intersection,
      :estimated_duration,
      :activated_at,
      :updated_at,
      :direction_id,
      :missed_stops,
      :connection_points,
      :missed_stops_text_only,
      :connection_points_text_only,
      :route_segments
    ]

    @spec from(Detour.t()) :: t()
    def from(%Detour{is_text_only: true} = detour) do
      %{
        base_report(detour)
        | missed_stops_text_only: get_in(detour.state, ["context", "typedDetour", "missedStops"]),
          connection_points_text_only:
            get_in(detour.state, ["context", "typedDetour", "connectionPoints"])
      }
    end

    def from(%Detour{is_text_only: false} = detour) do
      %{
        base_report(detour)
        | missed_stops: missed_stops!(detour),
          connection_points: connection_points!(detour),
          route_segments: route_segments(detour)
      }
    end

    defp base_report(%Detour{} = detour) do
      %__MODULE__{
        id: detour.id,
        route_id: detour.route_id,
        direction_id: get_in(detour.state, ["context", "routePattern", "directionId"]),
        reason: detour.reason,
        nearest_intersection: detour.nearest_intersection,
        estimated_duration: detour.estimated_duration,
        activated_at: naive_to_unix!(detour.activated_at),
        updated_at: naive_to_unix!(detour.updated_at)
      }
    end

    defp naive_to_unix!(naive_dt) do
      naive_dt
      |> DateTime.from_naive!("Etc/UTC")
      |> DateTime.to_unix()
    end

    defp missed_stops!(%Detour{} = detour) do
      case get_in(detour.state, ["context", "finishedDetour", "missedStops"]) do
        missed_stops when is_list(missed_stops) ->
          missed_stops
          |> Enum.map(&get_in(&1, ["id"]))
          |> Enum.reject(&is_nil/1)

        _ ->
          Logger.warning("detour #{detour.id} has invalid missed stops")
          raise ArgumentError
      end
    end

    defp connection_points!(%Detour{} = detour) do
      case get_in(detour.state, ["context", "finishedDetour", "connectionPoint"]) do
        connection_points when is_map(connection_points) ->
          ["start", "end"]
          |> Enum.map(&get_in(connection_points, [&1, "id"]))
          |> Enum.reject(&is_nil/1)

        _ ->
          Logger.warning("detour #{detour.id} has invalid connection points")
          raise ArgumentError
      end
    end

    defp route_segments(%Detour{
           state: %{
             "context" => %{
               "finishedDetour" => %{
                 "routeSegments" => %{
                   "beforeDetour" => before_detour,
                   "afterDetour" => after_detour,
                   "detour" => detour_segment
                 },
                 "detourShape" => %{"coordinates" => bypassed_segment}
               }
             }
           }
         }) do
      %{
        before_detour: before_detour,
        after_detour: after_detour,
        bypassed_segment: bypassed_segment,
        detour_segment: detour_segment
      }
    end
  end
end
```

lib/skate/detours/s3_exporter.ex
```diff
@@ -0,0 +1,270 @@
+defmodule Skate.Detours.S3Exporter do
+  @moduledoc """
+  Export detour information to S3.
+  """
+
+  require Logger
+
+  use Oban.Worker, max_attempts: 3
```
stevejrowe, line 8: Nice, agree with limiting the max_attempts

lib/skate/detours/s3_exporter.ex
```diff
@@ -0,0 +1,270 @@
...
+             # pass overrides to allow mocking aws during unit tests
+             Application.get_env(:ex_aws, :request_config_overrides, %{})
+           ) do
+      {:ok, length(converted)}
+    else
+      # missing required configuration value for s3 bucket
+      :error -> {:error, :missing_s3_bucket}
+      # aws s3 operation failed
+      {:error, reason} -> {:error, reason}
+    end
```
stevejrowe: This introduces a new error during local development. Can we skip running the job when the s3_bucket is not specified? That would also buy us an easy way to prevent the jobs if they ever give us trouble for some unknown reason.

lib/skate/detours/detour.ex
```diff
@@ -175,4 +175,140 @@ defmodule Skate.Detours.Detour do
...
+
+    defp base_report(%Detour{} = detour) do
+      %__MODULE__{
+        id: detour.id,
+        route_id: detour.route_id,
+        direction_id: get_in(detour.state, ["context", "routePattern", "directionId"]),
+        reason: detour.reason,
+        nearest_intersection: detour.nearest_intersection,
+        estimated_duration: detour.estimated_duration,
+        activated_at: naive_to_unix!(detour.activated_at),
```
stevejrowe: Sorry, I suggested bad code here. Unlike `updated_at`, `activated_at` is actually already DateTime type. So the dialyzer will pass when this is `DateTime.to_unix(detour.activated_at),`

stevejrowe (approved):
LGTM. Thank you for your flexibility and patience ❤️

PR #3229, opened 2026-07-20
fix: Text only detours - UX review feedback

stevejrowe:
Asana Ticket: [UX Review - View Past Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216722774255341?focus=true)

[UX Review - Detour Table](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216722671716555?focus=true)

--- 
1. The ::after text was Inter instead of monospace due to some bug with the way inherit works on readonly tags. Adds a css rule to overcome this.
2. Will now update the snapshots on typing.
3. Add a missing text-only alert in past detours.

### DIFF
--- a/assets/css/bootstrap/_extra_bespoke.scss
+++ b/assets/css/bootstrap/_extra_bespoke.scss
@@ -49,6 +49,10 @@
   font-family: $font-family-monospace;
 }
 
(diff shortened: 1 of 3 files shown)


vgu-mbta (approved):
Looks good to me! UX seems to match

PR #3231, opened 2026-07-21
fix: delete draft text only detour

stevejrowe:
Asana Ticket: [UX Review - Type Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216722774255339?focus=true)

note: the button will not appear unless the test_group `delete-draft-detours` is enabled


lubeso (approved):
🎯

PR #3238, opened 2026-07-23
fix: double text-only map alert

stevejrowe:
Asana Ticket: [UX Review - Type Detour Panel](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216722774255339?focus=true)

Feedback: https://mbta.slack.com/archives/C09RWAP5ZHP/p1784751753746549

### DIFF
--- a/assets/src/components/detours/diversionPage.tsx
+++ b/assets/src/components/detours/diversionPage.tsx
@@ -752,16 +752,14 @@ export const DiversionPage = ({
           )}
         </div>
         <div className="l-diversion-page__map position-relative">
-          {snapshot.context.isTextOnly &&
-            !snapshot.matches({ "Detour Drawing": "Share Detour" }) && (
-              <TextOnlyMapAlert />
-            )}
+          {snapshot.context.isTextOnly && <TextOnlyMapAlert />}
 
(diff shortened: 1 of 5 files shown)


assets/src/components/detours/diversionPage.tsx
```diff
@@ -751,11 +752,13 @@ export const DiversionPage = ({
           )}
         </div>
         <div className="l-diversion-page__map position-relative">
-          {snapshot.matches({ "Detour Drawing": "Share Detour" }) && (
+          {snapshot.context.isTextOnly ? (
+            <TextOnlyMapAlert />
+          ) : snapshot.matches({ "Detour Drawing": "Share Detour" }) ? (
```
vgu-mbta: given the multiple branches of logic, this may be more readable in its own function. But not a big deal

vgu-mbta (approved):

assets/src/components/detours/diversionPage.tsx
stevejrowe: Good point thanks, nested ternaries are never the best choice. I wanted to keep the logic inline though, so I just separated out the conditions.

PR #3240, opened 2026-07-27
Feat: Closed detours pagination api

vgu-mbta:
Asana Ticket: [Closed Detours Table | Backend Pagination Improvements ](https://app.asana.com/1/15492006741476/project/1216434230851197/task/1216326666772054?focus=true)

### DIFF
--- a/lib/skate_web/channels/detours_channel.ex
+++ b/lib/skate_web/channels/detours_channel.ex
@@ -48,6 +48,57 @@ defmodule SkateWeb.DetoursChannel do
   end
 
   @impl SkateWeb.AuthenticatedChannel
+  def handle_in_authenticated("paginate", %{"limit" => limit, "offset" => offset}, socket) do
+    with {:ok, limit} <- parse_integer(limit),
+         {:ok, offset} <- parse_integer(offset) do
+      detours = fetch_paginated_detours(socket, limit, offset)
+      {:reply, {:ok, %{data: detours}}, socket}
+    else
+      _ ->
+        {:reply, {:error, %{reason: :invalid_pagination}}, socket}
+    end
+  end
+
+  def handle_in_authenticated("paginate", _payload, socket) do
+    {:reply, {:error, %{reason: :invalid_pagination}}, socket}
+  end
+
+  defp parse_integer(value) when is_integer(value), do: {:ok, value}
+
+  defp parse_integer(value) when is_binary(value) do
+    case Integer.parse(value) do
+      {parsed, ""} -> {:ok, parsed}
+      _ -> :error
+    end
+  end
+
+  defp parse_integer(_), do: :error
+
+  defp fetch_paginated_detours(%{topic: "detours:active"} = socket, limit, offset) do
+    %{id: user_id} = Guardian.Phoenix.Socket.current_resource(socket)
+    Detours.detours_for_user(user_id, :active, limit, offset)
+  end
+
+  defp fetch_paginated_detours(%{topic: "detours:active:" <> route_id}, limit, offset) do
+    Detours.detours_for_route(route_id, :active, limit, offset)
+  end
+
+  defp fetch_paginated_detours(%{topic: "detours:past"}, limit, offset) do
+    Detours.detours_for_route("all", :past, limit, offset)
+  end
+
+  defp fetch_paginated_detours(%{topic: "detours:past:" <> route_id}, limit, offset) do
+    Detours.detours_for_route(route_id, :past, limit, offset)
+  end
+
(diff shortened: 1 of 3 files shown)


lib/skate_web/channels/detours_channel.ex
```diff
@@ -47,6 +47,58 @@ defmodule SkateWeb.DetoursChannel do
...
+  defp parse_integer(value) when is_integer(value), do: {:ok, value}
+
+  defp parse_integer(value) when is_binary(value) do
+    case Integer.parse(value) do
+      {parsed, ""} -> {:ok, parsed}
+      _ -> :error
+    end
+  end
+
+  defp parse_integer(_), do: :error
```
lubeso: (just a matter of preference)

```suggestion
  defp parse_integer(value) do
    case value do
      integer when is_integer(value) ->
        {:ok, integer}

      string when is_binary(value) ->
        case Integer.parse(string) do
          {parsed, _} -> {:ok, parsed}
          _ -> :error
        end

      _ -> :error
    end
  end
```

lib/skate_web/channels/detours_channel.ex
```diff
@@ -47,6 +47,58 @@ defmodule SkateWeb.DetoursChannel do
...
+    %{id: user_id} = Guardian.Phoenix.Socket.current_resource(socket)
+    Detours.detours_for_user(user_id, :active, limit, offset)
+  end
+
+  defp fetch_paginated_detours(%{topic: "detours:active:" <> route_id}, limit, offset) do
+    Detours.detours_for_route(route_id, :active, limit, offset)
+  end
+
+  defp fetch_paginated_detours(%{topic: "detours:past"}, limit, offset) do
+    Detours.detours_for_route("all", :past, limit, offset)
```
lubeso, line 93: why is this `Detours.detours_for_route` and not `Detours.detours_for_user` like in line 80?

(i'm not saying it should be... i'd like to clear my understanding)

test/skate_web/channels/detours_channel_test.exs
```diff
@@ -157,6 +157,129 @@ defmodule SkateWeb.DetoursChannelTest do
...
+    @tag :authenticated
+    test "paginates past detours with limit 10 and offset 1", %{socket: socket} do
+      now = DateTime.utc_now()
+
+      Enum.each(1..12, fn id ->
+        :detour
+        |> build()
+        |> deactivated
+        |> with_id(id)
+        |> with_updated_at(DateTime.add(now, -id, :minute))
```
lubeso: i understand what this line is doing, but would you consider using `i` instead of `id` for the variable/iterator variable?

i had to do a double take to understand why "negative id" (as opposed to $$-i$$ for example)

test/skate_web/channels/detours_channel_test.exs
lubeso, line 1: good testing strategy!

have you considered how might you parameterize the tests?

lubeso (commented):
great job! please review my comments

test/skate_web/channels/detours_channel_test.exs
vgu-mbta: Sure! Yes it gets a little confusing because the iterator i gets used for both the id and the minus i number of minutes

lib/skate_web/channels/detours_channel.ex
vgu-mbta: I'll check with product but I think it's based on this function earlier in the file. Lines 25-31. By default for closed detours I don't think it filters by user

```
 # Past
  @impl SkateWeb.AuthenticatedChannel
  def join_authenticated("detours:past", _message, socket) do
    SkateWeb.Endpoint.subscribe("detours:past")
    detours = Detours.detours_for_route("all", :past)
    {:ok, %{data: detours}, socket}
  end
```

test/skate_web/channels/detours_channel_test.exs
vgu-mbta: good question! I think there's definitely potential to parametrize the "input" list of detours and then matching against the ids of detours that are returned. If it's ok with you though I'll try to merge this PR so I can start testing the frontend with the backend. and then add more tests once they are both hooked up

lib/skate_web/channels/detours_channel.ex
lubeso: great thank you! makes sense to me

test/skate_web/channels/detours_channel_test.exs
lubeso: sounds good to me! this question was really just to probe the thought... please continue full speed ahead and we can consider this topic again later!

lubeso (approved):
🎯

PR #3242, opened 2026-07-30
chore(deps): Update storybook to v10

stevejrowe:
Asana Ticket: [Dependency - Storybook 7.5.3 to 10.2.17](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1213735939273254?focus=true)


lubeso (approved):
not all heroes wear capes... thank you for your service!

can you please reply in the thread with a comment explaining why you needed to add the `on*()` handlers in the `*.stories.tsx` files?

> just curious 💭

stevejrowe:
> not all heroes wear capes... thank you for your service!
> 
> can you please reply in the thread with a comment explaining why you needed to add the `on*()` handlers in the `*.stories.tsx` files?
> 
> > just curious 💭

The typing is stricter because on* callbacks were no longer implied in storybook 8. Even if we don't use them, we have to list them if they are required args.
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#implicit-actions-can-not-be-used-during-rendering-for-example-in-the-play-function

PR #3243, opened 2026-07-30
Feat: Calculate number of pages for pagination

vgu-mbta:
Asana Ticket: [Closed Detours Table | Backend Pagination Improvements](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216326666772054?focus=true)

Iterating on the previous PR https://github.com/mbta/skate/pull/3240

We need to know the # of pages in order to display the page range to the user. For example, if there are 140 detours and each page displays 10 detours, the user sees a range of page numbers they can click on: 1...14 (with some options in the middle)


test/skate_web/channels/detours_channel_test.exs
lubeso, line 1: what other options did/would you consider for the assertions (instead of hard-coded numbers)?

just probing the thought again!

lubeso (approved):

PR #3248, opened 2026-08-04
feat: Swiftly API timeout mitigations

stevejrowe:
Asana Ticket: [ID + implement Swiftly API timeout mitigations](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1216130621623936?focus=true)


stevejrowe:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

I was curious about this feature, but I'm not going to try again.

lib/skate/detours/db/detour.ex
```diff
@@ -22,6 +22,7 @@ defmodule Skate.Detours.Db.Detour do
 
     # When this detour was activated
     field :activated_at, :utc_datetime_usec
+    field :swiftly_id, :string
```
lubeso, line 25: is the `swiftly_id` attribute related to "When detour was activated"?

is there a more appropriate place within the `typed_schema` block to define the `swiftly_id` attribute?

what does the `swiftly_id` attribute represent?:

- just a foreign key identifier necessary when using Swiftly APIs
- represents some business/domain logic or event

lib/skate/detours/db/detour.ex
```diff
@@ -80,10 +81,22 @@ defmodule Skate.Detours.Db.Detour do
       })
 
     detour
-    |> change(%{activated_at: nil, estimated_duration: nil, reason: nil})
+    |> change(%{activated_at: nil, estimated_duration: nil, reason: nil, swiftly_id: nil})
     |> change(%{state: new_state})
```
lubeso, line 85: `activated_at`, `estimated_duration`, and `reason` are all related to each other...

is `swiftly_id` related to them too?

if not, consider having a separate line (similar to the other existing line with `new_state`):

```suggestion
    |> change(%{swiftly_id: nil})
    |> change(%{state: new_state})
```

lib/skate/detours/detours.ex
```diff
@@ -204,35 +204,34 @@ defmodule Skate.Detours.Detours do
   Update or insert a detour given a user id and a XState Snapshot.
   """
   def upsert_from_snapshot(author_id, %{} = snapshot) do
-    detour_changeset =
+    partial_changeset =
```
lubeso, line 207:
```suggestion
    detour_changeset_from_snapshot =
```

> consider which variable name conveys more meaning to the reader

priv/repo/async_migrations/20260716202203_backfill_detour_swiftly_id.exs
```diff
@@ -0,0 +1,65 @@
+defmodule Skate.Repo.Migrations.BackfillDetourSwiftlyId.MigratingSchema do
+  @moduledoc """
+  Detours database table schema frozen at this point in time.
+  """
+
+  use Skate.Schema
+
+  typed_schema "detours" do
+    # field :id, :integer, null: false
```
lubeso, line 9: is this commented line necessary?

priv/repo/async_migrations/20260716202203_backfill_detour_swiftly_id.exs
```diff
@@ -0,0 +1,65 @@
...
+  @batch_size 5
+  @throttle_ms 100
+
+  def up do
+    swiftly_id_map =
+      Skate.Detours.Detours.get_swiftly_adjustments()
+      |> then(fn {:ok, response} -> response end)
+      |> Map.new(fn response -> {String.to_integer(response.notes), response.id} end)
+
+    detour_ids = Map.keys(swiftly_id_map)
```
lubeso, line 31: matter of preference, but consider which is easier to read

```suggestion
    swiftly_adjustments =
      case Skate.Detours.Detours.get_swiftly_adjustments() do
        {:ok, adjustments} when is_list(adjustments) -> adjustments
        _ -> []
      end

    swiftly_adjustment_ids_by_detour_ids =
      swiftly_adjustments
      |> Enum.map(fn %{} = swiftly_adjustment ->
        with {:ok, serialized_detour_id} <- Map.fetch(swiftly_adjustment, :notes),
             {:ok, swiftly_id} <- Map.fetch(swiftly_adjustment, :id),
             {detour_id, _} <- Integer.parse(serialized_detour_id) do
          {detour_id, swiftly_id}
        else
          _ -> nil
        end
      end)
      |> Enum.reject(&is_nil/1)
      |> Map.new()

    detour_ids = Map.keys(swiftly_adjustment_ids_by_detour_ids)
```

lib/helpers.ex
```diff
@@ -45,4 +45,29 @@ defmodule Helpers do
...
+  ...>     0 ->
+  ...>       Process.put(:retry_n, 1)
+  ...>       {:error, :interrupt}
+  ...>
+  ...>     1 ->
+  ...>       :ok
+  ...>   end
+  ...> end
+  ...>
+  ...> Helpers.retry(fun, 2)
```
lubeso, line 63:
```suggestion
```

lib/helpers.ex
```diff
@@ -45,4 +45,29 @@ defmodule Helpers do
     |> Enum.filter(fn {_, v} -> f.(v) end)
     |> Map.new()
   end
+
+  @doc """
+  Run any given function `fun` again up to `n` times on {:error, _}
```
lubeso, line 50:
```suggestion
  Retry the function `fun` up to `n` times when `fun` returns `{:error, ...}`
```

lubeso: functions can also return just an `:error` atom (instead of a tuple) or raise an exception...

something to consider... but this helper is fine for now

lib/swiftly/api/service_adjustments.ex
```diff
@@ -186,7 +185,7 @@ defmodule Swiftly.API.ServiceAdjustments do
       |> Keyword.put_new(:createdBefore, DateTime.utc_now())
       |> Keyword.put_new(
         :createdAfter,
-        DateTime.new!(~D[2025-01-01], ~T[00:00:00], "America/New_York")
+        DateTime.new!(~D[2027-01-01], ~T[00:00:00])
```
lubeso, line 188: what are the potential side effects (if any) when removing the "America/New_York" (presumably timezone) positional argument?

why is January 1, 2027 correct instead of January 1, 2025? is there another way to get the first day of the next calendar year without hard-coding?

lib/swiftly/api/service_adjustments.ex
```diff
@@ -108,7 +107,7 @@ defmodule Swiftly.API.ServiceAdjustments do
 
   """
   @spec delete_adjustment_v1(Swiftly.Api.ServiceAdjustments.AdjustmentId.t(), keyword) ::
-          :ok | {:error, :bad_request | :not_found | :unknown}
+          {:ok, nil} | {:error, :bad_request | :not_found | :unknown}
```
lubeso, line 110: if delete operations do not have a body, just an atom is fine (to avoid updating the method definition)

```suggestion
          :ok | {:error, :bad_request | :not_found | :unknown}
```

test/swiftly/api/service_adjustments_test.exs
lubeso, line 1: you can revert the new `{:ok, nil}` changes if you follow my previous comment

lib/swiftly/api/fake_service_adjustments.ex
lubeso, line 1: you can revert the new `{:ok, nil}` changes if you follow my previous comment

lib/skate/detours/detours.ex
```diff
@@ -691,24 +681,27 @@ defmodule Skate.Detours.Detours do
...
+      )
+    end
+
+    with {:ok, response} <- Helpers.retry(retry_function, 5) do
+      adjustments =
+        response
         |> Map.get(:adjustments, [])
         |> Enum.filter(fn adjustment -> adjustment.feedId == service_adjustments_feed_id() end)
 
-      _ ->
```
lubeso, line 702: you changed the `case` statement to a `with`... but either way you should include a matching clause for the error state

lib/skate/detours/detours.ex
```diff
@@ -691,24 +681,27 @@ defmodule Skate.Detours.Detours do
...
-         ) do
-      {:ok, adjustments_response} ->
-        adjustments_response
+    retry_function = fn ->
+      adjustments_module.get_adjustments_v1(
+        Keyword.put(build_swiftly_opts(), :adjustmentTypes, "DETOUR_V0")
+      )
+    end
+
+    with {:ok, response} <- Helpers.retry(retry_function, 5) do
```
lubeso, line 690: matter of preference, but consider making the number of retries a function parameter (with a default value of 5, for example)

lib/skate/detours/db/detour.ex
```diff
@@ -80,10 +81,22 @@ defmodule Skate.Detours.Db.Detour do
...
+  end
+
+  def put_change_from_swiftly({:ok, nil}, changeset) do
+    put_change(changeset, :swiftly_id, nil)
+  end
+
+  def put_change_from_swiftly(:ok, changeset) do
+    changeset
+  end
+
```
lubeso, line 99: this function is used twice in the `lib/skate/detours/detours.ex` file

both cases are `with ...` statements, which are syntactic sugar for handling return values that follow the format `{:ok, ...}` for successes and `:error` or `{:error, ...}` for errors

therefore, consider this alternative:

```suggestion
  @spec put_change_from_swiftly(Ecto.Changeset.t(), map() | nil) :: Ecto.Changeset.t()
  def put_change_from_swiftly(changeset, %{} = swiftly_response) do
    put_change(changeset, Map.get(swiftly_response, :adjustmentId, nil))
  end

  def put_change_from_swiftly(changeset, _) do
    changeset
  end
```

lib/skate/detours/detours.ex
```diff
@@ -204,35 +204,34 @@ defmodule Skate.Detours.Detours do
...
-    detour_db_result
+  defp do_upsert_from_snapshot(changeset) do
+    changed_fields = if changeset.changes != %{}, do: Map.keys(changeset.changes), else: [:state]
+
+    Skate.Repo.insert(
+      changeset,
+      returning: true,
+      conflict_target: [:id],
+      on_conflict: {:replace, changed_fields}
+    )
```
lubeso, line 229: in the interest of:

* avoid defining a new private method just for the changed fields and insert result
* keep the same return value to avoid breaking anything
* refactor the logic to read more easily

```suggestion
    detour_changeset_from_snapshot =
      Skate.Detours.SnapshotSerde.deserialize(author_id, snapshot)

    detour = Ecto.Changeset.apply_changes(detour_changeset_from_snapshot)

    detour_changeset =
      case update_swiftly(detour_changeset_from_snapshot, detour) do
        {:ok, swiftly_response} when is_map(swiftly_response) ->
          Skate.Detours.Db.Detour.put_change_from_swiftly(
            swiftly_response,
            detour_changeset_from_snapshot
          )

        _ ->
          detour_changeset_from_snapshot
      end

    changed_fields =
      if detour_changeset.changes != %{} do
        Map.keys(detour_changeset.changes)
      else
        [:state]
      end

    upsert_result =
      Skate.Repo.insert(
        detour_changeset,
        returning: true,
        conflict_target: [:id],
        on_conflict: {:replace, changed_fields}
      )

    case upsert_result do
      {:ok, %Detour{} = new_record} ->
        handle_detour_updated(detour_changeset, new_record, author_id)

      _ ->
        nil
    end

    upsert_result
```

lib/skate/detours/detours.ex
```diff
@@ -253,14 +252,18 @@ defmodule Skate.Detours.Detours do
...
-    with {:ok, detour} <- fetch_detour_for_activation(detour_id, user_id),
+  def activate_detour(detour_id, author_id, selected_duration, selected_reason) do
+    with {:ok, detour} <- fetch_detour_for_activation(detour_id, author_id),
          :ok <- validate_detour_status(detour),
-         changeset <- build_activation_changeset(detour, selected_duration, selected_reason),
-         {:ok, updated_detour} <- Repo.update(changeset) do
-      handle_detour_updated(changeset, updated_detour, user_id)
-
-      {:ok, updated_detour}
+         partial_changeset <-
```
lubeso, line 258: consider more descriptive names:

```suggestion
         activation_changeset <-
```

or

```suggestion
         detour_changeset_from_activation <-
```

lib/skate/detours/detours.ex
```diff
@@ -497,61 +482,69 @@ defmodule Skate.Detours.Detours do
...
-        else
-          {:error, :not_found}
-        end
+  # Remove adjustment from swiftly when an activated drawn detour is updated to be text-only
+  defp update_swiftly(
+         %Ecto.Changeset{data: %Detour{swiftly_id: swiftly_id}, changes: %{is_text_only: true}},
+         %Detour{},
+         true
+       )
+       when swiftly_id != nil do
```
lubeso, line 508: my curiosity ... what's the difference between this and `!is_nil(swiftly_id)`?

lubeso (commented):
good job; please review my comments and i will approve after resolving all conversations

lib/skate/detours/db/detour.ex
stevejrowe: Yup, the swiftly ID is generated during detour activation, so to say something is activated is like saying it was sent to swiftly (aside from text-only detours).

I can move it down to activated properties though.

lib/skate/detours/db/detour.ex
stevejrowe: Yes, swiflty_id is related to them to. These are the db attributes which are all related to a detour being "active" - the ones that must be cleared out to make a detour that has been active into a draft that needs to be filled out and sent to swiftly.

lib/skate/detours/db/detour.ex
stevejrowe: I would rather keep it as written for two reasons:
1. I prefer the explicitly handling all three of the cases separately - one for inserting a new ID, one for deleting the ID, and one for making no change. I will make comments on each to further that clarity.
2. If there is an error, then it would assign the ID as nil and continue as if it were a success instead of surfacing the error.

lib/skate/detours/detours.ex
stevejrowe: I'm happy to make the variable name more descriptive, although I think the key thing to know about this variable is that it is not the full changeset - that comes later at line 213.

* We know it is from_snapshot, because it is in upsert_from_snapshot, and the result of `Skate.Detours.SnapshotSerde.deserialize(author_id, snapshot)`
* We know it is the detour_changeset, because there is no other changeset within skate/detours, and that is what the snapshot is of.

Are you ok with keeping it as is or is there any other reason to add either bit of information?

priv/repo/async_migrations/20260716202203_backfill_detour_swiftly_id.exs
stevejrowe: No, I'll remove it, thank you

lib/skate/detours/detours.ex
stevejrowe: Similarly I feel like there's no need to add `activation` or `detour` to the variable name because it is within activate_detour, and the important information to know about it is the fact that it is partial, not the complete changeset.

lib/skate/detours/detours.ex
stevejrowe: Seems like there is actually a lot written online about what is idiomatic in elixir for guard conditions. Although conditional operators are valid in guards, determining equality could be more tricky or could have unexpected results when dealing with custom structs.

It seems like the community is pretty consolidated around preferring built in type-check functions in guards, and usually in the positive rather than negative. I'll update this to be `is_binary(swiftly_id)` instead.

Thanks for pointing this out.

lib/swiftly/api/service_adjustments.ex
stevejrowe: 1. 2027 is very much not correct. That was an important typo to find. Thank you.
2. Removing the TZ from this doesn't really have consequence. We are not using createdAfter in any really dynamic way.
3. We could do some imprecise estimate based on the oldest activated_at for active detours - however this is used in the sync which has the goal of finding detours that should be deactivated, but are not - so it would defeat the purpose to go based on that condition.

I'll keep it hard coded, but correct the typo - there's a chance we'll even do without this sync if we end up closing most of the detours nightly.

lib/helpers.ex
stevejrowe: It's true, the error tuple is really the case where I would expect retry to be used. 

Error atoms are barely used in elixir, and I would never expect them to be the result of an asynchronous call that would sometimes succeed.

On the other hand, I think if something were to raise an exception, there's a good chance it signals that something should halt the process, and possibly not even retry.

I'll update the comment though.

lib/helpers.ex
stevejrowe: Why delete this?

lib/swiftly/api/service_adjustments.ex
stevejrowe: This update is intentional. update_swiftly() returns the response of either `create_adjustment_v1`, `delete_adjustment_v1`, or the atom `:ok` if no change was made. As create_adjustment_v1 returns {:ok, adjustment_id}, it's helpful to know an action was made, that is there is no longer an associated adjustment_id.

That way nil can be applied to the changeset on delete.

priv/repo/async_migrations/20260716202203_backfill_detour_swiftly_id.exs
stevejrowe: > case Skate.Detours.Detours.get_swiftly_adjustments() do
> {:ok, adjustments} when is_list(adjustments) -> adjustments
>  _ -> []

If get_swiftly_adjustments fails five times we want to know about it, and consider this migration a failure, not default to none found.

I'm also not inclined to replace Map.new with an Enum.map, the second arg of Map.new is a transform function that performs the same purpose.

lib/skate/detours/detours.ex
stevejrowe: I want to surface the error if it is unsuccessful instead of return an empty array.

lib/skate/detours/detours.ex
stevejrowe: Good suggestion!

lib/skate/detours/detours.ex
stevejrowe: We can have a conversation around coding styles, but I think we should try to write elixir code that is idiomatic. 

Your suggestion is clear, but I think the declared intermediate variables and handling each with conditionals is a very imperative style. Using with to define the "happy path", and letting errors break the chain of actions (and abandon the detour updates) I think is more consistent with the codebase.

Additionally, regarding 
>avoid defining a new private method just for the changed fields and insert result

I think defining small private functions to encapsulate code with a singular purpose  is a pattern widely used in functional programming.

All that said though, I appreciate the effort you've made to refactor this code, but I get the sense that we could use a conversation around what coding style we want to work toward as a team.

lib/helpers.ex
lubeso: i thought writing doctests was not an established convention in the codebase

lubeso (approved):

lib/helpers.ex
stevejrowe: Makes sense. I haven't seen many of them. Maybe they're more of a thing for helper functions?

PR #3257, opened 2026-08-17
Override default bootstrap color for pagination active state

vgu-mbta:
Asana Ticket: [QA Review: Pagination](https://app.asana.com/1/15492006741476/project/1148853526253420/task/1217493599510090?focus=true)

Later discussed with Brian: The purple for Detours button should match as well and both should be var(--Eggplant-500, #6D39AC);

### DIFF
--- a/assets/src/components/nav/paginationBar.tsx
+++ b/assets/src/components/nav/paginationBar.tsx
@@ -19,9 +19,13 @@ export const PaginationBar = ({
   onNext,
   onSelectPage,
 }: PaginationBarProps) => {
+  const paginationStyle = {
+    "--bs-pagination-active-bg": "var(--bs-link-color)",
+    "--bs-pagination-active-border-color": "var(--bs-link-color);",
+  } as React.CSSProperties
   return (
     <div className="d-flex justify-content-end mb-4">
-      <Pagination className="mb-0">
+      <Pagination className="mb-0" style={paginationStyle}>
         <Pagination.Prev
           disabled={pageNumber <= 1}
           aria-disabled={pageNumber <= 1}
--- a/assets/src/components/nav/paginationBar.tsx
+++ b/assets/src/components/nav/paginationBar.tsx
@@ -12,11 +12,6 @@ interface PaginationBarProps {
   style?: React.CSSProperties
 }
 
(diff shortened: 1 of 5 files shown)


assets/src/components/nav/paginationBar.tsx
```diff
@@ -19,9 +19,13 @@ export const PaginationBar = ({
   onNext,
   onSelectPage,
 }: PaginationBarProps) => {
+  const paginationStyle = {
+    "--bs-pagination-active-bg": "var(--bs-primary)",
+    "--bs-pagination-active-border-color": "var(--bs-primary);",
+  } as React.CSSProperties
```
lubeso: defining this variable within the component body implies:

* the variable cannot be reused across component render cycles
* the component style cannot be overridden by a caller

would you consider defining an optional prop with a default fallback value?:

```ts
interface PaginationBarProps {
  // ... same as before ...
  style?: React.CSSProperties
}

const DEFAULT_PAGINATION_STYLE = {
    "--bs-pagination-active-bg": "var(--bs-primary)",
    "--bs-pagination-active-border-color": "var(--bs-primary);",  
} as React.CSSProperties

export const PaginationBar = ({
  // ... same as before ...
  paginationStyle,
}: PaginationBarProps) => {
  return (
    <div className="d-flex justify-content-end mb-4">
      <Pagination className="mb-0" style={paginationStyle ?? DEFAULT_PAGINATION_STYLE}>
      // ... same as before ...
```

> [!NOTE]
> one last thing... did you try using this syntax before using type casting?:
>
> ```ts
> const paginationStyle: React.CSSProperties = { ... }
> ```
> 
> this syntax allows the compiler to identify any issues, whereas `... as ...` suppresses any &mdash; which is fine in the case that you know the type can never not be that type, but i personally think that should be a last resort

vgu-mbta: yes thanks for pointing that out. I tried 
`const paginationStyle: React.CSSProperties = { ... }`
but it gave me a typecheck error because React.CSSProperties type doesn't include CSS custom properties (those starting with --) so I'm going to try putting the style in a css file

lubeso (approved):

PR #3259, opened 2026-08-19
fix: Prevent unremovable detour

stevejrowe:

### Summary of changes
* `priv/repo`: db migrations to add the fields necessary to populate the snapshot from flattened fields.
* `snapshot_serde`: serialize_snapshot delivers data to the front end as a snapshot. It use to derive those values from the stored snapshot. Now it primarily uses the database fields and uses the snapshot as a backup. If we no longer receive missing value warnings, we can retire the snapshot entirely from the db.
* `lib/skate/detours/db/detour.ex`: Add all the new fields to the db schema and define their changeset behavior.
    * A refactor: There are several fields that should actually be nullable, and others where they shouldn't be. Refactored to clearly store those lists.
    * copy to draft was also reworked to not depend on the state value so much.
* various other files were updated (swiftly, s3_export, notifications) to use the database columns instead of the state values.

### DIFF
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -194,61 +194,60 @@ defmodule Skate.Detours.Db.Detour do
     end
   end
 
-  defp put_change_from_state(changeset, field, path) do
-    case {fetch_field(changeset, field), fetch_change(changeset, :state)} do
-      {{:data, table_value}, {:ok, state}} ->
-        context_value = get_in(state, path)
+  @state_fields %{
+    state_value: ["value"],
+    snapshot_children: ["children"],
+    undo_stack: ["context", "undoStack"],
+    estimated_duration: ["context", "selectedDuration"],
+    route_patterns: ["context", "routePatterns"],
+    garages: ["context", "route", "garages"],
+    reason: ["context", "selectedReason"],
+    direction_names: ["context", "route", "directionNames"],
+    edited_directions: ["context", "editedDirections"],
+    detour_shape: ["context", "detourShape"],
+    route_name: ["context", "route", "name"],
+    route_pattern_id: ["context", "routePattern", "id"],
+    route_pattern_name: ["context", "routePattern", "name"],
+    headsign: ["context", "routePattern", "headsign"],
+    is_text_only: ["context", "isTextOnly"],
+    route_id: ["context", "route", "id"]
+  }
 
-        if table_value != context_value and not is_nil(context_value) do
-          put_change(changeset, field, context_value)
-        else
-          changeset
-        end
+  @nullable_state_fields %{
+    nearest_intersection: ["context", "nearestIntersection"],
+    route_segments: ["context", "finishedDetour", "routeSegments"],
+    connection_points: ["context", "finishedDetour", "connectionPoint"],
+    missed_stops: ["context", "finishedDetour", "missedStops"],
+    start_point: ["context", "startPoint"],
+    end_point: ["context", "endPoint"],
+    waypoints: ["context", "waypoints"],
+    typed_detour: ["context", "typedDetour"],
+    coordinates: ["context", "detourShape", "ok", "coordinates"]
+  }
 
-      _ ->
-        changeset
+  defp put_change_from_state(changeset, field, path, allow_nil?) do
+    with {:data, table_value} <- fetch_field(changeset, field),
+         {:ok, state} <- fetch_change(changeset, :state),
+         context_value <- get_in(state, path),
+         true <- table_value != context_value,
+         true <- allow_nil? or not is_nil(context_value) do
+      put_change(changeset, field, context_value)
+    else
+      _ -> changeset
     end
+  end
+
+  defp put_changes_from_state(changeset, fields, allow_nil?) do
+    Enum.reduce(fields, changeset, fn {field, path}, acc ->
+      put_change_from_state(acc, field, path, allow_nil?)
+    end)
   end
 
   defp populate_fields_from_state(changeset) do
     changeset
-    |> put_change_from_state(:state_value, ["value"])
-    |> put_change_from_state(:snapshot_children, ["children"])
-    |> put_change_from_state(:undo_stack, ["context", "undoStack"])
-    |> put_change_from_state(:estimated_duration, ["context", "selectedDuration"])
-    |> put_change_from_state(:reason, ["context", "selectedReason"])
-    |> put_change_from_state(:nearest_intersection, ["context", "nearestIntersection"])
-    |> put_change_from_state(:route_patterns, ["context", "routePatterns"])
-    |> put_change_from_state(:garages, ["context", "route", "garages"])
-    |> put_change_from_state(:direction_names, ["context", "route", "directionNames"])
-    |> put_change_from_state(:edited_directions, ["context", "editedDirections"])
-    |> put_change_from_state(:detour_shape, ["context", "detourShape"])
-    |> put_change_from_state(:route_segments, ["context", "finishedDetour", "routeSegments"])
-    |> put_change_from_state(:connection_points, ["context", "finishedDetour", "connectionPoint"])
-    |> put_change_from_state(:missed_stops, ["context", "finishedDetour", "missedStops"])
-    |> put_change_from_state(:start_point, ["context", "startPoint"])
-    |> put_change_from_state(:end_point, ["context", "endPoint"])
-    |> put_change_from_state(:waypoints, ["context", "waypoints"])
-    |> put_change_from_state(:route_id, ["context", "route", "id"])
-    |> put_change_from_state(:route_name, ["context", "route", "name"])
-    |> put_change_from_state(:route_pattern_id, ["context", "routePattern", "id"])
-    |> put_change_from_state(:route_pattern_name, ["context", "routePattern", "name"])
-    |> put_change_from_state(:headsign, ["context", "routePattern", "headsign"])
-    |> put_change_from_state(:is_text_only, ["context", "isTextOnly"])
-    |> put_change_from_state(:typed_detour, ["context", "typedDetour"])
-    |> populate_coordinates_from_state()
+    |> put_changes_from_state(@state_fields, false)
+    |> put_changes_from_state(@nullable_state_fields, true)
     |> populate_direction_from_state()
-  end
-
-  defp populate_coordinates_from_state(changeset) do
-    case {fetch_field(changeset, :coordinates), fetch_change(changeset, :state)} do
-      {{:data, _},
-       {:ok, %{"context" => %{"detourShape" => %{"ok" => %{"coordinates" => coordinates}}}}}} ->
-        put_change(changeset, :coordinates, coordinates)
-
-      _ ->
-        changeset
-    end
   end
 
   defp populate_direction_from_state(changeset) do
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -46,8 +46,7 @@ defmodule Skate.Detours.Db.Detour do
     field :route_pattern_name, :string
     field :headsign, :string
     field :direction, :string
-    field :direction_id, :integer, virtual: true
-    field :route_pattern, :map, virtual: true
+    field :direction_id, :integer
 
     # Detour shape properties
     field :nearest_intersection, :string
@@ -203,6 +202,7 @@ defmodule Skate.Detours.Db.Detour do
     garages: ["context", "route", "garages"],
     reason: ["context", "selectedReason"],
     direction_names: ["context", "route", "directionNames"],
+    direction_id: ["context", "routePattern", "directionId"],
     edited_directions: ["context", "editedDirections"],
     detour_shape: ["context", "detourShape"],
     route_name: ["context", "route", "name"],
@@ -266,29 +266,6 @@ defmodule Skate.Detours.Db.Detour do
         changeset
     end
   end
-
-  def with_virtual_fields(detour) do
-    detour
-    |> put_route_pattern()
-    |> put_direction_id()
-  end
-
-  defp put_route_pattern(%{route_patterns: route_patterns, route_pattern_id: id} = detour)
-       when is_list(route_patterns) do
-    route_pattern =
-      Enum.find(route_patterns, &(&1["id"] == id))
-
-    %{detour | route_pattern: route_pattern}
-  end
-
-  defp put_route_pattern(detour), do: detour
-
-  defp put_direction_id(%{route_pattern: route_pattern} = detour) when is_map(route_pattern) do
-    direction_id = route_pattern["directionId"]
-    %{detour | direction_id: direction_id}
-  end
-
-  defp put_direction_id(detour), do: detour
 
   defmodule Queries do
     @moduledoc """
--- a/lib/skate/detours/db/detour.ex
+++ b/lib/skate/detours/db/detour.ex
@@ -143,11 +143,11 @@ defmodule Skate.Detours.Db.Detour do
   end
 
   defp add_status(changeset) do
-    case {fetch_field(changeset, :status), fetch_change(changeset, :state_value)} do
-      {{:data, :active}, {:ok, %{"Detour Drawing" => "Past"}}} ->
+    case {fetch_field(changeset, :status), fetch_change(changeset, :state)} do
+      {{:data, :active}, {:ok, %{"value" => %{"Detour Drawing" => "Past"}}}} ->
         put_change(changeset, :status, :past)
 
-      {{:data, :draft}, {:ok, %{"Detour Drawing" => %{"Active" => _}}}} ->
+      {{:data, :draft}, {:ok, %{"value" => %{"Detour Drawing" => %{"Active" => _}}}}} ->
         put_change(changeset, :status, :active)
 
       {{:data, nil}, {:ok, _state}} ->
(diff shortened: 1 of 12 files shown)


lib/skate/detours/detour.ex
```diff
@@ -229,17 +230,16 @@ defmodule Skate.Detours.Detour do
     def from!(%Detour{is_text_only: true} = detour) do
       %{
         base_report(detour)
-        | missed_stops_text_only: get_in(detour.state, ["context", "typedDetour", "missedStops"]),
-          connection_points_text_only:
-            get_in(detour.state, ["context", "typedDetour", "connectionPoints"])
+        | missed_stops_text_only: get_in(detour, ["typedDetour", "missedStops"]),
```
lubeso: **Bug: `get_in/2` on a struct with string keys always returns `nil`.**

`detour` here is a `%Detour{}` struct, not a plain map. `get_in(detour, ["typedDetour", ...])` will always return `nil` because struct access via string keys is not supported by `get_in/2`.

The intended reads should be:
```elixir
missed_stops_text_only: detour.typed_detour["missedStops"],
connection_points_text_only: detour.typed_detour["connectionPoints"]
```
This silently breaks text-only detour reporting on this branch.

lib/skate/detours/db/detour.ex
```diff
@@ -66,23 +72,51 @@ defmodule Skate.Detours.Db.Detour do
...
+
+  defp copy_to_draft_state(state) do
+    state
+    |> Map.put(
+      "context",
+      state["context"]
+      |> Map.merge(%{"status" => "draft"})
+      |> Map.drop(["selectedReason", "selectedDuration", "activatedAt"])
+    )
+    |> Map.put("value", %{SaveState: "Saved", "Detour Drawing": "Share Detour"})
```
lubeso: **Atom/string key mismatch in `copy_to_draft_state/1`.**

Line 114 writes `state["value"]` with atom keys:
```elixir
%{SaveState: "Saved", "Detour Drawing": "Share Detour"}
```
But line 100 sets `state_value` (and the rest of the codebase uses) string keys:
```elixir
%{"SaveState" => "Saved", "Detour Drawing" => "Share Detour"}
```
Atom-keyed and string-keyed maps are not equivalent in Elixir. Any downstream pattern match or `get_in` using string keys against `state["value"]` will fail for detours copied from a past detour via this path.

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
```diff
@@ -0,0 +1,115 @@
...
+      garages: get_in(state, ["context", "route", "garages"]),
+      direction_names: get_in(state, ["context", "route", "directionNames"]),
+      direction_id: get_in(state, ["context", "routePattern", "directionId"]),
+      edited_directions: get_in(state, ["context", "editedDirections"]),
+      detour_shape: get_in(state, ["context", "detourShape"]),
+      route_segments: get_in(state, ["context", "finishedDetour", "routeSegments"]),
+      connection_points: get_in(state, ["context", "finishedDetour", "connectionPoint"]),
+      missed_stops: get_in(state, ["context", "finishedDetour", "missedStops"])
+    }
+    |> Enum.reject(fn {_field, value} -> is_nil(value) end)
```
lubeso, line 98: **Backfill silently skips `state_value` for rows where `state["value"]` is `nil`.**

The `Enum.reject` on line 91 drops nil values from the changeset, so any detour row where `state["value"]` is absent will have `state_value` remain `NULL` after the backfill. The frozen schema on line 10 declares this field `null: false`, creating an inconsistency between schema expectation and actual DB state.

Consider either logging a warning for rows where `state["value"]` is nil, or writing an explicit default (e.g. `%{}`) instead of dropping the key.

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
```diff
@@ -0,0 +1,115 @@
...
+    |> repo().all(log: :info)
+    |> Enum.map(fn %Skate.Repo.Migrations.BackfillSerializingSnapshot.MigratingSchema{
+                     id: id,
+                     state: state
+                   } = detour ->
+      detour
+      |> Ecto.Changeset.change(map_fields(state))
+      |> repo().update()
+      |> case do
+        {:error, _changeset} -> raise "#{id} was not updated"
```
lubeso: **`raise` on a single row failure aborts the entire backfill.**

A single bad row will terminate the batch, leaving the migration in a partially-applied state with no automatic recovery path. Consider catching the error per-row, logging it, and continuing — or at minimum recording the failed `id` so it can be retried in isolation.

lubeso (commented):
Three correctness issues worth addressing before merge, ordered by severity.

lib/skate/detours/detour.ex
lubeso: **Bug: `get_in/2` on a struct raises at runtime.**

`detour` here is a `%Skate.Detours.Db.Detour{}` struct. `Skate.Detours.Db.Detour` does not implement the `Access` behaviour, so `get_in(detour, ["typedDetour", ...])` raises `UndefinedFunctionError` at runtime — it does not return `nil`.

Verified locally:
```
result: {:error, "%UndefinedFunctionError{module: Skate.Detours.Db.Detour, function: :fetch, arity: 2,
  reason: \"Skate.Detours.Db.Detour does not implement the Access behaviour\"}"}
```

The fix is:
```elixir
missed_stops_text_only: detour.typed_detour["missedStops"],
connection_points_text_only: detour.typed_detour["connectionPoints"]
```
This accesses the `:typed_detour` struct field directly (atom key), then indexes into the resulting map with the string key. This path has no test coverage on this branch.

lib/skate/detours/db/detour.ex
lubeso: **Inconsistency: `copy_to_draft_state/1` writes `state["value"]` with atom keys.**

Line 114 writes:
```elixir
%{SaveState: "Saved", "Detour Drawing": "Share Detour"}  # atom keys
```
Every other write to `state["value"]` in the codebase uses string keys — including `state_value` set on line 100 of the same function, `build_activation_changeset`, and `build_deactivation_changeset`.

In practice this does not currently cause a runtime failure: Jason serializes atom keys to strings, and DB round-trips normalize map keys to strings on decode. However, the in-memory struct returned by `Repo.insert()` inside `copy_to_draft_detour` will have `state["value"]` with atom keys until the next DB load. Any future code that pattern-matches on that field in the interim window (e.g. `%{"SaveState" => _}`) will silently fail to match.

Fix — use string keys consistent with every other callsite:
```elixir
|> Map.put("value", %{"SaveState" => "Saved", "Detour Drawing" => "Share Detour"})
```

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
lubeso, line 98: **Backfill silently skips `state_value` for rows with nil `state["value"]`.**

`map_fields/1` builds the field map and then `Enum.reject`s all nil values before applying the changeset. If a row has `state = nil` or is missing the `"value"` key, `state_value` is omitted from the changeset entirely — the column stays `NULL` for that row with no warning logged.

Note: the `null: false` annotation on `state_value` in the frozen schema (line 10) has no enforcement — the DDL migration adds the column as nullable and no Ecto validation is generated for it — so no runtime constraint violation occurs. The risk is silent data loss for the field that is the primary goal of this migration, making the backfill harder to audit.

Consider logging at the call site in `do_change/1` where the ID is in scope:
```elixir
|> Enum.map(fn %MigratingSchema{id: id, state: state} = detour ->
  if is_nil(state) or is_nil(get_in(state, ["value"])) do
    Logger.warning("backfill: detour id=#{id} has nil state_value, skipping field")
  end
  detour
  |> Ecto.Changeset.change(map_fields(state))
  ...
end)
```

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
lubeso: **`raise` on a single row failure halts all remaining rows in the migration.**

A failing `repo().update()` in `do_change/1` raises a `RuntimeError` that propagates through `throttle_change_in_batches` as an exception, bypassing the `error ->` match arm entirely (which is dead code for this path). Rows earlier in the same batch are already committed; all subsequent rows and batches are skipped. The migration must be re-run after addressing the failing row — `page_query` restarts from `id = 0` and re-applying already-set fields is idempotent, so re-running is safe.

Secondary: the `error -> raise error` branch in `throttle_change_in_batches` would crash with `ArgumentError` if a non-exception term were matched, obscuring the original error. Consider `reraise error, __STACKTRACE__` or removing the branch since `do_change` never returns a non-`{:ok, _}` value.

lubeso (commented):
Six issues identified via inline comments, two of which should block merge.

**Must fix before merge:**
- `state_from_detour` unconditional primary clause returns `nil` for all pre-existing rows during the async backfill window — every call to `get_detour_with_state!/1` produces a broken snapshot until backfill completes. (See comment on `snapshot_serde.ex` L155.)
- `get_in(detour, ["typedDetour", ...])` on a `%Db.Detour{}` struct raises `UndefinedFunctionError` at runtime — the text-only report path is broken with no test coverage. (See comment on `detour.ex` L233.)

**Should fix:**

| Location | Issue |
|---|---|
| `snapshot_serde.ex` L218, L281, L442 | Same unconditional-clause bug in `routepatterns_from_detour`, `detourshape_from_detour`, `snapshot_children_from_detour` |
| `db/detour.ex` L114 | `copy_to_draft_state/1` writes `state["value"]` with atom keys; all other callsites use string keys |

**Suggestions:**

| Location | Issue |
|---|---|
| `backfill_serializing_snapshot.exs` L91 | Rows with nil `state["value"]` silently skip `state_value` with no logging; log at call site with detour ID |
| `backfill_serializing_snapshot.exs` L67 | `raise` halts all remaining rows; `error -> raise error` branch is dead code that itself crashes with `ArgumentError` on a non-exception term |

lib/skate/detours/snapshot_serde.ex
```diff
@@ -152,7 +152,8 @@ defmodule Skate.Detours.SnapshotSerde do
     end
   end
 
-  # defp state_from_detour(%Detour{detour_state: state}), do: state
+  defp state_from_detour(%Detour{state_value: state}), do: state
```
lubeso: **Unconditional primary clause bypasses fallback when `state_value` is nil — breaks all detour snapshots during the backfill window.**

`state_value` is a newly added nullable column. Immediately after the DDL migration runs but before the async backfill completes, every pre-existing row has `state_value = NULL`. This clause matches all `%Detour{}` structs unconditionally and returns `nil`, making the fallback on lines 157–165 unreachable.

The snapshot's `"value"` key represents the XState machine state (e.g. `%{"SaveState" => "Saved", "Detour Drawing" => "Share Detour"}`). Returning `nil` for it means every call to `get_detour_with_state!/1` during the backfill window produces a broken snapshot — the frontend cannot restore the machine to its current state.

The pattern was correctly applied in `route_from_detour` (which guards `when not is_nil(route_id) and ...`) and `routepattern_from_detour` (which guards `when is_list(route_patterns)`). Add the same guard here:
```elixir
defp state_from_detour(%Detour{state_value: state}) when not is_nil(state), do: state
```

lib/skate/detours/snapshot_serde.ex
```diff
@@ -194,7 +215,8 @@ defmodule Skate.Detours.SnapshotSerde do
 
   defp routepattern_from_detour(_), do: nil
 
-  # defp routepatterns_from_detour(%Detour{route_patterns: route_patterns}), do: route_patterns
+  defp routepatterns_from_detour(%Detour{route_patterns: route_patterns}), do: route_patterns
```
lubeso: **Same unconditional-primary-clause issue in `routepatterns_from_detour`, `detourshape_from_detour` (L281), and `snapshot_children_from_detour` (L442).**

All three read newly added nullable columns (`route_patterns`, `detour_shape`, `snapshot_children`). All three match unconditionally and return `nil` for unbackfilled rows, bypassing their state fallbacks for the same backfill window reason described at line 155.

The contrast with the guarded counterparts is instructive: `routepattern_from_detour` (singular, immediately above) correctly uses `when is_list(route_patterns)`, which prevents matching when `route_patterns` is nil and lets the fallback fire. Apply the same pattern to all three:
```elixir
defp routepatterns_from_detour(%Detour{route_patterns: rp}) when not is_nil(rp), do: rp
defp detourshape_from_detour(%Detour{detour_shape: ds}) when not is_nil(ds), do: ds
defp snapshot_children_from_detour(%Detour{snapshot_children: sc}) when not is_nil(sc), do: sc
```

lib/skate/detours/db/detour.ex
stevejrowe: Thanks for spotting that oversight!

stevejrowe: Thanks for spotting that oversight! Should be fixed now.

lib/skate/detours/detour.ex
stevejrowe: Thanks! This should be resolved now!

stevejrowe: Thanks, this should be resolved now!

lib/skate/detours/snapshot_serde.ex
stevejrowe: Good idea! I made that change.

lib/skate/detours/snapshot_serde.ex
stevejrowe: Thanks, done!

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
stevejrowe: I made some changes regarding this

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
stevejrowe: That was originally the idea, however after thinking about it, I went for a lighter touch here. Let me know what you think of the changes.

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
```diff
@@ -0,0 +1,128 @@
...
+           {:ok, valid_changeset} <- validate_changeset(changeset),
+           {:ok, %{id: changed_id}} <-
+             repo().update(valid_changeset) do
+        changed_id
+      else
+        {:error, reason} ->
+          Logger.warning(
+            "backfill_migration: Row was not updated detour_id=#{id} reason=#{inspect(reason)}"
+          )
+      end
```
lubeso, line 76: **`Logger.warning` returns `:ok`, which corrupts `next_page` when the last row in a batch fails.**

`do_change` collects results via `Enum.map`, then `throttle_change_in_batches` computes `next_page = results |> Enum.reverse() |> List.first()`. When this `else` branch fires, `Logger.warning(...)` returns `:ok`, placing it in `results`. If the row with the highest ID in the batch fails validation or update, `next_page = :ok`. The subsequent `page_query(:ok)` generates `WHERE id > :ok` — Ecto raises a cast error, crashing the migration despite the intent to continue on per-row failures.

Return `id` from the `else` branch so pagination always advances past failed rows:
```suggestion
      else
        {:error, reason} ->
          Logger.warning(
            "backfill_migration: Row was not updated detour_id=#{id} reason=#{inspect(reason)}"
          )
          id
      end
```

lubeso (commented):
All six issues from the previous review are resolved. Brief confirmation of each:

1. **`detour.ex` L233 — `get_in` on struct (was: must fix):** Resolved. `detour.typed_detour["missedStops"]` correctly accesses the struct field then indexes into the resulting map — no longer raises `UndefinedFunctionError`.

2. **`db/detour.ex` L114 — atom keys in `copy_to_draft_state` (was: should fix):** Resolved. Line 114 now uses string keys `%{"SaveState" => "Saved", "Detour Drawing" => "Share Detour"}`, consistent with every other callsite.

3. **Backfill silent nil skip for `state_value` (was: suggestion):** Addressed. `validate_changeset/1` now requires both `state_value` and `route_patterns` to be present in the changeset changes, and logs a warning with the detour ID when either is absent.

4. **Backfill `raise` halts migration (was: suggestion):** Addressed. Per-row failures now log a warning and continue via `with/else`. The `error -> raise error` dead-code branch in `throttle_change_in_batches` is replaced with a proper `{:error, reason}` match that also logs.

5. **`snapshot_serde.ex` L155 — `state_from_detour` unconditional (was: must fix):** Resolved. `when not is_nil(state)` guard added; the state fallback is now reachable for unbackfilled rows.

6. **`snapshot_serde.ex` L218/281/442 — three unconditional clauses (was: should fix):** Resolved. `when not is_nil(...)` guards added to `routepatterns_from_detour`, `detourshape_from_detour`, and `snapshot_children_from_detour`.

**One new issue** introduced by the fix to item 4: the `else` branch in `do_change` returns `Logger.warning(...)` → `:ok`, which can corrupt pagination. See inline comment on the backfill migration.

priv/repo/async_migrations/20260817204135_backfill_serializing_snapshot.exs
stevejrowe: Thanks, made the change!

lubeso (approved):
The one remaining issue (pagination corruption when `else` returns `:ok`) is resolved — `id` is now returned from the `else` branch, ensuring `results` is always a list of integers and `next_page` is always a valid cursor.

All issues from the initial review are addressed. Approving.

PR #3261, opened 2026-08-20
fix: Prevent duplicate snapshot saves

stevejrowe:

This refactor has been long wanted:
* If you open up your network requests while making a detour you'll see many snapshot updates per action, including when you first load the page.
* The source of the drift between the context and database values may have been a result of snapshot updates racing against each other.
* Even before we had these bugs, there was a feeling that the no-save tags and logic in useDetour was getting unwieldy and unpredictable.
* We also only used the Unsaved / Saving / Saved states for the first save. 

Instead, we only save when the action is raised, and move from unsaved / saved into saving. The useEffect will look for the Saving and begin a save which will also prevent duplicate ones. It will be marked as "Saved" in the backend as well so that when a snapshot is loaded from the db, it will not immediate save on load.

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -54,6 +54,8 @@ export const useDetour = (useDetourProps: UseDetourInput) => {
               .can({ type: "detour.save.set-uuid", uuid: uuid.ok })
           ) {
             actorRef.send({ type: "detour.save.set-uuid", uuid: uuid.ok })
+          } else if (!isOk(uuid)) {
+            actorRef.send({ type: "detour.save.save-failed" })
           }
         })
       }
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -38,38 +38,26 @@ export const useDetour = (useDetourProps: UseDetourInput) => {
 
   // Record snapshots when changed
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


lubeso (changes requested):
The approach is sound and a clear improvement over the tag-based system. One reliability hole should be addressed before merging.

**[medium] assets/src/hooks/useDetour.ts** — If `putDetourUpdate` resolves with a non-`Ok` result, `set-uuid` is never sent. `SaveState` stays in `Saving`, `wasSaving` stays `true` on every subsequent tick, and no future edits will be auto-saved for the rest of the page session. A failure path needs to move the machine back out of `Saving`:

```ts
} else if (!isOk(uuid)) {
  actorRef.send({ type: 'detour.save.save-failed' })
}
```

**[medium] assets/src/models/createDetourMachine.ts** — `SaveState.Saving` has no transition for the failure case above. Without it, `save-failed` would be silently dropped. Add:

```ts
'detour.save.save-failed': {
  target: 'Unsaved',
},
```

Also add `| { type: 'detour.save.save-failed' }` to the `events` type union.

These two are two sides of the same gap — neither is useful without the other.

lubeso (approved):
Both requested changes are in — `save-failed` event added to the type union, `Saving → Unsaved` recovery transition in the state machine, and the matching `else if` branch in `useDetour.ts` to dispatch it. Thanks for turning these around quickly. ✓

PR #3268, opened 2026-08-24
fix: allow moving start and end points off route

stevejrowe:
Asana Ticket: [🐛 Detours of the 120 inbound bus route start on the wrong side of the road](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1217793948661121?focus=true)

Additional issues resolved:
* Dragging the waypoint off the line also made the cursor change back into the add-waypoint icon
* Releasing from that position also added a waypoint


assets/src/components/detours/detourMap.tsx
```diff
@@ -560,9 +566,18 @@ const StartOrEndMarker = ({
               const line = lineRef.current
               if (!marker || !line) return
 
-              const layerPoint = map.latLngToLayerPoint(marker.getLatLng())
-              const closestPoint = line.closestLayerPoint(layerPoint)
-              marker.setLatLng(map.layerPointToLatLng(closestPoint))
+              const markerLatLgn = marker.getLatLng()
+              const markerPoint = map.latLngToLayerPoint(marker.getLatLng())
```
lubeso: Typo in the variable name; the adjacent line also wasn’t updated to use it, so `getLatLng()` is called twice per frame.

```suggestion
              const markerLatLng = marker.getLatLng()
              const markerPoint = map.latLngToLayerPoint(markerLatLng)
```

assets/src/components/detours/detourMap.tsx
```diff
@@ -560,9 +566,18 @@ const StartOrEndMarker = ({
...
+
+              const closestLinePoint = line.closestLayerPoint(markerPoint)
+              const closestLineLatLng = map.layerPointToLatLng(closestLinePoint)
+
+              const bufferedPoint = closestWithBuffer(
+                closestLineLatLng,
+                markerLatLgn,
```
lubeso: Same rename from above.

```suggestion
                markerLatLng,
```

assets/src/components/detours/detourMap.tsx
```diff
@@ -571,6 +586,7 @@ const StartOrEndMarker = ({
 
               line?.setStyle({ opacity: 0 })
               onDragEnd?.(latLngLiteralToShapePoint(marker.getLatLng()))
+              setTimeout(() => toggleDraggingState(), 0)
```
lubeso, line 591: Without this comment, the next reader will remove the `setTimeout` thinking it’s dead code, reintroducing the bug it prevents.

```suggestion
              // defer until after the Leaflet click event that fires synchronously after dragend
              setTimeout(() => toggleDraggingState(), 0)
```

assets/src/util/geoUtils.ts
```diff
@@ -26,3 +26,16 @@ export const calculateGeographicCenter = (
...
+export const closestWithBuffer = (
+  p1: Leaflet.LatLng,
+  p2: Leaflet.LatLng,
+  buffer: number
+) => {
+  const d = p1.distanceTo(p2)
+  const ratio = Math.min(1, buffer / d)
+  const lat = p1.lat + (p2.lat - p1.lat) * ratio
+  const lng = p1.lng + (p2.lng - p1.lng) * ratio
+  return { lat, lng }
```
lubeso, line 40: `closestWithBuffer` is a new pure function with a non-obvious edge case (`d = 0` → `buffer/0 = Infinity` → `Math.min(1, Infinity) = 1` → returns `p2`, which equals `p1`). Suggest adding `assets/tests/util/geoUtils.test.ts`:

```typescript
import { describe, test, expect } from "@jest/globals"
import { LatLng } from "leaflet"
import { closestWithBuffer } from "../../src/util/geoUtils"

describe("closestWithBuffer", () => {
  // Kept separate: the assertion checks an approximate distance from p1,
  // which differs from the exact-equality check used in the parameterized cases below.
  test("clamps result to buffer meters from p1 when marker exceeds the buffer distance", () => {
    const p1 = new LatLng(42.35, -71.06)
    const p2 = new LatLng(42.35, -71.059) // ~82 m away at this latitude
    const result = closestWithBuffer(p1, p2, 7)
    expect(p1.distanceTo(new LatLng(result.lat, result.lng))).toBeCloseTo(7, 0)
  })

  // Parameterized together: both cases produce ratio = 1 (either d < buffer, or
  // d = 0 → buffer/0 = Infinity → Math.min(1, Infinity) = 1), so the function
  // always returns p2 exactly and the assertion is identical.
  test.each([
    {
      label: "marker is within the buffer distance",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06001), // ~1 m away
      buffer: 7,
    },
    {
      label: "p1 and p2 are the same point",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06),
      buffer: 7,
    },
  ])("returns p2 when $label", ({ p1, p2, buffer }) => {
    const result = closestWithBuffer(p1, p2, buffer)
    expect(result).toEqual({ lat: p2.lat, lng: p2.lng })
  })
})
```

lubeso (changes requested):
The approach is sound. `isDragging` + the `setTimeout(0)` correctly solve accidental waypoint insertion; `closestWithBuffer` cleanly replaces strict snap-to-line with a configurable tolerance.

| # | File | Issue | Urgency |
|---|------|-------|---------|
| 1 | `detourMap.tsx` | `setTimeout(0)` in `dragend` has no explanation | **Blocking** — the next reader will remove it thinking it’s dead code, reintroducing the bug |
| 2 | `detourMap.tsx` | Typo `markerLatLgn` + redundant `getLatLng()` call on adjacent line | Non-blocking — easy fix, also removes a semantic inconsistency in a hot-path handler |
| 3 | `geoUtils.ts` | `closestWithBuffer` has no tests | Non-blocking — new function with a non-obvious edge case; sets a bad precedent for `geoUtils` |

---

**A note on practices**

Two habits in this PR are worth reflecting on beyond the specific items above.

**Toggle semantics for set/clear state.** `toggleDraggingState` uses `!isDragging` to both open and close a condition that maps directly to `dragstart`/`dragend` lifecycle events. Toggles depend on perfectly balanced call pairs: any future disruption — an unhandled exception mid-drag, an extra call site, a component unmounting while dragging — silently flips the flag the wrong direction and permanently breaks map interaction, with nothing in the type system to catch it. When a flag has clearly named start and end events, prefer explicit `setIsDragging(true)` at `dragstart` and `setIsDragging(false)` at `dragend` (or equivalently, `onDragStart`/`onDragEnd` callbacks on `StartOrEndMarker`). Idempotent is better than balanced.

**Redundant API calls in hot-path handlers.** The `drag` handler fires every mouse-move frame, yet `marker.getLatLng()` is called twice on adjacent lines because the second line wasn’t updated when the new variable was introduced. Beyond the minor cost, the two calls could theoretically observe different positions, making the logic subtly incorrect. The fix for concern #2 addresses this directly — once the variable name is corrected, use it in both places.

assets/src/components/detours/detourMap.tsx
lubeso: Typo in the variable name; the adjacent line also wasn’t updated to use it, so `getLatLng()` is called twice per frame.

```suggestion
              const markerLatLng = marker.getLatLng()
              const markerPoint = map.latLngToLayerPoint(markerLatLng)
```

assets/src/components/detours/detourMap.tsx
lubeso: Same rename from above.

```suggestion
                markerLatLng,
```

assets/src/components/detours/detourMap.tsx
lubeso, line 591: Without this comment, the next reader will remove the `setTimeout` thinking it’s dead code, reintroducing the bug it prevents.

```suggestion
              // defer until after the Leaflet click event that fires synchronously after dragend
              setTimeout(() => toggleDraggingState(), 0)
```

assets/src/util/geoUtils.ts
lubeso, line 40: `closestWithBuffer` is a new pure function with a non-obvious edge case (`d = 0` → `buffer/0 = Infinity` → `Math.min(1, Infinity) = 1` → returns `p2`, which equals `p1`). Suggest adding `assets/tests/util/geoUtils.test.ts`:

```typescript
import { describe, test, expect } from "@jest/globals"
import { LatLng } from "leaflet"
import { closestWithBuffer } from "../../src/util/geoUtils"

describe("closestWithBuffer", () => {
  // Kept separate: the assertion checks an approximate distance from p1,
  // which differs from the exact-equality check used in the parameterized cases below.
  test("clamps result to buffer meters from p1 when marker exceeds the buffer distance", () => {
    const p1 = new LatLng(42.35, -71.06)
    const p2 = new LatLng(42.35, -71.059) // ~82 m away at this latitude
    const result = closestWithBuffer(p1, p2, 7)
    expect(p1.distanceTo(new LatLng(result.lat, result.lng))).toBeCloseTo(7, 0)
  })

  // Parameterized together: both cases produce ratio = 1 (either d < buffer, or
  // d = 0 → buffer/0 = Infinity → Math.min(1, Infinity) = 1), so the function
  // always returns p2 exactly and the assertion is identical.
  test.each([
    {
      label: "marker is within the buffer distance",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06001), // ~1 m away
      buffer: 7,
    },
    {
      label: "p1 and p2 are the same point",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06),
      buffer: 7,
    },
  ])("returns p2 when $label", ({ p1, p2, buffer }) => {
    const result = closestWithBuffer(p1, p2, buffer)
    expect(result).toEqual({ lat: p2.lat, lng: p2.lng })
  })
})
```

lubeso (changes requested):
The approach is sound. `isDragging` + the `setTimeout(0)` correctly solve accidental waypoint insertion; `closestWithBuffer` cleanly replaces strict snap-to-line with a configurable tolerance.

| # | File | Issue | Urgency |
|---|------|-------|---------|
| 1 | `detourMap.tsx` | `setTimeout(0)` in `dragend` has no explanation | **Blocking** — the next reader will remove it thinking it’s dead code, reintroducing the bug |
| 2 | `detourMap.tsx` | Typo `markerLatLgn` + redundant `getLatLng()` call on adjacent line | Non-blocking — easy fix, also removes a semantic inconsistency in a hot-path handler |
| 3 | `geoUtils.ts` | `closestWithBuffer` has no tests | Non-blocking — new function with a non-obvious edge case; sets a bad precedent for `geoUtils` |

---

**A note on practices**

Two habits in this PR are worth reflecting on beyond the specific items above.

**Toggle semantics for set/clear state.** `toggleDraggingState` uses `!isDragging` to both open and close a condition that maps directly to `dragstart`/`dragend` lifecycle events. Toggles depend on perfectly balanced call pairs: any future disruption — an unhandled exception mid-drag, an extra call site, a component unmounting while dragging — silently flips the flag the wrong direction and permanently breaks map interaction, with nothing in the type system to catch it. When a flag has clearly named start and end events, prefer explicit `setIsDragging(true)` at `dragstart` and `setIsDragging(false)` at `dragend` (or equivalently, `onDragStart`/`onDragEnd` callbacks on `StartOrEndMarker`). Idempotent is better than balanced.

**Redundant API calls in hot-path handlers.** The `drag` handler fires every mouse-move frame, yet `marker.getLatLng()` is called twice on adjacent lines because the second line wasn’t updated when the new variable was introduced. Beyond the minor cost, the two calls could theoretically observe different positions, making the logic subtly incorrect. The fix for concern #2 addresses this directly — once the variable name is corrected, use it in both places.

assets/src/components/detours/detourMap.tsx
lubeso: Typo in the variable name; the adjacent line also wasn’t updated to use it, so `getLatLng()` is called twice per frame.

```suggestion
              const markerLatLng = marker.getLatLng()
              const markerPoint = map.latLngToLayerPoint(markerLatLng)
```

assets/src/components/detours/detourMap.tsx
lubeso: Same rename from above.

```suggestion
                markerLatLng,
```

assets/src/components/detours/detourMap.tsx
lubeso, line 591: Without this comment, the next reader will remove the `setTimeout` thinking it’s dead code, reintroducing the bug it prevents.

```suggestion
              // defer until after the Leaflet click event that fires synchronously after dragend
              setTimeout(() => toggleDraggingState(), 0)
```

assets/src/util/geoUtils.ts
lubeso, line 40: `closestWithBuffer` is a new pure function with a non-obvious edge case (`d = 0` → `buffer/0 = Infinity` → `Math.min(1, Infinity) = 1` → returns `p2`, which equals `p1`). Suggest adding `assets/tests/util/geoUtils.test.ts`:

```typescript
import { describe, test, expect } from "@jest/globals"
import { LatLng } from "leaflet"
import { closestWithBuffer } from "../../src/util/geoUtils"

describe("closestWithBuffer", () => {
  // Kept separate: the assertion checks an approximate distance from p1,
  // which differs from the exact-equality check used in the parameterized cases below.
  test("clamps result to buffer meters from p1 when marker exceeds the buffer distance", () => {
    const p1 = new LatLng(42.35, -71.06)
    const p2 = new LatLng(42.35, -71.059) // ~82 m away at this latitude
    const result = closestWithBuffer(p1, p2, 7)
    expect(p1.distanceTo(new LatLng(result.lat, result.lng))).toBeCloseTo(7, 0)
  })

  // Parameterized together: both cases produce ratio = 1 (either d < buffer, or
  // d = 0 → buffer/0 = Infinity → Math.min(1, Infinity) = 1), so the function
  // always returns p2 exactly and the assertion is identical.
  test.each([
    {
      label: "marker is within the buffer distance",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06001), // ~1 m away
      buffer: 7,
    },
    {
      label: "p1 and p2 are the same point",
      p1: new LatLng(42.35, -71.06),
      p2: new LatLng(42.35, -71.06),
      buffer: 7,
    },
  ])("returns p2 when $label", ({ p1, p2, buffer }) => {
    const result = closestWithBuffer(p1, p2, buffer)
    expect(result).toEqual({ lat: p2.lat, lng: p2.lng })
  })
})
```

lubeso (changes requested):
The approach is sound. `isDragging` + the `setTimeout(0)` correctly solve accidental waypoint insertion; `closestWithBuffer` cleanly replaces strict snap-to-line with a configurable tolerance.

| # | File | Issue | Urgency |
|---|------|-------|---------|
| 1 | `detourMap.tsx` | `setTimeout(0)` in `dragend` has no explanation | **Blocking** — the next reader will remove it thinking it’s dead code, reintroducing the bug |
| 2 | `detourMap.tsx` | Typo `markerLatLgn` + redundant `getLatLng()` call on adjacent line | Non-blocking — easy fix, also removes a semantic inconsistency in a hot-path handler |
| 3 | `geoUtils.ts` | `closestWithBuffer` has no tests | Non-blocking — new function with a non-obvious edge case; sets a bad precedent for `geoUtils` |

---

**A note on practices**

Two habits in this PR are worth reflecting on beyond the specific items above.

**Toggle semantics for set/clear state.** `toggleDraggingState` uses `!isDragging` to both open and close a condition that maps directly to `dragstart`/`dragend` lifecycle events. Toggles depend on perfectly balanced call pairs: any future disruption — an unhandled exception mid-drag, an extra call site, a component unmounting while dragging — silently flips the flag the wrong direction and permanently breaks map interaction, with nothing in the type system to catch it. When a flag has clearly named start and end events, prefer explicit `setIsDragging(true)` at `dragstart` and `setIsDragging(false)` at `dragend` (or equivalently, `onDragStart`/`onDragEnd` callbacks on `StartOrEndMarker`). Idempotent is better than balanced.

**Redundant API calls in hot-path handlers.** The `drag` handler fires every mouse-move frame, yet `marker.getLatLng()` is called twice on adjacent lines because the second line wasn’t updated when the new variable was introduced. Beyond the minor cost, the two calls could theoretically observe different positions, making the logic subtly incorrect. The fix for concern #2 addresses this directly — once the variable name is corrected, use it in both places.

assets/src/components/detours/detourMap.tsx
stevejrowe: Good catch

assets/src/components/detours/detourMap.tsx
stevejrowe: Added the comment

assets/src/util/geoUtils.ts
stevejrowe: You're right, we should have tests for functions like this. I added one for the other function in that file as well. Thanks!

assets/src/components/detours/detourMap.tsx
```diff
@@ -560,9 +566,18 @@ const StartOrEndMarker = ({
               const line = lineRef.current
               if (!marker || !line) return
 
-              const layerPoint = map.latLngToLayerPoint(marker.getLatLng())
-              const closestPoint = line.closestLayerPoint(layerPoint)
-              marker.setLatLng(map.layerPointToLatLng(closestPoint))
+              const markerLatLng = marker.getLatLng()
+              const markerPoint = map.latLngToLayerPoint(marker.getLatLng())
```
lubeso: The typo was fixed but this line still wasn’t updated to use the corrected variable, leaving the redundant call in the hot path.

```suggestion
              const markerPoint = map.latLngToLayerPoint(markerLatLng)
```

assets/tests/util/geoUtils.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe("closestWithBuffer", () => {
+  test("clamps result to buffer meters from p1 when marker exceeds the buffer distance", () => {
+    const p1 = new LatLng(42.35, -71.06)
+    const p2 = new LatLng(42.35, -71.059) // ~82 m away at this latitude
+    const result = closestWithBuffer(p1, p2, 7)
+    expect(p1.distanceTo(new LatLng(result.lat, result.lng))).toBeCloseTo(7, 0)
+  })
+
+  // No buffer ratio is used when the second point is within the buffer
```
lubeso: The current comment — `// No buffer ratio is used when the second point is within the buffer` — only describes the first test case (`d < buffer`) and is technically imprecise (“no ratio is used” when ratio is actually computed and evaluates to 1). More importantly, it says nothing about why the `d = 0` case belongs in this group at all. A reader will look at `"p1 and p2 are the same point"` and have no idea why it shares an assertion with `"marker is within the buffer distance"` — the connection is the non-obvious `buffer/0 = Infinity → Math.min(1, Infinity) = 1` path, which is exactly what the original comment called out.

The comment should explain the shared reason both cases reach the same outcome, covering both paths:

```suggestion
  // In both cases ratio = 1 — either d < buffer, or d = 0 makes buffer/d = Infinity.
```

lubeso (commented):
The blocking concern and tests are addressed — thank you. Two items remain from concern #2.

lubeso (approved):
All concerns addressed promptly — thank you. Approving.

stevejrowe:
@lubeso Thanks for the review!

PR #3280, opened 2026-08-31
fix: prevent race condition between detour calculation and state saving

stevejrowe:

goal: eliminate race conditions between fetching async data and saving and reduce extraneous async calls.

1. unfinished detour should no longer be called when there is an endpoint (finished-detour is called within the machine)
2. finished detour is no longer called when loading an active detour or a complete one from draft (moved from invoke to part of save state)
3. nearest intersection is only called when moving or adding the start point
4. async calls (directions / finished-detour / intersection) are now made before the save is triggered.

Future improvements:
* explicitly handle unfinished-detour state
* clean up the detourShape response and a copy of it in finishedDetour

### DIFF
--- a/assets/src/hooks/useDetour.ts
+++ b/assets/src/hooks/useDetour.ts
@@ -84,12 +84,12 @@ export const useDetour = (useDetourProps: UseDetourInput) => {
 
   const { result: unfinishedDetour } = useApiCall({
     apiCall: useCallback(async () => {
-      if (startPoint && routePattern?.id) {
+      if (startPoint && routePattern?.id && !endPoint) {
         return fetchUnfinishedDetour(routePattern.id, startPoint)
       } else {
         return null
       }
-    }, [startPoint, routePattern]),
+    }, [startPoint, routePattern, endPoint]),
   })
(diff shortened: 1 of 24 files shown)


lubeso (approved):
### Summary of Changes & Assessment

The changes effectively resolve the race conditions between asynchronous data fetching and detour state persistence:
- **Sequential Pipeline**: Moving asynchronous invocations (`fetch-nearest-intersection`, `fetch-detour-directions`, and `fetch-finished-detour`) into a dedicated `Preparing` substate inside `SaveState` ensures saving is triggered only after data resolution.
- **Optimized Network Calls**: Reverse-geocoding is scoped strictly to start-point mutations via `detour.save.prepare-start-point`, and `useDetour` now guards `unfinishedDetour` against execution when an `endPoint` is already defined.
- **Failure Resilience**: The state machine includes fallback transitions to handle service failures without blocking state persistence.
- **Deterministic Testing**: Timer configurations and asynchronous act wrappers have been appropriately updated across the test suite.

No blocking issues identified. Ready for merge.

PR #3281, opened 2026-09-02
chore: block vulnerability scanning requests

vgu-mbta:
Asana Ticket: [Block Vulnerability Scanning Requests](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1218057073828438?focus=true)

Added block_scanner_requests.ex: a plug matching request path/query against a regex for .php/.asp/.cgi extensions, /wp-* WordPress paths, and ../ traversal sequences (raw or percent-encoded). Matches are short-circuited with an immediate 404 and demoted to :debug log level via Logster

### DIFF
--- a/lib/skate_web/plugs/block_scanner_requests.ex
+++ b/lib/skate_web/plugs/block_scanner_requests.ex
@@ -1,0 +1,32 @@
+defmodule SkateWeb.Plugs.BlockScannerRequests do
+  @moduledoc """
+  Short-circuits requests matching common vulnerability-scanner probes, e.g.
+  `/index.php` path-traversal attempts hunting for PHP/CGI misconfigurations
+  that don't exist in this Phoenix app. Skips the router/logger noise by
+  responding 404 immediately and demoting the request's log level.
+  """
+
+  import Plug.Conn
+
+  # Requests for PHP/CGI/ASP files, `wp-*` WordPress paths, or `..` traversal
+  # sequences (raw or percent-encoded) in the path or query string.
+  @scanner_regex ~r{(\.(php\d?|phtml|asp|aspx|cgi)$|/wp-|\.\.(/|\\|%2f|%5c))}i
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    if scanner_request?(conn) do
+      conn
+      |> put_private(:logster_log_level, :debug)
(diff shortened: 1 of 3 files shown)


lib/skate_web/plugs/block_scanner_requests.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  import Plug.Conn
+
+  # Requests for unsupported stacks, exposed metadata/credential files,
+  # traversal/null-byte probes, or fake file-fetch/proxy attempts.
+  @unsupported_stack_regex ~r{(\.(php\d?|phtml|phar|asp|aspx|cgi)(?:$|[/?#])|/(?:wp-admin|wp-login|wp-content|wp-includes)(?:$|[/?#]))}i
+  @traversal_regex ~r{(?:\.\.|%2e%2e)(?:/|\\|%2f|%5c)}i
+  @null_byte_regex ~r{(?:%00|\\u0000|\x00)}i
+  @sensitive_file_regex ~r{(?:^|/)(?:\.git/config|\.env(?:\.[a-z0-9_-]+)?|\.aws/credentials(?:\.[^/?#]+)?|\.azure/credentials(?:\.[^/?#]+)?|terraform\.tfstate(?:\.(?:orig|backup|bak|old))?|\.continue/config\.json|config/anthropic\.json|boot\.ini|win\.ini|etc/passwd)(?:$|[/?#])}i
+  @file_fetch_regex ~r{(?:^|[?&])(?:url|uri|path|endpoint)=.*(?:file:|file%3a|/root/|/etc/passwd|\.aws/credentials|\.azure/credentials)}i
```
stevejrowe, line 17: I know it's a very small computation to run 5 regular expressions on a string, but I have some concern that we're doing more than we need to here, and that doing the small amount of compute _on every request_ may add up.

Additionally, there is the risk that the filter accidentally blocks a legitimate request. I can't find anything in the regex that would lead to a valid request, but it might be a bit of a footgun. 

I would be more comfortable with this PR if you can find a way to downgrade the log_level after the router has determined it's a 404.

I think that using the scanner_request? filter within [register_before_send](https://plug.hexdocs.pm/Plug.Conn.html#register_before_send/2) might work, but I haven't tried

stevejrowe (changes requested):

PR #3283, opened 2026-09-03
feat(radio): 📻 add storybook components and interactive prototype for RTT queue

lubeso:
Asana Ticket: [📻 Storybook Components for RTT queue coded prototype](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1217834908728671?focus=true)

## Summary
Introduces the **Radio RTT (Request to Talk) Queue** component suite, accompanying custom state hooks, SCSS modular styles, Fishery test factories, unit tests, and interactive Storybook stories.

This provides dispatchers with an interactive queue to review, prioritize, respond to, and complete incoming driver RTT and Emergency calls.

---

## Review Guide & Architecture

To avoid fatigue reviewing ~3,600 LoC (over 50% of which is tests and Storybook data), review the changes in these 4 logical layers:

### Layer 1: Core Domain, State Machine & Utilities (~350 LoC)
- `assets/src/components/radio/rtts/types.ts`: Domain models (`RttCall`, `RttCallType`, `RttTab`, `RttStatus`).
- `assets/src/components/radio/rtts/sorting.ts`: Call prioritization (`Emergency > PRTT > RTT`) and reverse-chronological sorting.
- `assets/src/components/radio/rtts/rttQueueReducer.ts`: Pure state machine managing tab switching, call selection, responder assignment, and call completion.
- `assets/src/util/dateTime.ts`: Added `formattedTimeWithSeconds` for timestamp auditing.

### Layer 2: Custom Hooks (~300 LoC)
- `assets/src/components/radio/rtts/useRttQueueTabs.ts`: Tab selection and unread incoming badge count tracking.
- `assets/src/components/radio/rtts/useRttSelection.ts`: Selected vs. active call resolution and live state tracking.
- `assets/src/components/radio/rtts/useRttQueue.ts`: Composite hook coordinating state, memoized sorted lists, and dispatcher actions.

### Layer 3: Presentation Components & SCSS (~1,050 LoC)
- `assets/src/components/radio/rtts/activeBanner.tsx` & `_active_rtt_banner.scss`: Persistent banner when switching tabs while an active call is ongoing.
- `assets/src/components/radio/rtts/queueItem.tsx` & `_rtt_queue_item.scss`: Single queue row with keyboard navigation (`Enter`/`Space`), priority strip, and respond CTA.
- `assets/src/components/radio/rtts/detailsPanel.tsx` & `_rtt_details_panel.scss`: Selected call inspect panel with vehicle header, timestamps, and metadata.
- `assets/src/components/radio/rtts/queue.tsx` & `_rtt_queue.scss`: Composite tabbed container view.

### Layer 4: Tests & Storybook (~1,900 LoC)
- `assets/tests/factories/radio/rtt.ts`: Fishery factory for test fixtures.
- `assets/tests/components/radio/rtts/*`: 100% unit test coverage across reducer, sorting, hooks, and atomic/composite components using `@testing-library/react`.
- `assets/stories/skate-components/radio/rtts/*`: Atomic stories and a full queue lifecycle story with simulated incoming calls and play assertions.

---

## Verification
- [x] `npm --prefix assets run check` (TypeScript compilation, ESLint, Prettier)
- [x] `npm --prefix assets run test -- --watchAll=false` (214 suites passed, 2,130 tests)
- [x] `npm --prefix assets run build-storybook` (Storybook builds successfully)
- [x] Accessibility verified: WAI-ARIA `role="tablist"`, `role="list"`, `role="listitem"`, `role="region"`, and keyboard navigation on queue items.


assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+      type: "RESET"
+      payload?: Partial<RttQueueState>
+    }
+
+export type InitialRttQueueState = Partial<RttQueueState>
+
+export const createInitialRttQueueState = (
+  initialState: InitialRttQueueState = {}
+): RttQueueState => ({
+  tab: initialState.tab ?? "incoming",
```
vgu-mbta, line 33: Probably want to put this in a TAB_TYPE variable in the types.ts file

```
TAB_TYPE = {
   "incoming",
    ...
}
```

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+  incomingCalls: initialState.incomingCalls ?? [],
+  pastCalls: initialState.pastCalls ?? [],
+  selectedCallId: initialState.selectedCallId ?? null,
+  activeCallId: initialState.activeCallId ?? null,
```
vgu-mbta, line 37: Just for my knowledge: when is selectedCallId different from activeCallId?

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+  state: RttQueueState,
+  action: RttQueueAction
+): RttQueueState => {
+  switch (action.type) {
+    case "CHANGE_TAB": {
+      return {
+        ...state,
+        tab: action.tab,
+        newIncomingCount:
+          action.tab === "incoming" ? 0 : state.newIncomingCount,
```
vgu-mbta: Should use the TAB_TYPE.Incoming variable here I mentioned above so that we're not hard-coding "incoming"

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+
+      let newPastCalls = state.pastCalls
+      if (currentActiveId && currentActiveId !== targetCall.id) {
+        const priorCall = state.incomingCalls.find(
+          (c) => c.id === currentActiveId
+        )
+        if (priorCall) {
+          const completed: RttCall = {
+            ...priorCall,
+            status: "done",
```
vgu-mbta, line 76: Just checking: is this the behavior that we want? that when a dispatcher starts a response to a call, the prior call is automatically marked as done? Maybe they were in the middle of a call when a higher priority call came up

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+            markedDoneAt: now,
+          }
+          newPastCalls = [completed, ...newPastCalls]
+        }
+      }
+
+      const newIncomingCalls = state.incomingCalls
```
vgu-mbta, line 84: the double negative is hard to reason about on lines 86 and 89, though I now understand what it's doing. 

consider having a comment above to explain the logic, like
`# Remove previous active calls from the incoming list.`

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+    case "SELECT_CALL": {
+      return {
+        ...state,
+        selectedCallId: action.callId,
+      }
+    }
+
+    case "RESPOND_CALL": {
+      const now = action.answeredAt ?? new Date()
+      const currentActiveId = state.activeCallId
```
vgu-mbta, line 64: do we need to define these as new variables on lines 64-65? I think you could just use state.activeCallId and action.call and it would be less overhead- fewer variables to reason about

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+        incomingCalls: newIncomingCalls,
+        pastCalls: newPastCalls,
+        activeCallId: targetCall.id,
+        selectedCallId: targetCall.id,
+      }
+    }
+
+    case "MARK_DONE_CALL": {
+      const now = action.markedDoneAt ?? new Date()
+      const targetCall = action.call
```
vgu-mbta: same comment as before- not sure we need to rename a variable just to use it once

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+            : state.selectedCallId,
+      }
+    }
+
+    case "RECEIVE_CALL": {
+      return {
+        ...state,
+        incomingCalls: [action.call, ...state.incomingCalls],
+        newIncomingCount:
+          state.tab === "past"
```
vgu-mbta: Add to TAB_TYPES so that this variable is not hard-coded (my comment above)

vgu-mbta: I also don't understand how the incoming count is supposed to behave. wouldn't it increment every time you receive a new call?

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+    () => sortRttCalls(incomingCalls),
+    [incomingCalls]
+  )
+  const sortedPastCalls = useMemo(
+    () => sortRttCalls(pastCalls, { byPriority: false }),
+    [pastCalls]
+  )
+
+  const activeCallsList =
+    tab === "incoming" ? sortedIncomingCalls : sortedPastCalls
```
vgu-mbta: make this a TABS / TAB_TYPE constant (same comment as above)

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+export interface RttQueueActions {
+  selectCall: (call: RttCall) => void
+  respondCall: (call: RttCall) => void
+  markDoneCall: (call: RttCall) => void
+  receiveCall: (call: RttCall) => void
+  changeTab: (tab: RttTab) => void
+  reset: (payload?: Partial<RttQueueState>) => void
+}
+
+export interface UseRttQueueResult {
```
vgu-mbta, line 38: is this variable used anywhere? I didn't find references when I searched the repo

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+        options.selectedCallId !== undefined
+          ? options.selectedCallId
+          : options.initialState?.selectedCallId,
+      activeCallId:
+        options.activeCallId !== undefined
+          ? options.activeCallId
+          : options.initialState?.activeCallId,
+      newIncomingCount:
+        options.newIncomingCount ?? options.initialState?.newIncomingCount,
+    },
```
vgu-mbta: I noticed `createInitialRttQueueState` is only really used in the test code. Instead of this whole chunk of code, it would be less verbose just to define `createInitialRttQueueState` without passing in `initialState`

`export const createInitialRttQueueState = (): RttQueueState => ({
  tab: "incoming",
  incomingCalls: [],
  pastCalls: [],
  selectedCallId: null,
  activeCallId: null,
  newIncomingCount: 0,
})`

assets/src/components/radio/rtts/queueItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+    (call.respondedBy === currentDispatcherName ||
+      call.respondedBy === "YOU" ||
+      call.respondedBy === "You" ||
+      !call.respondedBy)
+
+  const responderDisplayName = isRespondedByCurrentUser
+    ? "YOU"
+    : call.respondedBy || "ACTIVE"
+
+  const modifier = CALL_TYPE_MODIFIERS[call.callType] ?? "rtt"
```
vgu-mbta: why not use CALL_TYPE_MODIFIERS.RTT here? so that "rtt" is not hard-coded

assets/src/components/radio/rtts/queueItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+          }
+        }}
+      >
+        <div className="c-rtt-queue-item__col c-rtt-queue-item__col--type">
+          <span className={`c-rtt-queue-item__type ${typeClass}`}>
+            {call.callType}
+          </span>
+        </div>
+
+        <div className="c-rtt-queue-item__col c-rtt-queue-item__col--time">
```
vgu-mbta, line 97: for maintainability, for these repeated hard-coded items like "c-rtt-queue-item__col" we should declare it as a variable as a CSS class

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: why do you recommend?

if the reason is to avoid providing incorrect string literals (e.g., `incomin` or some other typo), those types of errors are caught at compile time by the typechecker

i can implement your request if you deem it absolutely necessary, but my opinion is that this is sufficient and idiomatic

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+        currentDispatcherName: currentDispatcherName ?? "Current Dispatcher",
+      })
+      onRespondCall?.(call)
+    },
+    [currentDispatcherName, onRespondCall]
+  )
+
+  const handleMarkDoneCall = useCallback(
+    (call: RttCall) => {
+      dispatch({ type: "MARK_DONE_CALL", call })
```
vgu-mbta: In the RttQueueAction, MARK_DONE_CALL passes in markedDoneAt? Date but that variable is not passed through the dispatch call

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+    tab === "incoming" ? sortedIncomingCalls : sortedPastCalls
+
+  const { onRespondCall, onMarkDoneCall, currentDispatcherName } = options
+
+  const handleRespondCall = useCallback(
+    (call: RttCall) => {
+      dispatch({
+        type: "RESPOND_CALL",
+        call,
+        currentDispatcherName: currentDispatcherName ?? "Current Dispatcher",
```
vgu-mbta: "Current Dispatcher" is hardcoded in three separate places. This should be a single exported constant (e.g. DEFAULT_DISPATCHER_NAME in types.ts

assets/src/components/radio/rtts/useRttQueue.ts
```diff
@@ -0,0 +1,219 @@
...
+  const pastCalls = options.pastCalls ?? state.pastCalls
+  const selectedCallId =
+    options.selectedCallId !== undefined
+      ? options.selectedCallId
+      : state.selectedCallId
+  const activeCallId =
+    options.activeCallId !== undefined
+      ? options.activeCallId
+      : state.activeCallId
+  const newIncomingCount = options.newIncomingCount ?? state.newIncomingCount
```
vgu-mbta, line 97: `RttQueueState` seems to have duplicate fields of `UseRttQueueOptions`

I will address this in a pull request

assets/src/components/radio/rtts/rttQueueReducer.ts
vgu-mbta: ah I did not see this declaration `export type RttTab = "incoming" | "past"` so yes the typechecker will caught those types of errors

I'm not requiring you to do it. As best practice though- the code below would enable Visual Studio / your code editor to catch errors in the IDE
```
export type RttTab = "incoming" | "past"

export const TAB_TYPE = {
  INCOMING: "incoming",
  PAST: "past",
} as const satisfies Record<string, RttTab>
```

assets/src/components/radio/rtts/rttQueueReducer.ts
```diff
@@ -0,0 +1,158 @@
...
+export type RttQueueAction =
+  | { type: "CHANGE_TAB"; tab: RttTab }
+  | { type: "SELECT_CALL"; callId: string }
+  | {
+      type: "RESPOND_CALL"
+      call: RttCall
+      currentDispatcherName?: string
+      answeredAt?: Date
+    }
+  | { type: "MARK_DONE_CALL"; call: RttCall; markedDoneAt?: Date }
```
vgu-mbta: In the RttQueueAction, MARK_DONE_CALL passes in markedDoneAt? Date but that variable is not passed through the dispatch call in `useRttQueue.ts`. In general, if I'm passing a variable like "markedDoneAt" into a call, I'm expecting it to be updated in the database and this breaks that expectation. So I would recommend removing that variable from the call OR update it in the database if it is passed in as an optional parameter

vgu-mbta (commented):
Overall great job; I mainly pointed out some places where code maintainability could be improved. I'll have a small PR in response to one of the comments

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: interesting! for that case i would consider an enum, but thank you for sharing!

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: good question! the answer comes from [the description of the Asana ticket](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1217834908728671?focus=true):

> An RTT is considered "Done" when a) "Mark Done" is selected, or **_when the dispatcher whose active call it is selects "Respond" on a new call_**

it might be a good idea to bring this up as part of design review/discussion

assets/src/components/radio/rtts/useRttQueue.ts
lubeso: huh... this should have been specified as the return type for the `useRttQueue` hook...

before i add that, what do you think of just removing it? since it's not used anywhere else...

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: regarding the question about the incoming count, also from the Asana ticket:

> If a new RTT is received when the "Past" tab is open, a badge should appear in the incoming tab [e.g., "1 new"]

so, yes, it should increment every time a new call comes in while viewing the "Past" tab

---

regarding the `TAB_TYPES`, sure, i'll do an enum for that

assets/src/components/radio/rtts/queueItem.tsx
lubeso: for markup like this instance, i'm partial towards repeating the class name each time... but! if you could show an example of what you have in mind, i'll be happy to incorporate it (i don't have a good idea of how to implement what you mean...)

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: imagine a dispatcher has clicked "respond" on a given call

that is now their active and selected call

then, imagine they click on another call **but they do not click respond**

in that case, the selected call would be the one they just clicked on (and would appear in the details pane)

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: good catch! i think for now the better option is to just remove the variable... thanks!

lubeso: Removed `markedDoneAt` from `MARK_DONE_CALL` in commit e3661be9. The reducer now sets completion timestamp with `new Date()` internally.

lubeso: Removed `markedDoneAt` from `MARK_DONE_CALL` in commit e3661be9. The reducer now sets completion timestamp with `new Date()` internally.

lubeso: Removed `markedDoneAt` from `MARK_DONE_CALL` in commit e3661be9. The reducer now sets completion timestamp with `new Date()` internally.

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: Extracted `TAB_TYPE` (with `as const satisfies Record<string, RttTab>`) into `types.ts` and updated references across components, hooks, reducer, and tests in commit 42de989d.

assets/src/components/radio/rtts/useRttQueue.ts
lubeso: Extracted `DEFAULT_DISPATCHER_NAME` into `types.ts` and replaced all hardcoded occurrences in commit 42de989d.

assets/src/components/radio/rtts/useRttQueue.ts
lubeso: Added `UseRttQueueResult` as the explicit return type annotation for the `useRttQueue` hook in commit 0877674f.

assets/src/components/radio/rtts/queueItem.tsx
lubeso: Updated fallback to `CALL_TYPE_MODIFIERS.RTT` instead of `"rtt"` in commit ee67ea77.

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: Added an explanatory comment clarifying that this filter removes previous active calls from the incoming list when transitioning to a new active call in commit 2a00f6cd.

assets/src/components/radio/rtts/rttQueueReducer.ts
lubeso: Cleaned up the redundant local variable aliases in `MARK_DONE_CALL` and referenced `action.call` directly in commit 54505270.

lubeso:
👋 @vgu-mbta — thanks again for the thorough review and helpful feedback!

I've pushed an update addressing the items we discussed across 6 focused commits (`e3661be9..0877674f`):

1. **Extracted Constants:**
   - Defined `TAB_TYPE` (`as const satisfies Record<string, RttTab>`) and `DEFAULT_DISPATCHER_NAME` in `types.ts`, replacing hardcoded string literals across components, reducer, hook, and test files (`42de989d`).
   - Replaced the fallback string `"rtt"` with `CALL_TYPE_MODIFIERS.RTT` in `queueItem.tsx` (`ee67ea77`).

2. **Reducer Cleanups:**
   - Removed the unused `markedDoneAt?: Date` property from `MARK_DONE_CALL`; completion timestamps are now handled directly with `new Date()` internally (`e3661be9`).
   - Simplified single-use temporary variable aliases in the `MARK_DONE_CALL` handler (`54505270`).
   - Added an inline comment explaining the active call filter logic when transitioning between ongoing and incoming calls (`2a00f6cd`).

3. **Hook Return Type Annotation:**
   - Explicitly annotated `useRttQueue` with its return interface `UseRttQueueResult` (`0877674f`).

### Verification
- `npm --prefix assets run check` (TypeScript, ESLint, Prettier) passed with zero warnings.
- `npm --prefix assets run test -- rtt --watchAll=false` (9 test suites, 60 unit tests) all green.

All inline comments have been updated with references to their corresponding commits. Ready for your final look whenever you get a chance!

assets/src/components/radio/rtts/queueItem.tsx
vgu-mbta: opened up a separate PR. It's a small improvement but it helps with readability

vgu-mbta:
Thanks for your revisions! I opened up a small PR that you can approve and merge and then I'll approve this PR 

https://github.com/mbta/skate/pull/3302

vgu-mbta (approved):

vgu-mbta (approved):

PR #3284, opened 2026-09-04
feat: update snapshot logger info

stevejrowe:
Asana ticket: [Update Splunk detour data](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1218054347237280?focus=true)

note: I would have liked to add metadata to the Logster events themselves, but this was simpler.


lib/skate/detours/detours.ex
```diff
@@ -269,6 +269,7 @@ defmodule Skate.Detours.Detours do
     broadcast_detour(new_record, author_id)
     process_notifications(changeset, new_record)
     trigger_active_detour_s3_export_job(changeset, new_record)
+    log_metadata(changeset)
```
lubeso:
```suggestion
    log_metadata(changeset, new_record)
```
Pass `new_record` into `log_metadata/2` so the logger has direct access to the updated, persisted detour struct.

lib/skate/detours/detours.ex
```diff
@@ -649,6 +650,41 @@ defmodule Skate.Detours.Detours do
...
+           reason: reason
+         }
+       }) do
+    Logger.info(
+      "activate_detour id=#{id} estimated_duration=#{String.replace(estimated_duration, " ", "_")} copied_from_id=#{copied_from_id} reason=#{String.replace(reason, " ", "_")}"
+    )
+  end
+
+  defp log_metadata(_), do: nil
+
```
lubeso:
```suggestion
  defp log_metadata(%Ecto.Changeset{changes: %{status: :past}}, %Detour{} = detour) do
    now = DateTime.utc_now()
    minutes_active =
      if detour.activated_at,
        do: DateTime.diff(now, detour.activated_at, :minute),
        else: "unknown"

    Logger.info(
      "deactivate_detour id=#{detour.id} " <>
        "activated_at=#{format_datetime(detour.activated_at)} " <>
        "deactivated_at=#{DateTime.to_iso8601(now)} " <>
        "estimated_duration=#{format_str(detour.estimated_duration)} " <>
        "copied_from_id=#{detour.copied_from_id || "none"} " <>
        "reason=#{format_str(detour.reason)} " <>
        "minutes_active=#{minutes_active}"
    )
  end

  defp log_metadata(%Ecto.Changeset{changes: %{status: :active}}, %Detour{} = detour) do
    Logger.info(
      "activate_detour id=#{detour.id} " <>
        "activated_at=#{format_datetime(detour.activated_at)} " <>
        "estimated_duration=#{format_str(detour.estimated_duration)} " <>
        "copied_from_id=#{detour.copied_from_id || "none"} " <>
        "reason=#{format_str(detour.reason)}"
    )
  end

  defp log_metadata(_, _), do: nil

  defp format_datetime(%DateTime{} = dt), do: DateTime.to_iso8601(dt)
  defp format_datetime(_), do: "unknown"

  defp format_str(nil), do: "none"
  defp format_str(str) when is_binary(str), do: String.replace(str, " ", "_")
```

Refactoring `log_metadata/2` to match against `new_record` ensures we log the persisted record attributes (`estimated_duration`, `reason`, `activated_at`) reliably regardless of whether they were set prior in `changeset.data` or in `changeset.changes`. This also restores `activated_at` upon activation, adds explicit `deactivated_at`, and handles `nil` fallbacks cleanly.

lubeso (changes requested):
Thanks for putting this together, @stevejrowe!

The addition of `:snapshot` to `log_filter_params` in `config/config.exs` is a great improvement that keeps large canvas state payloads out of our HTTP access logs. The logging introduced in `log_metadata/1` also directly supports the PM team's need to track detour durations and copies via Splunk.

Before merging, there are a few items to address around log reliability and completeness:

### Summary of Feedback

1. **[Blocker] Source of truth for logged attributes (`changeset.data` vs. `new_record`)**
   `log_metadata/1` pattern matches exclusively on `changeset.data`. While this functions during the standard frontend flow (as intermediate modal steps autosave `estimated_duration` and `reason` beforehand), it is fragile if activation happens in a single step (such as via direct API calls, tests, or background jobs) where those fields reside in `changeset.changes`. In those scenarios, the pattern match fails and falls through to `defp log_metadata(_), do: nil`, causing activation events to be silently skipped without logging.
   👉 *See the inline suggestions below on `lib/skate/detours/detours.ex`.*

2. **[Nice-to-have] Log completeness (`activated_at` and `deactivated_at`)**
   - In commit `df9e3934`, `activated_at` was removed from the `activate_detour` log. Restoring `activated_at` on activation provides an immediate timestamp in Splunk for real-time visibility without requiring a join against later deactivation events.
   - Emitting an explicit `deactivated_at` ISO-8601 timestamp in addition to `minutes_active` in the deactivation event makes monthly exports more consistent.

3. **[Nice-to-have] Unit test coverage**
   Please consider adding test coverage using `ExUnit.CaptureLog` in `test/skate/detours/detours_test.exs` for both activation and deactivation. This will verify log emission and protect against future regressions as detour lifecycle transitions evolve.

4. **[Nit] Formatting of blank `copied_from_id`**
   When a detour is not copied, `copied_from_id` evaluates to `nil`, resulting in `copied_from_id=` in the log string. Depending on Splunk field extraction configurations, an empty trailing key-value pair can occasionally parse unpredictably; defaulting to `copied_from_id=none` provides a cleaner field extraction.

lib/skate/detours/detours.ex
```diff
@@ -649,6 +650,55 @@ defmodule Skate.Detours.Detours do
...
+    minutes_active =
+      if detour.activated_at,
+        do: DateTime.diff(now, detour.activated_at, :minute),
+        else: "unknown"
+
+    Logger.info(
+      "deactivate_detour id=#{detour.id} " <>
+        "activated_at=#{format_datetime(detour.activated_at)} " <>
+        "deactivated_at=#{format_datetime(now)} " <>
+        "minutes_active=#{minutes_active}" <>
```
lubeso:
```suggestion
        "minutes_active=#{minutes_active} " <>
```

adding space between this value and next label

lubeso (changes requested):
thank you for promptly

stevejrowe:
@lubeso Thanks for the review! I've implemented most of your suggestions. I thought we could get away with only using the changeset due to the narrow changes made during activation / deactivation - but if these functions are ever expanded it does make sense to use the Detour record.

test/skate/detours/detours_test.exs
```diff
@@ -161,6 +162,60 @@ defmodule Skate.Detours.DetoursTest do
...
+
+      log =
+        capture_log(fn ->
+          Skate.Detours.Detours.upsert_from_snapshot(author_id, with_id(snapshot, id))
+        end)
+
+      assert log =~ "deactivate_detour id=#{id}"
+    end
+  end
+
```
lubeso:
```suggestion
  describe "detour lifecycle logging" do
    setup do
      set_log_level(:info)
      reassign_env(:skate, :s3_bucket, nil)
      :ok
    end

    test "logs metadata when activating a detour" do
      %{id: id, author_id: author_id} =
        :detour
        |> build()
        |> insert()

      log =
        capture_log(fn ->
          Detours.activate_detour(id, author_id, "1 hour", "Construction")
        end)

      assert log =~ "activate_detour id=#{id}"
    end

    test "logs metadata when deactivating a detour" do
      %{id: id, author_id: author_id, state: snapshot} =
        :detour
        |> build()
        |> activated()
        |> insert()
        |> deactivated()

      log =
        capture_log(fn ->
          Skate.Detours.Detours.upsert_from_snapshot(author_id, with_id(snapshot, id))
        end)

      assert log =~ "deactivate_detour id=#{id}"
    end
  end
```

Consolidates test setup into a shared `setup` block. Setting `:s3_bucket` to `nil` via `reassign_env/3` safely disables the background S3 export job for these tests, removing the need to mock `ExAws.Request.HttpMock` and eliminating the duplicated mock definitions.

lubeso (approved):
Thanks for addressing the earlier feedback so quickly! The transition to `new_record`, the inclusion of explicit timestamps, and the null-safe helpers look great and resolve the core functional concerns.

I have two minor follow-up suggestions for polish before merge:

1. **Test Setup Deduplication & Decoupling from AWS S3:**  
   The two new tests currently duplicate `set_log_level(:info)` and the `ExAws.Request.HttpMock` expectation. Because the S3 export Oban job is triggered only when `:s3_bucket` is present, calling `reassign_env(:skate, :s3_bucket, nil)` within a shared `setup` block completely bypasses the S3 export pipeline for these tests. This eliminates the need to mock `ExAws` here entirely and removes the repetitive setup code.
   👉 *See inline suggestion on `test/skate/detours/detours_test.exs`.*

2. **Log String Formatting Consistency:**  
   Removed a trailing space at the end of the `deactivate_detour` log format string so it matches `activate_detour`.
   👉 *See inline suggestion on `lib/skate/detours/detours.ex`.*

lib/skate/detours/detours.ex
```diff
@@ -649,6 +650,55 @@ defmodule Skate.Detours.Detours do
...
+        else: "unknown"
+
+    Logger.info(
+      "deactivate_detour id=#{detour.id} " <>
+        "activated_at=#{format_datetime(detour.activated_at)} " <>
+        "deactivated_at=#{format_datetime(now)} " <>
+        "minutes_active=#{minutes_active} " <>
+        "estimated_duration=#{format_str(detour.estimated_duration)} " <>
+        "copied_from_id=#{detour.copied_from_id || "none"} " <>
+        "reason=#{format_str(detour.reason)} "
```
lubeso:
```suggestion
        "reason=#{format_str(detour.reason)}"
```

test/skate/detours/detours_test.exs
stevejrowe: Great suggestion!

PR #3293, opened 2026-09-10
chore: aggregate missing detour structure fields in logs

vgu-mbta:
Asana Ticket: [Unexpected Detour structure ~2k events/day](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1217686645710922?focus=true) part of our overall Splunk investigation which includes decreasing the frequency of redundant warnings

This PR aggregates the "Unexpected detour structure" warnings so that instead of logging an warning in splunk for every field that's missing, it logs one warning per request and lists all the fields that are missing 

Example: we are currently getting 15 splunk events from one detour change request. With this PR we would get one event: `Unexpected detour structure for detour_id=123 author_id=456. Using snapshot for fields: children, isTextOnly, undoStack, selectedReason, selectedDuration, editedDirections, finishedDetour, state`

*Note*: the last commit involves a change to *selectedduration_from_detour* which previously matched unconditionally on %Detour{state: state} instead of requiring state["context"]["selectedDuration"] to be present (unlike its sibling selectedreason_from_detour/1). The change requires state["context"]["selectedDuration"] to be present in order for it to fall back on that.

If you need a rundown on the logic in snapshot_serde.ex:
- Line 87 — Process.delete(@fallback_fields_key) inside serialize_snapshot/1 (line 84).
Runs first, before anything else. It clears out any stale value under the :snapshot_serde_fallback_fields key in the current process's dictionary, so leftover state from a previous call in the same process (e.g. a previous detour serialized in the same request/test) can't bleed into this one.
- Lines 90-93 — building the snapshot map calls state_from_detour/1, context_from_detour/1 (line 138), and snapshot_children_from_detour/1. Each of these may internally call log_fallback(field) (the macro defined at line 156).
- Lines 158-161 (inside the log_fallback macro) — Process.put(@fallback_fields_key, [unquote(field) | Process.get(@fallback_fields_key, [])]).
This runs once per fallback field encountered, each time one of the *_from_detour/1 getters falls back to the raw snapshot instead of a proper DB column. Each call reads whatever list is currently stored (or [] if none yet) and prepends the new field name, then stores it back. So if 8 fields fall back, this line executes 8 times in a row, building up a list like ["state", "finishedDetour", "editedDirections", ...] (in reverse order since each new one is prepended).
- Line 95 — log_fallback_summary(id, author_id), called after the whole snapshot map has been fully built (after every *_from_detour call has had a chance to run log_fallback).

### DIFF
--- a/lib/skate/detours/snapshot_serde.ex
+++ b/lib/skate/detours/snapshot_serde.ex
@@ -8,6 +8,10 @@ defmodule Skate.Detours.SnapshotSerde do
 
   alias Skate.Detours.Db.Detour
 
+  # Process dictionary key used to batch `log_fallback/1` warnings emitted
+  # while serializing a single detour into one aggregated log line.
+  @fallback_fields_key :snapshot_serde_fallback_fields
+
   @doc """
   Converts a XState JSON Snapshot to Detours Database Changeset
   """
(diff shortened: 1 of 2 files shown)


stevejrowe (approved):
Neat! This is a clever approach to reducing the noise. Hopefully we won't be firing off any of these warnings soon enough.

PR #3299, opened 2026-09-14
Fix: reapply detour change with new migration

stevejrowe:
[Prevent detour becoming stuck due to missing context values](https://app.asana.com/1/15492006741476/project/1205385723132845/task/1217159544354832?focus=true)

* the async-migration failed for 4000+ detours, with all of them due to missing fields in validation
  * from the sample I selected, all were missing route_patterns, and so the strategy of calculating route_pattern via route_pattern_id and route_patterns, will not work.
  * considering all fields are backed up via the snapshot in the serialize function, I decided to remove validation, and always insert rows. We can use the logging there to detect if there is missing data.

There are a few options, but I think the best is to store route_pattern (making route_pattern_id, and route_pattern_name unneeded), and not store route_patterns since it is very large, refetched if it is missing, and is only used in a narrow circumstance.

note:
* [3fa2a4d](https://github.com/mbta/skate/pull/3299/commits/3fa2a4dd204e93d541e70a03238fd7a7b810e736) is just reapplying the reverted commit
* [3fa2a4d..c1800b7](https://github.com/mbta/skate/pull/3299/changes/3fa2a4dd204e93d541e70a03238fd7a7b810e736..c1800b7b4c5159e00adc99de173e1615c2ea2eb2) is adding the summarized changes
* 283c171 there is a minor difference in detourShape (the directions include "Return to regular route") - I think it's safe to ignore.
* a new migration had to be made, because the last one ran already, and altering it does not change anything.


lib/skate/detours/snapshot_serde.ex
```diff
@@ -72,8 +72,15 @@ defmodule Skate.Detours.SnapshotSerde do
...
+      |> Map.drop(["context", "activatedAt"])
+      |> Map.drop(["context", "finishedDetour", "detourShape"])
+
+    cleaned_serialized_snapshot =
+      serialized_snapshot
+      |> Map.drop(["context", "activatedAt"])
+      |> Map.drop(["context", "finishedDetour", "detourShape"])
 
     MapDiff.diff(cleaned_state, cleaned_serialized_snapshot)
   end
```
lubeso, line 90: `Map.drop/2` only removes top-level keys; it doesn't follow a path like `pop_in`/`get_in` do. As written, `Map.drop(scoped_state, ["context", "activatedAt"])` removes the **entire `"context"` key** (verified: `Map.drop(%{"context" => %{...}, "value" => %{}}, ["context", "activatedAt"])` returns `%{"value" => %{}}`). The follow-up `Map.drop(..., ["context", "finishedDetour", "detourShape"])` is then a no-op since `context` is already gone.

Net effect: `compare_snapshots`/`validate_serialized_snapshot` no longer compares any route/waypoint/missed-stop/etc. data — the exact mechanism this PR's description says we'll use to detect missing data. Sentry/`Logger.error` alerting for context-level mismatches will never fire.

```suggestion
  defp diff_snapshots(scoped_state, serialized_snapshot) do
    cleaned_state =
      scoped_state
      |> pop_in(["context", "activatedAt"]) |> elem(1)
      |> pop_in(["context", "finishedDetour", "detourShape"]) |> elem(1)

    cleaned_serialized_snapshot =
      serialized_snapshot
      |> pop_in(["context", "activatedAt"]) |> elem(1)
      |> pop_in(["context", "finishedDetour", "detourShape"]) |> elem(1)

    MapDiff.diff(cleaned_state, cleaned_serialized_snapshot)
  end
```

lib/skate/detours/detours.ex
```diff
@@ -512,8 +520,8 @@ defmodule Skate.Detours.Detours do
               !is_nil(Ecto.Changeset.get_change(changeset, :estimated_duration)) ->
                 "active detour #{detour.id} estimated duration changed"
 
-              # ...or when saving changes...
-              is_map(get_in(detour.state, ["context", "savedContext"])) ->
+              # ...or when saving active detour edits (via updated_at changes)...
```
lubeso, line 523: Previously the S3 export only fired on a specific `"savedContext"` marker in `state`, meaning "the user hit final save." Now it fires whenever `updated_at` changes, and `add_updated_at/1` sets `updated_at` for essentially any multi-field change to an active detour. That looks like it broadens the export trigger to nearly every edit of an active detour, not just explicit saves. Is that intentional? If so, might be worth calling out in the PR description since it changes S3 export job frequency/volume.

priv/repo/async_migrations/20260914145515_retry_backfill_serializing_snapshot.exs
```diff
@@ -0,0 +1,123 @@
...
+            "backfill_migration: Row was not updated detour_id=#{id} reason=#{inspect(reason)}"
+          )
+
+          id
+      end
+    end)
+    |> (fn changed -> {:ok, changed} end).()
+  end
+
+  # func (field, func) -> Ecto.Changeset.put_change(field, func(changeset))
```
lubeso: Leftover dev note copied from the prior migration file — doesn't apply to the current code and can be dropped.

```suggestion
```

lubeso (changes requested):
**Summary:** This looks like the right overall direction — it fixes the stuck backfill by dropping the strict validation that was rejecting ~4,000 detours, and it moves detour fields off the `state` JSON blob onto typed columns. `mix test` passes for the touched files.

However, there's a correctness regression in [`SnapshotSerde.diff_snapshots/2`](lib/skate/detours/snapshot_serde.ex#L74-L86) that silently disables the drift-detection this PR explicitly relies on to catch missing/mismatched data going forward (see inline comment). I'd like that fixed before merging. I also left a couple of non-blocking nits/questions.

- 🔴 **Blocking:** `Map.drop/2` misuse strips the entire `context` key instead of a nested field — see comment on `lib/skate/detours/snapshot_serde.ex`.
- 🟡 **Suggestion:** `test/skate/detours/snapshot_serde_test.exs` isn't touched by this PR, but all of its `compare_snapshots` tests only assert the `true` (match) case — none assert that a real mismatch is caught, which is how the bug above passed CI. Worth adding a negative test, e.g.:
  ```elixir
  test "false: detects a genuine mismatch between state and serialized snapshot" do
    %{id: id, state: snapshot} =
      detour =
      :detour
      |> build()
      |> insert()

    snapshot = with_id(snapshot, id)
    mismatched_snapshot = put_in(snapshot, ["context", "route", "name"], "some other route")

    detour =
      detour
      |> Skate.Detours.Detours.change_detour(%{state: mismatched_snapshot})
      |> Skate.Repo.update!()

    assert {false, _diff} = SnapshotSerde.compare_snapshots(detour)
  end
  ```
- 🟡 **Question:** broadened S3-export trigger condition — see comment on `lib/skate/detours/detours.ex`.
- ⚪ **Nit:** stale dev comment — see comment on the retry backfill migration.

lib/skate/detours/detour.ex
```diff
@@ -229,17 +230,16 @@ defmodule Skate.Detours.Detour do
...
     def from!(%Detour{is_text_only: false} = detour) do
       %{
         base_report(detour)
-        | missed_stops: missed_stops!(detour),
-          connection_points: connection_points!(detour),
+        | missed_stops: detour.missed_stops,
+          connection_points: detour.connection_points,
           route_segments: route_segments(detour)
       }
     end
```
lubeso, line 245: The removed `missed_stops!/1` and `connection_points!/1` helpers didn't just relocate this lookup — they **transformed** the raw state data into flat ID lists:

```elixir
defp missed_stops!(%Detour{} = detour) do
  case get_in(detour.state, ["context", "finishedDetour", "missedStops"]) do
    missed_stops when is_list(missed_stops) ->
      missed_stops |> Enum.map(&get_in(&1, ["id"])) |> Enum.reject(&is_nil/1)
    _ -> raise ArgumentError
  end
end
```

`detour.missed_stops`/`detour.connection_points` (the new columns) store the **raw** objects as copied straight from state (see `@nullable_state_fields` in `db/detour.ex`) — full `Stop` maps (`id`/`name`/`lat`/`lon`, per `assets/src/models/finishedDetour.ts`), not extracted ids. So this changes the exported `detours/<status>.ndjson` schema:

- `missed_stops`: `["101", "205"]` → `[%{"id" => "101", "name" => ..., "lat" => ..., ...}, ...]`
- `connection_points`: `["101", "205"]` → `%{"start" => %{...}, "end" => %{...}}`

Any downstream consumer expecting the old flat ID-array shape will break or misparse the new payload. This also silently drops the old fail-fast behavior: previously a detour with missing/malformed finished-detour data raised (and was excluded + logged via `Logger.warning`) instead of being exported with null fields. Not covered by any test today — `s3_exporter_test.exs` only checks that the Oban job starts, not the exported payload shape.

Suggested fix (extract ids at the report-building boundary, matching prior behavior):

```suggestion
    def from!(%Detour{is_text_only: false} = detour) do
      %{
        base_report(detour)
        | missed_stops: missed_stop_ids(detour.missed_stops),
          connection_points: connection_point_ids(detour.connection_points),
          route_segments: route_segments(detour)
      }
    end

    defp missed_stop_ids(missed_stops) when is_list(missed_stops) do
      missed_stops |> Enum.map(&Map.get(&1, "id")) |> Enum.reject(&is_nil/1)
    end

    defp missed_stop_ids(_), do: nil

    defp connection_point_ids(%{"start" => _, "end" => _} = connection_points) do
      ["start", "end"]
      |> Enum.map(&get_in(connection_points, [&1, "id"]))
      |> Enum.reject(&is_nil/1)
    end

    defp connection_point_ids(_), do: nil
```

lib/skate/detours/detour.ex
```diff
@@ -263,45 +263,13 @@ defmodule Skate.Detours.Detour do
...
-           }
+           route_segments: %{
+             "beforeDetour" => before_detour,
+             "afterDetour" => after_detour,
+             "detour" => detour_segment
+           },
+           detour_shape: %{"ok" => %{"coordinates" => bypassed_segment}}
          }) do
       %{
         before_detour: before_detour,
```
lubeso, line 291: Minor/non-blocking: `bypassed_segment` now comes from the top-level `detour.detour_shape["ok"]["coordinates"]` instead of `state["context"]["finishedDetour"]["detourShape"]["coordinates"]`. Per the PR description these two aren't always identical ("there is a minor difference in detourShape... I think it's safe to ignore"), so exported `bypassed_segment` coordinates for some detours may differ slightly from what was exported historically. Flagging for visibility since it changes exported report data, even though you've already called out that the difference is expected/acceptable.

lubeso (changes requested):
**Follow-up:** found an additional correctness issue while reviewing how `s3_exporter.ex`'s `Report.from!/1` (in `lib/skate/detours/detour.ex`) was updated to read directly from the new `missed_stops`/`connection_points` DB columns instead of calling the removed `missed_stops!/1`/`connection_points!/1` helpers.

- 🔴 **Blocking:** the removed helpers didn't just relocate a lookup — they transformed raw state data into flat ID lists/arrays. Reading the new columns directly changes the exported JSON shape for every active/past detour. See inline comment for details and a suggested fix.
- ⚪ **Note:** `route_segments/1`'s `bypassed_segment` now sources from a different (but per the PR description, "safe to ignore") field than before — flagging for visibility, not blocking.

lib/skate/detours/detour.ex
stevejrowe: Thank you! This was an oversight.

lib/skate/detours/snapshot_serde.ex
stevejrowe: 😬 Ouch. I was actually tempted to remove the diffing, considering relying on the context at all could introduce problems with changes not being recorded simply because of a negligible difference with some secondary field. But I did not intend to actually do that in this PR.

stevejrowe: Implemented something pretty close to your suggestion!

lib/skate/detours/detour.ex
stevejrowe: It's good to call out, as it's a decision based on trying to not store redundant information in the database, and also not refactoring the front-end or changing the shape of the snapshot object while it is in this duplicated state.

We could store the detour_shape from finishedDetour separately from the detourShape that is lives on the first level of the context object, however we would be storing almost the same thing twice.

lib/skate/detours/detours.ex
stevejrowe: This was intentional, active detour edits are only sent server-side upon save. This change aligns the condition with the one for sending an updates to an active detour in swiftly

stevejrowe:
@lubeso I think I've addressed each point, let me know if I missed anything. Thanks!

lib/skate/detours/detour.ex
```diff
@@ -229,26 +230,41 @@ defmodule Skate.Detours.Detour do
...
+
+    defp missed_stop_ids(_), do: nil
+
+    defp connection_point_ids(connection_points) when is_list(connection_points) do
+      ["start", "end"]
+      |> Enum.map(&get_in(connection_points, [&1, "id"]))
+      |> Enum.reject(&is_nil/1)
+    end
+
+    defp connection_point_ids(_), do: nil
```
lubeso, line 261: **Impact:** Data accuracy (S3 export). `connection_points` is stored as a `:map` (`%{"start" => %{...}, "end" => %{...}}`, per the `connection_points` column and the frontend's `connectionPoint` model), never a list. `is_list(connection_points)` will never be true for real data, so this function unconditionally falls through to `connection_point_ids(_), do: nil` — meaning `connection_points` is now **always absent** from the exported detour report, not merely absent for malformed data as previously intended.

**Resolution:**

```suggestion
    defp connection_point_ids(connection_points) when is_map(connection_points) do
      ["start", "end"]
      |> Enum.map(&get_in(connection_points, [&1, "id"]))
      |> Enum.reject(&is_nil/1)
    end

    defp connection_point_ids(_), do: nil
```

lib/skate/detours/snapshot_serde.ex
```diff
@@ -367,7 +421,9 @@ defmodule Skate.Detours.SnapshotSerde do
 
   defp typeddetour_from_detour(_), do: nil
 
-  # defp selectedduration_from_detour(%Detour{estimated_duration: estimated_duration}), do: estimated_duration
+  defp selectedduration_from_detour(%Detour{estimated_duration: estimated_duration}),
+    do: estimated_duration
```
lubeso, line 425: **Impact:** Application stability / data accuracy. This clause matches unconditionally, regardless of whether `estimated_duration` is `nil`, which permanently disables the fallback clause below it (including its logic that defaults `nil` `selectedDuration` to `"Until further notice"` for active detours). See the "Systemic issue" section in the review body — the same fix is needed for `startpoint_from_detour/1`, `endpoint_from_detour/1`, `waypoints_from_detour/1`, `typeddetour_from_detour/1`, and `selectedreason_from_detour/1`.

**Resolution:**

```suggestion
  defp selectedduration_from_detour(%Detour{estimated_duration: estimated_duration})
       when not is_nil(estimated_duration),
       do: estimated_duration
```

lubeso (changes requested):
### Summary

This follow-up review covers commits pushed since the last round of feedback. One finding was resolved correctly; one fix introduced a new regression; and further review of the fallback-accessor pattern in `snapshot_serde.ex` surfaced a systemic issue that was not introduced by this specific set of commits, but is directly surfaced by this migration's design and warrants resolution before merge. Findings are detailed below and in inline comments where the diff allows it.

---

### 1. Resolved — Snapshot drift-detection (`diff_snapshots/2`)

The prior `Map.drop/2` misuse (which silently dropped the entire `context` key from comparison) has been correctly replaced with a path-aware `ignore_path/2` built on `pop_in`. No further action needed here.

---

### 2. Blocking — `connection_points` now unconditionally serializes as `nil` (Data Accuracy)

See inline comment on `lib/skate/detours/detour.ex`. The guard used to validate the shape of `connection_points` before extracting IDs is incorrect for the actual stored type, causing this field to be dropped from **every** active/past detour export, not just malformed ones.

---

### 3. Blocking — Missing `is_nil` guards silently disable the legacy-state fallback for several fields (Data Accuracy / Application Stability)

This is not a defect introduced by the author in isolation — it is a direct consequence of this migration's core design: splitting a single, schemaless `state` blob into typed columns while retaining a "read column, else parse legacy state" fallback for the transition period. That fallback requires each column-reading clause to explicitly guard on `not is_nil(...)`; where that guard is omitted, the column-reading clause matches unconditionally (Elixir cannot warn on this, since the two clauses are structurally distinct patterns), and the fallback clause becomes permanently unreachable dead code.

Confirmed instances in `lib/skate/detours/snapshot_serde.ex`:

| Function | Column read (no guard) | Fallback silently dead |
|---|---|---|
| `startpoint_from_detour/1` | `start_point` | yes |
| `endpoint_from_detour/1` | `end_point` | yes |
| `waypoints_from_detour/1` | `waypoints` | yes |
| `typeddetour_from_detour/1` | `typed_detour` | yes |
| `selectedduration_from_detour/1` | `estimated_duration` | yes (see inline comment) |
| `selectedreason_from_detour/1` | `reason` | yes |

By contrast, `route_from_detour/1`, `routepattern_from_detour/1`, `detourshape_from_detour/1`, and `snapshot_children_from_detour/1` correctly guard with `not is_nil(...)` — confirming this was the intended pattern throughout.

**Impact:** because this PR's backfill strategy deliberately inserts rows without validating that every derived column was populated, a meaningful share of historical detours will have `nil` in one or more of the columns above. For those rows, the intended safety net (reconstruct the value from the legacy `state` blob) will never execute. The most operationally significant case is `selectedduration_from_detour/1`: its dead fallback clause also contains the logic that defaults a missing `selectedDuration` to `"Until further notice"` for active detours. Historical active detours with an unpopulated `estimated_duration` column will now be served with **no** selected duration at all, a user-facing regression.

**Proposed resolution:** add `when not is_nil(<field>)` to the first clause of each function listed above, matching the pattern already used elsewhere in the file. A representative fix is proposed as an inline suggestion on `selectedduration_from_detour/1`; the same change should be applied to `startpoint_from_detour/1`, `endpoint_from_detour/1`, `waypoints_from_detour/1`, `typeddetour_from_detour/1`, and `selectedreason_from_detour/1`, e.g.:

```elixir
defp startpoint_from_detour(%Detour{start_point: start_point}) when not is_nil(start_point),
  do: start_point
```

---

### 4. Recommendation — Test fixtures cannot currently exercise the fallback path

`Skate.DetourFactory`'s `:detour` factory always populates the dedicated columns and the legacy `state` blob together and consistently. There is no fixture that produces a detour with a `nil` column but a populated legacy `state`, so no existing test can exercise (or protect) the fallback clauses discussed in finding #3 — including the `connection_points` regression in finding #2, which shipped without any test catching it.

**Proposed reference implementation**, to minimize room for the same class of mistake recurring:

**a. Factory helper** — add to `test/support/factories/detour_factory.ex`:

```elixir
@doc """
Sets the given dedicated column(s) to `nil` so `state` remains the only
populated source of truth. Use to exercise the "column missing, fall back
to legacy state" branches in `Skate.Detours.SnapshotSerde`.
"""
def without_columns(%Skate.Detours.Db.Detour{} = detour, fields) when is_list(fields) do
  Enum.reduce(fields, detour, fn field, acc -> Map.replace!(acc, field, nil) end)
end

def without_columns(%Skate.Detours.Db.Detour{} = detour, field) when is_atom(field) do
  without_columns(detour, [field])
end
```

`Map.replace!/3` is used deliberately instead of `Map.put/3`: it raises immediately if a caller passes a field name that doesn't exist on the struct (e.g. a typo), rather than silently doing nothing — the same "fail loudly on a bad assumption" property missing from the bugs found elsewhere in this PR.

**b. Regression test** — add to `test/skate/detours/snapshot_serde_test.exs`:

```elixir
describe "column/state fallback" do
  test "falls back to legacy state for start_point when the start_point column is nil" do
    detour =
      :detour
      |> build()
      |> insert()
      |> without_columns(:start_point)

    expected = get_in(detour.state, ["context", "startPoint"])
    refute is_nil(expected)

    snapshot = SnapshotSerde.serialize(detour)

    assert get_in(snapshot, ["context", "startPoint"]) == expected
  end
end
```

Repeat this exact shape (swapping the column name and state path) for `end_point` → `["context", "endPoint"]`, `waypoints` → `["context", "waypoints"]`, `typed_detour` → `["context", "typedDetour"]`, `estimated_duration` → `["context", "selectedDuration"]`, and `reason` → `["context", "selectedReason"]`. Each of these tests will fail against the current code until the corresponding `not is_nil(...)` guard from finding #3 is added — add one additional test for `connection_points` (asserting a non-`nil` result) to lock in the finding #2 fix as well.

---

### 5. Nits — resolved

The stale developer comment in the retry backfill migration has been removed as requested. No further action needed.

